PR #8, opened 2016-05-30
move NewUserQuotaInfo from kbfs-server to libkbfs

jinyangli:


strib:
Looks good to me, thanks!

PR #12, opened 2016-05-31
Basic support files for files with holes via truncates

taruti:


libkbfs/folder_block_ops.go
```diff
@@ -959,6 +961,21 @@ func (fbo *folderBlockOps) Read(
 		lastByteInBlock := startOff + blockLen
 
 		if nextByte >= lastByteInBlock {
+			if nextBlockOff > 0 {
+				fill := nextBlockOff - startOff
```
strib: I _think_ this should be:

```
fill := nextBlockOff - nextByte
```

Since `startOff` refers to the start of the block that contains the next byte to be read, but we don't necessarily want to read the block from the start.

libkbfs/folder_block_ops.go
```diff
@@ -1042,27 +1059,76 @@ func (fbo *folderBlockOps) pathFromNodeForBlockWriteLocked(
...
+	if err != nil {
+		return nil, "", err
+	}
+	return fblock, uid, nil
+}
+
+// createInderectBlockLocked creates a new indirect block and
+// pick a new id for the existing block, and use the existing block's ID for
+// the new indirect block that becomes the parent.
+func (fbo *folderBlockOps) createInderectBlockLocked(md *RootMetadata,
```
strib: s/Inderect/Indirect/

libkbfs/folder_block_ops.go
```diff
@@ -1256,36 +1296,113 @@ func (fbo *folderBlockOps) Write(
...
 	iSize := int64(size) // TODO: deal with overflow
-	ptr, parentBlock, indexInParent, block, more, startOff, err :=
+	ptr, parentBlock, indexInParent, block, nextBlockOff, startOff, err :=
 		fbo.getFileBlockAtOffsetLocked(
 			ctx, lState, md, file, fblock, iSize, blockWrite)
 
 	currLen := int64(startOff) + int64(len(block.Contents))
-	if currLen < iSize {
-		// if we need to extend the file, let's just do a write
+	if currLen+128*1024 < iSize {
```
strib: Would prefer the `128*1024` number in a constant somewhere.

strib:
I didn't do a super-careful review yet, but other than my small comments (and the fact that the new truncate test doesn't pass Windows testing), this looks like a good strategy.

I guess maybe no conflict resolution changes are needed after all!

taruti:
Ready for round two.

libkbfs/folder_block_ops.go
```diff
@@ -1042,27 +1059,76 @@ func (fbo *folderBlockOps) pathFromNodeForBlockWriteLocked(
...
+	if err != nil {
+		return nil, "", err
+	}
+	return fblock, uid, nil
+}
+
+// createIndirectBlockLocked creates a new indirect block and
+// pick a new id for the existing block, and use the existing block's ID for
+// the new indirect block that becomes the parent.
+func (fbo *folderBlockOps) createInderectBlockLocked(md *RootMetadata,
```
strib: Still need s/Inderect/Indirect/.

libkbfs/folder_block_ops.go
```diff
@@ -1089,67 +1155,53 @@ func (fbo *folderBlockOps) writeDataLocked(
...
 
 			// Make a new right block and update the parent's
 			// indirect block list
-			err := fbo.newRightBlockLocked(ctx, lState, file.tailPointer(),
+			err = fbo.newRightBlockLocked(ctx, lState, file.tailPointer(),
 				file.Branch, fblock, startOff+int64(len(block.Contents)), md)
 			if err != nil {
 				return WriteRange{}, nil, err
 			}
+		} else if nCopied < n && off+nCopied < nextBlockOff {
```
strib: Hmm, now I'm a little worried about something.  `BlockSplitterSimple.CopyUntilSplit()` currently assumes that the next block starts exactly `maxSize` bytes after this one does.  So I could imagine a scenario where we have these blocks:

```
[0, 512KB], [512KB, 600KB], <hole>, [800KB]
```

Now if someone were to try to write 400KB starting at offset 600KB, `BlockSplitterSimple` would let them, and now the second block would extend past the offset of the final block.  This seems a bit troubling.  We should at least add tests for this case, but we might need to fix it also.  Maybe we can only put holes exactly at block boundaries, or maybe `BlockSplitterSimple` needs to change to take the offset of the next block (if any).

strib:
The changes look good, but I'm worried about one thing mentioned inline.

libkbfs/folder_block_ops.go
taruti: Fixed the doc string but not the code, fixing.

libkbfs/folder_block_ops.go
taruti: Around line 1159 we calculate max and only copy max bytes with `BlockSplitterSimple.CopyUntilSplit()` which should take care of this.

strib: Ah sorry, forgot about that (was just looking at the new changes).  Sounds good, though would be good to have a specific test for that.

strib:
Cool, looks good to me.  I would just like another test or two, especially one that writes past the end of the original truncate hole (and the hole is not aligned on a perfect block boundary).  I'd also like to see a conflict resolution test involving a conflict on a file with a hole in it.

strib:
(Once those tests are in and passing, feel free to merge.)

libkbfs/folder_block_ops.go
```diff
@@ -1256,45 +1310,131 @@ func (fbo *folderBlockOps) Write(
...
-		return nil, nil, err
+		return WriteRange{}, nil, err
 	}
-	if !md.GetTlfHandle().IsWriter(uid) {
-		return nil, nil, NewWriteAccessError(md.GetTlfHandle(), username)
+
+	var dirtyPtrs []BlockPointer
+
+	fbo.log.CDebugf(ctx, "truncateExtendLocked: extending fblock %#v", fblock)
+	if !fblock.IsInd {
```
strib: Sorry, one last thing -- if `fblock.IsInd` is already true, what sets the data version to 2?

libkbfs/config_local.go
```diff
@@ -504,7 +504,16 @@ func (c *ConfigLocal) MetadataVersion() MetadataVer {
 
 // DataVersion implements the Config interface for ConfigLocal.
 func (c *ConfigLocal) DataVersion() DataVer {
-	return 1
+	//	TODO: FilesWithHolesDataVer
+	return FirstValidDataVer
```
strib: Oh and this needs to be updated.

libkbfs/folder_branch_ops.go
```diff
@@ -473,7 +473,8 @@ func (fbo *folderBranchOps) checkDataVersion(p path, ptr BlockPointer) error {
 	if ptr.DataVer < FirstValidDataVer {
 		return InvalidDataVersionError{ptr.DataVer}
 	}
-	if ptr.DataVer > fbo.config.DataVersion() {
+	// TODO: migrate back to fbo.config.DataVersion
+	if ptr.DataVer > FilesWithHolesDataVer {
```
strib: and here

libkbfs/folder_block_ops.go
```diff
@@ -1042,27 +1060,76 @@ func (fbo *folderBlockOps) pathFromNodeForBlockWriteLocked(
...
+		CommonBlock: CommonBlock{
+			IsInd: true,
+		},
+		IPtrs: []IndirectFilePtr{
+			{
+				BlockInfo: BlockInfo{
+					BlockPointer: BlockPointer{
+						ID:       newID,
+						KeyGen:   md.LatestKeyGeneration(),
+						DataVer:  dver,
```
strib: I believe this will just get overridden at https://github.com/keybase/kbfs/blob/master/libkbfs/folder_block_ops.go#L1734, in the `ReadyBlock` call.  It's probably best to just set the data version during `ReadyBlock` somehow, on the top-level indirect block if possible.

strib:
A few more last minute issues, sorry.  Can you please try this out manually where an old client tries to read a new file?  It should get a nice error message about needing to upgrade, rather than trying to read undefined data.

taruti:
Ready for more.

taruti:
Can we merge this?

strib:
I just got in.  I am planning to look it over quickly first before giving the final go-ahead.

libkbfs/config_local.go
```diff
@@ -504,7 +504,15 @@ func (c *ConfigLocal) MetadataVersion() MetadataVer {
 
 // DataVersion implements the Config interface for ConfigLocal.
 func (c *ConfigLocal) DataVersion() DataVer {
-	return 1
+	return FilesWithHolesDataVer
+}
+
+// DefaultNewBlockDataVersion returns the default data version for new blocks.
+func DefaultNewBlockDataVersion(c Config, holes bool) DataVer {
```
strib: Please move this to `block_types.go`.

libkbfs/folder_block_ops.go
```diff
@@ -1256,45 +1310,135 @@ func (fbo *folderBlockOps) Write(
...
+	de.Size = size
+	fbo.deCache[file.tailPointer().ref()] = de
+
+	// Always make the top block dirty, so we will sync its
+	// indirect blocks.  This has the added benefit of ensuring
+	// that any write to a file while it's being sync'd will be
+	// deferred, even if it's to a block that's not currently
+	// being sync'd, since this top-most block will always be in
+	// the fileBlockStates map.
+	for i := range fblock.IPtrs {
```
strib: The comment right above this refers to the `cacheBlockIfNotYetDirtyLocked` call below.  Should probably rearrange it so the comment is immediately above that call.

Also, this loop could use a separate comment.  I assume we're setting `Holes` to `true` on all of the pointers as a shortcut -- we only really need to set it on the one representing the hole.  The shortcut is fine -- I just think it should be commented.

strib:
Looks good modulo comments, thanks!  Feel free to merge when those are addressed.

PR #13, opened 2016-05-31
libdokan: Refuse private directory while errors are present

taruti:
Related to https://github.com/keybase/client/issues/2916


zanderz:
LGTM :thumbsup:
I will test this very soon.

libdokan/fs.go
```diff
@@ -535,6 +535,15 @@ func (r *Root) FindFiles(fi *dokan.FileInfo, callback func(*dokan.NamedStat) err
 		ns.FileAttributes = fileAttributeNormal
 		ns.FileSize = size
 		err = callback(&ns)
+		if err != nil {
```
taruti: Sure, but we do it in other code and it makes it less likely for refactorings to introduce errors.

PR #16, opened 2016-06-01
libfs: Only propagate stable errors

taruti:


zanderz:
Not quite sure what this does but let's try it
LGTM :thumbsup:

taruti:
It makes the following behaviour for remote status errors:
- Display on startup if no good state encountered
- Display while running if we have been 5 seconds in an error state

This avoids returning errors simply because mdserver returning a single error.

zanderz:
You already have LGTM from me, just had the question, let's try it.

PR #22, opened 2016-06-03
Support for TLF resets (pt. 1)

aalness:
This addresses the client-side of: https://keybase.atlassian.net/browse/KBFS-1051

The server-side isn't quite ready but I figured it shouldn't block initial review of this. I'm not planning to merge this until the server-side has been reviewed and is ready to be merged though.

pt. 2 will include addressing:
* https://keybase.atlassian.net/browse/KBFS-1053
* https://keybase.atlassian.net/browse/KBFS-1054


libkbfs/root_metadata.go
```diff
@@ -795,14 +823,39 @@ func (rmds *RootMetadataSigned) Version() MetadataVer {
...
+	}
+	err := rmds.MD.deepCopyInPlace(config.Codec(), false, &newRmds.MD)
+	if err != nil {
+		return nil, err
+	}
+	// clear the serialized data.
+	newRmds.MD.SerializedPrivateMetadata = nil
+	// set the final flag
+	newRmds.MD.Flags |= MetadataFlagFinal
+	// increment revision but keep the prev root
```
strib: What's the reasoning behind keeping the prev root field the same?  Isn't it still useful to have a complete linear chain, even if the field is not properly signed over?  Since this PR doesn't include the `MDOps` changes needed to consume this finalized MD, it's hard for me to tell how it will be used.

libkbfs/tlf_handle_extension.go
```diff
@@ -0,0 +1,218 @@
...
+	TlfHandleExtensionFinalized
+	// TlfHandleExtensionUnknown means the type is unknown.
+	TlfHandleExtensionUnknown
+)
+
+// String implements the fmt.Stringer interface for TlfHandleExtensionType
+func (et TlfHandleExtensionType) String() string {
+	switch et {
+	case TlfHandleExtensionConflict:
+		return "conflicted copy"
```
strib: I think "conflicted copy" and "finalized" should both be constants, since you use them in two places.  You could construct the regex from them as well.

libkbfs/tlf_handle_extension.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if extMap[ext.Type] {
+			// No duplicate extension types in the same suffix.
+			return nil, ErrTlfHandleExtensionInvalidString
+		}
+		extMap[ext.Type] = true
+		extensions = append(extensions, *ext)
+
```
strib: Extra blank line.

libkbfs/tlf_handle.go
```diff
@@ -126,12 +127,22 @@ func MakeBareTlfHandle(
 		sort.Sort(SocialAssertionList(unresolvedReadersCopy))
 	}
 
+	var conflictInfo, finalizedInfo *TlfHandleExtension
+	for _, extension := range extensions {
```
strib: Is there anything dangerous in the fact that a user can type either "(conflicted...) (finalized...)" and "(finalized...) (conflicted...)" on the command line and end up with the same TLF, because this function doesn't make sure the extensions are in the usual canonical order?  I can't think of anything, except possibly that if the user has both paths to the TLF active, the kernel could get confused about its cached when data is updated via one path.

It might avoid some confusion if we either enforce an order here, and/or return a non-canonical error from the TLF parsing function, to turn the unusual version into a symlink.

libkbfs/tlf_handle_extension_test.go
```diff
@@ -0,0 +1,170 @@
...
+	e, err := NewTlfHandleExtension(TlfHandleExtensionConflict, 1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	e2, err := NewTlfHandleExtension(TlfHandleExtensionFinalized, 2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	exts := []TlfHandleExtension{*e, *e2}
+	exts2, err := ParseTlfHandleExtensionSuffix(NewTlfHandleExtensionSuffix(exts))
```
strib: Can we please explicitly verify that `NewTlfHandleExtensionSuffix` returns the expected string?  I was expecting to see something of the form "(conflicted...) (finalized...)" among these tests and never did.

Also worth adding a test for the "(finalized...) (conflicted...)" case I mentioned above, to check whatever behavior we decide is best.

strib:
Looks good to me modulo comments/questions.  I don't see a problem with merging this before the server side is done.

libkbfs/root_metadata.go
aalness: I want the client to be able to verify the signature if masking out the final bit, decrementing the revision, and nulling out the finalized extension info. This way it can easily tell we didn't modify anything unexpected when creating the final block in the server. If I update the prev hash it would also be required to fetch final revision - 1 to verify the signature. I wasn't sure the extra code was worth it given it's not signed over. Don't feel strongly though.

>Since this PR doesn't include the MDOps changes needed to consume this finalized MD, it's hard for me to tell how it will be used.

Yeah, slightly out-of-order here. I need some sort of finalize method to proceed with the server-side changes although I could've stubbed it. Will add tests for this as part of https://keybase.atlassian.net/browse/KBFS-1054.

libkbfs/tlf_handle.go
aalness: I couldn't think of a reason but enforcing that to be safe isn't a problem.

libkbfs/root_metadata.go
strib: Ah right, I forgot about that.  Seems like a good reason, but please put the explanation here in the comments.

aalness:
@strib I think this is ready for another pass.

libkbfs/root_metadata.go
```diff
@@ -790,14 +814,44 @@ func (rmds *RootMetadataSigned) Version() MetadataVer {
...
+		return nil, err
+	}
+	// Copy the signature.
+	newRmds.SigInfo = rmds.SigInfo.deepCopy()
+	// Clear the serialized data.
+	newRmds.MD.SerializedPrivateMetadata = nil
+	// Set the final flag.
+	newRmds.MD.Flags |= MetadataFlagFinal
+	// Increment revision but keep the PrevRoot --
+	// We want the client to be able to verify the signature if masking out the final
```
strib: "if masking out" -> "by masking out"?

libkbfs/tlf_handle_extension.go
```diff
@@ -0,0 +1,249 @@
...
+func (l tlfHandleExtensionList) Less(i, j int) bool {
+	return l[i].Type < l[j].Type
+}
+
+func (l tlfHandleExtensionList) Swap(i, j int) {
+	l[i], l[j] = l[j], l[i]
+}
+
+func (l tlfHandleExtensionList) Splat() (ci, fi *TlfHandleExtension) {
```
strib: Mind commenting this one?  I've never heard of `Splat` before.

strib:
Looks good to me, thanks!  Maybe you already know this, but I just realized the verification step of a finalized TLF in `MDOps` won't work yet until CORE-3065 is in place and called by us.  But that's fine, since nothing will be creating finalized MDs until we turn it on at the servers. So this should be good to go mod small nits.

PR #28, opened 2016-06-07
libkbfs: avoid rekey loops on incomplete rekeys

strib:
If a reader handled a rekey request by setting the rekey bit, but
there was work it couldn't do on behalf of itself (like incrementing
the key generation due to a revoked device), it could enter into a
loop of continuous Put/SendRekeyNeeded messages back and forth with
the server.

Instead, detect if the rekeyer made any useful revisions, and if not,
don't bother putting a new revision.

Issue: KBFS-1119

### DIFF
--- a/libkbfs/key_manager.go
+++ b/libkbfs/key_manager.go
@@ -508,6 +508,7 @@ func (km *KeyManagerStandard) Rekey(ctx context.Context, md *RootMetadata, promp
 	// Figure out if we need to add or remove any keys.
 	// If we're already incrementing the key generation then we don't need to
 	// figure out the key delta.
+	addNewReaderDeviceForSelf := false
 	if !incKeyGen {
 		// See if there is at least one new device in relation to the
 		// current key bundle
@@ -526,6 +527,10 @@ func (km *KeyManagerStandard) Rekey(ctx context.Context, md *RootMetadata, promp
 		incKeyGen = len(wRemoved) > 0 || len(rRemoved) > 0
 
 		promotedReaders = make(map[keybase1.UID]bool, len(rRemoved))
+
+		// Before we add the removed devices, check if we are adding a
+		// new reader device for ourselves.
+		_, addNewReaderDeviceForSelf = newReaderUsers[uid]
 
 		for u := range rRemoved {
 			// FIXME (potential): this could cause a reader to attempt to rekey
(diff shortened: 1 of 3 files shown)


libkbfs/key_manager.go
```diff
@@ -641,8 +641,9 @@ func (km *KeyManagerStandard) Rekey(ctx context.Context, md *RootMetadata, promp
 	if !isWriter {
 		if len(newReaderUsers) > 0 || addNewWriterDevice || incKeyGen {
 			// If we're a reader but we haven't completed all the work, return
-			// RekeyIncompleteError
-			return false, nil, RekeyIncompleteError{}
+			// RekeyIncompleteError.
+			rekeyDone := len(newReaderUsers) > 0
```
aalness: Why does this mean it rekeyed anything?

libkbfs/key_manager_test.go
```diff
@@ -1077,6 +1077,92 @@ func TestKeyManagerReaderRekey(t *testing.T) {
 	}
 }
 
+func TestKeyManagerReaderRekeyAndRevoke(t *testing.T) {
```
jzila: Why do we need to test revocation as part of this case? I would have thought we can do the following:

1. Create u1d1 and u2d1.
2. Write file to u1#u2 from u1d1.
3. Create u1d2.
4. Create u2d2.
5. Rekey from u2d2.
6. Rekey again from u2d2, check that revision hasn't changed.

strib: I tried that at first, but it didn't work because the second Rekey returns an error [here](https://github.com/keybase/kbfs/blob/master/libkbfs/key_manager.go#L573).  To get into this state, the reader needs a revoked device.

(Side note: `ReadAccessError` is probably not the right error for the case I linked to above.)

libkbfs/key_manager.go
aalness: To be more clear: can't this be true yet the client didn't do any useful rekey work?

strib: Well, it means something has changed about the reader itself (a new device, or a revoked device).  It's true that in this case, when Rekey is called twice, the second Rekey won't do anything but rekeyDone will still be true.  Maybe I should pick a better name here.

jzila: `newReaderUsers` means that a different reader has devices that need to be keyed for this directory, and our permission model doesn't allow a reader to key for other readers.

jzila: So andy's point is valid in that the reader could have done nothing and is returning `RekeyIncompleteError`

libkbfs/key_manager_test.go
jzila: Actually the first rekey should be from u2d1, and the second can be from either u2d1 or u2d2 and the result should be the same.

I agree, the case you linked should probably return `RekeyIncompleteError`.

strib:
Ok, I added a commit that tries to be more correct about when a reader rekey was actually done.  Please take another look.

aalness:
👍

PR #37, opened 2016-06-12
folder_block_ops: when creating indirect block, mark it not dirty

strib:
Previously, we put the new block directly into the dirty cache.
However, with the new dirtyFile organization, we first check if it was
dirty before caching it, and the dirtyFile thought it was still marked
as dirty and so didn't re-cache the new indirect block.  As a result,
the sync got the old, direct block out of the cache and sync'd that.
The file size in the direct entry was roughly correct though, so any
reader would see zeroes when they read past the end of that block.

Issue: KBFS-1138


jzila:
Thanks for finding this. LGTM!

PR #40, opened 2016-06-13
Move max block size into a constant

jzila:


libkbfs/folder_branch_ops.go
```diff
@@ -46,6 +46,9 @@ const (
 
 // Constants used in this file.  TODO: Make these configurable?
 const (
+	// The maximum block size in bytes
+	MaxBlockSize = 512 << 10
```
strib: I would call this `MaxBlockSizeBytesDefault`, I think.  When we eventually make a DSL engine for the docker tests, we're going to need the ability to set the block size on the command line or via an environment variable (we already do that now in other DSL tests via the config object), so this constant just acts as a default.  And put "Bytes" in the name to clarify the units.

libkbfs/init.go
```diff
@@ -318,7 +318,7 @@ func Init(params InitParams, onInterruptFn func(), log logger.Logger) (Config, e
 	// Total history size for 1048576-byte blocks: 618945052672 bytes
 	// Total history size for 2097152-byte blocks: 1134341128192 bytes
 	// Total history size for 4194304-byte blocks: 2216672886784 bytes
-	bsplitter, err := NewBlockSplitterSimple(512*1024, 8*1024,
+	bsplitter, err := NewBlockSplitterSimple(MaxBlockSize, 8*1024,
```
strib: You should probably move the comment above to be above the new constant, since it explains why we chose that default value.

libkbfs/folder_branch_ops.go
```diff
@@ -46,6 +46,9 @@ const (
 
 // Constants used in this file.  TODO: Make these configurable?
 const (
+	// The maximum block size in bytes
```
strib: Since this is exported, it's probably good practice to use a standard `// <VarName> does XYZ`-style comment. I'm surprised govet didn't catch this actually.

strib:
Looks good modulo comments.  As I mentioned, I think eventually we'll want this to be easily settable from wherever kbfsfuse is launched, so at that point there may be less need for an exported const (since the launcher can just determine their own block size).  But for now this looks useful, thanks!

PR #43, opened 2016-06-15
libdokan: Fix non-creation of special new folders

taruti:
From KBFS-1086 testing.


zanderz:
Don't really grasp the import, but LGTM :thumbsup:

PR #44, opened 2016-06-15
Taruti/kbfs 1122

taruti:
Cache non-canonical folders with timeout.


zanderz:
LGTM :thumbsup:
I see this cache gets cleared out every hour. Is there a limit to its size? Any scenario where that could be a problem?

taruti:
Size is not limited, then again neither is the tlf-cache size for canonical tlfs.

I don't think there are realistic scenarios where the size will become a problem.

PR #52, opened 2016-06-16
Remove direct use of GlobalContext from package (use an interface)

gabriel:
Removes a direct dependency on GlobalContext for libkbfs package.

This is important for mobile which has issues with the libkb.G global and GlobalContext.

- Replaces libkb.GlobalContext with libkbfs.Context interface
- Replaces access of libkb.G with passed in var (context interface)
- Move construction of context (from GlobalContext) to alternate package (env), since it's used in kbfsfuse, kbfsdokan main packages (and test)
- The impl wraps GlobalContext (implementation hiding) for easy refactoring in the future

In general, we should be using interfaces for library packages like this rather than direct dependencies on something like GlobalContext which itself has a ton of dependencies.

In general we should be deprecating direct use of GlobalContext.

@strib 

cc @maxtaco

### DIFF
--- a/env/context.go
+++ b/env/context.go
@@ -10,10 +10,7 @@ import (
 
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/go-framed-msgpack-rpc"
-	"github.com/keybase/kbfs/libkbfs"
 )
-
-var _ libkbfs.Context = (*Context)(nil)
 
 // Context is an implementation for libkbfs.Context
 type Context struct {
--- a/env/context.go
+++ b/env/context.go
@@ -10,7 +10,10 @@ import (
 
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/go-framed-msgpack-rpc"
+	"github.com/keybase/kbfs/libkbfs"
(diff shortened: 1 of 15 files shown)


gabriel:
Fun fact... 3rd parties could use this package now without constructing a GlobalContext :)

env/context.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"net"
+	"sync"
+
+	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/go-framed-msgpack-rpc"
+)
+
+// Context is an implementation for libkbfs.Context
```
strib: Do you mean `libkb.Context`?  If so, can you please put a `var _ libkb.Context = (*Context)(nil)` somewhere in this file?

strib: Oh, nevermind, I see now the new `libkbfs.Context` class now.  Can you please put a `var _ libkbfs.Context = (*Context)(nil)` somewhere then?

strib:
Nice, this looks good to me, just the one suggestion.  Thanks!

PR #59, opened 2016-06-17
libdokan: SetFileTime for TLFs

taruti:
This adds support for chtimes on TLFs in addition to files/directories inside them.


zanderz:
LGTM :thumbsup:
What used to happen when SetFileTime was called on a TLF? or is this being used yet?

taruti:
It was not supported.

PR #60, opened 2016-06-17
Taruti/libdokan dir setattributes

taruti:
There are no attributes that are supported... but this is needed for SetFileTimes on recent Dokan1.


zanderz:
I'm a little confused as to how it works without these with Dokan1 now if they're required, but LGTM

PR #63, opened 2016-06-20
libkbfs: Support short reads rather than hitting the global timeout

taruti:
Make a smaller timeout for the read if there is an existing timeout and it has room for that.

The inner timeout will fire first and Read handles that returning to upstream which still has time in the global timeout and bubbles the short read up.


zanderz:
LGTM but maybe someone else should look too. @strib ?

strib:
Yes, please wait for my review, hopefully later today.

strib:
Looks good to me, thanks!

PR #64, opened 2016-06-20
libdokan: Add test for mkdir new folder

taruti:
Just a new test. Should catch recent regression fixed on friday in master.


zanderz:
LGTM :thumbsup:

PR #65, opened 2016-06-20
add TLF to favorites in Lookup; move `created` field out of Favorite

songgao:


songgao:
Windows build is failing because of the changed `Favorite` stuct; I'll get a Windows VM running and fix it.

strib:
I don't see any new reason for lookups to happen after the test is shutdown (and I can't reproduce it on either Linux or OS X 10.10).  My best guess is that OS X 10.11 helpfully spawns an extra lookup after the last operation in some test, which triggers the favorite add.

I think you should just go ahead and add something in `Favorites` that ignore new requests after `Favorites.Shutdown()` is called.

Let me know when it's ready for a real review?

songgao:
@strib  This is just for OS X; I haven't tested on Windows yet. Meanwhile, would a test case that spawns `mkdir`, `rmdir`, etc. commands be desired?

strib:
I'd like at least a test case that does a `mkdir` (using `os.Mkdir`) and then makes sure the favorite is in the favorites list after that.  If you want, you can follow that up with an `rmdir` and make sure that the favorite is indeed gone.

It might also be good to add something to `favorites_test.go` that does a regular add, and then a "created" add, and makes sure that the second add still goes through to the server.

libfuse/folderlist.go
```diff
@@ -24,6 +25,9 @@ type FolderList struct {
 
 	mu      sync.Mutex
 	folders map[string]*TLF
+
+	recentlyRemoved   map[libkbfs.CanonicalTlfName]bool
+	muRecentlyRemoved sync.RWMutex
```
strib: We usually put the mutex above the fields it protects (the recommended style for golang).

libfuse/folderlist.go
```diff
@@ -52,6 +56,26 @@ func (fl *FolderList) reportErr(ctx context.Context,
 	fl.fs.errLog.CDebugf(ctx, err.Error())
 }
 
+func (fl *FolderList) addToRecentlyRemove(name libkbfs.CanonicalTlfName) {
+	fl.muRecentlyRemoved.Lock()
+	if fl.recentlyRemoved == nil {
+		fl.recentlyRemoved = make(map[libkbfs.CanonicalTlfName]bool)
+	}
+	fl.recentlyRemoved[name] = true
+	fl.muRecentlyRemoved.Unlock()
```
strib: We find it easier to check for correctness when the Unlock happens in a `defer` right after the Lock.  In this case, I guess you don't want to hold the lock across the `execAfterDelay` call, so you can wrap this first part in an anonymous `func()` like:

```go
func() {
    fl.muRecentlyRemoved.Lock()
    defer fl.muRecentlyRemoved.Unlock()
    if fl.recentlyRemoved ...
    ...
}
fl.fs.execAfterDelay(time.Second, func() {
	fl.muRecentlyRemoved.Lock()
    defer fl.muRecentlyRemoved.Unlock()
	delete(fl.recentlyRemoved, name)
})
```

libfuse/fs.go
```diff
@@ -30,6 +30,10 @@ type FS struct {
 
 	// remoteStatus is the current status of remote connections.
 	remoteStatus libfs.RemoteStatus
+
+	// this is like time.AfterFunc, except that in some tests this can be
+	// overiden to executing f without any delay.
```
strib: "overiden" -> "overridden"

libfuse/folderlist.go
```diff
@@ -67,21 +91,48 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp
...
 	}
 
 	h, err := libkbfs.ParseTlfHandle(
 		ctx, fl.fs.config.KBPKI(), req.Name, fl.public)
+
+	addToFavorite := func() (err error) {
+		cName := h.GetCanonicalName()
+
+		// when a lookup happens (e.g. `mkdir` does a lookup before trying to
+		// make the directory), if the lookup is canonical name of a TLF, the
```
strib: "is canonical" -> "is the canonical"

libfuse/folderlist.go
```diff
@@ -67,21 +91,48 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp
...
+		// `rmdir` command on macOS does a lookup after removing the dir. if the
+		// TLF is recently removed, it's likely that this lookup is issued by the
+		// `rmdir` command, and the lookup should indicate the dir does not exist
+		// and not add the dir to favorites.
+		if !fl.isRecentlyRemoved(cName) {
+			fl.fs.log.CDebugf(ctx, "adding %s to favorites", cName)
+			// internally this calls getOps, which adds the TLF to favorite
+			fl.fs.config.KBFSOps().AddFavorite(ctx, h.ToFavorite())
+		} else {
+			fl.fs.log.CDebugf(ctx, "recently removed; will skip adding %s to favorites and return nil", cName)
```
strib: You're not returning `nil`, you're returning an error.

libfuse/folderlist.go
```diff
@@ -67,21 +91,48 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp
...
-
 	// Shortcut for dreaded extraneous OSX finder lookups
 	if strings.HasPrefix(req.Name, "._") {
 		return nil, fuse.ENOENT
 	}
 
 	h, err := libkbfs.ParseTlfHandle(
 		ctx, fl.fs.config.KBPKI(), req.Name, fl.public)
+
+	addToFavorite := func() (err error) {
```
strib: I think this would be better as a real function on `FolderList` because a) it would keep `Lookup` smaller and more readable, and b) it's only used in the `nil` case below, so we wouldn't waste time constructing the closure for all the other cases.

libfuse/folderlist.go
```diff
@@ -161,26 +212,39 @@ func (fl *FolderList) Remove(ctx context.Context, req *fuse.RemoveRequest) (err
...
-	// TODO trying to delete non-canonical folder handles
-	// could be skipped.
-	//
-	// TODO how to handle closing down the folderbranchops
-	// object? Open files may still exist long after removing
-	// the favorite.
-	return fl.fs.config.KBFSOps().DeleteFavorite(ctx, req.Name, fl.public)
+	switch err := err.(type) {
+	case nil:
+
```
strib: extra blank line.

libfuse/folderlist.go
```diff
@@ -52,6 +56,26 @@ func (fl *FolderList) reportErr(ctx context.Context,
 	fl.fs.errLog.CDebugf(ctx, err.Error())
 }
 
+func (fl *FolderList) addToRecentlyRemove(name libkbfs.CanonicalTlfName) {
```
strib: `addToRecentlyRemove` -> `addToRecentlyRemoved`

libfuse/folderlist.go
```diff
@@ -161,26 +212,39 @@ func (fl *FolderList) Remove(ctx context.Context, req *fuse.RemoveRequest) (err
...
-		if tlf, ok := fl.folders[req.Name]; ok {
-			// Fake future attr calls for this TLF until the user
-			// actually opens the TLF again, because some OSes (*cough
-			// OS X cough*) like to do immediate lookup/attr calls
-			// right after doing a remove, which would otherwise end
-			// up re-adding the favorite.
-			tlf.clearStoredDir()
-		}
-	}()
+	h, err := libkbfs.ParseTlfHandle(
```
strib: I don't see a reason to parse the handle here.  We can just try removing whatever name we're given (like the code did previously), right?

libfuse/mount_test.go
```diff
@@ -603,6 +612,50 @@ func TestCreateExecutable(t *testing.T) {
...
+		"jdoe": nil,
+	})
+	if b, err := exec.Command("/bin/mkdir", "-p", p).CombinedOutput(); err != nil {
+		t.Fatalf("error executing mkdir -p %s: %s (%s)", p, err.Error(), string(b))
+	}
+	checkDir(t, path.Join(mnt.Dir, PrivateName), map[string]fileInfoCheck{
+		"jdoe":         nil,
+		"janedoe,jdoe": nil,
+	})
+	if b, err := exec.Command("/bin/rmdir", p).CombinedOutput(); err != nil {
```
strib: `os.Remove` works for directories too, I'd prefer that rather that using a shell.

libfuse/mount_test.go
```diff
@@ -603,6 +612,50 @@ func TestCreateExecutable(t *testing.T) {
...
+	config := libkbfs.MakeTestConfigOrBust(t, "jdoe", "janedoe")
+	defer libkbfs.CheckConfigAndShutdown(t, config)
+	mnt, _, cancelFn := makeFS(t, config)
+	defer mnt.Close()
+	defer cancelFn()
+	p := path.Join(mnt.Dir, PrivateName, "janedoe,jdoe")
+	checkDir(t, path.Join(mnt.Dir, PrivateName), map[string]fileInfoCheck{
+		"jdoe": nil,
+	})
+	if b, err := exec.Command("/bin/mkdir", "-p", p).CombinedOutput(); err != nil {
```
strib: `os.MkdirAll`?

libkbfs/favorites.go
```diff
@@ -142,6 +160,7 @@ func (f *Favorites) loop() {
 
 // Shutdown shuts down this Favorites instance.
 func (f *Favorites) Shutdown() {
+	atomic.StoreInt32(&f.shutdown, 1)
```
strib: I've never used `atomic` before, and the documentation says it probably shouldn't be used by high-level code.  I agree this all looks ok, but maybe we should use a `sync.Mutex` and a `bool` to be consistent with the documentation and our other code?  I don't feel too strongly about it though.

libkbfs/interfaces.go
```diff
@@ -102,7 +102,10 @@ type KBFSOps interface {
 	// DeleteFavorite deletes the favorite from both the server and
 	// the local cache.  Idempotent, so it succeeds even if the folder
 	// isn't favorited.
-	DeleteFavorite(ctx context.Context, name string, public bool) error
+	DeleteFavorite(ctx context.Context, fav Favorite) error
+	// AddFavorite adds the favorite to both the server and
+	// the local cache.
+	AddFavorite(ctx context.Context, fav Favorite) error
```
strib: Super-nit: please place this above `DeleteFavorite`.

strib:
Looks good to me overall, thanks @songgao.  I think basically all of my comments are nits.

Let me know when your libdokan test is ready and I can take a quick look at that if you want too.

songgao:
Thanks @strib for the detailed review! I'll address all the comments in next commits.

libfuse/folderlist.go
songgao: This is to address a TODO below

> `TODO trying to delete non-canonical folder handles could be skipped.`

I was relying on `ParseTlfHandle` to determine whether the name is canonical. Is there a better way to do this?

strib: Ah, I missed that TODO.  Yes, `ParseTlfHandle` is the right way to do this.  I was worried this would slow down deletes, and make it hard to recover from bugs where we end up with non-canonical names in the favorites list for whatever reason.  But I guess it's the right thing to do, so I'm ok with it.

songgao: Got you. Thanks @strib!

libfuse/mount_test.go
songgao: I was under the impression that it was `rmdir` command rather than the syscall that did the lookup after removing the directory. (same for `mkdir`). However, I just tested it on my macOS and it turns out it's the syscalls that did those. So I guess `TestMkdirAndRmdirTLFCommand` is actually not necessary and should be merged into `TestMkdirTLF`.

songgao:
@strib I finally got the `libdokan` test pass. I also added a small fix for `libfuse`. It's not causing problems now but just to be safe ...

libdokan/folderlist.go
```diff
@@ -86,7 +93,15 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string)
 		}
 		fl.mu.Unlock()
 
+		h, er := libkbfs.ParseTlfHandle(
```
strib: `er` -> `err`?

libdokan/folderlist.go
```diff
@@ -86,7 +93,15 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string)
...
+			ctx, fl.fs.config.KBPKI(), name, fl.public)
+
 		if ok {
+			if er == nil {
+				if err = fl.addToFavorite(ctx, h); err != nil {
```
strib: Not sure why you need this call.  In this case, the node already exists in `fl.folders`, so shouldn't it have already been added to favorites?  If there's a good reason, please add a comment.

libdokan/folderlist.go
songgao: hmm.. This line was moved to above several other calls, and `err` would be overridden by other statements.

strib: Oh I see.  I think it might be confusing to not check an error code immediately after receiving it.  For example, you probably shouldn't be calling `addToFavorite` in the `ok` case if `er` is not `nil` and is not the "canonical name" error.  Hopefully you can just eliminate the first `addToFavorite` call and not need this anymore.  But if you still need it, I would call it something more descriptive (`parseHandleErr`) and check it for fatal error conditions immediately.

libdokan/folderlist.go
songgao: Without this, second call to `mkdir` after `rmdir` won't go through and test fails [here](https://github.com/keybase/kbfs/pull/65/files#diff-bacfce32314e3bbe1bc75470e97a335fR473). I'll add the comment in this as well as `libfuse`.

strib: Ah got it.  Ok.

libfuse/folderlist.go
```diff
@@ -67,21 +114,28 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp
...
-		return child, nil
-	}
-
 	// Shortcut for dreaded extraneous OSX finder lookups
 	if strings.HasPrefix(req.Name, "._") {
 		return nil, fuse.ENOENT
 	}
 
-	h, err := libkbfs.ParseTlfHandle(
+	h, er := libkbfs.ParseTlfHandle(
```
strib: Same comment here about the error name and checking the error immediately.

strib:
Looks good to me mod comments.  Thanks!

libdokan/folderlist.go
songgao: I see. I'm thinking maybe I can move this call back to where it was (right before the switch-case), and use `name` (or `req.name` in `libfuse`) here instead. But this assumes that `fl.folders` never contains non-Canonical names. Do you think that's a safe assumption?

strib: Yes, that is definitely the case.

songgao: Thanks @strib. I ended up changing `er` to more descriptive name, since `AddFavorite` takes `libkbfs.Favorite` which is derived from `h`.

PR #66, opened 2016-06-20
libkbfs: Protect staller block ops delegate by a lock in tests

taruti:
Spotted on https://travis-ci.org/keybase/kbfs/jobs/138941150

Fixes a data race in testing code.


zanderz:
LGTM but maybe better to have someone look who's familiar with staller. @strib ?

strib:
Looks good to me, thanks!

PR #73, opened 2016-06-22
don't populate LogTags field in RPC calls

songgao:


strib:
Looks good to me, thanks!

songgao:
Thanks @strib!

PR #74, opened 2016-06-22
libkbfs: support verification with unverifiable keys

aalness:
This adds support to fetch all known public keys associated with the account w/o any sigchain verification. This is used only to verify finalized folders which will have some TBD warning in the UI.


libkbfs/interfaces.go
```diff
@@ -365,7 +371,9 @@ type KBPKI interface {
 	// HasVerifyingKey returns nil if the given user has the given
 	// VerifyingKey, and an error otherwise.
 	HasVerifyingKey(ctx context.Context, uid keybase1.UID,
-		verifyingKey VerifyingKey, atServerTime time.Time) error
+		verifyingKey VerifyingKey, atServerTime time.Time,
+	        checkUnverified bool) error
```
strib: I have a slight preference for a separate interface method for checking unverified keys, to make it really hard to do it by accident.  Though the use of this bool in the code is clear (it's not just a `true`, it's a named var), so I'm ok with this if you prefer it.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -538,6 +549,36 @@ func (k *KeybaseDaemonRPC) processUserPlusKeys(upk keybase1.UserPlusKeys) (
 	return u, nil
 }
 
+// LoadUnverifiedKeys implements the KeybaseDaemon interface for KeybaseDaemonRPC.
+func (k *KeybaseDaemonRPC) LoadUnverifiedKeys(ctx context.Context, uid keybase1.UID) (
+	[]VerifyingKey, []CryptPublicKey, error) {
+	cachedUserInfo := k.getCachedUserInfo(uid)
+	keys := len(cachedUserInfo.UnverifiedVerifyingKeys) +
+		len(cachedUserInfo.UnverifiedCryptPublicKeys)
+	if keys != 0 {
```
strib: Do you think it's worth distinguishing the case where we've already checked, and the user legitimately has 0 unverified keys?  For example, if this is a shared, finalized folder, but the user you're checking wasn't the one that reset their account.  Maybe that will never happen though, so this isn't a big deal?

libkbfs/interfaces.go
strib: If you keep it here, you should explain unverified keys (and their dangerousness) in the comment above.

libkbfs/md_ops_test.go
```diff
@@ -96,7 +96,13 @@ func verifyMDForPrivate(config *ConfigMock, rmds *RootMetadataSigned) {
 		gomock.Any(), TLFCryptKey{}).Return(&rmds.MD.data, nil)
 
 	config.mockKbpki.EXPECT().HasVerifyingKey(gomock.Any(), gomock.Any(),
-		gomock.Any(), gomock.Any()).AnyTimes().Return(nil)
+		gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes().Return(nil)
+
+	if rmds.MD.IsFinal() {
+		config.mockCodec.EXPECT().Decode(gomock.Any(), gomock.Any()).Return(nil)
```
strib: These are from the MD deepCopy in VerifyRootMetadata?  Maybe you can add a comment explaining?

libkbfs/root_metadata_test.go
```diff
@@ -631,3 +631,70 @@ func TestRootMetadataFinal(t *testing.T) {
...
+	err = rmds2.VerifyRootMetadata(config.Codec(), config.Crypto())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// verify the writer metadata too for good measure
+	err = rmds2.MD.VerifyWriterMetadata(config.Codec(), config.Crypto())
+	if err != nil {
+		t.Fatal(err)
+	}
```
strib: It would also be good to check the server changing a field it's not allowed to change, and making sure the check fails.

strib:

libkbfs/keybase_daemon_rpc.go
aalness: Hmm. I guess I was thinking if we're checking to see if they owned the key that signed a metadata block then they must've had a valid key at some point -- or the mdserver is hacked or broken. So this should be 0 only in the uncached case but I may as well be explicit about it.

aalness:
@strib incorporated feedback. Mind giving it a quick skim when you have a moment?

libkbfs/md_ops.go
```diff
@@ -46,12 +46,19 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
...
-	ctx context.Context, rmds *RootMetadataSigned) error {
+	ctx context.Context, rmds *RootMetadataSigned, isTlfFinal bool) error {
 	if !rmds.MD.IsWriterMetadataCopiedSet() {
-		err := md.config.KBPKI().HasVerifyingKey(ctx,
-			rmds.MD.LastModifyingWriter,
-			rmds.MD.WriterMetadataSigInfo.VerifyingKey,
-			rmds.untrustedServerTimestamp)
+		var err error
+		if isTlfFinal {
+			err = md.config.KBPKI().HasUnverifiedVerifyingKey(ctx,
```
strib: Is there any chance the TLF could be signed with a non-unverified key?  I.e., should we be doing the `else` clause first, and do this if `err != nil && isTlfFinal`?  Or are we guaranteed the verifying key will be from before the reset?

strib: (If the latter, might be good to add a comment here explaining why that is.)

libkbfs/md_ops.go
```diff
@@ -46,12 +46,19 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
 }
 
 func (md *MDOpsStandard) verifyWriterKey(
-	ctx context.Context, rmds *RootMetadataSigned) error {
+	ctx context.Context, rmds *RootMetadataSigned, isTlfFinal bool) error {
```
strib: Do we need the extra parameter?  Can we just check for `rmds.MD.IsFinal()`?

strib:

libkbfs/md_ops.go
aalness: It's possible the key is verified. But I'm not sure if it matters. I guess it depends on how we handle the UX? We could only alert someone if the client detects it had to verify metadata with an unverified key. But if the UX is, "Warning: this folder is a folder where we can't completely verify all of the metadata, use at your own risk." then it doesn't seem worthwhile.

aalness: I should also add, unverified keys are a super set of verified keys.

strib: But when we ask for the list of unverified keys from the service, does it give us ALL the keys, even verified ones?  Or just the unverified ones?  If the latter, then this check won't pass if the key is verified.

strib: Ah ok, we raced on the comment.  Put a comment somewhere about that?  The name is misleading.

libkbfs/md_ops.go
aalness: Yeah, the metadata block may not be final (pre-reset) but the passed handle may be.

PR #80, opened 2016-06-23
Test with Keybase-built dokan 1.0.0-RC4.3 in CI

zanderz:
@taruti


taruti:
LGTM, but I think we should use versions like 1.0-RC-Keybase-N to be clear.

PR #81, opened 2016-06-24
stop devices from continuously registering and fetching merged updates

songgao:
(by letting devices track latest merged revision on server)b

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -542,6 +542,10 @@ func (fbo *folderBranchOps) setHeadLocked(
 		// Use uninitialized for the merged branch; the unmerged
 		// revision is enough to trigger conflict resolution.
 		fbo.cr.Resolve(md.Revision, MetadataRevisionUninitialized)
+	} else if md.MergedStatus() == Merged {
+		// if we are already merged through this write, the revision would be the
+		// latestMergedRevision on server
+		fbo.setLatestMergedRevisionLocked(ctx, lState, md.Revision)
 	}
 
 	fbo.head = md
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -543,8 +543,8 @@ func (fbo *folderBranchOps) setHeadLocked(
 		// revision is enough to trigger conflict resolution.
 		fbo.cr.Resolve(md.Revision, MetadataRevisionUninitialized)
 	} else if md.MergedStatus() == Merged {
-		// if we are already merged through this write, the revision would be the
-		// latestMergedRevision on server
+		// If we are already merged through this write, the revision would be the
+		// latestMergedRevision on server.
 		fbo.setLatestMergedRevisionLocked(ctx, lState, md.Revision)
 	}
 
@@ -3275,8 +3275,6 @@ func (fbo *folderBranchOps) reembedForFBM(ctx context.Context,
 
 type applyMDUpdatesFunc func(context.Context, *lockState, []*RootMetadata) error
 
-var errUnmerged = errors.New("fbo is on an unmerged local revision")
-
 func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 	lState *lockState, rmds []*RootMetadata) error {
 	fbo.mdWriterLock.AssertLocked(lState)
@@ -3298,7 +3296,7 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 			}
 			fbo.cr.Resolve(unmergedRev, rmds[len(rmds)-1].Revision)
 		}
-		return errUnmerged
+		return UnmergedError{}
 	}
 
 	// Don't allow updates while we're in the dirty state; the next
@@ -3406,7 +3404,7 @@ func (fbo *folderBranchOps) setLatestMergedRevisionLocked(ctx context.Context, l
 	fbo.headLock.AssertLocked(lState)
 
 	fbo.latestMergedRevision = rev
-	fbo.log.CDebugf(ctx, "updated latestMergedRevision to %d", rev)
+	fbo.log.CDebugf(ctx, "Updated latestMergedRevision to %d.", rev)
 }
 
 // Assumes all necessary locking is either already done by caller, or
@@ -3856,8 +3854,6 @@ func (fbo *folderBranchOps) ctxWithFBOID(ctx context.Context) context.Context {
 	return ctxWithRandomID(ctx, CtxFBOIDKey, CtxFBOOpID, fbo.log)
 }
 
-var errShutdownHappened = errors.New("Shutdown happened")
-
 // Run the passed function with a context that's canceled on shutdown.
 func (fbo *folderBranchOps) runUnlessShutdown(fn func(ctx context.Context) error) error {
 	ctx := fbo.ctxWithFBOID(context.Background())
@@ -3872,7 +3868,7 @@ func (fbo *folderBranchOps) runUnlessShutdown(fn func(ctx context.Context) error
 	case err := <-errChan:
 		return err
 	case <-fbo.shutdownChan:
-		return errShutdownHappened
+		return ShutdownHappenedError{}
 	}
 }
 
@@ -3904,7 +3900,7 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 					}
 				}
 				err = fbo.waitForAndProcessUpdates(newCtx, updateChan)
-				if err == errUnmerged {
+				if _, ok := err.(UnmergedError); ok {
 					// skip the back-off timer and continue directly to next
 					// registerForUpdates
 					return nil
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -3271,6 +3271,8 @@ func (fbo *folderBranchOps) reembedForFBM(ctx context.Context,
 
 type applyMDUpdatesFunc func(context.Context, *lockState, []*RootMetadata) error
 
+var errUnmerged = errors.New("fbo is on an unmerged local revision")
+
 func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 	lState *lockState, rmds []*RootMetadata) error {
 	fbo.mdWriterLock.AssertLocked(lState)
@@ -3292,7 +3294,7 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 			}
 			fbo.cr.Resolve(unmergedRev, rmds[len(rmds)-1].Revision)
 		}
-		return errors.New("Ignoring MD updates while local updates are staged")
+		return errUnmerged
 	}
 
 	// Don't allow updates while we're in the dirty state; the next
@@ -3898,6 +3900,11 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 					}
 				}
 				err = fbo.waitForAndProcessUpdates(newCtx, updateChan)
+				if err == errUnmerged {
+					// skip the back-off timer and continue directly to next
+					// registerForUpdates
+					return nil
+				}
 				select {
 				case <-ctx.Done():
 					// Shortcut the retry, we're done.
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -279,6 +279,9 @@ type folderBranchOps struct {
 	// rekey with a paper key prompt, if enough time has passed.
 	// Protected by mdWriterLock
 	rekeyWithPromptTimer *time.Timer
+
+	// latestMergedRevision tracks the latest heard merged revision on server
+	latestMergedRevision MetadataRevision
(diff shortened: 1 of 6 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -279,6 +279,11 @@ type folderBranchOps struct {
 	// rekey with a paper key prompt, if enough time has passed.
 	// Protected by mdWriterLock
 	rekeyWithPromptTimer *time.Timer
+
+	// latestRemoteMergedRevision tracks the latest know merged revision of on
+	// server
+	muLatestRemoteMergedRevision sync.RWMutex
```
strib: I'd rather you rely on the `headLock` mutex instead of introducing a new one.  See below for more details.

libkbfs/folder_branch_ops.go
```diff
@@ -279,6 +279,11 @@ type folderBranchOps struct {
...
+	latestRemoteMergedRevision   MetadataRevision
```
strib: No need for the word "Remote" here -- it might not even be true if the server got an update since you last heard about it.

libkbfs/folder_branch_ops.go
```diff
@@ -3253,6 +3258,18 @@ func (fbo *folderBranchOps) applyMDUpdates(ctx context.Context,
 	return fbo.applyMDUpdatesLocked(ctx, lState, rmds)
 }
 
+func (fbo *folderBranchOps) getLatestRemoteMergedRevision() MetadataRevision {
```
strib: If you go with my `headLock` suggestion, you can suffix this function name with "Locked", have it take an `lState`, and then `fbo.headLock.AssertLocked(lState)` here.

libkbfs/folder_branch_ops.go
```diff
@@ -3253,6 +3258,18 @@ func (fbo *folderBranchOps) applyMDUpdates(ctx context.Context,
...
+	fbo.muLatestRemoteMergedRevision.RLock()
+	defer fbo.muLatestRemoteMergedRevision.RUnlock()
+	return fbo.latestRemoteMergedRevision
+}
+
+func (fbo *folderBranchOps) setLatestRemoteMergedRevision(rev MetadataRevision) {
```
strib: If you go with my `headLock` suggestion, you can suffix this function name with "Locked", have it take an `lState`, and then `fbo.headLock.AssertLocked(lState)` here.

libkbfs/folder_branch_ops.go
strib: If you go with my `headLock` suggestion, you can suffix this function name with "Locked", have it take an `lState`, and then `fbo.headLock.AssertAnyLocked(lState)` here.

strib: Actually scratch that.  For this one, since you only use it in a place where the lock isn't yet taken, you can just take an `lState`, and then `fbo.headLock.RLock(lState)` here.

libkbfs/folder_branch_ops.go
```diff
@@ -3264,6 +3281,10 @@ func (fbo *folderBranchOps) getAndApplyMDUpdates(ctx context.Context,
 		return err
 	}
 
+	if len(rmds) > 0 {
+		fbo.setLatestRemoteMergedRevision(rmds[len(rmds)-1].Revision)
```
strib: You can move this into `applyMDUpdatesLocked()`, so that the proper lock is already taken.

strib: (Before the `isMasterBranch` check.)

libkbfs/folder_branch_ops.go
```diff
@@ -3253,6 +3258,18 @@ func (fbo *folderBranchOps) applyMDUpdates(ctx context.Context,
...
+func (fbo *folderBranchOps) getLatestRemoteMergedRevision() MetadataRevision {
+	fbo.muLatestRemoteMergedRevision.RLock()
+	defer fbo.muLatestRemoteMergedRevision.RUnlock()
+	return fbo.latestRemoteMergedRevision
+}
+
+func (fbo *folderBranchOps) setLatestRemoteMergedRevision(rev MetadataRevision) {
+	fbo.muLatestRemoteMergedRevision.Lock()
+	defer fbo.muLatestRemoteMergedRevision.Unlock()
+	fbo.latestRemoteMergedRevision = rev
```
strib: A debug log message here is probably a good idea, to say what the new merged head is.

libkbfs/mdserver_local.go
```diff
@@ -50,6 +50,29 @@ type MDServerLocal struct {
...
 }
 
+type registerForUpdateRecord struct {
+	id       TlfID
+	currHead MetadataRevision
+}
+
+type mdServerLocalRecordingRegisterForUpdate struct {
+	*MDServerLocal
+	ch chan registerForUpdateRecord
```
strib: This can be a `chan<-`.

libkbfs/mdserver_local.go
```diff
@@ -50,6 +50,29 @@ type MDServerLocal struct {
...
+}
+
+// NewMDServerLocalRecordingRegisterForUpdate returns a wrapper of
```
strib: "New" -> "new"

libkbfs/kbfs_cr_test.go
```diff
@@ -517,6 +517,115 @@ func TestBasicCRNoConflict(t *testing.T) {
...
+
+	chForEnablingUpdates <- struct{}{}
+
+	equal := false
+
+	// check for at most 8 times. This should be sufficiently long for client get
+	// latest merged revision and register with that
+	for i := 0; i < 8; i++ {
+		record := <-chForMdServer2
+		mergedRev, err := config2.MDServer().(mdServerLocalRecordingRegisterForUpdate).MDServerLocal.getCurrentMergedHeadRevision(ctx, rootNode2.GetFolderBranch().Tlf)
```
strib: There is a race here, since `getCurrentMergedHeadRevision` assumes the lock is held.  You probably want a new accessor that takes the lock.

strib: Also, this line is really long!  We don't have a hard limit on line sizes, but I think you can probably break it up a little.  You can just use the `mdServ` variable, right?

libkbfs/mdserver_local.go
```diff
@@ -50,6 +50,29 @@ type MDServerLocal struct {
 	shutdownLock *sync.RWMutex
 }
 
+type registerForUpdateRecord struct {
+	id       TlfID
+	currHead MetadataRevision
+}
+
+type mdServerLocalRecordingRegisterForUpdate struct {
```
strib: You should move this test code into a `_test.go` file, either in `kbfs_cr_test.go` or a new one you make just for this new struct.

libkbfs/kbfs_cr_test.go
```diff
@@ -517,6 +517,115 @@ func TestBasicCRNoConflict(t *testing.T) {
...
+	var userName1, userName2 libkb.NormalizedUsername = "u1", "u2"
+	config1, _, ctx := kbfsOpsConcurInit(t, userName1, userName2)
+	defer CheckConfigAndShutdown(t, config1)
+
+	config2 := ConfigAsUser(config1.(*ConfigLocal), userName2)
+	mdServ, chForMdServer2 := newMDServerLocalRecordingRegisterForUpdate(config2.MDServer().(*MDServerLocal))
+	config2.SetMDServer(mdServ)
+	defer CheckConfigAndShutdown(t, config2)
+
+	clock, _ := newTestClockAndTimeNow()
```
strib: I don't think you need a fake clock for this test.

libkbfs/mdserver_local.go
```diff
@@ -50,6 +50,29 @@ type MDServerLocal struct {
...
+// NewMDServerLocalRecordingRegisterForUpdate returns a wrapper of
+// MDServerLocal that records RegisterforUpdate calls.
+func newMDServerLocalRecordingRegisterForUpdate(mdServerRaw *MDServerLocal) (mdServer MDServer, records <-chan registerForUpdateRecord) {
+	ret := mdServerLocalRecordingRegisterForUpdate{mdServerRaw, make(chan registerForUpdateRecord, 8)}
+	return ret, ret.ch
+}
+
+func (md mdServerLocalRecordingRegisterForUpdate) RegisterForUpdate(ctx context.Context, id TlfID,
+	currHead MetadataRevision) (<-chan error, error) {
+	defer func() { md.ch <- registerForUpdateRecord{id: id, currHead: currHead} }()
```
strib: Why does this need to be deferred?  It's a buffered channel, right?

strib:
Looks like the test failed with a timeout on OS X.  On my Linux desktop, the test takes about 13% of the entire test time of `libkbfs`, so I think there's something very slow happening.  Is the test still relying on the exp. backoff somehow?

I'm not sure if the OS X timeout is your fault, it could just be slow, but it seems suspicious that it times out during this new test.  In any case, this test should execute much faster so let's try to figure out where the time is going.

strib:
Oh, one more thing:  you should also set the latest known merged head [here](https://github.com/keybase/kbfs/blob/e42c5c9ea3d21a42e706b5d7431e4d26a9473d5f/libkbfs/folder_branch_ops.go#L574), when `md.MergeStatus() == Merged`.

(Finished my first pass.  Thanks for working on this!)

songgao:
Yes I think it's still using the back-off timer. Should I add ability to replace the timer in this (or a separate) PR?

strib:
> Yes I think it's still using the back-off timer. Should I add ability to replace the timer in this (or a separate) PR?

I was wondering why it was even hitting the backoff time at all, but I think I see now.  We expect to hit it one time, after we fetch the merged updates.

I don't think you need to replace the backoff timer in this case.  I would make a new error type that gets returned from `applyMDUpdatesLocked` when we are unmerged, and then when `registerAndWaitForUpdates` sees that error it can return `nil` instead, skipping directly to the next `registerForUpdates` call.  This is a good fix to have for production too, I think.

songgao:
Got you @strib. Thanks for the detailed review!

> We expect to hit it one time, after we fetch the merged updates.

I think I was seeing it hit more than once. Could be *before* fetching the merged updates though. Is this normal / should I look into it?

strib:
> I think I was seeing it hit more than once. Could be before fetching the merged updates though. Is this normal / should I look into it?

That seems possible, since the merged side has more than one update.  So it's probably normal/expected.

songgao:
Looks like the new commits have addressed all comments. Please take another look when you have time. Thanks @strib!

libkbfs/folder_branch_ops.go
```diff
@@ -3142,13 +3149,19 @@ func (fbo *folderBranchOps) reembedForFBM(ctx context.Context,
 
 type applyMDUpdatesFunc func(context.Context, *lockState, []*RootMetadata) error
 
+var errUnmerged = errors.New("fbo is on an unmerged local revision")
```
strib: We have a bunch of error types in `errors.go`.  You could add a new type for this, and that way the check would be a simple type check, rather than a string check.  It probably doesn't matter much, but it would be a little more consistent with the rest of our code.

libkbfs/folder_branch_ops.go
```diff
@@ -3253,6 +3266,20 @@ func (fbo *folderBranchOps) applyMDUpdates(ctx context.Context,
...
+	defer fbo.headLock.RUnlock(lState)
+	return fbo.latestMergedRevision
+}
+
+// caller should have held fbo.headLock
+func (fbo *folderBranchOps) setLatestMergedRevisionLocked(ctx context.Context, lState *lockState, rev MetadataRevision) {
+	fbo.headLock.AssertLocked(lState)
+
+	fbo.latestMergedRevision = rev
+	fbo.log.CDebugf(ctx, "updated latestMergedRevision to %d", rev)
```
strib: super nit: I think most comments start with a capital letter, i.e. "updated" -> "Updated".

libkbfs/folder_branch_ops.go
songgao: Sounds good. There's an [`errShutdownHappened`](https://github.com/keybase/kbfs/blob/d8cd1200bde1ee33a2b9909bdf6925501dfac560/libkbfs/folder_branch_ops.go#L3730) as well. Should I change that as well?

libkbfs/folder_branch_ops.go
```diff
@@ -569,6 +572,10 @@ func (fbo *folderBranchOps) setHeadLocked(ctx context.Context,
 		// Use uninitialized for the merged branch; the unmerged
 		// revision is enough to trigger conflict resolution.
 		fbo.cr.Resolve(md.Revision, MetadataRevisionUninitialized)
+	} else if md.MergedStatus() == Merged {
+		// if we are already merged through this write, the revision would be the
```
strib: Another super nit: comments that are complete sentences usually start in a capital letter and end in punctuation.

libkbfs/folder_branch_ops.go
strib: Ah, sure, thanks!

strib:
Looks good to me, mod comments.  Thanks!

PR #90, opened 2016-06-26
Taruti/kbfs 1170

taruti:
Even though we don't support cross-tlf renames we can force explorer to do the right thing.


zanderz:
this fixes it? Great!
LGTM :thumbsup:

taruti:
Yes, after this it works on Windows and still fails on *nix.

PR #95, opened 2016-06-28
Taruti/kbfs 1157

taruti:


zanderz:
LGTM but another KBFS person should look @strib

taruti:
I'll wait for @strib, especially the libkbfs/kbfs_ops.go could use some eyes  :)

libdokan/tlf.go
```diff
@@ -165,11 +165,16 @@ func (tlf *TLF) open(ctx context.Context, oc *openContext, path []string) (dokan
...
 		return tlf, true, nil
 	}
-	dir, exitEarly, err := tlf.loadDirAllowNonexistent(ctx, "open")
+	// If it is a creation then we need the dir for real.
+	dir, exitEarly, err := tlf.loadDirHelper(ctx, "open", !oc.isCreation())
 	if err != nil {
 		return nil, false, err
 	}
 	if exitEarly {
+		if node := openSpecialFile(lastStr(path), tlf.folder); node != nil {
```
strib: Not sure it does much good to allow the user to open a special file like `.kbfs_status` here, if there's no TLF to populate the data. What's the reasoning behind having this check?

libfs/tlf.go
```diff
@@ -10,6 +10,13 @@ import (
 	"golang.org/x/net/context"
 )
 
+// TlfDoesNotExist is a shortcut error for the cases a TLF does not exist and
+// an early successfull exit via FilterTLFEarlyExitError is wished.
```
strib: "successfull" -> "successful"

libfuse/mount_test.go
```diff
@@ -2715,6 +2715,10 @@ func TestUnstageFile(t *testing.T) {
 	}
 }
 
+type TP struct {
```
strib: What's this?  It seems unused.

libkbfs/kbfs_ops.go
```diff
@@ -247,6 +246,12 @@ func (fs *KBFSOpsStandard) GetOrCreateRootNode(
 			return nil, EntryInfo{}, err
 		}
 	}
+
+	if !create && md.data.Dir.Type != Dir {
+		kbpki := fs.config.KBPKI()
+		return nil, EntryInfo{}, identifyHandle(ctx, kbpki, kbpki, h)
```
strib: I don't think we need an `identifyHandle` call here.  The user has not yet needed to trust any data in the folder. Was there a specific reason for the identify?

libfuse/dir.go
```diff
@@ -134,6 +134,26 @@ func (f *Folder) forgetNode(node libkbfs.Node) {
 
 var _ libkbfs.Observer = (*Folder)(nil)
 
+func (f *Folder) resolve(ctx context.Context) (*libkbfs.TlfHandle, error) {
```
strib: You could factor this out into `libfs` and have it take a `libkbfs.Observer` in order to call `TlfHandleChange`.

strib:
Looks good besides my comments.  However, the test failure looks legitimate: https://travis-ci.org/keybase/kbfs/jobs/141155437#L270.  I can't figure out what the bug is, but it seems like the `resolve()` call failed to invoke `TlfHandleChange` the same way it did before.  The code seems pretty identical though, so I'm not sure exactly what's wrong.

strib:
Or maybe the first lookup (resulting in the expected error) in that test used to create the folder handle mapping in the mdserver, and now it no longer does, and that changes what happens when the assertion is added?  Hmm.

strib:
@taruti are you planning to get back to this?  We have more complaints about the current behavior: https://github.com/keybase/keybase-issues/issues/2353

libfs/tlf.go
taruti: Done.

libfuse/mount_test.go
taruti: Was a residue of a debugging thing to get test printouts even in the event of a crash, removed.

libfuse/dir.go
taruti: I want to refactor the common bits on a higher level but want to avoid that as part of this pull request. I.e. probably have a libfs.FolderCommon and resolve etc being methods on that.

libdokan/tlf.go
taruti: Some tests require it, e.g. TestErrorFile, and syncing with .kbfs_sync_from_server in many cases.

libkbfs/kbfs_ops.go
taruti: That is required for TestReaddirPublicFailedIdentifyViaOSCall. Is there a better way to do it?

taruti:
I think this hits the pathname resolution cache on OS X :(

libdokan/tlf.go
strib: Ah ok thanks.  The error file thing makes sense.

libkbfs/kbfs_ops.go
strib: Bleh, I forgot about that OS X `readdir` bug.  Ugly.

Ok, well we should add a comment here about the fact that we need to do this to make sure identify errors are returned on open on OS X, since `readdir` doesn't return errors.

Also, `identifyHandle` is fairly expensive and only needs to be run once every 6 hours or so for a given folder.  It's probably worth caching successful identifies in `KBFSOps` (normally they are cached in `folderBranchOps`).

strib:
> I think this hits the pathname resolution cache on OS X :(

Can you describe the issue a little more?  I don't follow what the problem is, but I am slow on Mondays.

strib:
@taruti: Jenkins is now running osxfuse 3.4.2, which has the rm+lookup locking fix.

taruti:
After rebasing:
+ Old failure from `TestSBSNewlyResolvedWritersPublic` is fixed by OSXFuse 3.4.2
+ Half-a-dozen new failures looking into what causes these

taruti:
@strib Now it passes CI - just rebasing the old code with new OSXFuse :)

libfuse/dir.go
```diff
@@ -260,6 +280,7 @@ func (f *Folder) batchChangesInvalidate(ctx context.Context,
 // TlfHandleChange is called when the name of a folder changes.
 func (f *Folder) TlfHandleChange(ctx context.Context,
 	newHandle *libkbfs.TlfHandle) {
+	f.fs.log.CDebugf(ctx, "TlfHandleChange called")
```
strib: Maybe include `newHandle.GetCanonicalName()` in the log message to make it more useful?

libfuse/mount_test.go
```diff
@@ -2759,11 +2759,19 @@ func TestSimpleCRNoConflict(t *testing.T) {
 		t.Skip("Old FUSE protocol")
 	}
 
-	// both users read the root dir first
 	root1 := path.Join(mnt1.Dir, PrivateName, "user1,user2")
 	root2 := path.Join(mnt2.Dir, PrivateName, "user1,user2")
-	checkDir(t, root1, map[string]fileInfoCheck{})
-	checkDir(t, root2, map[string]fileInfoCheck{})
+	// both users should mutate the dir first
```
strib: Why is it required for both users to mutate?  Shouldn't it be enough for one user to mutate, and the other user to read it?  (I don't mind this way, I'm just making sure I understand what's going on.)

libkbfs/folder_branch_ops.go
```diff
@@ -1055,8 +1055,14 @@ func (fbo *folderBranchOps) GetTLFCryptKeys(ctx context.Context,
 func (fbo *folderBranchOps) GetOrCreateRootNode(
 	ctx context.Context, h *TlfHandle, branch BranchName) (
 	node Node, ei EntryInfo, err error) {
-	err = errors.New("GetOrCreateRootNode is not supported by " +
-		"folderBranchOps")
+	err = errors.New("GetOrCreateRootNode is not supported by folderBranchOps")
```
strib: Might as well fix these bare returns by making them `return errors.New(...` directly.

strib:

libfuse/dir.go
taruti: Done. (although for debugging whether TlfHandleChange is called at all was sufficient).

libfuse/mount_test.go
taruti: Sure that would also work. Modifying the comment.

libkbfs/folder_branch_ops.go
taruti: Sure, they used to be bare and tried to keep the same style - making them direct.

PR #103, opened 2016-06-30
favorites: fix a possible hang, and handle deletions better

strib:
I hit a KBFS hang when Emacs canceled a context while KBFS was trying to send a favorites request onto the channel, leaving it in the `inFlightAdds` map forever and blocking and future requests for that same folder.  This PR cleans up the map in that case.

Also, a user noticed that they couldn't delete certain favorites.  It turns out that can happen when a `folderBranchOps` is made, but `head` is never set (e.g., due to identify or MD fetch errors)

@songgao: want to try reviewing?

Issue: KBFS-1235
Issue: keybase/client#3263

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -160,7 +160,12 @@ func (fs *KBFSOpsStandard) DeleteFavorite(ctx context.Context,
 		return fs.opsByFav[fav]
 	}()
 	if ops != nil {
-		return ops.deleteFromFavorites(ctx, fs.favs)
+		err := ops.deleteFromFavorites(ctx, fs.favs)
+		if _, ok := err.(OpsCantHandleFavorite); !ok {
+			return err
+		}
+		// If the ops couldn't handle the delete, fall through to
+		// going directly via Favorites.
 	}
 
 	if isLoggedIn {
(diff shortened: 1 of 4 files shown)


libkbfs/kbfs_ops.go
```diff
@@ -160,7 +160,12 @@ func (fs *KBFSOpsStandard) DeleteFavorite(ctx context.Context,
 		return fs.opsByFav[fav]
 	}()
 	if ops != nil {
-		return ops.deleteFromFavorites(ctx, fs.favs)
+		err := ops.deleteFromFavorites(ctx, fs.favs)
+		if _, ok := err.(OpsCantHandleFavorite); !ok {
```
songgao: Could the `deleteFromFavorites` happen when `head == nil` because it's not initialized yet somehow but not because TLF doesn't exist? I was concerned if this would make it possible in any way to have favorites seemingly deleted on a device while it's not actually happened on server.

strib: Good question.  But actually either way the request will go to the server.  The only reason we prefer to go through `folderBranchOps` is because there might be some local state to clean up in that object when it's removed from favorites.  (Right now there isn't any, but was trying to plan ahead.)  I believe the different ways to delete favorites are equivalent.

songgao:
LGTM!

libkbfs/kbfs_ops.go
songgao: Thanks for the explanation!

PR #105, opened 2016-06-30
add isExcl param to CreateFile

songgao:
[KBFS-1217](https://keybase.atlassian.net/browse/KBFS-1217)

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -640,6 +640,28 @@ func (et EntryType) String() string {
 	return "<invalid EntryType>"
 }
 
+// EXCL indicates whether O_EXCL is set on a fuse call
+type EXCL bool
+
+const (
+	// NoEXCL indicates O_EXCL is not set
+	NoEXCL EXCL = false
+
+	// WithEXCL indicates O_EXCL is set
+	WithEXCL EXCL = true
+)
+
+func (o EXCL) String() string {
+	switch o {
+	case NoEXCL:
+		return "O_EXCL unset"
+	case WithEXCL:
+		return "O_EXCL set"
+	default:
+		return "<invalid EXCL>"
+	}
+}
+
 // EntryInfo is the (non-block-related) info a directory knows about
 // its child.
 //
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1790,7 +1790,7 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
 }
 
 func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
-	lState *lockState, md *RootMetadata, bps *blockPutState) (err error) {
+	lState *lockState, md *RootMetadata, bps *blockPutState, excl EXCL) (err error) {
 	fbo.mdWriterLock.AssertLocked(lState)
 
 	// finally, write out the new metadata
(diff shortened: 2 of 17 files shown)


libkbfs/data_types.go
```diff
@@ -640,6 +641,28 @@ func (et EntryType) String() string {
...
 
+// EXCL indicates whether O_EXCL is set on a fuse call
+type EXCL uint32
+
+const (
+	// NoEXCL indicates O_EXCL is not set
+	NoEXCL EXCL = 0
+
+	// WithEXCL indicates O_EXCL is set
+	WithEXCL EXCL = syscall.O_EXCL
```
strib: This is a neat trick, but I wonder if it will make the eventual Windows implementation trickier.  Maybe just use `iota` for these and add a simple function to make an `EXCL` from a syscall uint32?

libkbfs/interfaces.go
```diff
@@ -141,7 +141,7 @@ type KBFSOps interface {
 	// logged-in user has write permission to the top-level folder.
 	// Returns the new Node for the created file, and its new
 	// entry info.  This is a remote-sync operation.
-	CreateFile(ctx context.Context, dir Node, name string, isEx bool) (
+	CreateFile(ctx context.Context, dir Node, name string, isExec bool, excl EXCL) (
```
strib: Please add a comment explaining the `excl` parameter.

libkbfs/folder_branch_ops.go
```diff
@@ -1782,34 +1782,38 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
...
+			if excl == WithEXCL {
+				// If this was caused by an exclusive create, we shouldn't do an
+				// UnmergedPut, but rather wait until we are on master branch.
+				return err
+			}
+		} else if err != nil {
+			return err
 		}
+	} else if excl == WithEXCL {
+		return UnmergedError{}
```
strib: A more specific error might be nice here.  Either a new `UnmergedEXCLError`, or you can just use `errors.New()` since this is a pretty unexpected case and I doubt we'll ever need to check for the error type.

libkbfs/folder_branch_ops.go
```diff
@@ -1782,34 +1782,38 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
...
-		// the same (at least) as what we thought our new revision
-		// should be.  Otherwise, just leave it at uninitialized and
-		// let the resolver sort it out.
-		if doUnmergedPut {
+
+		if doUnmergedPut = fbo.isRevisionConflict(err); doUnmergedPut {
 			fbo.log.CDebugf(ctx, "Conflict: %v", err)
 			mergedRev = md.Revision
+
+			if excl == WithEXCL {
```
strib: I'm not 100% sure we should do this yet.  I'm worried it will start causing errors for programs that use `O_EXCL` that wouldn't happen prior to this PR.  Arguably it is better for them to fail since they requested exclusivity, but previous to this we didn't fail them so it might be weird if we start failing now.  What do you think?

strib:
First pass done.  Code looks good, but we should think about whether we should merge this as is, or if we should still allow EXCL creates to cause conflicts for now.

libkbfs/folder_branch_ops.go
songgao: Hmm.. my understanding is that, it's possible that local revision is not sync'ed with server yet, so it's possible that previous check (will be in #109) passes, but the `Put()` results in a conflict. In this case, if `O_EXCL` is set, we should not continue with unmerged Put right?

Are we gonna have any other operation that supports `O_EXCL` or is this a pretty specialized scenario? If latter, maybe we could probably make `EXCL` carry information regarding what ops call this is and only honor it if it's from `CreateFile`?

libkbfs/data_types.go
songgao: Got you. I'm in for making a simple function to make EXCL from fuse/dokan calls. But since it's still the O_EXCL semantics, do you think we can keep the value to align with `O_EXCL` on Linux?

libkbfs/folder_branch_ops.go
strib: Sorry, I'm saying that this is code is correct, but that it should appear in #109, not in this PR.  If we check it in without the rest of #109, it will start causing failures in a brand new way that didn't exist before.

songgao: Ah I see. Makes sense!

libkbfs/data_types.go
strib: It doesn't matter to me what the value is, I just don't see what advantage that gives over the standard `iota` way for constants.  It's not like we ever need to pass this value into a file system call.

PR #106, opened 2016-07-01
fix slow test introduced by KBFS-1206

songgao:


strib:
Looks good to me, thanks!

PR #109, opened 2016-07-01
[KBFS-1218 & KBFS-1219] wait for conflicts to resolve if O_EXCL is set

songgao:
(includes changes from #105 )


songgao:
Moving to #110

PR #110, opened 2016-07-01
O_EXCL Create

songgao:
Relevant JIRA Epic: [KBFS-1216](https://keybase.atlassian.net/browse/KBFS-1216)

### DIFF
--- a/libdokan/dir.go
+++ b/libdokan/dir.go
@@ -383,17 +383,16 @@ func openSymlink(ctx context.Context, oc *openContext, parent *Dir, rootDir *Dir
 	return rootDir.open(ctx, oc, dst)
 }
 
-func getEXCLFromOpenContext(oc *openContext) libkbfs.EXCL {
-	return libkbfs.EXCL(oc.CreateDisposition == dokan.FileCreate)
+func getExclFromOpenContext(oc *openContext) libkbfs.Excl {
+	return libkbfs.Excl(oc.CreateDisposition == dokan.FileCreate)
 }
 
 func (d *Dir) create(ctx context.Context, oc *openContext, name string) (f dokan.File, isDir bool, err error) {
-	d.folder.fs.log.CDebugf(ctx, "Dir Create %s with openContext: %#v, %#v",
-		name, oc, oc.CreateData)
+	d.folder.fs.log.CDebugf(ctx, "Dir Create %s", name)
 	defer func() { d.folder.reportErr(ctx, libkbfs.WriteMode, err, nil) }()
 
 	isExec := false // Windows lacks executable modes.
-	excl := getEXCLFromOpenContext(oc)
+	excl := getExclFromOpenContext(oc)
 	newNode, _, err := d.folder.fs.config.KBFSOps().CreateFile(
 		ctx, d.node, name, isExec, excl)
 	if err != nil {
--- a/libdokan/dir.go
+++ b/libdokan/dir.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-	"syscall"
 
 	"github.com/keybase/kbfs/dokan"
 	"github.com/keybase/kbfs/libfs"
@@ -385,11 +384,12 @@ func openSymlink(ctx context.Context, oc *openContext, parent *Dir, rootDir *Dir
 }
 
 func getEXCLFromOpenContext(oc *openContext) libkbfs.EXCL {
-	return libkbfs.EXCL(oc.CreateDisposition == syscall.CREATE_NEW)
+	return libkbfs.EXCL(oc.CreateDisposition == dokan.FileCreate)
 }
 
 func (d *Dir) create(ctx context.Context, oc *openContext, name string) (f dokan.File, isDir bool, err error) {
-	d.folder.fs.log.CDebugf(ctx, "Dir Create %s", name)
+	d.folder.fs.log.CDebugf(ctx, "Dir Create %s with openContext: %#v, %#v",
+		name, oc, oc.CreateData)
 	defer func() { d.folder.reportErr(ctx, libkbfs.WriteMode, err, nil) }()
 
 	isExec := false // Windows lacks executable modes.
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1865,6 +1865,8 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
 
 func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 	lState *lockState, md *RootMetadata, bps *blockPutState, excl EXCL) (err error) {
+	fbo.log.CDebugf(ctx, "finalizeMDWriteLocked: excl=%s", excl)
+	defer func() { fbo.deferLog.CDebugf(ctx, "Done: %v", err) }()
 	fbo.mdWriterLock.AssertLocked(lState)
 
 	// finally, write out the new metadata
@@ -2286,7 +2288,8 @@ func (fbo *folderBranchOps) CreateDir(
 func (fbo *folderBranchOps) CreateFile(
 	ctx context.Context, dir Node, path string, isExec bool, excl EXCL) (
 	n Node, ei EntryInfo, err error) {
-	fbo.log.CDebugf(ctx, "CreateFile %p %s", dir.GetID(), path)
+	fbo.log.CDebugf(ctx, "CreateFile %p %s isExec=%v EXCL=%s",
+		dir.GetID(), path, isExec, excl)
 	defer func() {
 		if err != nil {
 			fbo.deferLog.CDebugf(ctx, "Error: %v", err)
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1723,7 +1723,8 @@ func isRecoverableBlockErrorForRemoval(err error) bool {
 }
 
 func isRetriableError(err error, retries int) bool {
-	recoverable := isRecoverableBlockError(err)
+	_, isEXCLOnUnmergedError := err.(EXCLOnUnmergedError)
+	recoverable := isRecoverableBlockError(err) || isEXCLOnUnmergedError
 	return recoverable && retries < maxRetriesOnRecoverableErrors
 }
 
@@ -1882,14 +1883,19 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 
 			if excl == WithEXCL {
 				// If this was caused by an exclusive create, we shouldn't do an
-				// UnmergedPut, but rather wait until we are on master branch.
-				return err
+				// UnmergedPut, but rather try to get newest update from server, and
+				// retry afterwards.
+				err = fbo.getAndApplyMDUpdates(ctx, lState, fbo.applyMDUpdatesLocked)
+				if err != nil {
+					return err
+				}
+				return EXCLOnUnmergedError{}
 			}
 		} else if err != nil {
 			return err
 		}
 	} else if excl == WithEXCL {
-		return UnmergedError{}
+		return EXCLOnUnmergedError{}
 	}
 
 	if doUnmergedPut {
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1880,13 +1880,21 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 			fbo.log.CDebugf(ctx, "Conflict: %v", err)
 			mergedRev = md.Revision
 
+			if excl == WithEXCL {
+				// If this was caused by an exclusive create, we shouldn't do an
+				// UnmergedPut, but rather wait until we are on master branch.
+				return err
+			}
 		} else if err != nil {
 			return err
 		}
+	} else if excl == WithEXCL {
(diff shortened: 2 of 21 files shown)


songgao:
I think this is ready for a review; Thanks!

OS X test is failing but we'll need to wait until osxfuse supports passing in O_EXCL flag.

libdokan/dir.go
```diff
@@ -385,11 +384,12 @@ func openSymlink(ctx context.Context, oc *openContext, parent *Dir, rootDir *Dir
...
 
 func getEXCLFromOpenContext(oc *openContext) libkbfs.EXCL {
-	return libkbfs.EXCL(oc.CreateDisposition == syscall.CREATE_NEW)
+	return libkbfs.EXCL(oc.CreateDisposition == dokan.FileCreate)
 }
 
 func (d *Dir) create(ctx context.Context, oc *openContext, name string) (f dokan.File, isDir bool, err error) {
-	d.folder.fs.log.CDebugf(ctx, "Dir Create %s", name)
+	d.folder.fs.log.CDebugf(ctx, "Dir Create %s with openContext: %#v, %#v",
+		name, oc, oc.CreateData)
```
strib: We need to be careful about logging data since it could get shipped back to our servers.  Should we really be logging `CreateData` here?

libkbfs/folder_branch_ops.go
```diff
@@ -1683,7 +1683,8 @@ func isRecoverableBlockErrorForRemoval(err error) bool {
 }
 
 func isRetriableError(err error, retries int) bool {
-	recoverable := isRecoverableBlockError(err)
+	_, isEXCLOnUnmergedError := err.(EXCLOnUnmergedError)
+	recoverable := isRecoverableBlockError(err) || isEXCLOnUnmergedError
```
strib: super minor optimization: put `isEXCLOnUnmergedError` first to avoid a function call if it's already true.

libkbfs/folder_branch_ops.go
```diff
@@ -1824,6 +1825,8 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
 
 func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 	lState *lockState, md *RootMetadata, bps *blockPutState, excl EXCL) (err error) {
+	fbo.log.CDebugf(ctx, "finalizeMDWriteLocked: excl=%s", excl)
```
strib: Please just log this in `CreateFile`, as part of the existing log line, so we don't spam the logs too badly.

libkbfs/folder_branch_ops.go
```diff
@@ -1824,6 +1825,8 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
...
+	defer func() { fbo.deferLog.CDebugf(ctx, "Done: %v", err) }()
```
strib: Don't think we need this -- `doMDWriteWithRetry` already logs errors.

songgao: Sorry for the excessive logs! I added those to help with debugging; will clean them up

libkbfs/folder_branch_ops.go
```diff
@@ -1840,13 +1843,26 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 			fbo.log.CDebugf(ctx, "Conflict: %v", err)
 			mergedRev = md.Revision
 
+			if excl == WithEXCL {
```
strib: I know it's a pain, but I think we should consider moving all the `EXCL` constants to say `Excl` instead.  To me it really stands out, since it's not an acronym and none of our other consts use the all caps convention (which is usually frowned-upon in golang).

I know we're borrowing it from POSIX, but since this isn't exposed anywhere that other POSIX flags are exposed, I think it makes the most sense to change everything to `Excl` instead.  What do you think?

songgao: Sure; I think that makes sense! `gorename` makes it super easy to do it so no worries :D

libkbfs/folder_branch_ops.go
```diff
@@ -2231,6 +2248,12 @@ func (fbo *folderBranchOps) CreateFile(
 		entryType = File
 	}
 
+	if excl == WithEXCL {
+		if err = fbo.cr.Wait(ctx); err != nil {
```
strib: Do you think we should also retry, after ` cr.Wait()` call, if we get an `EXCLOnUnmergedError`?  Since we don't hold any locks here, it's very possible for another operation to slip in after this check and make us unmerged before taking the lock as part of `doMDWriteWithRetryUnlessCanceled`.  Not sure if it's worth doing or not.

test/cr_simple_test.go
```diff
@@ -976,6 +976,82 @@ func TestCrUnmergedBothRmfile(t *testing.T) {
...
+			reenableUpdates(),
+			lsdir("a/", m{"b$": "FILE"}),
+		),
+		as(alice,
+			lsdir("a/", m{"b$": "FILE"}),
+		),
+	)
+}
+
+// alice and bob both exclusively create the same file, but neither write to it.
```
strib: Comment copied from above I think, could be more descriptive.

test/cr_simple_test.go
```diff
@@ -976,6 +976,82 @@ func TestCrUnmergedBothRmfile(t *testing.T) {
...
+	test(t,
+		users("alice", "bob"),
+		as(alice,
+			mkdir("a"),
+		),
+		as(bob,
+			lsdir("a/", m{}),
+		),
+		as(bob, stallOnMDPut()),
+		parallel(
```
strib: Whoa cool!

test/dsl_test.go
```diff
@@ -55,21 +58,51 @@ func test(t testing.TB, actions ...optionOp) {
 	}
 }
 
+func parallel(actions ...optionOp) optionOp {
+	return func(o *opt) {
+		wg := new(sync.WaitGroup)
```
strib: We usually do `var wg sync.WaitGroup`.

test/dsl_test.go
```diff
@@ -232,7 +265,7 @@ func (o *opt) failf(format string, objs ...interface{}) {
 
 func (o *opt) expectSuccess(reason string, err error) {
 	if err != nil {
-		o.t.Fatalf("Error: %s: %v", reason, err)
+		o.t.Errorf("Error: %s: %v", reason, err)
```
strib: Was this just for testing?  Seems like it should be `Fatalf`.

test/dsl_test.go
```diff
@@ -364,6 +398,19 @@ func notExists(filename string) fileOp {
 	}, Defaults}
 }
 
+func mkfileexcl(name string) fileOp {
+	return fileOp{func(c *ctx) error {
+		dir, file := filepath.Split(name)
+		d, _, err := c.getNode(dir, noCreate, resolveAllSyms)
```
strib: It might be better to make a `createFileExcl createType` flag to use here, so you can still let the `getNode` function do the filepath stuff.  Any reason not to do it that way?

test/dsl_test.go
```diff
@@ -364,6 +398,19 @@ func notExists(filename string) fileOp {
...
+		if err != nil {
+			return err
+		}
+		err = c.engine.CreateFileEXCL(c.user, d, file)
+		c.t.Logf("mkfileexcl-CreateFileEXCL: %v", err)
```
strib: Remove.

strib:
Finished a pass.  Looks basically ready, just a few nits and suggestions.

test/dsl_test.go
songgao: This was because calling `Fatalf` in test in a non-main routine (is an issue)[https://golang.org/pkg/testing/#T]. For me it just hang after the error. `Errorf` doesn't exit right away, but still fails the test and is usable in a different goroutine.

test/dsl_test.go
songgao: Hmm... with a value type, doesn't it allow accidentally copying the `wg` object? All the methods defined on `WaitGroup` are pointer types and it looks like [there'd be issues copying it](https://golang.org/src/sync/waitgroup.go?s=522:830#L8).

strib: You don't copy it, though, you're using a closure.  But you're right, I guess in theory someone could change the code to copy it.  If you want to use a pointer, `wg := sync.WaitGroup{}` is more standard in our code base.

test/dsl_test.go
strib: Ah, because of the new `parallel` thing.  Maybe you can add an indicator on an `opt` that this is being executed in a different goroutine, and only use `Errorf` in that case?  I'd rather not change the existing behavior too much.  It makes the test output harder to decipher when the test continues past a fatal error.

test/dsl_test.go
songgao: It seemed to me that `getNode` was not supposed to error if file existed. But I guess we could treat the `createFileExcl` differently. Will do!

test/dsl_test.go
songgao: Got you; I'll change it (assuming you meant `wg := &sync.WaitGroup{}`)

strib: Oops, yeah that's what I meant, sorry.

libkbfs/folder_branch_ops.go
songgao: Good point! I think it's worth doing; if we support `O_EXCL`, it should be a guarantee.

test/dsl_test.go
songgao: No worries!

test/dsl_test.go
songgao: Sounds good!

songgao:
@strib I think all comments are addressed. Please take another look when you have time? Thanks!

libkbfs/folder_branch_ops.go
```diff
@@ -2248,14 +2246,13 @@ func (fbo *folderBranchOps) CreateFile(
...
-	if excl == WithEXCL {
-		if err = fbo.cr.Wait(ctx); err != nil {
-			return nil, EntryInfo{}, err
-		}
-	}
-
 	err = fbo.doMDWriteWithRetryUnlessCanceled(ctx,
 		func(lState *lockState) error {
+			if excl == WithExcl {
+				if err = fbo.cr.Wait(ctx); err != nil {
```
strib: Putting it here is problematic because you will be holding the lock over the Wait call.  If the conflict resolver needs the lock to finalize the resolution, it will deadlock.

It might be better to do it right after this call: https://github.com/keybase/kbfs/blob/master/libkbfs/folder_branch_ops.go#L2225

Obviously would be nice to have a test that covers it as well.  I guess you could use `parallel` and a staller to cook something up, but it might be messy and not worth doing?

test/dsl_test.go
```diff
@@ -255,17 +256,13 @@ func noSync() fileOp {
...
-}
-
-func (o *opt) failf(format string, objs ...interface{}) {
-	o.t.Fatalf(format, objs...)
-}
-
 func (o *opt) expectSuccess(reason string, err error) {
 	if err != nil {
-		o.t.Errorf("Error: %s: %v", reason, err)
+		if o.isParallel {
```
strib: Maybe a comment here explaining why we're dooing this.

test/dsl_test.go
```diff
@@ -695,19 +689,42 @@ func (c *ctx) getNode(filepath string, create createType, sym symBehavior) (
...
 		node, symPath, err = c.engine.Lookup(c.user, parent, name)
 		c.t.Log("getNode:", i, name, node, symPath, err)
-		if err != nil && create != noCreate {
-			if create == createFile && i+1 == len(components) {
+
+		if i+1 == len(components) { // last element in path
+			switch {
+			case err == nil:
+				if create == createFileExcl {
+					err = libkbfs.NameExistsError{}
```
strib: Why not just `return` directly here?

strib:

strib:
Oh wait, one last comment -- if the tests fail on OS X, shouldn't we be skipping them for now?  You can skip with something in the test like:

```go
var skipOp fileOp
if runtime.GOOS == "darwin" {
        skipOp = skip("fuse", "osxfuse doesn't pass through O_EXCL yet"),
} else {
        skipOp = custom(func (f fileOp) error { return nil })
}
test(
        skipOp,
        ...
}
```

strib:
(Oh and please fix the compile errors that CI caught too!)

libkbfs/folder_branch_ops.go
```diff
@@ -2135,7 +2149,7 @@ func (fbo *folderBranchOps) createEntryLocked(
 }
 
 func (fbo *folderBranchOps) doMDWriteWithRetry(ctx context.Context,
-	lState *lockState, fn func(lState *lockState) error) error {
+	lState *lockState, waitForCR bool, fn func(lState *lockState) error) error {
```
songgao: @strib Do you think this is OK? Although not many functions use `doMDWriteWithRetry` directly so it's probably fine. But I started to wonder if makes sense to put Excl flag into ctx?

strib: I was just thinking you could do the wait only if you got an `ExclOnUnmergedError`, without needing to pass any additional flags.  Any reason that wouldn't work?

songgao: Ah that makes sense! No that would work; I'll update the PR

songgao:
CI passed :joy: I'll cleanup and rebase the branch

PR #116, opened 2016-07-05
libdokan: Wibble remount wait times

taruti:


taruti:
This will probably make triggering 3304 less likely (cannot reproduce it..).

zanderz:
This is a purely speculative fix for one user with BSOD? Should we build one just for him first?

taruti:
It has no negative effects though.

zanderz:
LGTM :thumbsup:

PR #127, opened 2016-07-06
delete blocks in a constantly running goroutine; also use a `chan` instead of a `map` for blocks to be deleted

songgao:
selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -30,6 +30,11 @@ const (
 	numMaxRevisionsPerQR = 100
 )
 
+type blocksToDelete struct {
+	md     *RootMetadata
+	blocks []BlockPointer
+}
+
 // folderBlockManager is a helper class for managing the blocks in a
 // particular TLF.  It archives historical blocks and reclaims quota
 // usage, all in the background.
@@ -51,14 +56,17 @@ type folderBlockManager struct {
 	archiveCancelLock sync.Mutex
 	archiveCancel     context.CancelFunc
 
-	// blocksToDeleteAfterError is a list of blocks, for a given
+	// blocksToDeleteChan is a list of blocks, for a given
 	// metadata revision, that may have been Put as part of a failed
-	// MD write.  These blocks should be deleted as soon as we know
+	// MD write. These blocks should be deleted as soon as we know
 	// for sure that the MD write isn't visible to others.
-	// The lock should only be held immediately around accessing the
-	// list.  TODO: Persist these to disk?
-	blocksToDeleteLock       sync.Mutex
-	blocksToDeleteAfterError map[*RootMetadata][]BlockPointer
+	// TODO: Persist these to disk?
+	blocksToDeleteChan      chan blocksToDelete
+	blocksToDeletePauseChan chan (<-chan struct{})
+	blocksToDeleteWaitGroup RepeatedWaitGroup
+
+	blocksToDeleteCancelLock sync.Mutex
+	blocksToDeleteCancel     context.CancelFunc
 
 	// forceReclamation forces the manager to start a reclamation
 	// process.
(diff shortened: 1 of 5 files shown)


libkbfs/folder_block_manager.go
```diff
@@ -320,82 +335,59 @@ func (fbm *folderBlockManager) deleteBlockRefs(ctx context.Context,
...
-				toDeleteAgain[md] = ptrs
-			}
+	_, err = fbm.deleteBlockRefs(ctx, toDelete.md, toDelete.blocks)
+	// Ignore permanent errors
+	_, isPermErr := err.(BServerError)
+	_, isNonceNonExistentErr := err.(BServerErrorNonceNonExistent)
+	if err != nil {
+		fbm.log.CWarningf(ctx, "Couldn't delete some ref in batch %v: %v", toDelete.blocks, err)
+		if !isPermErr && !isNonceNonExistentErr {
+			fbm.enqueueBlockToDelete(toDelete)
```
strib: I _think_ we'll want a `enqueueBlockToDeleteNoWait()` function (like `archiveUnrefBlocksNoWait()`, to make absolutely sure we don't cause deadlock when `blocksToDeleteChan` is full, but isn't being drained because the deleting goroutine is stuck trying to send on the channel!

libkbfs/folder_block_manager.go
```diff
@@ -504,6 +490,35 @@ func (fbm *folderBlockManager) archiveBlocksInBackground() {
 	}
 }
 
+func (fbm *folderBlockManager) deleteBlocksInBackground() {
+	for {
+		select {
+		case toDelete := <-fbm.blocksToDeleteChan:
+			fbm.runUnlessShutdown(func(ctx context.Context) (err error) {
+				if err := fbm.processBlocksToDelete(ctx, toDelete); err != nil {
```
strib: Just like the archiver, I think you'll want to make a new context with a cancel function here, and cancel it in the `shutdown()` method, to make sure any deletes happening in the background get shut down when a test is over.  Limiting it to the `backgroundTaskTimeout` is probably a good idea too.

strib:
Thanks for doing this!  Finished a pass, just a few things left to do.

libkbfs/folder_block_manager.go
songgao: Whoa I didn't think that far. Thanks for catching that!

What do you think if we just do a `go fbm.enqueueBlockToDelete(toDelete)`? It would 1) still insert it to the end of queue, allowing more time before it tries again (Maybe this can be useful when a mdserver or bserver is temporarily under high load and not responding); 2) avoid having long recursive calls when the `chan` is full.

Although for 2), it's not very convincing because it's essentially trading that by spawning a lot of goroutines. I don't know which is worse.

strib: I think it should do the same thing as the archiver, which is to try to send and only use a goroutine when it's full.  That way in the common case we still get ordered deletes without a bunch of extra goroutines.

songgao: Ah right! Sorry I jumped the gun too fast without reading `archiveUnrefBlocksNoWait`'s implementation

songgao:
Tests passed! Could you take another look when you have time? @strib Thanks!

libkbfs/folder_block_manager.go
```diff
@@ -61,9 +61,11 @@ type folderBlockManager struct {
...
 	// for sure that the MD write isn't visible to others.
 	// TODO: Persist these to disk?
-	blocksToDeleteChan      chan blocksToDelete
-	blocksToDeletePauseChan chan (<-chan struct{})
-	blocksToDeleteWaitGroup RepeatedWaitGroup
+	blocksToDeleteChan       chan blocksToDelete
+	blocksToDeletePauseChan  chan (<-chan struct{})
+	blocksToDeleteWaitGroup  RepeatedWaitGroup
+	blocksToDeleteCancel     context.CancelFunc
+	blocksToDeleteCancelLock sync.Mutex
```
strib: Please put the lock above the thing it protects (with maybe a blank line above it to make things clear).

libkbfs/folder_block_manager.go
```diff
@@ -186,6 +208,30 @@ func (fbm *folderBlockManager) enqueueBlockToDelete(toDelete blocksToDelete) {
 	fbm.blocksToDeleteChan <- toDelete
 }
 
+// enqueueBlockToDeleteNoWait enqueues a block to be deleted just like
```
strib: "a block" -> "blocks".

Actually, the function names of this and the one above should be "Block" -> "Blocks" as well.

strib:

songgao:
Thanks @strib!

PR #133, opened 2016-07-08
folder_block_manager: don't re-enqueue deletes immediately

strib:
Since that could cause a near-infinite loop, spinning on waiting for
the next update to make it to the server.

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -209,6 +209,13 @@ func (fbm *folderBlockManager) enqueueBlocksToDelete(toDelete blocksToDelete) {
 	fbm.blocksToDeleteChan <- toDelete
 }
 
+func (fbm *folderBlockManager) enqueueBlocksToDeleteAfterShortDelay(
+	toDelete blocksToDelete) {
+	fbm.blocksToDeleteWaitGroup.Add(1)
+	time.AfterFunc(10*time.Millisecond,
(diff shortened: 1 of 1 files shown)


strib:
Self-reviewing (after passing CI) because the issue this fixes has the potential to chew up serious CPU and I don't think it should go into the morning release.

libkbfs/folder_block_manager.go
```diff
@@ -209,6 +209,13 @@ func (fbm *folderBlockManager) enqueueBlocksToDelete(toDelete blocksToDelete) {
 	fbm.blocksToDeleteChan <- toDelete
 }
 
+func (fbm *folderBlockManager) enqueueBlocksToDeleteAfterShortDelay(
+	toDelete blocksToDelete) {
+	fbm.blocksToDeleteWaitGroup.Add(1)
+	time.AfterFunc(10*time.Millisecond,
```
songgao: Shall we maybe put this in a `const` like `backgroundTaskTimeout`?

songgao: Oops just noticed it's already merged. NVM :D

strib: Yeah sorry, I merged it without review because I was seeing a lot of log spam and CPU usage last night due to a near-infinite receive-send loop.  Probably would be better as a const, I'll clean it up one day.

songgao: No worries; Thanks!

PR #135, opened 2016-07-08
move delay for block deletion retry to const

songgao:


strib:
Looks good to me, thanks!

songgao:
Thx Jeremy!

PR #137, opened 2016-07-08
set latest merged revision in CheckForNewMDAndInit

songgao:


strib:
Code looks good.  Maybe we can test this by adding something like this to `TestUnmergdAfterRestart`, right before the call to `UnstageForTesting`:

```go
        ops1B := getOps(config1B, fileNode1B.GetFolderBranch().Tlf)
        ops2B := getOps(config2B, fileNode1B.GetFolderBranch().Tlf)
        lState := makeFBOLockState()
        if ops1B.getLatestMergedRevision(lState) != ops2B.getCurrMDRevision(lState) {
                t.Fatalf(...)
        }
```

songgao:
@strib Is this for testing the changes in this PR? Because it looks like it doesn't fail without the changes from this PR

strib:
Hrm, yeah it was supposed to be a regression test.  Oh I think I see: the `registerAndWaitForUpdates` probably gets to run and sets the merged head that way, before your test check happens.  I don't see any easy way around that, so just go ahead and merge as is.

songgao:
Thanks Jeremy!

PR #139, opened 2016-07-11
dokan: Load dokan1.dll dynamically

taruti:
This allows for more freedom in specifying the location of dokan1.dll


zanderz:
Who calls LoadDokanDLL()? How do we get the path to Dokan to check? It's different on 64 vs 32

taruti:
LoadDokanDLL can be called manually but if it is not defined then dokan1.dll is loaded with LOAD_LIBRARY_SEARCH_SYSTEM32 which automatically gets the right path.

In the future we can call it manually from kbfsdokan like: `LoadDokanDLL(keybaseInstallLocationDokanDll)`.

taruti:
The automatic call is inside the C code.

zanderz:
LGTM :thumbsup:

PR #145, opened 2016-07-12
TLF reset bug fixes

aalness:
This along with https://github.com/keybase/kbfs-server/pull/64 fixes issues seen when integration testing in the staging environment.

Note: It's not obvious from the diff but when the API server returns unverified keys they aren't tagged as sibling or subkeys so I needed to stop filtering them as such (they all ended up as subkeys w/filtering.)

### DIFF
--- a/libkbfs/tlf_handle.go
+++ b/libkbfs/tlf_handle.go
@@ -637,19 +637,19 @@ func (h TlfHandle) ResolvesTo(
 
 	// Check the conflict extension.
 	var conflictAdded, finalizedAdded bool
-	if h.conflictInfo == nil && other.conflictInfo != nil {
+	if !h.IsConflict() && other.IsConflict() {
 		conflictAdded = true
 		// Ignore the added extension for resolution comparison purposes.
 		other.conflictInfo = nil
 	}
 
 	// Check the finalized extension.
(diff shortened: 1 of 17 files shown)


libkbfs/tlf_handle.go
```diff
@@ -626,8 +632,29 @@ func (pr partialResolver) Resolve(ctx context.Context, assertion string) (
 // resolved except for unresolved assertions in other; this should
 // equal other if and only if true is returned.
 func (h TlfHandle) ResolvesTo(
-	ctx context.Context, codec Codec, resolver resolver, other *TlfHandle) (
+	ctx context.Context, codec Codec, resolver resolver, other TlfHandle) (
 	resolvesTo bool, partialResolvedH *TlfHandle, err error) {
+
+	// Check the conflict extension.
+	var conflictAdded, finalizedAdded bool
+	if h.conflictInfo == nil && other.conflictInfo != nil {
```
strib: Maybe `TlfHandle` should get `IsConflict()` and `IsFinal()` methods too, like `BareTlfHandle`?

strib:
Looks good to me, just one suggestion, thanks!

aalness:
Thanks!

PR #156, opened 2016-07-13
fix the issue caused by squash merge

songgao:


strib:
Looks good to me, thanks!

PR #157, opened 2016-07-14
add staller support to dsl tests

songgao:
This PR is pretty much #151 rebased to master, but created against master rather than `songgao/KBFS-1221`.

GitHub, please close #151 when this is merged.

### DIFF
--- a/libkbfs/test_stallers.go
+++ b/libkbfs/test_stallers.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"fmt"
+	"sync"
 
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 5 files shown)


songgao:
How could it fail?

download

libkbfs/test_stallers.go
```diff
@@ -38,6 +39,156 @@ const (
 	StallableMDPutUnmerged           StallableMDOp = "PutUnmerged"
 )
 
+const stallKeyStallEveryThing = ""
```
akalin-keybase: I think `Everything` instead of `EveryThing` is more idiomatic

libkbfs/test_stallers.go
```diff
@@ -38,6 +39,156 @@ const (
...
+
+// NaïveStaller is used to stall certain ops in BlockOps or MDOps. Unlike
+// StallBlockOp and StallMDOp which provides a way to precisely control which
+// particular op is stalled by passing in ctx with corresponding stallKey,
+// NaïveStaller simply stalls all instances of specified op.
+type NaïveStaller struct {
+	config Config
+
+	mu             sync.RWMutex
+	blockOpsStalls map[StallableBlockOp]*naïveStallInfo
```
akalin-keybase: My rationale for preferring value types here is that returning pointers to lock-protected structures carry the danger that something might inadvertently modify them outside the lock.

But I don't think it matters too much in this case, so I'm okay with leaving it as pointers.

akalin-keybase:
lgtm

libkbfs/test_stallers.go
songgao: Oops; no idea why I put in the upper case T :D Thanks Fred

songgao:
Thanks for the fast response Fred!

libkbfs/test_stallers.go
songgao: Got you; I'll keep that in mind for future reference!

PR #168, opened 2016-07-17
revert the behavior that Lookup handler adds TLF to favorite

songgao:
Tested locally in my zsh with `syntax-highlighting` enabled. Resolves https://github.com/keybase/client/issues/3503

### DIFF
--- a/libdokan/folderlist.go
+++ b/libdokan/folderlist.go
@@ -93,20 +93,7 @@ func (fl *FolderList) open(ctx context.Context, oc *openContext, path []string) 
 		}
 		fl.mu.Unlock()
 
-		h, errParseTlfHandle := libkbfs.ParseTlfHandle(
-			ctx, fl.fs.config.KBPKI(), name, fl.public)
-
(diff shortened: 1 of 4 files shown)


strib:
I think this will bring back the problems with `rmdir` on OS X again.  Because the `folderBranchOps` object for the removed object is still in the list of folders, the subsequent `ls` that `rmdir` does will succeed and get plumbed through the `folderBranchOps` object to do a `Stat`, which will cause it to get re-added to favorites.  I think we might need to keep the `recentlyRemoved` stuff around.  Unless there's something I'm missing?

songgao:
@strib I thought `recentlyRemoved` existed because we did "add TLF to favorites in Lookup" and that OS X always does a lookup at the end of `rmdir`. If we don't add TLF to favorites anymore, it should be fine right?

strib:
`folderBranchOps` itself adds it to favorites if it gets a `Stat` call, which it would if the TLF is still around.

PR #170, opened 2016-07-18
Create logger before establishing connection to avoiding racing with OnConnect

jzila:


strib:
Looks good to me once CI passes, thanks!

PR #171, opened 2016-07-18
qr_test: add workaround to TestQRLargePointerSet for osxfuse bug

strib:
If a Lookup gets issued from outside the test (by Finder?)
concurrently with the Remove, it could end up adding the file name
back into the osxfuse lookup cache, and a subsequent lookup by the
test can succeed incorrectly when it hits the cache.

This PR works around that by simply using a unique name for each
iteration, while looking into the osxfuse issue separately.

Issue: KBFS-1253

### DIFF
--- a/test/qr_test.go
+++ b/test/qr_test.go
@@ -7,6 +7,7 @@
 package test
 
 import (
+	"fmt"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


strib:
@jzila maybe you can take a look since it's a CI flake (and Fred is out today)?

jzila:
Lgtm

PR #172, opened 2016-07-19
dokan*: Disable timeouts and allow customizing mount options from command line

taruti:
This disables timeouts on the Dokan level - they are still handled with context.

`-mount-flags` as follows:
0 - nothing
3 - extra debugging
32 - removable
64 - mount manager
128 - only current session

Can be combined: e.g. removable + mount manager + only current session = 224.


zanderz:
Tested, BitDefender BSOD avoided with related driver change, but still occurs with Symantec.
LGTM :thumbsup:

PR #173, opened 2016-07-19
Context Hack to Support Critical Awareness

songgao:
This is an attempt to fix a use case (e.g. `git`) where a O_EXCL create canceled (interrupted) due to a signal (e.g. SIGALRM) returns EINTR while the creation succeeds. Relevant JIRA Ticket: [KBFS-1123](https://keybase.atlassian.net/browse/KBFS-1123)


songgao:
@strib @tv42  This fixes the checkout part of `git clone`, but in the end, it gets a `bus error (core dumped)` (most of time). I suppose this has something to do with internal assumptions in fuse or kernel?

jzila:
Please rebase to get CI working

strib:
Ignoring cancel across the entire operation is too much, I think.  The operating system, not to mention the user, really needs a way to cancel outstanding operations when possible.  On OS X in particular, the OS enforces a strict amount of time an operation can run, and will send an interrupt after that time is up.

In our case, "stat", "lookup" and "mkdir" are not fast operations, and we can't really afford to ignore completely a user- or OS-initiated cancel.  For example, if we are blocking waiting for a user to enter their password, we will be stuck forever until the user notices the pinentry window.  What if we just gave it an extra epsilon time to complete (e.g., 1 or 2 seconds)?  It wouldn't hurt usability too much and it should solve most of the git problems you've experienced.  Basically, you'd have to copy the context like you're doing here, and then kick off a goroutine that would listen for `ctx.Done()` on the original context, and wait an extra timeout before hitting cancel on the new context.  It's ugly but it's the best I can think of.  What do you think?

As for fixing the "create" case, it might be better to use a similar strategy but just around the `MDOps.Put()`/`MDOps.PutUnmerged()` call. If one of those calls returns with a `context.Canceled` error, we can start a new context (with a timeout of 1 or 2 seconds) that checks for the latest head revision from the MD server to see if our operation succeeded or not.  We'd have to restructure the goroutine wrapping the create call a little bit to always wait for the return error if the finalize logic has reached the Put call.  The goal would be to minimize the amount of time we ignore the interrupt, and if it does happen at the worst possible time, still try to return the right answer.  Thoughts?

songgao:
Thanks @strib. I like the idea of 2-3s grace period rather than either ignore the interrupt all together or canceling right away. This way in bad network conditions, user would experience the same error, but in good network conditions, git should work properly. Although since it depends on network conditions, shall we make this configurable somehow, so people can choose the tolerance timeout based on their network conditions? We could also provide an option of an adaptive setting which automatically adjusts this timeout based on recent performance measurements (which we'd need to add).

About the Create path, is the reason for checking latest head revision that, so we can returning EINTR if head revision is not properly updated to reflect the change, and `nil` error if it does?

strib:
> shall we make this configurable somehow, so people can choose the tolerance timeout based on their network conditions?

I think a future where we expose this knob to people is a future where we've failed to build a usable product.  If you want, make it an option to `NewFS` or expose it in the `libkbfs.Config` interface, but I wouldn't bother making a command line or env variable option for it.

> About the Create path, is the reason for checking latest head revision that, so we can returning EINTR if head revision is not properly updated to reflect the change, and nil error if it does?

Yes, exactly.

songgao:
> I think a future where we expose this knob to people is a future where we've failed to build a usable product. If you want, make it an option to NewFS or expose it in the libkbfs.Config interface, but I wouldn't bother making a command line or env variable option for it.

How about the adaptive thing? Although it'd be a separate PR

strib:
> How about the adaptive thing? Although it'd be a separate PR

It's a good idea, but in my opinion not worth spending time on right now.  I'd say add a TODO and a JIRA ticket and move on.

songgao:
I'm still working on fixing the tests, but the latest commit adds new context helpers (in `delayed_cancellation.go`) which introduce the a `criticalAwareness` object into context, that enables communicating from operations to associated context about when the operations have entered/exited a critical state that is not safe to cancel. This help us to maintain a consistent metadata state when interrupt (mostly concerning SIGALRM since it's pretty common with `git`) happens during writes.

`git clone` works now! Well, almost; tested with `keybase/client` and everything's fine except a [coredump](https://gist.github.com/songgao/5978fa5f1cc61df1f3da834d7807329f) in the end. Any ideas?

songgao:
Also this is still a pretty hacky way to "fix" it. At some point in the future, can we perhaps try to get rid of `runUnlessCanceled`? The `runUnlessCanceled` makes sure the function returns as soon as context is canceled for improved responsiveness for, e.g. Ctrl-C. Maybe we can instead add the context check before every remote operation or locking attempt? This way the execution flow is never interrupted, and the handlers have more flexibility in determining when it is safe to cancel vs when it is not.

strib:
re: coredump, you might need to compile git from source to get more info.  It looks like git doesn't have an ubuntu package with debug symbols: https://bugs.launchpad.net/ubuntu/+source/git/+bug/911693

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,205 @@
...
+		return nil
+	}
+	return NoCriticalAwarenessError{}
+}
+
+// ExitCritical can be called on a "critical aware" context produced by
+// NewContextWithCriticalAwareness, to indicate that the operation(s)
+// associated with the context has exited a critical state and it is safe to be
+// canceled.
+func ExitCritical(ctx context.Context) error {
```
strib: Shouldn't this function stop `c.timer` if it's not `nil`?

libkbfs/folder_branch_ops.go
```diff
@@ -1863,19 +1863,95 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
...
+	if bid == NullBranchID {
+		errPut = fbo.config.MDOps().Put(ctx, md)
+	} else {
+		errPut = fbo.config.MDOps().PutUnmerged(ctx, md, bid)
+	}
+	mdID, err = fbo.config.Crypto().MakeMdID(&md.BareRootMetadata)
+	if err != nil {
+		return mdID, err
+	}
+	if errPut == context.Canceled {
```
strib: Checking for `context.DeadlineExceeded` probably makes sense too.

libkbfs/folder_branch_ops.go
```diff
@@ -1863,19 +1863,95 @@ func (fbo *folderBranchOps) isRevisionConflict(err error) bool {
...
+		// way for user to cancel now.
+		ctx, err = NewContextWithReplayFrom(ctx)
+		if err != nil {
+			panic(err) // we shouldn't let this happen
+		}
+		ctx, _ = context.WithTimeout(ctx, fbo.config.GracePeriod())
+
+		var rmds []ImmutableRootMetadata
+		if bid == NullBranchID {
+			rmds, err = getMergedMDUpdates(ctx, fbo.config, fbo.id(), oldRevision+1)
```
strib: Why do you need `oldRevision`?  Why not just use `md.Revision`?

libkbfs/folder_branch_ops.go
```diff
@@ -1911,7 +1987,7 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 		} else {
 			bid = fbo.bid
 		}
-		err := mdops.PutUnmerged(ctx, md, bid)
+		mdID, err = fbo.doPutWithCancellationCheck(ctx, lState, md, bid)
```
strib: There are a few calls after this that use `ctx`, even though it might be canceled now.  I don't think any of them check for a canceled context, but if they did, or if we added something that did, the whole operation would fail anyway.  Maybe `doPutWithCancellationCheck` should return the new context and we can use that for the rest of it.  Technically probably `finalizeMDWriteLocked` should also return the new context for its callers to use, etc.

libfuse/dir.go
```diff
@@ -347,6 +353,12 @@ func (d *Dir) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.Lo
 	d.folder.fs.log.CDebugf(ctx, "Dir Lookup %s", req.Name)
 	defer func() { d.folder.reportErr(ctx, libkbfs.ReadMode, err) }()
 
+	err = libkbfs.EnterCriticalWithTimeout(
```
strib: There are no matching `ExitCritical` calls for any of these.  Shouldn't there be, to clean up state?

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,205 @@
...
+		return nil, err
+	}
+	c := newCriticalAwareness()
+	newCtx = NewContextReplayable(newCtx,
+		func(ctx context.Context) context.Context {
+			return context.WithValue(ctx, CtxCriticalAwarenessKey, c)
+		})
+	newCtx, cancel := context.WithCancel(newCtx)
+	go func() {
+		<-ctx.Done()
```
strib: If `ctx` didn't have a canceled function, or it's never called, then this will leak goroutines.  Maybe you could save `cancel` in the struct, call it in `ExitCritical`, and select on it here?

strib:
The overall approach here looks great to me! Nice work!

libkbfs/delayed_cancellation.go
songgao: Yes; nice catch!

libkbfs/folder_branch_ops.go
songgao: It's "critical aware" so it can only be `context.Canceled` right? But yet, no harm to check `context.DeadlineExceeded` as well

strib: I view the deadline as just an automated way of canceling something -- seems very much in line with the critical awareness to me.

songgao: hmm.. What I mean was in `delayed_concellation.go`, it disconnects the old cancellations, and only use `cancel()` when timeout hits, so `context.DeadlineExceeded` wouldn't happen here.

(just making sure we are on the same page!)

libkbfs/folder_branch_ops.go
songgao: I was under the impression that `Put`/`PutUnmerged` would update `md.Revision`. Looks like it's not true? Is it only updated through `setHeadSuccessorLocked`?

libkbfs/folder_branch_ops.go
songgao: It's the same context that is already "critical aware", which had a timeout on cancellation (if any) equal to `GracePeriod()`. Thinking about this together with the comment below on adding `ExitCritical`, shall we maybe do `ExitCritical` at the end of `finalizeMDWriteLocked`?

libkbfs/folder_branch_ops.go
strib: I don't think that's quite true.  `NewContextWithCriticalAwareness` calls `NewContextWithReplayFrom`, which could set a timeout on the context.  (`libfuse.FS.WithContext` does exactly that.)  So certainly the new context could still timeout.  Right?

libfuse/fs.go
```diff
@@ -82,33 +82,49 @@ func (f *FS) LaunchNotificationProcessor(ctx context.Context) {
...
+			if runtime.GOOS == "darwin" {
+				// Timeout operations before they hit the osxfuse time limit,
+				// so we don't hose the entire mount (Fixed in OSXFUSE 3.2.0).
+				// The timeout is 60 seconds, but it looks like sometimes it
+				// tries multiple attempts within that 60 seconds, so let's go
+				// a little under 60/3 to be safe.
+				//
+				// It should be safe to ignore the CancelFunc here because our
+				// parent context will be canceled by the FUSE serve loop.
+				ctx, _ = context.WithTimeout(ctx, 19*time.Second)
```
strib: This isn't quite right anymore, because the replay could technically happen much later.  Instead, we should probably calculate a deadline outside of this anonymous function, and set a deadline here instead of a timeout.

libfuse/dir.go
songgao: It has a timeout, which is why I didn't put an `ExitCritical` explicitly. But now thinking about it more, do we have code that relies on `<-ctx.Done()` to cleanup, and requres us to make sure `<-ctx.Done()` happens in time?

libkbfs/folder_branch_ops.go
strib: Correct, it's already fixed here.

libkbfs/folder_branch_ops.go
strib: I'm talking about the `if errPut == context.Canceled {` case above in `finalizeMDWriteLocked`.  If we hit that case, then the grace period is already over.  But you're already put a lot of effort into possibly returning a `nil` error for that case, which could in theory be discarded by future checks of `ctx.Done()`.

libkbfs/delayed_cancellation.go
songgao: That's a good point! But on the other hand, `ExitCritical` being called doesn't mark the end of a context. There might be more ops (relying on the same context) that need to work if the parent cancellation isn't issued yet. Also `EnterCriticalWithTimeout` could be called again. How about instead we make sure `cancel()` is always called on parent ctx when a handler is finished?

libkbfs/folder_branch_ops.go
strib: And yes, we should also exit the critical section in `finalizeMDWriteLocked`, ideally using a `defer`.

libfuse/fs.go
songgao: Nice catch; Thanks Jeremy!

libfuse/dir.go
strib: Yes possibly, plus in tests the goroutines/memory from previous tests could stick around a long time, making it more difficult and confusing to debug stacktraces.

libkbfs/delayed_cancellation.go
strib: Oh right, good point.  I'm not sure what you mean by your followup though.  We might not have any control of the parent context, right?  From a libkbfs perspective.

Hrmm.  Maybe `NewContextWithCriticalAwareness` should return the cancel function and require that the caller always calls it?

libkbfs/folder_branch_ops.go
songgao: Ah I see!

Damn it I just realize the checks done in `doPutWithCancellationCheck` after the `Put` doesn't make sense since if the `errPut` is `context.Canceled`, it means the context is canceled after all (even with the `EnterCriticalWithTimeout`). At that point, the `runUnlessCanceled` would have returned early and we still have  problems.

songgao: But I does it still help with keeping a consistent state internally?

strib: Hmm.  It's a little bit useless -- it basically just doubles the grace period.  Maybe we can do without it.  The only case I can think of where it would help is if the Put call completely hangs on the server due to some bug, and doesn't return quickly even though the Put succeeded.  Unless you can think of a more realistic bug this would fix (that wouldn't also be fixed by just increasing the grace period), we can probably drop that logic.

libkbfs/folder_branch_ops.go
songgao: Ah, you are right. I think I intended to never have `WithTimeout`/`WithCancel` replayable, but that OS X thing is a special case and somehow I overlooked it when adding the "critical aware" stuff

libfuse/fs.go
songgao: Hold on; but if we move it out of the anonymous function, it'd get disconnected by `NewContextWithCriticalAwareness`. Maybe keep it inside but use a deadline (calculated from `time.Now()` outside the anonymous function) rather than timeout?

strib: Yep, that's exactly what I was suggesting.  Sorry if I wasn't clear.

strib: But please  use `Config.Clock().Now()` instead of `time.Now()`.

libkbfs/delayed_cancellation.go
songgao: I was thinking about always calling `cancel()` when a request is done. In this case, this would have to be in fuse, which seems to [be there](https://github.com/bazil/fuse/blob/b3d5a5514a27076c63844bde6d60703a7aea4adc/fs/serve.go#L771) already?

libkbfs/folder_branch_ops.go
songgao: Good call; that'd simplify `doPutWithCancellationCheck` (also rename) a lot as well!

libkbfs/delayed_cancellation.go
strib: Right, but we can't control that from within `libkbfs` so it seems dangerous to rely on it.  For example, a lot of tests don't even call `WithCancel`, etc.  If you just return `cancel` from here and document that the caller has to cancel it when they're done with the context, that should be good enough.

libfuse/fs.go
songgao: :ok_hand:

libkbfs/delayed_cancellation.go
songgao: @strib OK I've added a `CleanupContextAwareContext` for this, but not calling it yet, considering that fuse is cancelling that right now. I was thinking we can add this call if they decide to not cancel upon completion of each request, but I'm fine with adding it now to every fuse handler as well if you think it's better/safer that way. Just let me know :)

songgao:
Two tests are still failing (102d8cf disables them). Will look into fixing them!

songgao:
Hmm.. cloning `keybase/kbfs` succeeded, but `keybase/client` failed:

```
songgao at Song-ThinkPad in /tmp/kbfstmp/private/strib 
$ git clone https://github.com/keybase/kbfs
Cloning into 'kbfs'...
remote: Counting objects: 20237, done.
remote: Compressing objects: 100% (214/214), done.
remote: Total 20237 (delta 127), reused 3 (delta 3), pack-reused 20020
Receiving objects: 100% (20237/20237), 24.76 MiB | 783.00 KiB/s, done.
Resolving deltas: 100% (14739/14739), done.
Checking connectivity... done.
Checking out files: 100% (1268/1268), done.

songgao at Song-ThinkPad in /tmp/kbfstmp/private/strib 
$ git clone https://github.com/keybase/client
Cloning into 'client'...
remote: Counting objects: 96372, done.
remote: Compressing objects: 100% (208/208), done.
remote: Total 96372 (delta 99), reused 3 (delta 3), pack-reused 96161
Receiving objects: 100% (96372/96372), 154.91 MiB | 699.00 KiB/s, done.
Resolving deltas: 100% (71513/71513), done.
Checking connectivity... done.
error: unable to create file go/libkb/tablify.go: File exists
error: unable to create file go/protocol/notify_ctl.go: File exists
error: unable to create file osx/KBKit/KBKit/UI/KBImageLabel.h: File exists
error: unable to create file osx/KBKit/KBKit/Utils/KBWaitFor.h: File exists
[1]    17654 bus error (core dumped)  git clone https://github.com/keybase/client
```

 The latter one took much longer, and it *seems* the fixes were ineffective. Could it be the ops are taking too long to finish that they just hit the grace period timeout? Also seeing a lot of `Getxattr ... "security.capabilities"` in logs. Do they slow things down?

songgao:
> Could it be the ops are taking too long to finish that they just hit the grace period timeout?

Well, changing grace period from 2 minutes to 2 hours didn't "fix" the "file exists" error, so that's not the case. I'll look into it later today or tomorrow ...

songgao:
New coredump's not from `libz`. Hmm...

```
Aug 09 16:54:01 Song-ThinkPad systemd-coredump[3720]: Process 3619 (git) of user 1000 dumped core.
                                                      
                                                      Stack trace of thread 3619:
                                                      #0  0x00000000005631d2 unpack_object_header_buffer (git)
                                                      #1  0x0000000000563c18 unpack_object_header (git)
                                                      #2  0x0000000000564011 packed_object_info (git)
                                                      #3  0x0000000000565b57 sha1_object_info_extended (git)
                                                      #4  0x000000000057195d istream_source (git)
                                                      #5  0x0000000000571a0c open_istream (git)
                                                      #6  0x000000000057267f stream_blob_to_fd (git)
                                                      #7  0x00000000004e4002 streaming_write_entry (git)
                                                      #8  0x00000000004e4112 write_entry (git)
                                                      #9  0x00000000004e4766 checkout_entry (git)
                                                      #10 0x0000000000587e82 check_updates (git)
                                                      #11 0x000000000058a01f unpack_trees (git)
                                                      #12 0x000000000042f7a5 checkout (git)
                                                      #13 0x00000000004308f6 cmd_clone (git)
                                                      #14 0x0000000000405854 run_builtin (git)
                                                      #15 0x0000000000405a42 handle_builtin (git)
                                                      #16 0x0000000000405b66 run_argv (git)
                                                      #17 0x0000000000405ce3 cmd_main (git)
                                                      #18 0x00000000004a0bd8 main (git)
                                                      #19 0x00007f6370305741 __libc_start_main (libc.so.6)
                                                      #20 0x0000000000404709 _start (git)
``

File: libfuse/fs.go [hunk 8]
```diff
@@ -82,33 +82,52 @@ func (f *FS) LaunchNotificationProcessor(ctx context.Context) {
...
-		//
-		// It should be safe to ignore the CancelFunc here because our
-		// parent context will be canceled by the FUSE serve loop.
-		ctx, _ = context.WithTimeout(ctx, 19*time.Second)
+	// context.WithDeadline uses clock from `time` package, so we are not using
+	// f.config.Clock() here
+	start := time.Now()
+	ctx, err := libkbfs.NewContextWithCancellationDelayer(
+		libkbfs.NewContextReplayable(ctx, func(ctx context.Context) context.Context {
+
```
strib: Extra blank line.

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+// ExitCritical are called on a ctx without Critical Awareness
+type NoCancellationDelayerError struct{}
+
+func (e NoCancellationDelayerError) Error() string {
+	return "Context doesn't have critical awareness or CtxCancellationDelayerKey " +
+		"already exists in ctx but is not of type *cancellationDelayer"
+}
+
+// ContextAlreadyHasCancellationDelayerError is returned when
+// NewContextWithCancellationDelayer is called second time on the same ctx, which
```
strib: "called second time" -> "called for the second time"

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (e ContextAlreadyHasCancellationDelayerError) Error() string {
+	return "Context already has critical awareness; only one layer is supported."
+}
+
+// NewContextReplayable creates a new context from ctx, with change applied. It
+// also makes this change replayable by NewContextWithReplayFrom. When
+// replayed, the resulting context is replayable as well.
+//
+// It is important that all WithValue-isu mutations on ctx is done "replayably"
```
strib: "isu" -> "ish"?

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+// (with NewContextReplayable) if any delayed cancellation is used, e.g.
+// through EnableDelayedCancellationWithGracePeriod,
+func NewContextReplayable(
+	ctx context.Context, change CtxReplayFunc) context.Context {
+	var replay CtxReplayFunc
+	replay = func(ctx context.Context) context.Context {
+		ctx = change(ctx)
+		// _ is necessary to avoid panic if nil
+		replays, _ := ctx.Value(CtxReplayKey).([]CtxReplayFunc)
+		replays = append(replays, replay)
```
strib: Whoa, this is a little crazy!  You're appending the same function you're defining?  Why can't it just hold the slice of `change`s?  Then `NewContextWithReplayFrom` can be responsible for explicitly adding the slice of `change`s back into the new context after creating it.

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+	delayEnabled bool
+	timer        *time.Timer
+	canceled     bool
+	done         chan struct{}
+}
+
+func newCancellationDelayer() *cancellationDelayer {
+	return &cancellationDelayer{
+		Cond:         sync.NewCond(&sync.Mutex{}),
+		delayEnabled: false,
```
strib: No need to explicitly assign the default zero values for `delayEnabled`, `timer`, or `canceled`.

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+		timer:        nil,
+		canceled:     false,
+		done:         make(chan struct{}),
```
strib: Might be better to give this a capacity of 1, in case someone tries calling `CleanupCancellationDelayer` after the original context was cancelled.

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+	return NoCancellationDelayerError{}
+}
+
+func (c *cancellationDelayer) disableDelay() {
+	c.L.Lock()
+	defer c.L.Unlock()
+	c.disableDelayLocked()
+}
+
+func (c *cancellationDelayer) disableDelayLocked() {
```
strib: I don't see another caller of this function.  Can you combine it with the one above?

libkbfs/delayed_cancellation.go
```diff
@@ -0,0 +1,224 @@
...
+	if c.delayEnabled {
+		c.delayEnabled = false
+		c.Broadcast()
+	}
+}
+
+// CleanupCancellationDelayer cleans up a context (ctx) that is cancellation
+// delayable and makes the go routine spawned in
+// NewContextWithCancellationDelayer exit. As part of the cleanup, this also
+// causes the cancellation delayable context to be canceled, no matter the
```
strib: "no matter the" -> "no matter whether the"

libkbfs/folder_branch_ops.go
```diff
@@ -2845,6 +2859,32 @@ func (fbo *folderBranchOps) Read(
 		return 0, err
 	}
 
+	filePath, err := fbo.pathFromNodeForRead(file)
+	if err != nil {
+		return 0, err
+	}
+
+	{
+		// It seems git isn't handling EINTR from some of read calls (likely
```
strib: "some of read" -> "some of its read"

libkbfs/folder_branch_ops.go
```diff
@@ -4160,13 +4197,22 @@ const (
 const CtxFBOOpID = "FBOID"
 
 func (fbo *folderBranchOps) ctxWithFBOID(ctx context.Context) context.Context {
-	return ctxWithRandomID(ctx, CtxFBOIDKey, CtxFBOOpID, fbo.log)
+	return ctxWithRandomIDReplayable(ctx, CtxFBOIDKey, CtxFBOOpID, fbo.log)
+}
+
+func (fbo *folderBranchOps) newCtxWithFBID() (context.Context, context.CancelFunc) {
```
strib: "FBID" -> "FBOID"

libkbfs/folder_branch_ops.go
```diff
@@ -4160,13 +4197,22 @@ const (
...
+	ctx := NewContextReplayable(context.Background(), fbo.ctxWithFBOID)
```
strib: This will assign a random FBOID on a replay, right?  It should be static across replays I think.

libkbfs/folder_branch_ops.go
```diff
@@ -3871,7 +3906,7 @@ func (fbo *folderBranchOps) UnstageForTesting(
 		// notifications if we do.  But we still want to wait for the
 		// context to cancel.
 		c := make(chan error, 1)
-		ctxWithTags := fbo.ctxWithFBOID(context.Background())
+		ctxWithTags, cancel := fbo.newCtxWithFBID()
```
strib: `cancel` is overridden on the next line.  I don't think we need the next line anymore actually.

libkbfs/interfaces.go
```diff
@@ -1325,6 +1325,22 @@ type Config interface {
 	// setting the rekey bit, before prompting for a paper key.
 	RekeyWithPromptWaitTime() time.Duration
 
+	// GracePeriod specifies a grace period for which a delayed cancellation
+	// waits before actual cancels the context. This is useful for giving
+	// critical portion of a slow remote operation some extra time to finish as
+	// an effort to avoid conflicting. Example include an O_EXCL Create call
+	// interrupted by ALRM signal actually makes it to the server, while
+	// application assumes not since EINTR is returned. A delayed cancellation
+	// allows us to dinstinct between successful cancel (where remote operation
```
strib: "dinstinct" -> "distinguish"

libkbfs/interfaces.go
```diff
@@ -1325,6 +1325,22 @@ type Config interface {
...
+	// didn't make to server) or failed cancel (where remote operation made to
+	// the server). However, the optimal value of this depends on the network
+	// conditions. A long grace period for really good network condition would
+	// just unnecessarily slow down Ctrl-C.AuthTokenMinRefreshSeconds
```
strib: Is "AuthTokenMinRefreshSeconds" supposed to be there?

libkbfs/folder_branch_ops.go
strib: Nevermind I see how this works now.

libfuse/dir.go
```diff
@@ -343,6 +343,12 @@ func (d *Dir) Attr(ctx context.Context, a *fuse.Attr) (err error) {
 	d.folder.fs.log.CDebugf(ctx, "Dir Attr")
 	defer func() { d.folder.reportErr(ctx, libkbfs.ReadMode, err) }()
 
+	err = libkbfs.EnableDelayedCancellationWithGracePeriod(
```
strib: I think we need some comments here explaining why these particular file system calls need delayed cancellation.  Actually i forget the exact reason why.  Maybe POSIX says that they should always return successfully even when an interrupt/SIGALRM is received, and some programs like git rely on that?

libkbfs/folder_branch_ops.go
```diff
@@ -1914,6 +1915,19 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
 
 	var mdID MdID
 
+	if err = EnableDelayedCancellationWithGracePeriod(
```
strib: A comment here would be good, explaining that this acts somewhat like a critical section, to help us know for sure whether the MD put made it to the server or not even when the operation is cancelled.  This is useful even when `excl != WithExcl`, since it helps us avoid self-conflicts.  But it comes at the cost of a less-responsive ctrl-C once the operation has reached this point.

strib:
This is quite a PR!  Thanks for all your patience and hard work getting it to this point.

Have you tested that ctrl-c definitely works (after the grace period delay) when the MD server is completely unresponsive?  It would be nice to have a real test for that (with a very short grace period), and also a test that cancels an O_EXCL create during the critical section and makes sure it still succeeds.  Have you thought about that yet?  How hard would it be?  Unit tests fo `delayed_cancellation.go` would also be nice.  I just want to make sure this stuff doesn't break in the future, since it's so complicated.

That said, this looks good to me mod comments and tests if possible.  Thanks again!

libkbfs/interfaces.go
```diff
@@ -1325,6 +1325,22 @@ type Config interface {
...
+	//
+	// TODO: make this adaptive and self-change over time based on network
+	// conditions.
+	GracePeriod() time.Duration
```
strib: I know it's wordy. but maybe `DelayedCancellationGracePeriod()` would be more clear? Or `CancellationDelay`?

libkbfs/interfaces.go
```diff
@@ -1325,6 +1325,22 @@ type Config interface {
...
+	SetGracePeriod(time.Duration)
```
strib: Also i think you should run `./gen_mocks.sh` to pick up these changes in the (not very useful) `MockConfig` definition.

libkbfs/folder_branch_ops.go
songgao: Actually I think you were right with your first (deleted) comment! This attaches as a replayable function, which means it gets executed every times it replays. Although in `ctxWithRandomIDReplayable` it attaches a reaplayable using "static" value, that generation of "static" value is itself in replayable as well, causing the FBOID set for multiple times.

So this should instead be just `ctx := fbo.ctxWithFBOID()`. Does it make sense?

libkbfs/folder_branch_ops.go
songgao: Good catch; thanks!

libkbfs/delayed_cancellation.go
songgao: I think that is similar my first approach, but moved to this approach at some point. Although I can't remember off the top of my head exactly why. Let me think about this for a while and I'll let you know!

libkbfs/delayed_cancellation.go
songgao: I was thinking ideally `CleanupCancellationDelayer` would be called at most once. Maybe we can make it closing the channel instead of sending `struct{}` into the channel, to guarantee the function can only be called once. What do you think?

libkbfs/folder_branch_ops.go
strib: Ah right.  I think that makes sense.

libkbfs/delayed_cancellation.go
strib: It doesn't matter if it's only called once -- if the CancelFunc of the original context was called, nothing will be receiving on the channel anymore, so the send will block if there's 0 channel capacity when `CleanupCancellationDelayer` is called.

songgao: Ah that makes sense! Using `close` would be fine though? Since with capacity one, the (presumably accidental) secret be call will block

libkbfs/delayed_cancellation.go
songgao: Will do!

libkbfs/delayed_cancellation.go
strib: Yep, `close` would be fine.

libkbfs/folder_branch_ops.go
songgao: Yeah I think it's because Git doesn't expect EINTR from some of I/O calls. Taru mentioned some manual pages where those calls don't have EINTR listed as possible errors, so Git's behavior is expected in this case.

I'll add comments!

libkbfs/interfaces.go
songgao: Oops! Must be an accidental middle-click :D

songgao:
@strib No problem! Thanks for the extensive review and constant feedback!

Adding tests for ctrl-C is a good idea; Will do! I'll hopefully address all comments above and update the PR within today

libkbfs/folder_branch_ops.go
songgao: Oops this comment was supposed to respond to the one above

libkbfs/delayed_cancellation.go
songgao: I couldn't thinking of a reason why not so I implemented your idea. Thanks Jeremy!

songgao:
@strib I added unit tests for `delayed_cancellation.go` to make sure cancellation actually happens. Since Go doesn't have things like `nextTick`, the order that events happen is not deterministic unless we add things like `WaitGroup` just for test, which does not seem ideal. So I ended up adding some `time.Sleep`, which adds < 100 milliseconds delay to the test suite. Does that seem reasonable to you?

Also I think there should still be tests with Ctrl-C, but would be in `kbfs-server` docker test. I guess we could also put in some quick git tests? I'll make a PR for that later.

libkbfs/delayed_cancellation_test.go
```diff
@@ -0,0 +1,135 @@
...
+func TestDelayedCancellationCancelWhileNotEnabled(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancel := makeContextWithDelayedCancellation(t)
+
+	cancel()
+
+	// wait until the go routine spawned in NewContextWithCancellationDelayer
+	// actually cancels
+	time.Sleep(10 * time.Millisecond)
```
strib: Why do you need a sleep here?  Can't you just `<-ctx.Done()`?  It won't fire until the goroutine reacts to the cancel, right?  If you want to keep the test failure timeout short, you can also select on a `time.After()`.  Maybe I'm missing something.

libkbfs/delayed_cancellation_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	ctx, _ := makeContextWithDelayedCancellation(t)
+
+	if err := CleanupCancellationDelayer(ctx); err != nil {
+		t.Fatalf("calling CleanupCancellationDelayer error: %s", err)
+	}
+
+	// wait until the go routine spawned in NewContextWithCancellationDelayer
+	// actually cancels
+	time.Sleep(10 * time.Millisecond)
```
strib: Same.

libkbfs/delayed_cancellation_test.go
```diff
@@ -0,0 +1,135 @@
...
+	// in NewContextWithCancellationDelayer actually cancels
+	time.Sleep(time.Millisecond * 10)
+	select {
+	case <-ctx.Done():
+		t.Fatalf("Cancellation is not delayed")
+	default:
+	}
+
+	// wait a bit longer and cancellation should happen
+	time.Sleep(time.Millisecond * 10)
```
strib: Same here.

libkbfs/delayed_cancellation_test.go
songgao: Ah you're right! This way there's no delay if tests are passing. I'll update this soon!

libkbfs/delayed_cancellation_test.go
```diff
@@ -0,0 +1,135 @@
...
+	t.Parallel()
+
+	ctx, cancel := makeContextWithDelayedCancellation(t)
+	EnableDelayedCancellationWithGracePeriod(ctx, 15*time.Millisecond)
+
+	cancel()
+
+	// wait until any (presumably) wrong cancellation in the go routine spawned
+	// in NewContextWithCancellationDelayer actually cancels
+	time.Sleep(time.Millisecond * 10)
```
strib: This one doesn't really matter since `<-ctx.Done()` will block for 15 seconds anyway, but you can probably get rid of it too if you want, since it seems just as good to check `ctx.Done` immediately after the cancel to ensure that the context hasn't cancelled yet.

songgao: From my test earlier earlier it seemed that the cancellation wouldn't fire right away even with out delay, since the routing that fires the actual cancellation needs to do some locking and singnaling stuff which makes it slower. But I'll change it to use `time.After` as well to be consistent!

songgao:
Just got rid of all `time.Sleep`s!

strib:
Cool thanks!  As for other tests, I think it would be pretty easy to write one like [TestKBFSOpsConcurCanceledSyncSucceeds](https://github.com/keybase/kbfs/blob/88a6091f01c37aff8d051c9875cd9c9151632a11/libkbfs/kbfs_ops_concur_test.go#L1230), which cancels a context after the Put successfully hits the mdserver. If you copy that test and change it to do an O_EXCL Create instead of a Sync, and make sure the Create returns a `nil` error in that case, that would exercise the code path we care about pretty well, I think.

songgao:
@strib sounds good; will do! Thanks for the pointers :)

libkbfs/kbfs_ops_concur_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestKBFSOpsMultiBlockWriteDuringRetriedSync(t *testing.T) {
 	}
 }
 
+// This tests the situation where cancellation happens when MD writes has
```
strib: "MD writes has started" -> "the MD write has already started"

libkbfs/kbfs_ops_concur_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestKBFSOpsMultiBlockWriteDuringRetriedSync(t *testing.T) {
...
+	close(putUnstallCh)
+
+	// We expect no canceled error
+	err = <-errChan
+	if err != nil {
+		t.Fatalf("Create returned error: %v", err)
+	}
+
+	// SyncFromServerForTesting and make sure it worked.
+	err = kbfsOps.SyncFromServerForTesting(ctx, rootNode.GetFolderBranch())
```
strib: It's a bit weird to be using the cancelled `ctx` here.  Also, this shouldn't be necessary because the previous operation succeeded, so I would just get rid of this call.

libkbfs/kbfs_ops_concur_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestKBFSOpsMultiBlockWriteDuringRetriedSync(t *testing.T) {
...
+		t.Fatalf("Couldn't sync from server: %v", err)
+	}
+
+	if _, _, err = kbfsOps.Lookup(
+		BackgroundContextWithCancellationDelayer(), rootNode, "a"); err != nil {
+		t.Fatalf("Lookup returned error: %v", err)
+	}
+}
+
+// This tests the situation where cancellation happens when MD writes has
```
strib: Same here.

strib:
New tests look good to me, mod one comment.  Thanks!

Looks like you have a compile error though, due to your master merge.

songgao:
Thanks Jeremy! Just fixed the merge error (hopefully) and made changes per review comments

PR #174, opened 2016-07-19
keybase_daemon_rpc: use KeyFamilyChanged instead of UserChanged

strib:
Issue: KBFS-1238

### DIFF
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -52,7 +52,7 @@ type KeybaseDaemonRPC struct {
 
 var _ keybase1.NotifySessionInterface = (*KeybaseDaemonRPC)(nil)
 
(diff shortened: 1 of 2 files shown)


strib:
@patrickxb does this look right?  In terms of clearing out cached `keybase1.UserPlusKeys` (and unverified keys) state, `KeyfamilyChanged` should work as a drop-in replacement for `UserChanged`, right?

patrickxb:
Yes, looks right to me.  Will this also trigger a rekey check?

strib:
@patrickxb: Yep, see the whole function here: https://github.com/keybase/kbfs/blob/b62fc0ea9d78c83e824aee864a7b05b8f80d7507/libkbfs/keybase_daemon_rpc.go#L261-L275.

patrickxb:
Great, LGTM.

PR #176, opened 2016-07-19
[KBFS-1293] wait for block deletion in parallel test

songgao:
This should fix https://ci.appveyor.com/project/keybase/kbfs/build/1.0.4659

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -707,6 +707,10 @@ func (c *ConfigLocal) Shutdown() error {
 				if err := fbo.fbm.waitForArchives(context.Background()); err != nil {
 					return err
 				}
+				if err := fbo.fbm.waitForDeletingBlocks(
+					context.Background()); err != nil {
+					return err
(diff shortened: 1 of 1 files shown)


strib:
@songgao the test shutdown method already waits on SyncFromServer for each user, before running the consistency checks, so I don't think this will fix the issue.

strib:
Oh crap I bet it needs to be added here: https://github.com/keybase/kbfs/blob/master/libkbfs/config_local.go#L707

That's where we make sure all users have finished archiving before we turn off the block server.  Because by the time we call config.Shutdown for the second user, the block server could already be shut down.  D'oh!  Sorry for not catching this before.

songgao:
Ah that make sense! Thanks @strib!

songgao:
@strib Here it is. I also made sure `waitForDeletingBlocks` is called whenever `waitForArchives` is called

strib:
Looks good to me, thanks!

PR #180, opened 2016-07-20
Retry batchDowngrade in bserver_remote using exponential backoff

jinyangli:
The bug of KBFS-1287 is due to downgrade RPC not retried (when there's only one block nonce in the batch request). I cleaned up the retry logic and am now doing proper exponential backup on throttle error.

I also need to fix a few checks in bserver_test to make sure retried RPCs still result in test passing.
see https://github.com/keybase/kbfs-server/pull/73

### DIFF
--- a/libkbfs/bserver_remote.go
+++ b/libkbfs/bserver_remote.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/logger"
 	keybase1 "github.com/keybase/client/go/protocol"
@@ -339,12 +340,12 @@ func (b *BlockServerRemote) ArchiveBlockReferences(ctx context.Context,
 func (b *BlockServerRemote) batchDowngradeReferences(ctx context.Context,
 	tlfID TlfID, contexts map[BlockID][]BlockContext, archive bool) (
 	doneRefs map[BlockID]map[BlockRefNonce]int, finalError error) {
-	tries := 0
 	doneRefs = make(map[BlockID]map[BlockRefNonce]int)
 	notDone := b.getNotDone(contexts, doneRefs)
-	var res keybase1.DowngradeReferenceRes
-	var err error
-	for len(notDone) > 0 {
+
+	throttleErr := backoff.Retry(func() error {
+		var res keybase1.DowngradeReferenceRes
+		var err error
 		if archive {
 			res, err = b.client.ArchiveReferenceWithCount(ctx, keybase1.ArchiveReferenceWithCountArg{
 				Refs:   notDone,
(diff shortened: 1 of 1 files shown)


libkbfs/bserver_remote.go
```diff
@@ -405,25 +380,39 @@ func (b *BlockServerRemote) batchDowngradeReferences(ctx context.Context,
...
+		if err != nil {
+			// if error is of type throttle, retry
+			if _, ok := err.(BServerErrorThrottle); ok {
+				return err
+			}
+			// non-throttle error, do not retry here
+			finalError = err
+		}
+		return nil
+	}, backoff.NewExponentialBackOff())
```
strib: Can we put a `MaxElapsedTime` on this, to make sure we don't loop forever on systems where there is no context timeout?

strib: Maybe 30s?

libkbfs/bserver_remote.go
```diff
@@ -405,25 +380,39 @@ func (b *BlockServerRemote) batchDowngradeReferences(ctx context.Context,
...
+		// check whether to backoff and retry
+		if err != nil {
+			// if error is of type throttle, retry
+			if _, ok := err.(BServerErrorThrottle); ok {
+				return err
+			}
+			// non-throttle error, do not retry here
+			finalError = err
+		}
+		return nil
```
strib: Are we guaranteed that the server will only return a `nil` when the completed list matches the `Refs` list exactly? If not, there could be a bug in returning early before everything is complete.  If so, maybe that's worth double checking here that `len(notDone) == 0` when `err == nil`, just to be safe?

strib: I see, you're checking it down below actually.  It might make more sense here in the loop.

libkbfs/bserver_remote.go
```diff
@@ -405,25 +380,39 @@ func (b *BlockServerRemote) batchDowngradeReferences(ctx context.Context,
...
+	}, backoff.NewExponentialBackOff())
+
+	// if backoff has given up retrying, return error
+	if throttleErr != nil {
+		return doneRefs, throttleErr
```
strib: This is not a change in behavior but I notice that the interface function comments for `RemoveBlockReference` and `ArchiveBlockReferences` don't mention that `liveCounts` can be non-nil when there's a fatal throttle error.  Can you please add that to the comments (and in `BlockOps` too) so callers will know they are getting some information back about previously-completed call?

strib:
Ok, I think I see.  I guess @akalin-keybase and I introduced the bug when we introduced the `makeProgress` stuff.  We assumed it wasn't possible to get a throttle message with no completed references, but I guess that's not the case.  Sorry!

I agree this solves the infinite loop issue (KBFS-1234) in a better way with exponential backoff.  So, mod comments, this looks good to me, thanks!

PR #186, opened 2016-07-21
mdserver: calculate the time offset from the local md server

strib:
This will let us figure out how "old" a server-timestamped MD update is.

I expect the docker tests to fail until the corresponding server update is merged.

Depends on keybase/client#3569.

Issue: KBFS-1312


aalness:
LGTM. I wonder if it's worth having some sanity check to throw out times where the RTT is insane?

strib:
Good idea, I'll avoid setting the offset time if one is already set, and the ping time was > 5 seconds.

PR #190, opened 2016-07-25
dokan*: Change default mount flags

taruti:
Some antivirus products behave better with these flags.


zanderz:
LGTM :thumbsup:, but what driver version is this supposed to match?

taruti:
This is as good or better with all driver versions. I did most of my latest testing with https://github.com/marinkobabic/dokany/tree/FixBsodMounting which appears to be excellent with the latest updates.

PR #197, opened 2016-07-25
Taruti/kbfs 1215 dokan stabler api

taruti:
* Semi-automatic reorganization to move context.Context into dokan from libdokan
* Make dokan API more finalized
* Allow compiling libdokan on *nix


libdokan/dummy.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
```
zanderz: Shouldn't this be named something like "new_folder_name_nix.go"?

zanderz:
LGTM :thumbsup:

PR #199, opened 2016-07-27
Volume icon for KBFS mount (macOS)

gabriel:
We specify 2 special files in the root dir:

- `.VolumeIcon.icns`: The icon file to use
- `._.`: Extended attributes for current (.) dir, which tell Finder to look for volume icon


gabriel:
2016-07-27 at 3 03 pm

gabriel:
Including github.com/kardianos/osext vendor. We use this to find the icon file in the app bundle.

gabriel:
BTW, this code is also useful in the future for supporting custom icons for folders, etc. Since that will also use special external files and finder extended attributes :)

libfs/constants.go
```diff
@@ -50,3 +50,12 @@ const FlushJournalFileName = ".kbfs_flush_journal"
 // DisableJournalFileName is the name of the journal-disabling
 // file. It can be reached anywhere within a top-level folder.
 const DisableJournalFileName = ".kbfs_disable_journal"
+
```
akalin-keybase: `libfs` is for code that's shared between dokan and libfuse, so I think these macOS-specific constants belong more in libfuse.

libfuse/special_xattr.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"time"
+)
+
+// finderInfoExtendedAttribute is for extended attributes for the Finder
```
akalin-keybase: why embed into the executable rather than having this as an external file like the volume icon? My main worry is that the gzipped bytes aren't that easy to inspect and verify, and a naive reader of the code might think something nefarious is going on.

libfuse/fs.go
```diff
@@ -257,3 +274,26 @@ func (r *Root) ReadDirAll(ctx context.Context) (res []fuse.Dirent, err error) {
 	}
 	return res, nil
 }
+
+func (r *Root) log() logger.Logger {
+	return r.private.fs.log
+}
+
+var loadVolIconOnce sync.Once
```
akalin-keybase: no need for this to be a global. It can live inside `Root`. Or better yet, this can be made part of `newExternalFile` (see comment above)

libfuse/external_file.go
```diff
@@ -0,0 +1,55 @@
...
+func newByteFile(data []byte, t time.Time) *SpecialReadFile {
+	return &SpecialReadFile{
+		read: func(context.Context) ([]byte, time.Time, error) {
+			return data, t, nil
+		},
+	}
+}
+
+func newExternalFile(path string) (*SpecialReadFile, error) {
+	if path == "" {
```
akalin-keybase: You can move the `load file only once` logic in here, with something like:

```
var once sync.Once;
var data []byte;
var fileTime ...;
return &SpecialReadFile{
  read: func(...) ... {
    var err error;
    once.Do(func() {
      data, err = ioutil.ReadFile(path);
      fileTime = ...;
      
    })
    return data, fileTime, err;
  }
}
```

libfuse/special_xattr.go
```diff
@@ -0,0 +1,62 @@
...
+// on macOS/darwin. It is here as gzipped bytes so we can embed in the
+// go binary. For more details on extended attributes for the Finder,
+// see http://gotofritz.net/blog/geekery/os-x-extended-attibutes/.
+var finderInfoExtendedAttribute = []byte("\x1f\x8b\x08\x00\x00\x09\x6e\x88\x00\xff\x62\x60\x15\x63\x67\x60\x62\x60\xf0\x4d\x4c\x56\xf0\x0f\x56\x88\x50\x80\x02\x90\x18\x03\x27\x10\x1b\x31\x30\xf0\x6d\x00\xd2\x40\x3e\xdf\x23\x06\x06\x46\x39\x06\x28\x60\x61\xc0\x05\x1c\x43\x42\x82\x5e\xef\x7d\xa9\x0f\xd1\xc1\x50\x81\x53\xe1\x28\x18\x05\xa3\x60\x14\x8c\x82\x51\x30\x0a\x46\xc1\x28\x18\x05\xa3\x60\x14\x8c\x82\x51\x40\x45\xc0\x08\xc5\x60\x20\x17\x92\x91\x59\xac\x50\x94\x5a\x9c\x5f\x5a\x94\x9c\xaa\x90\x96\x5f\x94\xad\x90\x99\x57\x92\x9a\x57\x92\x99\x9f\x97\x98\x93\x53\xa9\x90\x93\x9a\x56\xa2\x90\x94\x93\x98\x97\x0d\xea\x07\x0f\x03\x80\xea\x7f\xb8\xb0\x0c\x83\xdc\xff\xff\x80\x00\x00\x00\xff\xff\x37\x65\x97\xf3\x00\x10\x00\x00")
+
+var finderInfoExtendedAttributeRaw []byte
```
akalin-keybase: these global variables are racy, since FUSE requests run on their on goroutines. You can do something similar to my suggestion for `newExternalFile` instead, if you stick with embedding this file.

libfuse/fs.go
```diff
@@ -257,3 +274,26 @@ func (r *Root) ReadDirAll(ctx context.Context) (res []fuse.Dirent, err error) {
 	}
 	return res, nil
 }
+
+func (r *Root) log() logger.Logger {
```
akalin-keybase: may as well make the other functions for `Root` use this helper function

libfuse/fs.go
```diff
@@ -221,6 +223,21 @@ func (r *Root) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.L
 		return &SpecialReadFile{r.private.fs.remoteStatus.NewSpecialReadFunc}, nil
 	}
 
+	// Special files for macOS/darwin
```
akalin-keybase: Can you move all the platform-specific code to files like `platform_(something_osx.go`, and replace `volIcon` with a `PlatformRoot` type? I think that'll be cleaner.

akalin-keybase:
First pass, some suggestions.

gabriel:
Great feedback. I think I fixed everything you mentioned.

libfuse/constants_darwin.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
```
akalin-keybase: may as well make this +build darwin, so we can catch errors where we try to use a darwin-only constant in a non-darwin file

libfuse/fs_darwin.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
```
akalin-keybase: need +build darwin here

libfuse/external_file.go
```diff
@@ -0,0 +1,63 @@
...
+				}
+				fileTime = info.ModTime()
+				data, err = ioutil.ReadFile(path)
+			})
+			return data, fileTime, err
+		},
+	}, nil
+}
+
+func newExternalBundleResourceFile(path string) (*SpecialReadFile, error) {
```
akalin-keybase: this function and the one below are darwin-only, so probably belongs better in fs_darwin.go

libfuse/fs_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+package libfuse
+
+import (
+	"bazil.org/fuse"
+	"bazil.org/fuse/fs"
+	"golang.org/x/net/context"
+)
+
+func (r *Root) platformLookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.LookupResponse) fs.Node {
```
akalin-keybase: this should probably return `(fs.Node, err)`, so that any error in loading the external file will be propagated up.

libfuse/fs_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+	switch req.Name {
+	case VolIconFileName:
+		volIcon, err := newExternalBundleResourceFile("KeybaseFolder.icns")
+		if err != nil {
+			r.log().CErrorf(ctx, "Error getting bundle resource path: %s", err)
+		} else if volIcon != nil {
```
akalin-keybase: I think you can just do `return volIcon, nil` here. Since these files never change, we can leave `EntryValid` to the default, I think. (setting it to 0 is only for special files that can be updated at any time)

libfuse/fs_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+			resp.EntryValid = 0
+			return volIcon
+		}
+	case ExtendedAttributeSelfFileName:
+		xattrSelf, err := newExternalBundleResourceFile("ExtendedAttributeFinderInfo.bin")
+		if err != nil {
+			r.log().CErrorf(ctx, "Error getting bundle resource path: %s", err)
```
akalin-keybase: bundle resource path -> extended attributes file

libfuse/fs_darwin.go
```diff
@@ -0,0 +1,33 @@
...
+	case VolIconFileName:
+		volIcon, err := newExternalBundleResourceFile("KeybaseFolder.icns")
+		if err != nil {
+			r.log().CErrorf(ctx, "Error getting bundle resource path: %s", err)
+		} else if volIcon != nil {
+			resp.EntryValid = 0
+			return volIcon
+		}
+	case ExtendedAttributeSelfFileName:
+		xattrSelf, err := newExternalBundleResourceFile("ExtendedAttributeFinderInfo.bin")
```
akalin-keybase: same deal as above here

akalin-keybase:
few more comments

libfuse/constants_darwin.go
gabriel: if you name a file `_darwin.go` the build line is implied..
http://blog.ralch.com/tutorial/golang-conditional-compilation/#file-suffixes:0765dbe650f7259dfb0634d2c3467e31

akalin-keybase: Ah, forgot about that! My mistake.

libfuse/fs_darwin.go
gabriel: The error isn't really fatal and I don't want a missing external file to cause a real error.
But the platformLookup method should match the Lookup method signature.
And I'm changing those errors to Warning log messages, so it's clear they aren't fatal, and will add a comment.

libfuse/external_file.go
gabriel: Sure, moved.

libfuse/fs_darwin.go
gabriel: Filename implies +build darwin

libfuse/fs_darwin.go
gabriel: Actually nevermind, that erroring would be really weird so it's ok to propagate.

libfuse/fs_darwin.go
gabriel: ok, yeah EntryValid was confusing to me... now am just returning newExternalBundleResourceFile result directly.

gabriel:
OK I think I got everything fixed you mentioned (except +build lines aren't required).

A comment about lookup... it seems error prone to have fuse.LookupResponse mutated at any point in lookup code path, it seems we may want to refactor that. Also EntryValid is really confusing but I guess that's a bazil thing.

akalin-keybase:
lgtm! Yeah, I think having the responses be output pointer variables is due to bazil's interface. I'm guessing it's so that it can inherit defaults? @tv42 would know!

tv42:
A few reasons.

- historical: it's what I inherited from rsc
- defaults: some of the responses have prepopulated fields, while still allowing the handler to override them
- existing style: that's how e.g. net/rpc does it
- performance: this lets the framework be in control of when a response is allocated, and decrease gc work

PR #201, opened 2016-07-28
libdokan: Support PRIVATE/PUBLIC for case insensitive link destinations

taruti:
This should fix https://github.com/keybase/keybase-issues/issues/2248

### DIFF
--- a/kbfsdokan/README.md
+++ b/kbfsdokan/README.md
@@ -69,3 +69,25 @@ dir
 mkdir foo
 notepad bar.txt
 ```
+
+# Symlink destinations from outside KBFS to KBFS
+
+Windows makes paths case insentive in symlink destination. KBFS has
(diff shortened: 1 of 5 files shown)


zanderz:
LGTM, pending CI :thumbsup:
- The capitalized case ("Public") is separate from this, right?
- I learned recently that junctions don't require admin (mklink /j) - did you test this one?

taruti:
Tested only `mklink /d`, will test also `/j`. Yes, `Public` would be different.

The all caps version is used  because if the link is to the root (`K:\`) then if case insensitive lookups are used `public` will be translated to `PUBLIC` in requests to the fs.

taruti:
Junctions require all participants to be local (NTFS) volumes. So no `/J` for us - `/D` still works.

PR #215, opened 2016-08-04
Crypto client for KBFS init

gabriel:
Allows us to specify a engine based (non-rpc) crypto provider for mobile, directly in the client.

See https://github.com/keybase/client/pull/3723

- Breaks up `CryptoClient` into `CryptoClient` and `CryptoClientRPC`
- Change `KeybaseServiceFn` constructor function into a `KeybaseServiceCn` "factory-style" interface
- Include `NewCrypto` method in this interface so we can supply our own non-rpc crypto provider

There should be no change in behavior, should be just some shuffling, separation of code/interfaces.


gabriel:
@maxtaco @akalin-keybase @strib

libkbfs/crypto_client.go
```diff
@@ -10,110 +10,54 @@ import (
...
-	rpc "github.com/keybase/go-framed-msgpack-rpc"
 	"golang.org/x/net/context"
 )
 
-// CryptoClient implements the Crypto interface by sending RPCs to the
-// keybase daemon to perform signatures using the device's current
-// signing key.
+// CryptoClientProvider should match keybase1.CryptoClient for signing
+// and unboxing.
+type CryptoClientProvider interface {
```
strib: Please put this in `interfaces.go`.

strib: Wait, this is literally identical to `keybase1.CryptoClient`.  Why are we redefining it?  Are you trying to avoid importing the protcol files somewhere?

strib: Ah yes, I meant `keybase1.CryptoInterface`.

gabriel: Ahhh thats what I was looking for, ok yeah will switch to that.

gabriel: K fixed.. yeah I was looking at the struct and totally forgot that protocol makes corresponding interfaces too

libkbfs/keybase_daemon.go
```diff
@@ -13,10 +13,10 @@ import (
 	"github.com/keybase/client/go/logger"
 )
 
-// KeybaseServiceFn defines a constructor for a KeybaseService
-type KeybaseServiceFn func(config Config, params InitParams, ctx Context, log logger.Logger) (KeybaseService, error)
+// keybaseDaemonCn is the default
```
strib: Incomplete comment?  Also, what does `Cn` stand for?

libkbfs/keybase_service.go
```diff
@@ -0,0 +1,14 @@
...
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import "github.com/keybase/client/go/logger"
+
+// KeybaseServiceCn defines methods needed to construct KeybaseService
+// and Crypto implementations.
+type KeybaseServiceCn interface {
```
strib: This can go in `interfaces.go`.

strib:
Looks good overall, just a few comments, thanks.

libkbfs/keybase_daemon.go
gabriel: Changed to `keybaseDaemon`. I think I was going for "Constructor functions" like how golang puts `Fn` at the end of types that are functions. Mostly cause I didn't want to use "Factory" or "Builder".

gabriel:
ok, I think ready now

strib:
Looks good to me.

PR #224, opened 2016-08-09
tlf_edit_history: update cached history in response to new MD updates

strib:
This PR adds a background processor to `TlfEditHistory` that processes batches of `RootMetadata`s and updates the cached edit history in response, if one has already been cached.  It adds edits for newly-created and newly-edited files, and it also removes deleted files and renames moved files.

Issue: KBFS-912

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -3249,6 +3249,7 @@ func (fbo *folderBranchOps) notifyBatchLocked(
 
 	lastOp := md.data.Changes.Ops[len(md.data.Changes.Ops)-1]
 	fbo.notifyOneOpLocked(ctx, lState, lastOp, md)
+	fbo.editHistory.UpdateHistory(ctx, []ImmutableRootMetadata{md})
 }
 
 // searchForNode tries to figure out the path to the given
@@ -3550,6 +3551,7 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 			fbo.notifyOneOpLocked(ctx, lState, op, rmd)
 		}
 	}
+	fbo.editHistory.UpdateHistory(ctx, rmds)
 	return nil
 }
 
@@ -3609,6 +3611,7 @@ func (fbo *folderBranchOps) undoMDUpdatesLocked(ctx context.Context,
 			fbo.notifyOneOpLocked(ctx, lState, io, rmd)
 		}
 	}
+	// TODO: update the edit history?
 	return nil
 }
 
@@ -4115,6 +4118,9 @@ func (fbo *folderBranchOps) SyncFromServerForTesting(
 	if err := fbo.fbm.waitForDeletingBlocks(ctx); err != nil {
 		return err
 	}
+	if err := fbo.editHistory.Wait(ctx); err != nil {
+		return err
+	}
 	return fbo.fbm.waitForQuotaReclamations(ctx)
 }
 
@@ -4400,6 +4406,7 @@ func (fbo *folderBranchOps) finalizeResolutionLocked(ctx context.Context,
 	for _, op := range newOps {
 		fbo.notifyOneOpLocked(ctx, lState, op, irmd)
 	}
+	fbo.editHistory.UpdateHistory(ctx, []ImmutableRootMetadata{irmd})
 	return nil
 }
 
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -284,7 +284,7 @@ type folderBranchOps struct {
 	// Protected by mdWriterLock
 	rekeyWithPromptTimer *time.Timer
 
-	editHistory TlfEditHistory
+	editHistory *TlfEditHistory
 }
 
 var _ KBFSOps = (*folderBranchOps)(nil)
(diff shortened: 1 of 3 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -3612,6 +3611,7 @@ func (fbo *folderBranchOps) undoMDUpdatesLocked(ctx context.Context,
 			fbo.notifyOneOpLocked(ctx, lState, io, rmd)
 		}
 	}
+	// TODO: update the edit history?
```
jzila: if it's being undone, why would we need to update the history?

strib: @jzila: You might be undoing file edits/creates, which have already been added to the history.  So in theory those need to be undone in the history as you undo them on the file system.

libkbfs/tlf_edit_history.go
```diff
@@ -77,6 +77,37 @@ func (we TlfWriterEdits) isComplete() bool {
 	return true
 }
 
+// updateOldEdits removes edits from `we` belonging to files that have
+// been removed, and renames ones that have been renamed.
+func (we *TlfWriterEdits) updateOldEdits(removed map[string]bool,
```
jzila: Maps are implicitly a pointer receiver:
https://play.golang.org/p/-DWM2d2wgj

libkbfs/tlf_edit_history.go
```diff
@@ -348,26 +434,168 @@ func (teh *TlfEditHistory) GetComplete(ctx context.Context,
...
-		sort.Sort(list)
-		if len(list) > desiredEditsPerWriter {
-			list = list[len(list)-desiredEditsPerWriter:]
+	teh.setEdits(ctx, currEdits, rmds)
+	return teh.getEditsCopyLocked(), nil
+}
+
+func (teh *TlfEditHistory) updateHistory(ctx context.Context,
+	rmds []ImmutableRootMetadata) error {
+	defer teh.wg.Done()
```
jzila: I'm having trouble finding the corresponding `teh.wg.Wait`

strib: See `TlfEditHistory.Wait()`.

libkbfs/tlf_edit_history.go
```diff
@@ -348,26 +434,168 @@ func (teh *TlfEditHistory) GetComplete(ctx context.Context,
...
+	removed := make(map[string]bool)
+	// TODO: can I used chains.deletedOriginals instead?  It's hard to
+	// get the full path that way.
+	for _, chain := range chains.byOriginal {
+		for _, op := range chain.ops {
+			rop, ok := op.(*rmOp)
+			if !ok {
+				continue
+			}
+			path := rop.getFinalPath().String() + "/" + rop.OldName
```
jzila: Might be useful to have an `rop.getOldPath()` method here

jzila: Or `getPathWithSuffix(suffix string)` based on later calls

libkbfs/tlf_edit_history.go
```diff
@@ -348,26 +434,168 @@ func (teh *TlfEditHistory) GetComplete(ctx context.Context,
...
+		}
+		newPath := newParentChain.ops[0].getFinalPath().String() + "/" +
+			ri.newName
+
+		// Ignore any previous rmOps.
+		delete(removed, oldPath)
+		// If a file was overwritten, ignore all the old edits.
+		removed[newPath] = true
+		// Rename the file.
+		renamed[oldPath] = newPath
```
jzila: What happens if you have multiple renames for the same file in one edit batch?

libkbfs/tlf_edit_history.go
```diff
@@ -348,26 +434,168 @@ func (teh *TlfEditHistory) GetComplete(ctx context.Context,
...
+	return nil
+}
+
+func (teh *TlfEditHistory) process() {
+	var ctx context.Context
+	func() {
+		teh.lock.Lock()
+		defer teh.lock.Unlock()
+		ctx, teh.cancel = context.WithCancel(context.Background())
+	}()
```
jzila: I've noticed this pattern a few times for code clarity. Is this kind of closure cost-free? Would it be better to lock, do something, then unlock in the method scope itself?

libkbfs/tlf_edit_history.go
```diff
@@ -124,21 +155,57 @@ type TlfEditHistory struct {
...
+}
+
+// NewTlfEditHistory makes a new TLF edit history.
+func NewTlfEditHistory(config Config, fbo *folderBranchOps,
+	log logger.Logger) *TlfEditHistory {
+	teh := &TlfEditHistory{
+		config:   config,
+		fbo:      fbo,
+		log:      log,
+		rmdsChan: make(chan []ImmutableRootMetadata, 100),
```
jzila: If you create `cancel` here then you don't need to lock it for initialization in `process`, and you don't have to nil check it in `Shutdown`

jzila:
First pass done, mostly looks good. A few bits and questions.

libkbfs/tlf_edit_history.go
strib: We already had a `path.ChildPathNoPtr()` call, so using that instead.

libkbfs/tlf_edit_history.go
strib: `crChains` takes care of that magically for you, by eliminating all the interior renames.  So `chains.renamedOriginals` will only have the original start and final end name for any given file.

libkbfs/tlf_edit_history.go
strib: It's not just for clarity -- if `context.WithCancel` hit a panic, we would still be able to unlock the lock using the closure, just in case the caller recovered from the panic.  So I always use a closure unless the thing being locked over is very simple.

I've been assuming that the compiler can make this cost free if it was smart enough, so even if it's not cost-free today it could be some day.  (Haven't measure it though.)

libkbfs/tlf_edit_history.go
strib: I can do this without storing `ctx`; I can just pass it directly to `process`.  Will do.

strib:
Implemented your suggestions, ready for another look, thanks!

jzila:
LGTM

PR #226, opened 2016-08-09
export TLF crypt keys through RPC

songgao:


maxtaco:
Does this create the TLF if it doesn't already exist?

songgao:
TLFs exist implicitly so this would work if the TLF has never been accessed before. Although it doesn't cause it to be added to favorites if it's not favorited. Should it?

maxtaco:
ok, great.  is this true even for sharing before signup?  if so this is great news, one fewer moving part.

maxtaco:
(no to the favorites issue)

songgao:
Just tried "sharing before signup", and it seems to be working fine! Although thinking about it more, do we maybe also need a addToFavorite RPC, since we'll have a "favorite" list in chat as well?

maxtaco:
I'm going to defer to @strib on the rest of this review!

libkbfs/interfaces.go
```diff
@@ -107,6 +107,12 @@ type KBFSOps interface {
 	// isn't favorited.
 	DeleteFavorite(ctx context.Context, fav Favorite) error
 
+	// GetOrInitializeNewMDMaster gets the existing MD or initialize a new MD for
+	// master branch
+	GetOrInitializeNewMDMaster(ctx context.Context, mdops MDOps, h *TlfHandle) (
```
strib: Previously we worked to remove all references from metadata from the `KBFSOps` interface -- we don't really want outside libraries to know about that concept.  So ideally neither `ImmutableRootMetadata` or the term "MD" would appear in this interface.

I can think of two possible alternatives:
* Define some new interface containing just this method, make it available on `Config`, and set it to the same `*KBFSOpsStandard` instance whenever a `Config` is made.
* When `KeybaseDaemonRPC` is constructed, pass it a function that calls that method on `KBFSOpsStandard`.  (And maybe don't capitalize the method name.)

What do you think?  Other options?  I slightly prefer the second option.

strib: Actually a third option would be making a method on the `KBFSOps` interface that just returns the keys as byte slices.  That might be useful outside this library.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -305,3 +306,46 @@ func (k *KeybaseDaemonRPC) Shutdown() {
...
+	res.CanonicalName = keybase1.CanonicalTlfName(tlfHandle.GetCanonicalName())
+
+	var rmd ImmutableRootMetadata
+	var id TlfID
+	_, rmd, id, err = k.config.KBFSOps().GetOrInitializeNewMDMaster(
+		ctx, k.config.MDOps(), tlfHandle)
+	if err != nil {
+		return res, err
+	}
+	res.TlfID = keybase1.TLFID(id.String())
```
strib: Why do we need to return the TLF ID?  (Sorry, I'm missing some context.)  If the caller can get this some other way (say from the API call that returns the favorites), then we could take that as an input parameter instead of a name, and avoid the TLF handle all together.

libkbfs/interfaces.go
songgao: I like the third option; will go with that approach. Thanks!

libkbfs/key_manager.go
```diff
@@ -53,6 +53,21 @@ func (km *KeyManagerStandard) GetTLFCryptKeyForBlockDecryption(
 	return km.getTLFCryptKeyUsingCurrentDevice(ctx, kmd, blockPtr.KeyGen, true)
 }
 
+// GetTLFCryptKeyFromAllGenerations implements the KeyManager interface for
+// KeyManagerStandard.
+func (km *KeyManagerStandard) GetTLFCryptKeyFromAllGenerations(
+	ctx context.Context, kmd KeyMetadata) (keys []TLFCryptKey, err error) {
+	for g := KeyGen(FirstValidKeyGen); g <= kmd.LatestKeyGeneration(); g++ {
+		var key TLFCryptKey
+		key, err = km.getTLFCryptKeyUsingCurrentDevice(ctx, kmd, g, false)
```
strib: I _think_ it's safe to set the `cache` parameter to `true` here.  We don't have it set to `true` in `GetTLFCryptKeyForEncryption` because there used to be some danger that `Rekey` would call that method before the MD was finalized.  But when getting all the key generations on a finalized MD, it seems safe to cache them.

libkbfs/interfaces.go
```diff
@@ -491,6 +497,12 @@ type KeyManager interface {
 	GetTLFCryptKeyForBlockDecryption(ctx context.Context, kmd KeyMetadata,
 		blockPtr BlockPointer) (TLFCryptKey, error)
 
+	// GetTLFCryptKeyFromAllGenerations gets the crypt keys of all generations
+	// for current devices. keys contains crypt keys from all generations, in
+	// order, starting from FirstValidKeyGen.
+	GetTLFCryptKeyFromAllGenerations(ctx context.Context, kmd KeyMetadata) (
```
strib: I would do either s/From/For/ or s/From/Of/.

libkbfs/keybase_daemon_rpc.go
songgao: I agree we should handle this way (using TlfID as input). But we don't have an RPC call for getting TlfID yet. Maybe I'll add a ticket for it?

libkbfs/keybase_daemon_rpc.go
```diff
@@ -305,3 +306,46 @@ func (k *KeybaseDaemonRPC) Shutdown() {
...
+
+	var keys []TLFCryptKey
+	keys, err = k.config.KeyManager().GetTLFCryptKeyFromAllGenerations(ctx, rmd)
+	if err != nil {
+		return res, err
+	}
+
+	res.FirstValidKeyGen = FirstValidKeyGen
```
strib: It might be more futureproof to return a map of keygen->key instead of implicitly relying on key generation numbers starting at 1 and increasing by 1 every time?  I don't feel too strongly though.

libkbfs/keybase_daemon_rpc.go
strib: Actually on second thought maybe getting it by name is better, since it means the caller can run Identify calls to make sure the participants on the folder are the same users they are trying to chat with.

But then I'm still wondering why the caller needs to know the ID.

libkbfs/keybase_daemon_rpc.go
songgao: I thought about using a slice of `{generation, key}` struct, but decided to use to save a few bytes. It's probably not worth it. Are you OK with using struct slice here (so that keys are sorted by generation) as well?

libkbfs/keybase_daemon_rpc.go
songgao: Perhaps other calls can use TlfID for other purposes? I was thinking maybe a RPC call like `GetTlfID(tlfName string)`. The caller can run Identify before this call as well?

libkbfs/keybase_daemon_rpc.go
strib: Fine with me.  I guess you want to avoid `map<bytes>` so that the key isn't a string?

libkbfs/keybase_daemon_rpc.go
strib: I meant specifically why is it being returned with this RPC call?  If you have some other use for it, adding a `GetTlfID` call makes sense. I don't care much one way or the other, I'm just wondering what it will be used for.

strib:
Finished a pass.

libkbfs/keybase_daemon_rpc.go
songgao: I don't have an idea yet about how it'd be used; sorry! But I also wasn't sure it's not useful at all. So returning the TlfID in this RPC call is to have it (and not blocking other tickets) when it's needed, but make it trivial to remove if it turns out not useful. My understanding is if it becomes widely used in chat, this would be split into a separate `GetTlfID` call.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -305,3 +306,43 @@ func (k *KeybaseDaemonRPC) Shutdown() {
 		k.shutdownFn()
 	}
 }
+
+// GetTLFCryptKeys implements the TlfKeysInterface interface for
+// KeybaseDaemonRPC.
+func (k *KeybaseDaemonRPC) GetTLFCryptKeys(
+	ctx context.Context, tlfName string) (res keybase1.TLFCryptKeys, err error) {
+
```
strib: Extra blank line.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -328,24 +329,20 @@ getHandle:
 
 	res.CanonicalName = keybase1.CanonicalTlfName(tlfHandle.GetCanonicalName())
 
-	var rmd ImmutableRootMetadata
+	var keys []TLFCryptKey
```
strib: Why do you need to pre-declare `keys` and `id`?

libkbfs/interfaces.go
```diff
@@ -107,11 +107,10 @@ type KBFSOps interface {
 	// isn't favorited.
 	DeleteFavorite(ctx context.Context, fav Favorite) error
 
-	// GetOrInitializeNewMDMaster gets the existing MD or initialize a new MD for
-	// master branch
-	GetOrInitializeNewMDMaster(ctx context.Context, mdops MDOps, h *TlfHandle) (
-		initialized bool, md ImmutableRootMetadata,
-		id TlfID, err error)
+	// GetTLFCryptKeys gets crypt key of all generations as well as
+	// TLF ID for tlfHandle.
```
strib: Mention that the keys are ordered, starting with the key for `FirstValidKeyGen`.

strib:

songgao:
Thanks @strib !

PR #227, opened 2016-08-09
folder_block_ops: fix missing block when switching to indirect

strib:
If a series of writes exactly fills up a one-block file, and then a
new write comes in, we transition to an file with a top-level indirect
block, that has two direct file blocks under it -- the original direct
file block, plus a new empty one that receives the write.  However, we
were failing to mark this new block as dirty, because in this case
nCopied == 0 on the first write attempt (because the original block
was full).

In this case, we should dirty that first block, since it needs to be
re-uploaded under a new BlockPointer.

Noticed this while running our benchmark tests.

Issue: KBFS-1364

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -1315,6 +1315,7 @@ func (fbo *folderBlockOps) writeDataLocked(
 		// generalize, just annoying
 
 		// if we need another block but there are no more, then make one
+		switchToIndirect := false
 		if nCopied < n && nextBlockOff < 0 {
 			// If the block doesn't already have a parent block, make one.
 			if ptr == file.tailPointer() {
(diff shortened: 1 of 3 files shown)


jzila:
Think value to creating a corner case test for this?

jzila:
Lgtm other than a potential test

strib:
Well `BenchmarkWriteSeq512` and friends already cover it, which is how I found it.  We just don't run those much.  Not sure it's worth building another one but I certainly can if you think it's important!

jzila:
Ooh can you factor out the test part of the benchmark, so that the test is always run?

strib:
Ok, made a simple test.  Didn't bother refactoring the benchmark since it uses way more data than it needs to.  Verified the new test fails on master with the same error I saw in the benchmark.

strib:
(Will push when CI passes.)

jzila:
Perfect thx

PR #229, opened 2016-08-10
dirty_bcache: don't increase buffer size on small syncs

strib:
If the sync was smaller than the current buffer size, it doesn't
really convey any info about how much data the connection can handle.

Add a benchmark that mixes big and small files, which fails without
this change.

Issue: KBFS-1316

### DIFF
--- a/test/benchmark_test.go
+++ b/test/benchmark_test.go
@@ -211,3 +211,46 @@ func BenchmarkWriteBigFileBigBandwidth(b *testing.B) {
 	benchmarkWriteWithBandwidthPlusWarmup(b, 1<<30, /* 1 GB */
 		1<<20 /*1 MB writes */, 100*1024/8 /* 100 Mbps */)
 }
+
+func BenchmarkWriteMixedFilesNormalBandwidth(b *testing.B) {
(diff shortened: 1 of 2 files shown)


strib:
cc: @maxtaco

test/benchmark_test.go
```diff
@@ -211,3 +211,46 @@ func BenchmarkWriteBigFileBigBandwidth(b *testing.B) {
...
+		fileSizes = append(fileSizes, 2<<20)
+	}
+	fileSizes = append(fileSizes, 100<<20)
+	var totalSize int64
+	for _, size := range fileSizes {
+		totalSize += size
+	}
+	b.SetBytes(totalSize)
+
+	test(silentBenchmark{b},
```
maxtaco: is this considered a "DSL" test?

strib: Yes, although this is a dirty example since we use `customOp`.  But yes in the sense that this can run over libkbfs, fuse, or dokan as is.

maxtaco:
Cool! So glad to play a bit part in making our test suite bigger and better!

jzila:
LGTM

PR #236, opened 2016-08-12
enable OS X tests for O_EXCL

songgao:


strib:
Looks good to me, thanks!

PR #243, opened 2016-08-14
libfuse should not ignore errors from libkfs init procedure, this res…

krillr:
libfuse should not ignore errors from libkfs init procedure, this results in runtime panics in certain situations.

### DIFF
--- a/libfuse/start.go
+++ b/libfuse/start.go
@@ -24,7 +24,10 @@ type StartOptions struct {
 // Start the filesystem
 func Start(mounter Mounter, options StartOptions, kbCtx libkbfs.Context) *libfs.Error {
 	// InitLog errors are non-fatal and are ignored.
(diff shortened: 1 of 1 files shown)


strib:
Thanks @krillr!  @taruti it looks like you did this on purpose in e40c5adf8baec4267c8747ccfdea4e779f6870ed, but I'm not sure why.  Do you remember?

taruti:
@strib `libkbfs.InitLog` failure used to be non-fatal and things could proceed (without logging) even after failure. But if it now causes panics down the line somewhere then aborting is the right thing to do.

krillr:
Basically what ends up happening is that log is null, but the code tries to
use it anyway, and it panics.

taruti:
We used to return non-nil log even if InitLog failed. However I see this has changed making the errors fatal.

@strib This should be merged, I'll update also libdokan accordingly.

strib:
Works for me.  Thanks @krillr!

PR #245, opened 2016-08-15
libdokan: Handle InitLog error as in PR#243

taruti:
The same as the libfuse change.


strib:
Looks good to me, thanks!

PR #251, opened 2016-08-16
Fix govet

jzila:


jzila:
@strib: any risk here?

strib:
Hmm, assuming CI failure is unrelated, looks good to me.  I don't see an open issue for `TestSBSConflicts` though, unless it's the same authentication issue as the other tests.

PR #254, opened 2016-08-17
KBFS-1284 per file metadata file

taruti:


zanderz:
LGTM :thumbsup:
so, does this write special secret metadata files for the shell extension to read?

libfuse/dir.go
```diff
@@ -429,6 +431,18 @@ func (d *Dir) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.Lo
 		return child, nil
 	}
 
+	if strings.HasPrefix(req.Name, ".kbfs_fileinfo_") {
```
strib: ".kbfs_fileinfo_" should probably a const in `libfs`.

libkbfs/data_types.go
```diff
@@ -941,3 +941,8 @@ func SessionInfoFromProtocol(session keybase1.Session) (SessionInfo, error) {
 		VerifyingKey:   verifyingKey,
 	}, nil
 }
+
+// NodeMetadata has metadata about a node needed for higher level operations.
+type NodeMetadata struct {
```
strib: Maybe we can just put the entire `DirEntry` in here?  I can imagine it might be useful for debugging to get the block ID/refnonce, encrypted size, key gen, data version, etc.  What do you think?

libkbfs/data_types.go
```diff
@@ -941,3 +941,8 @@ func SessionInfoFromProtocol(session keybase1.Session) (SessionInfo, error) {
...
+	LastWriter keybase1.UID
```
strib: Didn't we talk about naming this `UnverifiedLastWriter`, to clarify that a modified client could fake this info?

Although one caveat to that, if the file is actually readable, then the block server would have enforced that the writer was really the one who wrote the block in the first place.

libkbfs/data_types.go
strib: Actually I guess `DirEntry.EntryInfo` isn't useful, since all that is available with `ls -l` already.  But anything in `DirEntry.BlockInfo` is potentially interesting.

libkbfs/mocks_test.go
```diff
@@ -4,15 +4,14 @@
...
-	reflect "reflect"
-	time "time"
-
 	gomock "github.com/golang/mock/gomock"
 	libkb "github.com/keybase/client/go/libkb"
 	logger "github.com/keybase/client/go/logger"
 	protocol "github.com/keybase/client/go/protocol"
 	go_metrics "github.com/rcrowley/go-metrics"
 	context "golang.org/x/net/context"
+	reflect "reflect"
```
strib: Do you have an up-to-date `mockgen`?  I don't these imports should change.

libkbfs/data_types.go
strib: Also, let's make this a `libkb.NormalizedUserName` so its human-readable.

strib:
> so, does this write special secret metadata files for the shell extension to read?

@zanderz: some users want to see who last wrote a file, and we don't expose that info any other way right now.  I think, with the other things I commented about, it'll be useful for debugging as well (to see BlockIDs, etc).  The encoded size may also be of interest to users.

strib:
First pass done.  Looks mostly good, with some questions/comments.

@taruti: Why did you choose this way, rather than xattrs?  Did you decide that racing with renames and really long file names weren't a big issue?  Might be worth commenting in the code about it.

Also, why not expose this in libdokan as well?

Finally it would be good to have a simple test in `mount_test.go` for this.

libfuse/dir.go
taruti: Done.

libkbfs/data_types.go
taruti: As long as we don't end with people complaining when the contents changes. Done.

libkbfs/data_types.go
taruti: Done.

taruti: Using `LastWriterAccordingToTLFLastWriter`

taruti:
I struggled whether to use xattrs and streams or separate files.

In the end Windows CLI programs and consoles are too horrible with streams and quoting and filename parsing to make it worth using alternate streams at this point of time.

Adding some tests.

taruti:
Rebased and adderessed things and added Dokan.

taruti:
ping @strib

libkbfs/data_types.go
```diff
@@ -960,3 +960,9 @@ func SessionInfoFromProtocol(session keybase1.Session) (SessionInfo, error) {
 		VerifyingKey:   verifyingKey,
 	}, nil
 }
+
+// NodeMetadata has metadata about a node needed for higher level operations.
+type NodeMetadata struct {
+	LastWriterAccordingToTLFLastWriter libkb.NormalizedUsername
```
strib: I like the intent here, but I'm not sure we should bake the term "TLF" into this data structure.  I think "LastWriterUnverified" (along with a comment saying why it's unverified, plus documentation we'll add on the website too) is clear enough.

libkbfs/interfaces.go
```diff
@@ -268,6 +268,9 @@ type KBFSOps interface {
 	GetEditHistory(ctx context.Context, folderBranch FolderBranch) (
 		edits TlfWriterEdits, err error)
 
+	// GetNodeMetadata gets meta data associated with a Node.
```
strib: "meta data" -> "metadata"

libfuse/dir.go
```diff
@@ -469,6 +470,18 @@ func (d *Dir) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.Lo
 		return node, nil
 	}
 
+	if strings.HasPrefix(req.Name, libfs.FileInfoPrefix) {
```
strib: Both here and in libdokan, it would be a little clearer to either have a comment above here explaining what this is doing, or to move it out into a helper function, so it doesn't confuse readers of the Lookup function too much.

strib:
Just very minor nits, otherwise looks good to me, thanks!

PR #257, opened 2016-08-17
libdokan: Add .kbfs_unmount to root

taruti:
Any access to K:\.kbfs_unmount will cause os.Exit(0)

### DIFF
--- a/libdokan/fs.go
+++ b/libdokan/fs.go
@@ -6,6 +6,7 @@ package libdokan
 
 import (
 	"errors"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


zanderz:
LGTM :thumbsup:

zanderz:
I have a question about this: if the filesystem isn't mounted, will it work? For installation, kbfsdokan.exe may need to be killed even if unmounted.

taruti:
If it is not mounted:
- kbfsdokan exits if it is unable to mount (will take 1-10 seconds)
- without a K:\ of course accessing K:\.kbfs_unmount does nothing

maxtaco:
Steve come to think of it with this approach you don't need to change the service. You can just touch this file from the updater or reinstall.

zanderz:
But the client is where the name of the mounted drive is configured, currently. I am leery of spreading that dependency around, or making assumptions or hardcoding it in a way that will cause pain later.

maxtaco:
Which client do you mean here?  If you mean the service, then you can ask the service where the mounted drive is (via `keybase config`) and then make a file out of that?  That way you won't have an additional hard-coded path.

zanderz:
Yeah I meant keybase.exe. Isn't config only one of the possible places for that to live, as per https://github.com/keybase/client/blob/master/go/libkb/env.go#L1013 ? The cat seems to be partly out the bag already. 

It would be great if there were some durable UNC path we could rely on - like I guess Fuse has? but we can't.

PR #260, opened 2016-08-18
Fix reader rekey bug

jzila:


jzila:
@strib: looks like this fixes the issue! That said, it doesn't address the scenario where there are other clients online that _should_ be able to rekey.

jzila:
Oh actually I think it does make sense; the reader successfully obtained the lock to rekey, but disabled the rekey bit so no rekey gets picked up by anything else.

libkbfs/folder_branch_ops.go
```diff
@@ -3951,7 +3951,13 @@ func (fbo *folderBranchOps) rekeyLocked(ctx context.Context,
 		}
 		// Clear the rekey bit if any.
 		md.Flags &= ^MetadataFlagRekey
-		md.clearLastRevision()
+		_, uid, err := fbo.config.KBPKI().GetCurrentUserInfo(ctx)
+		if err != nil {
+			return err
+		}
+		if md.GetTlfHandle().IsWriter(uid) {
```
strib: Add a brief comment here explaining why readers aren't allowed / don't need to do this.

strib:
There's not really a lock, but yeah if the reader's MD Put succeeds, everything should be fine.

Looks good to me.  I guess the test is going to be in kbfs-server?

jzila:
@strib yeah the test is already there: https://github.com/keybase/kbfs-server/pull/94/commits/bb303112ec8cd3f4e6cf6dc683c33bc45e23de29

PR #278, opened 2016-08-23
lib{dokan,fuse}: Small cleanup

taruti:
+ Cleanup code in libdokan and libfuse
+ Ensure special files set `resp.EntryValid = 0` in `libfuse/dir.go`
+ Add `isNoSuchNameError` helper to libfuse to avoid some type assertions


zanderz:
LGTM :thumbsup:

PR #280, opened 2016-08-24
folder_branch_ops: panic if no sync progress

strib:
It seems there's a bug that runs into an infinite loop because a Sync
seems to make no progress, and it's instantly retried.  Not sure
what's going on, but it spams the logs and causes them to roll over to
fast to see the debugging.  So detect when it happens and panic
instead.

Issue: KBFS-1332


strib:
cc: @maxtaco

aalness:
LGTM

PR #283, opened 2016-08-25
libfs: Fix remote status

taruti:
Bugfix.

### DIFF
--- a/libfs/remote_status.go
+++ b/libfs/remote_status.go
@@ -36,8 +36,6 @@ type RemoteStatus struct {
 // Init a RemoteStatus and register it with libkbfs.
 func (r *RemoteStatus) Init(ctx context.Context, log logger.Logger, config libkbfs.Config) {
 	r.failingServices = map[string]error{}
-	// A time in the far past that is not IsZero
-	r.failingSince.Add(time.Second)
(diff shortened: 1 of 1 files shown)


taruti:
Related to https://github.com/keybase/client/issues/4031

libfs/remote_status.go
```diff
@@ -109,7 +107,7 @@ func (r *RemoteStatus) ExtraFileName() string {
 	r.Lock()
 	defer r.Unlock()
 
-	if r.extraFileName == "" || time.Since(r.failingSince) > failureDisplayThreshold {
+	if r.extraFileName == "" || time.Since(r.failingSince) < failureDisplayThreshold {
```
zanderz: We return "" until it has failed for awhile? As long as this is the intent, but it seems counterintuitive, like what if there is no error?

zanderz:
Oh, it fixes that guy's problem? great.
LGTM :thumbsup:

libfs/remote_status.go
taruti: We return "" (= no error) if either:
+ There is no error (r.extraFilename == "")
+ There is an error but it might be a flake

Only start erroring out when there is a stable error that has presisted for failureDisplayThreshold.

PR #284, opened 2016-08-26
libdokan: Nicer logging of favorite removals

taruti:
Just makes logging more precise and informative.


zanderz:
LGTM :thumbsup:

PR #287, opened 2016-08-27
kbfs_ops: don't "create" a favorite except on TLF initialization

strib:
Otherwise the service could allocate multiple invites for an SBS user.
This came about when we started allowing empty TLFs for read
operations when the TLF hadn't yet been created.

Issue: keybase/client#4082


songgao:
LGTM!

(Is `create` only used for user interactions like sending notification emails?)

strib:
Yes, right now it's only used for popping up a window with an invite code for a user that doesn't yet have a Keybase account.

PR #290, opened 2016-08-29
libfs: Start with more eager errors

taruti:
Related to https://github.com/keybase/client/issues/4091

Error out even more eagerly on startup.


zanderz:
Not sure what the import is for users - this is the timeout thing?
LGTM :thumbsup:

taruti:
This changes things so that on startup we use the first error to enter error state. After that we need 5 seconds of errors to enter the error state. A single Status without any errors will get us out of the error state.

This means much less "why does kbfsdokan hang for me" when the user is not properly logged in.

PR #292, opened 2016-08-29
dirty_bcache: reset sync buf capacity after 5 minutes

strib:
If no sync as large as the current buffer size was completed in the
last five minutes, this will reset the buffer size back to the minimum
and let it grow again according to the normal algorithm.  This lets us
adapt to suddenly changing network conditions (for example, a laptop
going offline and coming back online with a worse network).

Issue: KBFS-1414

### DIFF
--- a/libkbfs/dirty_bcache.go
+++ b/libkbfs/dirty_bcache.go
@@ -26,6 +26,10 @@ type dirtyReq struct {
 	deadline time.Time
 }
 
+const (
+	resetBufferCapTimeDefault = 5 * time.Minute
+)
+
 // DirtyBlockCacheStandard implements the DirtyBlockCache interface by
 // storing blocks in an in-memory cache.  Dirty blocks are identified
 // by their block ID, branch name, and reference nonce, since the same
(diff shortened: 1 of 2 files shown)


jzila:
What if instead of resetting you cut it in half, floored at the minimum sync buffer cap? Seems like it's a bit aggressive to just reset it.

strib:
@jzila: my reasoning is that if its a laptop, and the laptop suspends right after a sync on a good network, and then comes up hours later on a bad network and then the timer fires, we want to reset completely right away -- halving isn't good enough.

If you still think this is too aggressive, maybe when the timer fires we can measure how much time has passed and lower it proportionally?

In my opinion this is still the safer way though, in case network or OS conditions change quickly.

jzila:
In that case the 5 minute timer should be reset any time there's communication of any kind. So it's a true "offline for 5 minutes" check.

strib:
> In that case the 5 minute timer should be reset any time there's communication of any kind. So it's a true "offline for 5 minutes" check.

A successful ping between the client and the mdserver in the last 5 minutes doesn't imply anything about the connection throughput.  And if a user starts up some other bandwidth-intensive process on the same machine, our effective bandwidth could easily fall instantaneously even if the laptop hasn't been suspended for more than 5 minutes.

I'm trying to figure out the right balance between graceful decay and building the buffer size back up when it's reset.  I think building it back up from a complete reset is likely to settle on the right bandwidth faster (when the user writes a big file) than decaying-by-half once every five minutes, when the bandwidth drops.  But I don't have hard evidence for that.  It just seems like a more conservative approach (in terms of avoiding timeouts) and something we can tweak later if performance is an issue.  Whereas decaying-by-half every five minutes is much more likely to result in timeouts on "sudden" network changes (from the app's perspective).

jzila:
I'm convinced :). 🚢 !

strib:
Appveyor issue fixed in #304.

PR #297, opened 2016-08-31
Wipe the key cache on logout

jzila:
r? @aalness


strib:
If you're going to change this to handle all caches, an easier way to do it would just be to call `config.ResetCaches()` on the loggedout notification.

aalness:
LGTM

jzila:
@strib: like so?

strib:
Yep!  Looks good to me.

PR #307, opened 2016-09-01
Skip unembedded block change check for reader rekeys

jzila:
r? @strib


strib:
Looks good to me pending CI, thanks!

PR #311, opened 2016-09-02
md_ops: serialize getRange calls in verifyWriterKey

strib:
Don't let concurrent goroutines try to fetch the same ranges
concurrently when looking for the original MD that was copied into a
later revision.

Issue: KBFS-1483


aalness:
LGTM

PR #315, opened 2016-09-02
fix broken/flaky test introduced by TestKBFSOpsCanceledCreateDelayTimeoutErrors

songgao:


libkbfs/kbfs_ops_concur_test.go
```diff
@@ -1311,6 +1312,7 @@ func TestKBFSOpsCanceledCreateDelayTimeoutErrors(t *testing.T) {
 	// cancellation should have been enabled.
 	<-onPutStalledCh
 	cancel()
+	time.Sleep(20 * time.Millisecond) // make sure the cancellation is relayed
```
strib: Instead of a hard delay, what if you just did `<-ctx.Done()`, which will guarantee that the `DelayCancellation` object cancelled the context passed to `CreateFile`?

strib:
Looks good with one suggestion for speeding it up.

libkbfs/kbfs_ops_concur_test.go
songgao: Sounds good!

strib:
Oh also please fix the commit subject/message before you merge.

PR #333, opened 2016-09-09
cleanup cancellation delayer in tests

songgao:
CR contexts are still not cleaned as they are longer term runners


strib:
Looks good to me, thanks!

PR #340, opened 2016-09-13
Circle CI (for mobile)

gabriel:
Copy of circle from client, with tiny change to swap repo names


gabriel:
This is temporary until we do jenkins, but will let you know if any KBFS changes will break mobile build

@strib ?

strib:
Looks good to me, thanks!

PR #346, opened 2016-09-14
config_local: journalize clean block cache as well

strib:
Also add a comment about dirty bcache resetting.

FYI: `journalizeDirtyBcache()` is something I added to the `journalDirtyBlockCache` PR after you reviewed it, when I realized a call to `ResetCaches()` would delete the journal dirty bcache entirely.  I just realized we needed the same treatment for the regular bcache as well, so that `CheckForKnownPtr` is still disabled after a reset.


strib:
Appveyor CI failure is the one I'm trying to track down in #345, unrelated.

PR #347, opened 2016-09-14
fix flaky test for cancellation delayer

songgao:


strib:
Looks good to me pending CI, thanks!

PR #351, opened 2016-09-15
dokan: add parenthesises in bridge.c

taruti:


taruti:
Noticed while debugging, not directly causing any issue however it seems.

taruti:
Appveyor failure seems unrelated.

PR #355, opened 2016-09-16
Write access notifications

gabriel:
Trying to create a file or directory in:

- /keybase (root)
- /keybase/public or /keybase/private (folderlist)
- /keybase/public/notyou (tlf)

will result in a WriteAccessError with Filename set to the file path trying to be created. This also gives a nicer "Permission denied" response when doing things in the terminal.

```sh
kbmb ➜ /keybase touch foo.txt
touch: foo.txt: Permission denied

kbmb ➜ /keybase cd public/

kbmb ➜ /keybase/public touch foo.txt
touch: foo.txt: Permission denied

kbmb ➜ /keybase/public cd testergabe

kbmb ➜ /keybase/public/testergabe touch foo.txt
touch: foo.txt: Permission denied

kbmb ➜ /keybase/public/testergabe cd ..

kbmb ➜ /keybase/public mkdir foo
mkdir: foo: File exists

kbmb ➜ /keybase/public mkdir foo.txt
mkdir: foo.txt: Permission denied

kbmb ➜ /keybase/public cd ..

kbmb ➜ /keybase mkdir foo
mkdir: foo: Permission denied
```

2016-09-15 at 5 34 pm (Ignore devel icon)

See https://github.com/keybase/client/pull/4294 for the client side.

In the future, I think it would be good to nuke PublicTopLevelFolder from FSNotification and have the Filename always be the canonical file path. The tlfname, public pair is a bit gnarly to keep around just cause it's not very future proof (if we add another folder type) and doesn't apply to things not in a tlf.

### DIFF
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -98,15 +98,17 @@ func (fl *FolderList) addToFavorite(ctx context.Context, h *libkbfs.TlfHandle) (
 // Create implements the fs.NodeCreater interface for FolderList.
 func (fl *FolderList) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (_ fs.Node, _ fs.Handle, err error) {
 	fl.fs.log.CDebugf(ctx, "FL Create")
-	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, "", err) }()
-	return nil, nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, libkbfs.CanonicalTlfName(req.Name)))
+	tlfName := libkbfs.CanonicalTlfName(req.Name)
+	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
+	return nil, nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, string(tlfName)))
 }
 
 // Mkdir implements the fs.NodeMkdirer interface for FolderList.
 func (fl *FolderList) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (_ fs.Node, err error) {
 	fl.fs.log.CDebugf(ctx, "FL Mkdir")
-	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, "", err) }()
-	return nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, libkbfs.CanonicalTlfName(req.Name)))
+	tlfName := libkbfs.CanonicalTlfName(req.Name)
+	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
+	return nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, string(tlfName)))
 }
 
 // Lookup implements the fs.NodeRequestLookuper interface.
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -108,7 +108,7 @@ func (fl *FolderList) Create(ctx context.Context, req *fuse.CreateRequest, resp 
 	fl.fs.log.CDebugf(ctx, "FL Create")
 	tlfName := libkbfs.CanonicalTlfName(req.Name)
 	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
-	return nil, nil, fl.fs.writeUnsupportedError(ctx, libkbfs.BuildCanonicalPath(fl.PathType(), string(tlfName)))
+	return nil, nil, libkbfs.NewWriteUnsupportedError(libkbfs.BuildCanonicalPath(fl.PathType(), string(tlfName)))
 }
 
 // Mkdir implements the fs.NodeMkdirer interface for FolderList.
@@ -116,7 +116,7 @@ func (fl *FolderList) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (_ fs.N
 	fl.fs.log.CDebugf(ctx, "FL Mkdir")
 	tlfName := libkbfs.CanonicalTlfName(req.Name)
 	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
-	return nil, fl.fs.writeUnsupportedError(ctx, libkbfs.BuildCanonicalPath(fl.PathType(), string(tlfName)))
+	return nil, libkbfs.NewWriteUnsupportedError(libkbfs.BuildCanonicalPath(fl.PathType(), string(tlfName)))
 }
 
 // Lookup implements the fs.NodeRequestLookuper interface.
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -95,12 +95,20 @@ func (fl *FolderList) addToFavorite(ctx context.Context, h *libkbfs.TlfHandle) (
 	return nil
 }
 
+// PathType returns PathType for this folder
+func (fl *FolderList) PathType() libkbfs.PathType {
+	if fl.public {
+		return libkbfs.PublicPathType
+	}
+	return libkbfs.PrivatePathType
+}
+
 // Create implements the fs.NodeCreater interface for FolderList.
 func (fl *FolderList) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (_ fs.Node, _ fs.Handle, err error) {
 	fl.fs.log.CDebugf(ctx, "FL Create")
 	tlfName := libkbfs.CanonicalTlfName(req.Name)
 	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
-	return nil, nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, string(tlfName)))
+	return nil, nil, fl.fs.writeUnsupportedError(ctx, libkbfs.BuildCanonicalPath(fl.PathType(), string(tlfName)))
 }
 
 // Mkdir implements the fs.NodeMkdirer interface for FolderList.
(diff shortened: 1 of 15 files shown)


gabriel:
@strib @akalin-keybase or whoever ❤️

libfuse/folderlist.go
```diff
@@ -95,6 +95,20 @@ func (fl *FolderList) addToFavorite(ctx context.Context, h *libkbfs.TlfHandle) (
 	return nil
 }
 
+// Create implements the fs.NodeCreater interface for FolderList.
+func (fl *FolderList) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (_ fs.Node, _ fs.Handle, err error) {
+	fl.fs.log.CDebugf(ctx, "FL Create")
+	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, "", err) }()
+	return nil, nil, fl.fs.writeAccessError(ctx, nil, libkbfs.BuildCanonicalPath(fl.public, libkbfs.CanonicalTlfName(req.Name)))
```
strib: I don't think "Permission denied" is the right error, here in the folder list or in the `/keybase` operations -- it implies that there might be some user who has permission to do this thing, when in fact the operation isn't supported at all.

libfuse/tlf.go
```diff
@@ -163,7 +168,9 @@ func (tlf *TLF) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.
 }
 
 // Create implements the fs.NodeCreater interface for TLF.
-func (tlf *TLF) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (fs.Node, fs.Handle, error) {
+func (tlf *TLF) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (_ fs.Node, _ fs.Handle, err error) {
+	tlf.log().CDebugf(ctx, "TLF Create %s", req.Name)
+	defer func() { tlf.folder.reportErr(ctx, libkbfs.WriteMode, err) }()
```
strib: If the code reaches `dir.Create`, two errors will be reported for the same operation.  Maybe you should only report this one when the `loadDir` fails?

libfuse/tlf.go
```diff
@@ -172,8 +179,9 @@ func (tlf *TLF) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.
 }
 
 // Mkdir implements the fs.NodeMkdirer interface for TLF.
-func (tlf *TLF) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (
-	fs.Node, error) {
+func (tlf *TLF) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (_ fs.Node, err error) {
+	tlf.log().CDebugf(ctx, "TLF Mkdir %s", req.Name)
+	defer func() { tlf.folder.reportErr(ctx, libkbfs.WriteMode, err) }()
```
strib: Same here.

libkbfs/tlf_handle.go
```diff
@@ -522,20 +522,20 @@ func (h *TlfHandle) GetCanonicalName() CanonicalTlfName {
...
+func BuildCanonicalPath(public bool, tlfName CanonicalTlfName) string {
 	var folderType string
 	if public {
 		folderType = "public"
 	} else {
 		folderType = "private"
 	}
-	// TODO: Handle windows paths?
-	return fmt.Sprintf("/keybase/%s/%s", folderType, canonicalName)
+	return fmt.Sprintf("/keybase/%s/%s", folderType, tlfName)
```
strib: Why did you get rid of the TODO?  Still seems valid to me.

libkbfs/folder_branch_ops.go
```diff
@@ -2221,7 +2229,7 @@ func (fbo *folderBranchOps) createEntryLocked(
 	}
 
 	// verify we have permission to write
-	md, err := fbo.getMDForWriteLocked(ctx, lState)
+	md, err := fbo.getMDForWriteLockedForName(ctx, lState, name)
```
strib: This doesn't seem right.  If you did `touch /keybase/private/strib/a/b/c/name`, the error would be for `/keybase/private/strib/name`, right?  I think you need to include the full path using `dir`.

libkbfs/tlf_handle.go
gabriel: I think we want this /keybase/private/username string as the canonical universal path that is representative of a path on a platform... we would convert it to windows path on windows? Otherwise it's not canonical.

`/keybase/private/username` - Canonical path

`k:\private\username` - Windows (if k: configured, could be another drive letter)
`/keybase/private/username` - Linux and darwin (happens to be same as canonical)
`sftp://keybase.priv/username` - Some theoretical other thing

gabriel: Added a comments on the method to say this.

strib: > we would convert it to windows path

You mean the GUI would convert it?  I just think it's an open issue what to do about displaying canonical path names on Windows, so it still seems like a TODO.  But maybe your new comment clarifies things.

strib: Ah ok I see it now.  Looks good.

libkbfs/folder_branch_ops.go
gabriel: Yeah... I created a new method:
```go
func (fbo *folderBranchOps) canonicalPath(ctx context.Context, dir Node, name string) (string, error)
```
which returns canonical path for a dir node and name.

libfuse/tlf.go
gabriel: Reverted this since folder reportErr will report.

libfuse/tlf.go
gabriel: Reverted this since folder reportErr will report.

libfuse/folderlist.go
gabriel: To me I think it makes sense to return permission denied (EACCES) if a program (or user) tries to create /keybase/foo.txt or /keybase/private/foo.txt ... It seems more appropriate than ENOENT which implies not found?

gabriel: Made an WriteUnsupportedError for these non-tlf cases.

gabriel:
k, I did another pass and think is ready to review again...

libkbfs/path.go
```diff
@@ -0,0 +1,42 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
```
strib: We have a `path` type in `data_types.go` already.  Maybe it makes sense to move that in here (along with `pathNode`), or else rename this file to `canonical_path.go`?

libfuse/tlf.go
```diff
@@ -163,7 +168,7 @@ func (tlf *TLF) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.
 }
 
 // Create implements the fs.NodeCreater interface for TLF.
-func (tlf *TLF) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (fs.Node, fs.Handle, error) {
+func (tlf *TLF) Create(ctx context.Context, req *fuse.CreateRequest, resp *fuse.CreateResponse) (_ fs.Node, _ fs.Handle, err error) {
```
strib: Any reason to name `err` here (and below) anymore?

libfuse/fs.go
```diff
@@ -201,6 +201,22 @@ func (f *FS) Statfs(ctx context.Context, req *fuse.StatfsRequest, resp *fuse.Sta
 	return nil
 }
 
+func (f *FS) writeAccessError(ctx context.Context, h *libkbfs.TlfHandle, filename string) error {
```
strib: There aren't any more callers for this, right?  Can we remove it?

libkbfs/errors.go
```diff
@@ -165,18 +165,32 @@ func (e ReadAccessError) Error() string {
...
 
 // Error implements the error interface for WriteAccessError
 func (e WriteAccessError) Error() string {
-	return fmt.Sprintf("%s does not have write access to directory %s",
-		e.User, buildCanonicalPath(e.Public, e.Tlf))
+	if e.Tlf != "" {
+		return fmt.Sprintf("%s does not have write access to directory %s",
+			e.User, buildCanonicalPath(e.Public, e.Tlf))
+	}
+	return fmt.Sprintf("%s does not have write access to %s", e.User, e.Filename)
```
strib: Are there any remaining uses that hit this case?  I couldn't find any.  If not, can we revert it back (and get rid of "Filename")?

libkbfs/errors.go
```diff
@@ -165,18 +165,32 @@ func (e ReadAccessError) Error() string {
...
+
+// WriteUnsupportedError indicates an error when trying to write a file
+type WriteUnsupportedError struct {
+	User     libkb.NormalizedUsername
+	Filename string
+}
+
+// Error implements the error interface for WriteAccessError
+func (e WriteUnsupportedError) Error() string {
+	return fmt.Sprintf("%s can't write to %s", e.User, e.Filename)
```
strib: I don't think we need the username in here, since it's not supported for anyone.  Maybe something like simple like `("Writing to %s is unsupported", e.Filename)`, and get rid of the need for `User` (and the `GetCurrentUserInfo` call in `writeUnsupportedError`).

libkbfs/tlf_handle.go
```diff
@@ -522,15 +522,13 @@ func (h *TlfHandle) GetCanonicalName() CanonicalTlfName {
 	return h.name
 }
 
-func buildCanonicalPath(public bool, canonicalName CanonicalTlfName) string {
-	var folderType string
+// buildCanonicalPath returns a canonical path for a tlf.
+func buildCanonicalPath(public bool, tlfName CanonicalTlfName) string {
```
strib: Do we need this method anymore?  Seems confusing to have it around with an identical but lower case name.  I think we can replace existing callers pretty easily.  Or maybe just rename it to something like `translatePublicAndBuildCanonicalPath()` and move it into the same file as `BuildCanonicalPath()`, if you don't feel like dealing with that.

libkbfs/path.go
gabriel: Moved path and pathNode to path.go

libfuse/tlf.go
gabriel: Nope removed these named vars

libfuse/fs.go
gabriel: Removed

libkbfs/errors.go
gabriel: Yeah we always have tlf for access error now. Removed.

libkbfs/errors.go
gabriel: Yeah we don't use it, we can add in if we need it. Removed.

libkbfs/tlf_handle.go
gabriel: It's still useful in a couple spots but is out of place. Renamed to `buildCanonicalPathForTlfName` and moved to path.go.

PR #358, opened 2016-09-19
Add GetPublicCanonicalTLFNameAndID to implement KeybaseServiceBase

patrickxb:
To support public chats, need this info from KBFS.

r? @songgao 

cc @strib @maxtaco @mmaxim


libkbfs/keybase_service_base.go
```diff
@@ -619,3 +619,25 @@ func (k *KeybaseServiceBase) GetTLFCryptKeys(ctx context.Context,
 
 	return res, nil
 }
+
+// GetTLFCanonicalID implements the TlfKeysInterface interface for
+// KeybaseServiceBase.
+func (k *KeybaseServiceBase) GetTLFCanonicalID(ctx context.Context,
```
strib: What does "canonical ID" mean?  I don't think we have different ways to present a TLF ID.  I guess you're referring to the name?  Should this be `GetCanonicalTLFNameAndID`?

libkbfs/keybase_service_base.go
```diff
@@ -619,3 +619,25 @@ func (k *KeybaseServiceBase) GetTLFCryptKeys(ctx context.Context,
...
+	arg keybase1.GetTLFCanonicalIDArg) (res keybase1.TLFCanonicalID, err error) {
+	ctx = ctxWithRandomIDReplayable(ctx, CtxKeybaseServiceIDKey, CtxKeybaseServiceOpID,
+		k.log)
+	tlfHandle, err := k.getHandleFromFolderName(ctx, arg.TlfName, arg.Public)
```
strib: Can we return an error if `!arg.Public`?  I really want to discourage direct use of the TLF ID whenever possible, so ideally we'd only allow this used in the strictest possible way. (I realize `GetTLFCryptKeys` also returns the ID, which also makes me uncomfortable. but I think that's slightly better since the name implies it should only be used for something specific.)

songgao:
Would it help to rename the function by adding a suffix (DoNotCacheID)?

patrickxb:
It's for public chats where we don't need encryption key, just tlfid and canonical tlf name.

strib:
@patrickxb: right, but what do you _do_ with the TLF ID after you get it?  Ideally you wouldn't store it in any persistent chat data structures, where it could later be used by some other client as the primary identifier of the TLF.

patrickxb:
It is sent to Gregor. Same as private chats. Nothing in service stores it but chat server needs tlfid

strib:
Ok sounds good, thanks.  I think the name or the avdl documentation around the function should explain what's allowed to happen with the ID.

libkbfs/keybase_service_base.go
patrickxb: Ok

libkbfs/keybase_service_base.go
patrickxb: I don't find the `GetTLFCryptKeys` name to be specific.  It sounds very generic to me and doesn't even mention the TLF ID that you are concerned about.

patrickxb:
PTAL

songgao:
LGTM!

PR #367, opened 2016-09-21
bug fix in delayed cancellation and caching EntryInfo for Create

songgao:


songgao:
So part of this PR is mainly for the situation where a context can get canceled right after `Create` handler enters the critical portion (MD Write), and it takes too long for the rest to finish. Since for `open()` that creates a file, an `Attr` is followed after the `Create`. This can cause `Attr` to fail.

I couldn't come up with an easy way to test this, but I did try cancelling at the end of the `Create` handler, which results in failure before the change and success after the change.

songgao:
Actually on a second look, I think the issue was probably caused by the delayed cancellation bug, where `canceled` was set to try right after original cancellation was cancelled, even though it had not cancelled the actual context yet. This caused the `Enable...` call afterwards for `Attr` to get a `context.Canceled` error even though the context might not have been canceled.

In other words, it seems the `eiCache` is probably not absolutely necessary unless we are worried about MD write taking longer than the grace period. What do you think? @strib

libfuse/file.go
```diff
@@ -11,19 +11,41 @@ import (
...
+}
+
 // Attr implements the fs.Node interface for File.
 func (f *File) Attr(ctx context.Context, a *fuse.Attr) (err error) {
 	f.folder.fs.log.CDebugf(ctx, "File Attr")
 	defer func() { f.folder.reportErr(ctx, libkbfs.ReadMode, err) }()
 
+	if f.eiCache != nil && ctx.Value(CtxIDKey) == f.eiCache.rid {
+		fillAttrWithMode(&f.eiCache.ei, a)
+		f.eiCache = nil
```
strib: You might need a lock for this?  Though I guess it's likely not necessary since there will only ever be one call with the same `rid`.  Maybe a comment explaining why this is ok?

libfuse/file.go
```diff
@@ -11,19 +11,41 @@ import (
 	"golang.org/x/net/context"
 )
 
+type eiCache struct {
+	ei  libkbfs.EntryInfo
+	rid string
```
strib: What does the "r" stand for?  "request"?  `reqID` might be clearer.

libfuse/file.go
```diff
@@ -11,19 +11,41 @@ import (
...
+}
+
 // File represents KBFS files.
 type File struct {
 	folder *Folder
 	node   libkbfs.Node
+
+	// for the Attr call after Create
+	eiCache *eiCache
```
strib: It might make sense to set this to `nil` on every modifying operation, including invalidates if that's easy.  Just in case another operation slips in between the create and the attr call. (In which case, you'd definitely need a lock.)

libkbfs/delayed_cancellation.go
```diff
@@ -192,17 +186,18 @@ func NewContextWithCancellationDelayer(
 // delayed cancellation for ctx. This is useful to indicate that the
 // operation(s) associated with the context has entered a critical state, and
 // it should not be canceled until after timeout or CleanupCancellationDelayer
-// is called.
+// is called. Note that if EnableDelayedCancellationWithGracePeriod is called
+// for the second time, and the grace period has started due to a cancellation,
+// the grace period would not be extended (i.e. timeout has no effect in this
+// case).
```
strib: Maybe add something like ", although no error will be returned" at the end of the sentence?  It's a little surprising that the new timeout will be ignored without an error being thrown.

songgao:
> I thought we checked in the logs and saw that the write was indeed taking a long time?

I was not sure about that, which was part of reasons why I brought it up. But yeah, sounds good!

libfuse/file.go
songgao: Yes; sounds good!

libfuse/file.go
songgao: But the request ID is "guarding" this right? So it means the cached EntryInfo is only valid for this particularly request ID. I'll add a comment for it?

libkbfs/delayed_cancellation.go
songgao: Sounds good!

libfuse/file.go
strib: I mean if this happens:

1. Bazil sends Create call, and it completes.
2. Another thread/process writes to the file somehow.
3. Bazil sends the corresponding Attr call.

I'm not sure if it's possible for (2) to sneak in between (1) and (3) -- maybe there's no way to get a file handle before the overall Create syscall returns.  But it might be possible, and maybe worth guarding against.

libfuse/file.go
songgao: Yeah I'll add a comment. Although as a pre-caution in case bazil changes, I guess we could use `atomic` for loading/storing `f.eiCache`. It wouldn't make it race-free (although I think the race is fine here since it's just setting `f.eiCache` to nil, which is not conflicting with presumably race situation where f.eiCache is already nil before line 45) but it'd eliminate the possibility of dereferencing null-pointer `f.ei.Cache` here

songgao: ah, nvm. I'll just add a lock!

libfuse/file.go
songgao: I see! I'll push another commit to set it to nil when modified

libfuse/file.go
```diff
@@ -5,24 +5,58 @@
...
+	cache *eiCache
+}
+
+func (c *eiCacheHolder) destroy() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.cache = nil
+}
+
+func (c *eiCacheHolder) getAndDestroy(reqID string) (ei *libkbfs.EntryInfo) {
```
strib: Maybe `getAndDestroyIfMatches`?  Or something similar?  Otherwise it's a little confusing when `Attr` calls this with a wrong request ID.

libfuse/file.go
```diff
@@ -5,24 +5,58 @@
...
+
+// eiCacheHolder caches the EntryInfo for a particular reqID. It's used for the
+// Attr call after Create. This should only be used for operations with same
+// reqID.
+type eiCacheHolder struct {
+	mu    sync.Mutex
+	cache *eiCache
+}
+
+func (c *eiCacheHolder) destroy() {
```
strib: You should also call this in `libfuse.Folder.invalidateNodeDataRange()` and maybe in the `default` case of `libfuse.Folder.batchChangesInvalidate()`.

songgao: OK; I'll update it before merging!

libfuse/file.go
songgao: Sounds good!

PR #370, opened 2016-09-22
Taruti/kbfs 1466 number of open handles

taruti:
* Add tracking of number of open files per mount to dokan
* Add a `K:\.kbfs_number_of_handles` that exposes it as an integer


zanderz:
oh wait, and maybe also CI?

taruti:
Good note on the locking, I simplified it so we only hold one lock at a time.

taruti:
CI appears to be nonrelated.

PR #373, opened 2016-09-23
dokan: Try to load dll on ancient machines without KB2533623

taruti:
This is bad. Users really should install security fixes for Windows that are >5 years old.


zanderz:
LGTM :thumbsup:

PR #384, opened 2016-09-28
libdokan: Handle deletions after renames properly

taruti:
Fix bug.

### DIFF
--- a/libdokan/dir.go
+++ b/libdokan/dir.go
@@ -104,13 +104,12 @@ func (f *Folder) getFolderBranch() libkbfs.FolderBranch {
 }
 
 // forgetNode forgets a formerly active child with basename name.
-func (f *Folder) forgetNode(node libkbfs.Node) {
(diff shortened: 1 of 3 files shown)


libdokan/dir.go
```diff
@@ -486,6 +486,8 @@ func (d *Dir) Cleanup(ctx context.Context, fi *dokan.FileInfo) {
 	defer func() { d.folder.reportErr(ctx, libkbfs.WriteMode, err) }()
 
 	if fi != nil && fi.IsDeleteOnClose() && d.parent != nil {
+		d.folder.fs.renameAndDeletionLock.Lock()
```
zanderz: Unlocking with defer means that this lock is still held curing the call to forgetNode(), below, which I see attempts to acquire `f.mu`, which looks dangerous to me.

taruti:
Added documentation about the order of locking.

taruti:
Jenkins failure not related.

zanderz:
Thank you! LGTM :thumbsup:

PR #402, opened 2016-10-04
libkbfs: Throttling should not cause error state

taruti:
Don't push the connection status error down if it is just throttling.


taruti:
Related to https://github.com/keybase/client/issues/4476

PR #406, opened 2016-10-05
run identify in cryptkeys and plumb through IdentifyTrackBreaks

songgao:
Hi there,

This PR is to:

1. make `GetTLFCryptKeys` to run identify;
2. support chat specific behavior of identify, i.e. not fail but plumb through the proof failures

There's virtually no tests for the new stuff yet, but I wanted to see how people feel about this. If this look good, I'll add some tests before merging!

### DIFF
--- a/libkbfs/identify_util.go
+++ b/libkbfs/identify_util.go
@@ -13,14 +13,7 @@ import (
 	"golang.org/x/net/context"
 )
 
-func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
-	identifier identifier, uid keybase1.UID, isPublic bool) error {
-	_, _, err := identifyUIDWithIdentifyBehavior(ctx,
-		nug, identifier, uid, isPublic, keybase1.TLFIdentifyBehavior_DEFAULT_KBFS)
-	return err
-}
-
-func identifyUIDWithIdentifyBehavior(ctx context.Context,
+func identifyUIDWithBehavior(ctx context.Context,
 	nug normalizedUsernameGetter, identifier identifier, uid keybase1.UID,
 	isPublic bool, identifyBehavior keybase1.TLFIdentifyBehavior) (
 	libkb.NormalizedUsername, *keybase1.IdentifyTrackBreaks, error) {
@@ -45,27 +38,20 @@ func identifyUIDWithIdentifyBehavior(ctx context.Context,
 		fallthrough
 	case keybase1.TLFIdentifyBehavior_DEFAULT_KBFS:
 		userInfo, err = identifier.Identify(ctx, username.String(), reason)
-		if err != nil {
-			// Convert libkb.NoSigChainError into one we can report.  (See
-			// KBFS-1252).
-			if _, ok := err.(libkb.NoSigChainError); ok {
-				return "", nil, NoSigChainError{username}
-			}
-			return "", nil, err
-		}
 	case keybase1.TLFIdentifyBehavior_CHAT_GUI:
 		userInfo, breaks, err = identifier.IdentifyForChat(
 			ctx, username.String(), reason)
-		if err != nil {
-			// Convert libkb.NoSigChainError into one we can report.  (See
-			// KBFS-1252).
-			if _, ok := err.(libkb.NoSigChainError); ok {
-				return "", nil, NoSigChainError{username}
-			}
-			return "", nil, err
-		}
 	default:
 		return "", nil, errors.New("unknown identifyBehavior")
+	}
+
+	if err != nil {
+		// Convert libkb.NoSigChainError into one we can report.  (See
+		// KBFS-1252).
+		if _, ok := err.(libkb.NoSigChainError); ok {
+			return "", nil, NoSigChainError{username}
+		}
+		return "", nil, err
 	}
 
 	if userInfo.Name != username {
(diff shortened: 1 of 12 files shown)


maxtaco:
I really think @strib or @akalin-keybase or someone on KBFS should weight in too

libkbfs/identify_util.go
```diff
@@ -5,68 +5,139 @@
...
 
-func identifyUID(ctx context.Context, nug normalizedUsernameGetter, identifier identifier, uid keybase1.UID, isPublic bool) error {
+func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
+	identifier identifier, uid keybase1.UID, isPublic bool) error {
+	_, _, err := identifyUIDWithIdentifyBehavior(ctx,
+		nug, identifier, uid, isPublic, keybase1.TLFIdentifyBehavior_DEFAULT_KBFS)
+	return err
+}
+
+func identifyUIDWithIdentifyBehavior(ctx context.Context,
```
strib: Having a double "identify" in the name is confusing.  Maybe just `identifyUIDWithBehavior`?

libkbfs/identify_util.go
```diff
@@ -5,68 +5,139 @@
...
+	"errors"
 	"fmt"
 
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
 	"golang.org/x/net/context"
 )
 
-func identifyUID(ctx context.Context, nug normalizedUsernameGetter, identifier identifier, uid keybase1.UID, isPublic bool) error {
+func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
```
strib: Also, this is a small style nit, but I try to organize functions so that within a file they only call functions that are above them in the file.  Not a style requirement, but you'd make me happy if you reorganized these functions that way!

libkbfs/interfaces.go
```diff
@@ -298,6 +298,9 @@ type KeybaseService interface {
 	// popups spawned.
 	Identify(ctx context.Context, assertion, reason string) (UserInfo, error)
 
+	IdentifyForChat(ctx context.Context, assertion, reason string) (
```
strib: I'd prefer a name less chat-specific, since later on we could have other app types that want the same behavior.  Maybe `IdentifyAndAllowTrackBreaks`?

strib: Also, please include an explanatory comment.

libkbfs/interfaces.go
```diff
@@ -378,6 +381,8 @@ type identifier interface {
 	// necessary.  The reason string is displayed on any tracker
 	// popups spawned.
 	Identify(ctx context.Context, assertion, reason string) (UserInfo, error)
+	IdentifyForChat(ctx context.Context, assertion, reason string) (
```
strib: Same here.

libkbfs/tlf_handle.go
```diff
@@ -931,40 +919,78 @@ func ParseTlfHandle(
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		return nil, nil, TlfNameNotCanonical{name, string(h.GetCanonicalName())}
+	case keybase1.TLFIdentifyBehavior_CHAT:
+		breaks, err := identifyHandleWithIdentifyBehavior(ctx, kbpki, kbpki, h,
+			keybase1.TLFIdentifyBehavior_CHAT)
+		if err != nil {
+			return nil, breaks, err
```
strib: Should return a `nil` breaks, here and below, when there's an error.

libkbfs/tlf_handle.go
```diff
@@ -931,40 +919,78 @@ func ParseTlfHandle(
...
+		}
+
+		if string(h.GetCanonicalName()) != name {
```
strib: Why does this come after the identifies in this case?  I guess you're trying to force identifies to happen during parsing, to avoid plumbing `behavior` all the way through `KBFSOps`?  I don't really think that strategy is going to work, since in the case where we're initializing a new folder that didn't exist before, KBFS will do another identify with the default behavior, which will break if there are broken tracker statements.

I think it would be better to still short-circuit the identifies here if the name is canonical, and instead plumb `behavior` into `KBFSOpsStandard.getOrInitializeNewMDMaster`.  Then, you'd have to add an explicit call to identify somewhere if a) the TLF already existed on the servers, and b) there isn't yet a `folderBranchOps` with `identifyDone=true`.  And when creating a new folder, you'd have to plumb the behavior through to `folderBranchOps.SetInitialHeadToNew`.  I can't see a way around this.

If it's too painful to plumb the behavior forward and the breaks backwards through all those codepaths, it probably makes sense to use the context strategy like we discussed on the phone call.

strib: Also, if KBFS has already done the full identify (e.g., `folderBranchOps.identifyDone = true`), this PR does extra work on every `GetTLFCryptKeys` RPC.

songgao: Ah, I didn't realize that; somehow I was thinking that was the only place that runs identify. Although what's the purpose of running identify here originally (before returning canonical name if it's not canonical)?

songgao: Also I have a few more questions:

1) what should the behavior be like if the TLF is first time being accessed? Should identify fail to avoid creating it at all, or go ahead and create it but just don't set `identifyDone` to `true`, so that next time a normal KBFS operation happens, it'd run identify with KBFS behavior again?

2) if the answer for 1) is latter, is it what "And when creating a new folder, you'd have to plumb the behavior through to folderBranchOps.SetInitialHeadToNew. " is about? If the answer for 1) is former, could you explain that a bit?

Thanks!

strib: > Ah, I didn't realize that; somehow I was thinking that was the only place that runs identify. Although what's the purpose of running identify here originally (before returning canonical name if it's not canonical)?

The issue as I recall (@akalin-keybase should correct me if I'm wrong) is that the resolve calls (which turn an `@` assertion into a username) don't actually check the proofs, it only relies on the Keybase database.  So if there were any assertions in the original name, we have to do an identify here to make sure it is safe to return the canonicalized name.

> 1) what should the behavior be like if the TLF is first time being accessed? Should identify fail to avoid creating it at all, or go ahead and create it but just don't set identifyDone to true, so that next time a normal KBFS operation happens, it'd run identify with KBFS behavior again?

Not sure I follow your question.  If `IdentifyForChat` fails, then the folder should definitely not be created.  If `IdentifyForChat` succeeds but with track breaks, then KBFS still needs to do a regular identify later -- whether or not you make a `folderBranchOps` or not doesn't make much difference to me.  If `IdentifyForChat` succeeds without any track breaks, then you can create a `folderBranchOps` and set `identifyDone=true`.

> 2) if the answer for 1) is latter, is it what "And when creating a new folder, you'd have to plumb the behavior through to folderBranchOps.SetInitialHeadToNew. " is about? If the answer for 1) is former, could you explain that a bit?

`GetTLFCryptKeys` can end up calling `folderBranchOps.SetInitialHeadToNew` if the TLF doesn't exist yet on the servers, and that will eventually call `Identify2`.  If you don't plumb anything through, it will do the regular identify, which you don't want.  If the folder already existed on the servers, you should add an explicit identify in `getOrInitializeNewMDMaster` when it is being called for chat purposes.

songgao: Got it; Thanks Jeremy!

songgao:
Closing in favor of https://github.com/keybase/kbfs/pull/411

PR #414, opened 2016-10-10
folder_branch_ops: check last flush error when waiting for journal

strib:
Issue: KBFS-1609


strib:
@akalin-keybase: FYI I will probably merge this when CI passes, since we just ran into it and needed to manually repair a TLF.

PR #415, opened 2016-10-11
Push KBFS to docker hub if triggered by client master

jzila:
r? @songgao


jzila:
@songgao 1. is incorrect, 2. is correct. We can make 1. true, but it'll slow down client builds by a lot (because it won't start KBFS tests until client tests have passed)

songgao:
@jzila that makes sense. But doesn't this make it possible to push a kbfs image with broken client?

jzila:
Yes

songgao:
Got you; we'll just need to watch out master builds then :)

PR #416, opened 2016-10-11
kbfs_ops_test: fix ReIdentify test flake

strib:
This test tried to get around the fact that reidentifies were
asynchronous by sending twice on the reidentify channel.
Unfortunately, that led to a race, where identifyDone could be set
back to false by the second send, after the final identify was done.

Instead, the reidentify loop can explicitly signal when it's finished.

Issue: KBFS-1553

### DIFF
--- a/libkbfs/kbfs_ops_test.go
+++ b/libkbfs/kbfs_ops_test.go
@@ -337,14 +337,15 @@ func TestKBFSOpsGetRootNodeReIdentify(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, fboIdentityDone(ops))
 
-	// The channel is not buffered, when the second value is received
-	// the first round of marking for reidentify will be done.
+	// Mark everything for reidentifying, and wait for it to finish
+	// before checking.
 	kop := config.KBFSOps().(*KBFSOpsStandard)
-	kop.reIdentifyControlChan <- struct{}{}
-	kop.reIdentifyControlChan <- struct{}{}
+	returnCh := make(chan struct{})
+	kop.reIdentifyControlChan <- returnCh
+	<-returnCh
 	assert.False(t, fboIdentityDone(ops))
 
-	// Trigger identify.
+	// Trigger new identify.
 	lState = makeFBOLockState()
 	_, err = ops.getMDLocked(ctx, lState, mdReadNeedIdentify)
 	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


libkbfs/kbfs_ops_test.go
```diff
@@ -337,14 +337,15 @@ func TestKBFSOpsGetRootNodeReIdentify(t *testing.T) {
...
-	// The channel is not buffered, when the second value is received
-	// the first round of marking for reidentify will be done.
+	// Mark everything for reidentifying, and wait for it to finish
+	// before checking.
 	kop := config.KBFSOps().(*KBFSOpsStandard)
-	kop.reIdentifyControlChan <- struct{}{}
-	kop.reIdentifyControlChan <- struct{}{}
+	returnCh := make(chan struct{})
+	kop.reIdentifyControlChan <- returnCh
+	<-returnCh
```
jzila: Is this the only place the channel is read?

libkbfs/kbfs_ops_test.go
```diff
@@ -337,14 +337,15 @@ func TestKBFSOpsGetRootNodeReIdentify(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, fboIdentityDone(ops))
 
-	// The channel is not buffered, when the second value is received
-	// the first round of marking for reidentify will be done.
+	// Mark everything for reidentifying, and wait for it to finish
+	// before checking.
 	kop := config.KBFSOps().(*KBFSOpsStandard)
-	kop.reIdentifyControlChan <- struct{}{}
-	kop.reIdentifyControlChan <- struct{}{}
```
jzila: here we wrote to `reIdentifyControlChan` twice, and now we write once. Why does this work?

libkbfs/kbfs_ops_test.go
strib: Yes, the control channel is only for tests, and this is the only test that uses it.

libkbfs/kbfs_ops_test.go
strib: I tried to explain it in the PR/commit description -- the second write was just a hacky way of knowing that the results of the first write were finished, but that had unintended consequences.

jzila: I see, yeah I understood the idea from the message, just didn't understand how it could work properly (in terms of who was reading the write)

strib: It's just being read by a loop in `KBFSOpsStandard`, and each time there's something on the channel it makes `identifyDone = false`.

PR #438, opened 2016-10-19
mdserver_remote: support MDv3

aalness:


aalness:
This is ready for review. I think CI is failing because `kbfs-server:master` can't build with this version of the protocol. I won't merge this until `kbfs-server` is updated in any case.

libkbfs/mdserver_remote.go
```diff
@@ -836,9 +860,68 @@ func (md *MDServerRemote) backgroundRekeyChecker(ctx context.Context) {
...
+	}
+
+	response, err := md.client.GetKeyBundles(ctx, arg)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var wkb *TLFWriterKeyBundleV3
+	if response.WriterBundle.Bundle != nil {
+		if response.WriterBundle.Version != int(SegregatedKeyBundlesVer) {
```
jzila: `<`?

libkbfs/mdserver_remote.go
```diff
@@ -836,9 +860,68 @@ func (md *MDServerRemote) backgroundRekeyChecker(ctx context.Context) {
...
+			err = fmt.Errorf("Unsupported writer bundle version: %d",
+				response.WriterBundle.Version)
+			return nil, nil, err
+		}
+		wkb := new(TLFWriterKeyBundleV3)
```
jzila: You're shadowing `wkb` above, which won't be set by this, and thus you'll return `nil` below. Should be `=` instead of `:=`.

libkbfs/mdserver_remote.go
```diff
@@ -836,9 +860,68 @@ func (md *MDServerRemote) backgroundRekeyChecker(ctx context.Context) {
...
+		if bundleID != wkbID {
+			err = fmt.Errorf("Expected writer bundle ID %s, got: %s",
+				wkbID, bundleID)
+			return nil, nil, err
+		}
+	}
+
+	var rkb *TLFReaderKeyBundleV3
+	if response.ReaderBundle.Bundle != nil {
+		if response.ReaderBundle.Version != int(SegregatedKeyBundlesVer) {
```
jzila: `<`?

libkbfs/mdserver_remote.go
```diff
@@ -836,9 +860,68 @@ func (md *MDServerRemote) backgroundRekeyChecker(ctx context.Context) {
...
+			err = fmt.Errorf("Unsupported reader bundle version: %d",
+				response.ReaderBundle.Version)
+			return nil, nil, err
+		}
+		rkb := new(TLFReaderKeyBundleV3)
```
jzila: As above, should be `=` instead of `:=`.

libkbfs/root_metadata.go
```diff
@@ -694,7 +699,15 @@ func (md *RootMetadata) fillInDevices(crypto Crypto,
...
 func (md *RootMetadata) finalizeRekey(
 	crypto cryptoPure, prevKey, currKey kbfscrypto.TLFCryptKey) error {
-	return md.bareMd.FinalizeRekey(crypto, prevKey, currKey, md.extra)
+	wkbID := md.bareMd.GetTLFWriterKeyBundleID()
+	rkbID := md.bareMd.GetTLFReaderKeyBundleID()
+	err := md.bareMd.FinalizeRekey(crypto, prevKey, currKey, md.extra)
+	if err == nil && md.extra != nil {
+		includeWkb := wkbID != md.bareMd.GetTLFWriterKeyBundleID()
+		includeRkb := rkbID != md.bareMd.GetTLFReaderKeyBundleID()
+		md.extraNew = md.extra.Copy(includeWkb, includeRkb)
```
aalness: It's a simple filter for which key bundles are new and need uploading so setting the pointers is fine.

libkbfs/mdserver_remote.go
aalness: For now I'm assuming if there are external key bundles they must be v3.

libkbfs/mdserver_remote.go
aalness: Same here re: assuming v3.

aalness:
I think this is ready for another pass too

PR #440, opened 2016-10-19
Dokan be stricter about files vs directories

taruti:
+ Sometimes returning either a file or directory is ok
+ Sometimes returning only a file is ok
+ Sometimes returning only a directory is ok

Thus:
+ Add extra checks for debug builds
+ Add two helper methods to dokan
+ Add more checking to libdokan


zanderz:
LGTM :thumbsup:

PR #462, opened 2016-10-25
mdserver_remote: support MDv3

aalness:


aalness:
Note this just re-adds the previous change but makes the following change ontop: https://github.com/keybase/kbfs/pull/462/commits/10cf2da5c367feab89a3fe5fae2404dbe5588528

PR #463, opened 2016-10-25
Fix flakes

jzila:


jzila:
Merging since this seems to have fixed most of the flakes, determined by running it many times.

strib:
@jzila: I would have liked a chance to review/discuss this.  I don't understand why this fixes anything.  As far as I know, `BServerRemote.RefreshAuthToken` isn't called by anything -- `blockServerRemoteClientHandler` is what's passed in as the callback handler to `NewAuthToken`.  I think I just forgot to delete `BServerRemote.RefreshAuthToken` when I changed the code to have multiple tokens.  Are you sure this is the fix?

jzila:
As mentioned in slack, it's called [here](https://github.com/keybase/kbfs/blob/550baee383edfb483c1e062354879c75662d79de/libkbfs/keybase_service_util.go#L33-L44).

PR #468, opened 2016-10-27
Include filename in ReadAccessError

gabriel:
We show the path that failed to read in the notification. This includes it in the ReadAccessError.


gabriel:
Ready for review.

PR #469, opened 2016-10-27
Use service RPC to get the mount dir

zanderz:
Is this the right way to hook up a new client RPC?
There is a chicken and egg issue with the matching client changes on branch zanderz/CORE-4000
I think the drive letter one should be renamed so it could be more meaningful on other platforms.
@strib @taruti @maxtaco


taruti:
Creating the mounter inside libdokan.Start instead of kbfsdokan.main would allow to do the rpc before creating the mounter. Thus removing the need for SetDir.

genmocks.sh needs to be run inside libkbfs.

GetMountDir needs a godoc comment in all three definitions to keep linters happy.

Are we happy with the term DriveLetter in function names? How about MountPoint instead?

kbfsdokan/main.go
```diff
@@ -25,6 +25,7 @@ var mountType = flag.String("mount-type", defaultMountType, "mount type: default
 var version = flag.Bool("version", false, "Print version")
 var mountFlags = flag.Int64("mount-flags", int64(libdokan.DefaultMountFlags), "Dokan mount flags")
 var dokandll = flag.String("dokan-dll", "", "Absolute path of dokan dll to load")
+var servicemount = flag.Bool("service-mount", false, "get mount path from serfice")
```
strib: "serfice" -> "service"

kbfsdokan/main.go
```diff
@@ -33,14 +34,14 @@ To run against remote KBFS servers:
   kbfsdokan [-debug] [-cpuprofile=path/to/dir]
     [-bserver=%s] [-mdserver=%s]
     [-runtime-dir=path/to/dir] [-label=label] [-mount-type=force]
-    [-log-to-file] [-log-file=path/to/file]
+    [-log-to-file] [-log-file=path/to/file][-service-mount]
     /path/to/mountpoint
```
strib: Now "/path/to/mount" is optional, depending on whether "-service-mount" is given, right? So I think it should be surrounded with square brackets, or else have another usage line showing "-service-mount" without an explicit mount passed in.

kbfsdokan/main.go
strib: Also, flags shared among the different binaries should go in `AddFlags` in `libkbfs/init.go`.

libdokan/mounter.go
```diff
@@ -113,24 +119,28 @@ type NoopMounter struct {
 }
 
 // NewNoopMounter creates a mounter that does nothing.
-func NewNoopMounter() NoopMounter {
-	return NoopMounter{c: make(chan struct{}, 1)}
+func NewNoopMounter() *NoopMounter {
```
strib: Why did you make this return a pointer?  I don't think there's any reason to for `NoopMounter` to have pointer receivers, since none of its fields are ever modified.

libkbfs/interfaces.go
```diff
@@ -349,6 +349,8 @@ type KeybaseService interface {
 	// instance. No other methods may be called after this is
 	// called.
 	Shutdown()
+
+	GetMountDir(ctx context.Context) (string, error)
```
strib: Please add a comment explaining what this does.

kbfsfuse/main.go
```diff
@@ -24,6 +24,7 @@ var runtimeDir = flag.String("runtime-dir", os.Getenv("KEYBASE_RUNTIME_DIR"), "r
 var label = flag.String("label", os.Getenv("KEYBASE_LABEL"), "label to help identify if running as a service")
 var mountType = flag.String("mount-type", defaultMountType, "mount type: default, force, none")
 var version = flag.Bool("version", false, "Print version")
+var servicemount = flag.Bool("service-mount", false, "get mount path from serfice")
```
maxtaco: There's a typo here --- `serfice` ?

maxtaco: Let's just keep this feature windows-only for now, and we can add it to linux/OSX if we need it.

@strib the feature is that we need to enable the changing of the "drive letter" from the Electron interface.  So therefore need a way to update the configuration and then relaunch kbfsfuse with the new parameter.  The complication is that a process called the "watchdog" launches kbfsfuse on windows and can't conveniently access keybase configuration.  So this technique seemed simplest --- to ask for the service where to mount on startupt.

zanderz: That was my question, if we ever want to change it on other platforms, "drive letter" will be the wrong thing to call it.

zanderz:
@taruti The mounter is needed before connecting to the service because of `onInterruptFn`. I got rid of SetDir() but maybe there is another way.

zanderz:
Sorry about all the thrashing, I think this is done finally. @strib ?

libkbfs/choosedefaultmount_windows.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func chooseDefaultMount(dirs []string) (string, error) {
+	var dir string
+	if len(dirs) == 0 {
+		return "K:", errors.New("chooseDefaultMount fails - nothing to choose from")
+	}
+	for _, dir = range dirs {
+		// Try to use a drive at K or later
+		if strings.Compare(strings.ToUpper(dir), "K:") >= 0 {
```
strib: You can just use `if strings.ToUpper(dir) >= "K:"` (according to the [docs](https://golang.org/pkg/strings/#Compare)).

libkbfs/interfaces.go
```diff
@@ -349,6 +349,10 @@ type KeybaseService interface {
 	// instance. No other methods may be called after this is
 	// called.
 	Shutdown()
+
+	// EstablishMountDir asks the service for the current mount path
+	// and sets it if not establishe
```
strib: "establishe" -> "established." (with the period).

libkbfs/interfaces.go
```diff
@@ -349,6 +349,10 @@ type KeybaseService interface {
...
+	EstablishMountDir(ctx context.Context) (string, error)
```
strib: Also, we like to make `Shutdown` the last thing in an interface, so please move this above it.

libkbfs/keybase_daemon_local.go
```diff
@@ -495,3 +495,9 @@ func newKeybaseDaemonLocal(codec kbfscodec.Codec,
 		favoriteStore: favoriteStore,
 	}
 }
+
+// EstablishMountDir asks the service for the current mount path
+// and sets it if not established
+func (k *KeybaseDaemonLocal) EstablishMountDir(ctx context.Context) (string, error) {
```
strib: Please group this with the other `KeybaseDaemonLocal` functions (above `Shutdown`).

libkbfs/keybase_daemon_local.go
```diff
@@ -495,3 +495,9 @@ func newKeybaseDaemonLocal(codec kbfscodec.Codec,
 		favoriteStore: favoriteStore,
 	}
 }
+
+// EstablishMountDir asks the service for the current mount path
```
strib: It's a stupid convention, but interface implementors in KBFS use comments of this form instead:

```go
// EstablishMountDir implements the KeybaseDaemon interface for KeybaseDaemonLocal.
```

libkbfs/keybase_service_base.go
```diff
@@ -681,3 +684,19 @@ func (k *KeybaseServiceBase) GetPublicCanonicalTLFNameAndID(
...
 }
+
+// EstablishMountDir asks the service for the current mount path
+func (k *KeybaseServiceBase) EstablishMountDir(ctx context.Context) (
+	string, error) {
+	dir, err := k.kbfsMountClient.GetCurrentMountDir(ctx)
+	if err != nil || dir == "" {
+		dirs, _ := k.kbfsMountClient.GetAllAvailableMountDirs(ctx)
+		// chooseDefaultMount can handle an emtpy argument
+		dir, err = chooseDefaultMount(dirs)
```
strib: It's strange to ignore the error here, but pass it along to the caller (who fails when they get the error). Is the overall mount operation supposed to succeed when `dirs` is empty or is it supposed to fail?  If the former, you'll need to fix up the caller of this function.  If the later, you should return immediately when you hit an error.

libkbfs/keybase_service_base.go
```diff
@@ -681,3 +684,19 @@ func (k *KeybaseServiceBase) GetPublicCanonicalTLFNameAndID(
...
 	return res, nil
 }
+
+// EstablishMountDir asks the service for the current mount path
+func (k *KeybaseServiceBase) EstablishMountDir(ctx context.Context) (
+	string, error) {
+	dir, err := k.kbfsMountClient.GetCurrentMountDir(ctx)
+	if err != nil || dir == "" {
+		dirs, _ := k.kbfsMountClient.GetAllAvailableMountDirs(ctx)
+		// chooseDefaultMount can handle an emtpy argument
```
strib: "emtpy" -> "empty"

libkbfs/keybase_service_measured.go
```diff
@@ -165,3 +165,9 @@ func (k KeybaseServiceMeasured) FlushUserUnverifiedKeysFromLocalCache(
 func (k KeybaseServiceMeasured) Shutdown() {
 	k.delegate.Shutdown()
 }
+
+// EstablishMountDir asks the service for the current mount path
+// and sets it if not established
```
strib: Please use the convention of the comments of the other methods in this file.

libkbfs/keybase_service_base.go
```diff
@@ -681,3 +684,19 @@ func (k *KeybaseServiceBase) GetPublicCanonicalTLFNameAndID(
 
 	return res, nil
 }
+
+// EstablishMountDir asks the service for the current mount path
+func (k *KeybaseServiceBase) EstablishMountDir(ctx context.Context) (
+	string, error) {
+	dir, err := k.kbfsMountClient.GetCurrentMountDir(ctx)
+	if err != nil || dir == "" {
```
strib: Are these two cases really equivalent? If err != nil shouldn't that be a fatal error and we should return immediately?

libkbfs/keybase_service_base.go
```diff
@@ -681,3 +684,19 @@ func (k *KeybaseServiceBase) GetPublicCanonicalTLFNameAndID(
...
+		dirs, _ := k.kbfsMountClient.GetAllAvailableMountDirs(ctx)
```
strib: Please don't ignore the error here.

libkbfs/choosedefaultmount_windows.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"errors"
+	"strings"
+)
+
+func chooseDefaultMount(dirs []string) (string, error) {
+	var dir string
+	if len(dirs) == 0 {
+		return "K:", errors.New("chooseDefaultMount fails - nothing to choose from")
```
zanderz: I intended for this to fall back to what we do now, in case the detection/availability logic failed. K: may accidentally work, where "" definitely will not - I don't see the disadvantage of using K: here. But I'll change it.

libkbfs/keybase_service_base.go
zanderz: This won't be true anymore without the K: default anyway

libkbfs/choosedefaultmount_windows.go
strib: I'm fine with using "K:" as a default, but in that case the return error should be `nil`, or the function shouldn't return an error at all.  It's bad Go to return a usable value along with an error.

libkbfs/keybase_service_base.go
zanderz: The job here is "if we can't get the current mount dir, pick one from those available and save it", so I don't see why we would bail out on an error here, but I will make it do so.

strib: I'm worried that `err` could be something fatal, like "You don't have permission to connect to the service" or "This is not a Windows system so clearly you are trying to run a Windows binary on the wrong system".  If there's a specific error that should be treated non-fatally, you should check for it explicitly.

PR #486, opened 2016-11-04
MDv3: use KBFS prefixed signatures for v3 metadata

aalness:
See also: https://github.com/keybase/client/pull/4792


kbfscrypto/signature.go
```diff
@@ -92,6 +94,19 @@ func (k SigningKey) Sign(data []byte) SignatureInfo {
 	}
 }
 
+// SignForKBFS signs the given data with the KBFS prefix and returns a SignatureInfo.
+func (k SigningKey) SignForKBFS(data []byte) (SignatureInfo, error) {
+	sigInfo, err := k.kp.SignV2(data, libkb.SignaturePrefixKBFS)
```
jzila: Why isn't this using the crypto client? Is this only for tests?

aalness: Yup. Just for `CryptoLocal`.

PR #488, opened 2016-11-07
MDv3: fixes for issues exposed by server-side testing

aalness:


libkbfs/root_metadata_test.go
```diff
@@ -739,3 +740,33 @@ func TestRootMetadataUpconversionPublic(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, handle, handle2)
 }
+
+func TestRootMetadataV3NoPanicOnWriterMismatch(t *testing.T) {
```
strib: We want this because the server will be reusing `IsLastModifiedBy` and we don't want a client to be able to construct an MD that will crash the server?  Might be worth mentioning this as a test comment.

PR #489, opened 2016-11-09
crypto_client: save correct signature type when signing for KBFS

aalness:


aalness:
The server rejects the metadata block as having an invalid signature. I think the only good test for it will be with kbfsdocker since that's the only time we actually use the real client.

strib:
Cool.  Seems like our in-memory servers should do those same checks, so this would be caught in our `kbfs` repo tests.  But no need to do that before merging this.

aalness:
Those would use `CrytpoLocal` though which doesn't have the bug.

strib:
Ahhh makes sense.  Thanks.

PR #490, opened 2016-11-09
Revert "Revert Use preferred names for tlf names - user's own usernam…

taruti:
Here we go again.

This reintroduces the reverted preferred names. Synchronous changes to kbfs-server are needed.


taruti:
Will delete the Jenkinsfile commit before merge.

taruti:
CI passes will now delete the additional commit and merge both kbfs and kbfs-server - might be 2 minutes of CI not passing.

PR #491, opened 2016-11-09
libkbfs: use MDv3 in staging env

aalness:
This actually does 3 things:

* Fixes a bug where we were dereferencing nil `ExtraMetadata` for public folders.
* Adds a cmdline option for which metadata version to use
* Makes the default version of the above v3 when connecting to staging.

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -553,8 +553,8 @@ func (c *ConfigLocal) SetConflictRenamer(cr ConflictRenamer) {
 
 // MetadataVersion implements the Config interface for ConfigLocal.
 func (c *ConfigLocal) MetadataVersion() MetadataVer {
-	c.lock.Lock()
-	defer c.lock.Unlock()
+	c.lock.RLock()
+	defer c.lock.RUnlock()
 	return c.metadataVersion
 }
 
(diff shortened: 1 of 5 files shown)


libkbfs/config_local.go
```diff
@@ -553,9 +553,18 @@ func (c *ConfigLocal) SetConflictRenamer(cr ConflictRenamer) {
 
 // MetadataVersion implements the Config interface for ConfigLocal.
 func (c *ConfigLocal) MetadataVersion() MetadataVer {
+	c.lock.Lock()
```
jzila: `RLock()`/`RUnlock()`

libkbfs/init.go
```diff
@@ -94,6 +98,18 @@ func GetDefaultMDServer(ctx Context) string {
...
 
+// GetDefaultMetadataVersion returns the default metadata version per run mode.
+func GetDefaultMetadataVersion(ctx Context) MetadataVer {
+	switch ctx.GetRunMode() {
+	case libkb.StagingRunMode:
+		return SegregatedKeyBundlesVer
+	case libkb.ProductionRunMode:
+		return InitialExtraMetadataVer
+	default:
+		return InitialExtraMetadataVer
```
jzila: What about dev? Doesn't that need to be set to `SegregatedKeyBundlesVer` for kbfsdocker?

aalness: Hmm. I didn't want to switch it over yet while we're still using v2 in prod. To run kbfsdocker I manually hardcoded to v3.

jzila: Given that we're preparing for it to be in prod, I'd say that it might be a good idea to enable kbfsdocker for md v3

aalness: Is the default run mode dev? Because I don't want to enable it for all of the libkbfs tests yet.

aalness: (merging this now can continue discussing enabling it for dev)

PR #495, opened 2016-11-10
FSNotification filename should use canonical path

gabriel:
Sends full canonical path for FSNotifications.

The desktop app has a fix to use this since it knowns FSNotification.Filename is always the full canonical path: https://github.com/keybase/client/issues/4833


gabriel:
@strib?

PR #511, opened 2016-11-17
don't always hit remote in GetTLFCryptKeys

songgao:


libkbfs/kbfs_ops.go
```diff
@@ -256,26 +256,53 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
...
+		lState := makeFBOLockState()
+		rmd = fbo.getHead(lState)
+	}
+	if rmd == (ImmutableRootMetadata{}) {
+		_, rmd, id, err = fs.getOrInitializeNewMDMaster(
+			ctx, fs.config.MDOps(), tlfHandle, true)
+		if err != nil {
+			return ImmutableRootMetadata{}, tlf.ID{}, err
+		}
+		if fbo != nil {
```
strib: To get any benefit, I think if `fbo == nil && !initialized`, you're going to have to make a new `fbo` here (e.g.:
```go
fb := FolderBranch{Tlf: id, Branch: MasterBranch}
fbo := fs.getOpsByHandle(ctx, tlfHandle, fb)
```

Otherwise you'll still be fetching from the server each time, unless you are creating the TLF for the first time.

libkbfs/kbfs_ops.go
```diff
@@ -256,26 +256,53 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
...
+func (fs *KBFSOpsStandard) getMDByHandle(ctx context.Context,
+	tlfHandle *TlfHandle) (rmd ImmutableRootMetadata, id tlf.ID, err error) {
+	fbo := func() *folderBranchOps {
+		fs.opsLock.Lock()
+		defer fs.opsLock.Unlock()
+		return fs.opsByFav[tlfHandle.ToFavorite()]
+	}()
+	if fbo != nil {
+		lState := makeFBOLockState()
+		rmd = fbo.getHead(lState)
```
strib: You need to also set `id` in this case, I think that's what's breaking the kbfsdocker tests.

songgao: Ended up removing the `id` field ...

strib:
On 11/17/2016 04:52 PM, Song Gao wrote:
>
> I just realized |getOrInitializeNewMDMaster| already does
> |SetInitialHeadToNew|, so this seems unnecessary whether |fbo == nil|
> or not. Sounds correct?
>

No, this isn't correct.  It only sets the head if you're creating the
folder for the first time ever.  You need to set the head as you were
doing before for folders that weren't newly-initialized.

songgao:
@strib Make sense. Thanks for explaining it!

PR #516, opened 2016-11-21
clean block cache capacity: add CLI flag; set default based on total RAM

songgao:


libkbfs/bcache.go
```diff
@@ -141,6 +142,18 @@ func (b *BlockCacheStandard) CheckForKnownPtr(tlf tlf.ID, block *FileBlock) (
 	return ptr, nil
 }
 
+// SetCleanBytesCapacity implements the BlockCache interface for
+// BlockCacheStandard.
+func (b *BlockCacheStandard) SetCleanBytesCapacity(capacity uint64) {
```
strib: The `init.go` code calls this unconditionally.  Shouldn't this function (or maybe the caller) check that `capacity != 0`?

libkbfs/config_local.go
```diff
@@ -208,6 +209,15 @@ func MakeLocalUsers(users []libkb.NormalizedUsername) []LocalUser {
 	return localUsers
 }
 
+func getDefaultCleanBlockCacheCapacity() uint64 {
+	vmstat, err := mem.VirtualMemory()
+	if err == nil {
+		// By default,  we try to cap at 1/8 of available memory
```
strib: Super minor nits: looks like you have two spaces after "By default,", when you only need one.  Also. the comment should end in a period.

libkbfs/config_local.go
```diff
@@ -208,6 +209,15 @@ func MakeLocalUsers(users []libkb.NormalizedUsername) []LocalUser {
...
+		return vmstat.Total / 8
```
strib: I'm not sure we should get too greedy here.  For example, on my desktop, this would take 4 GB for KBFS.  Maybe we should cap the memory at `MaxBlockSizeBytesDefault * 1024`?  I don't feel too strongly about it though, maybe it's ok to use that much if we have lots of memory.

libkbfs/config_local.go
```diff
@@ -643,7 +653,7 @@ func (c *ConfigLocal) resetCachesWithoutShutdown() DirtyBlockCache {
 	c.kcache = NewKeyCacheStandard(defaultMDCacheCapacity)
 	c.kbcache = NewKeyBundleCacheStandard(defaultMDCacheCapacity * 2)
 	// Limit the block cache to 10K entries or 1024 blocks (currently 512MiB)
-	c.bcache = NewBlockCacheStandard(10000, MaxBlockSizeBytesDefault*1024)
+	c.bcache = NewBlockCacheStandard(10000, getDefaultCleanBlockCacheCapacity())
```
strib: Can you please print out what you're setting it to?  You can use `c.MakeLogger().CDebugf(nil, ...)`.

libkbfs/init.go
```diff
@@ -305,6 +310,8 @@ func Init(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn, onI
 
 	config := NewConfigLocal()
 
+	config.BlockCache().SetCleanBytesCapacity(params.CleanBlockCacheCapacity)
```
strib: Please print out here if we're overriding the default.

libkbfs/bcache.go
songgao: Oops yes thanks!

libkbfs/config_local.go
songgao: Got you; I think it's reasonable. If use wants to increase to cap, she can always set it from commandline. Since we are capping with `MaxBlockSizeBytesDefault * 1024` too, I'm changing `/8` to `/4` to loosen up the lower-end. Let me know if you think it's too big!

songgao: Ah with the dirty bcache also set to the same value, I guess it makes more sense to keep `/4`. Never mind!

PR #535, opened 2016-11-29
don't bust if tlfBreak is nil

songgao:


libkbfs/identify_util.go
```diff
@@ -58,22 +61,29 @@ func (ei *extendedIdentify) makeTlfBreaksIfNeeded(
...
 
 // getTlfBreakOrBust returns a keybase1.TLFBreak. This should only be called
 // for behavior.WarningInsteadOfErrorOnBrokenTracks() == true, and after
-// makeTlfBreaksIfNeeded is called. Otherwise it panics.
-func (ei *extendedIdentify) getTlfBreakOrBust() keybase1.TLFBreak {
+// makeTlfBreaksIfNeeded is called, to make sure user proof breaks get
+// populated in GUI mode.
+//
+// If called otherwise, we don't panic here anymore, since we can't panic on
+// nil ei.tlrBreaks. The reason is if a previous successful identify has
```
strib: `tlrBreaks` -> `tlfBreaks`

libkbfs/identify_util.go
```diff
@@ -58,22 +61,29 @@ func (ei *extendedIdentify) makeTlfBreaksIfNeeded(
...
+// already happened recently, it could cause this identify to be skipped, which
+// means ei.tlfBreaks is never populated. In this case, it's safe to return an
+// empty keybase1.TLFBreak.
+func (ei *extendedIdentify) getTlfBreakAndClose() keybase1.TLFBreak {
 	ei.tlfBreaksLock.Lock()
 	defer ei.tlfBreaksLock.Unlock()
-	close(ei.userBreaks)
-	return *ei.tlfBreaks
+	if ei.userBreaks != nil {
```
strib: The comment implies that `tlfBreaks` would be `nil`, but you're actually checking `userBreaks`.  This seems to say that the context was not returned from `makeExtendedIdentify` or that `behavior.WarningInsteadOfErrorOnBrokenTracks()` was `false`.  Could it be that the GUI behavior you were testing doesn't have the correct behavior?

songgao:
Thanks Strib!

PR #537, opened 2016-11-29
Windows DACLs and winacl debugging

taruti:
+ Return a short DACL if requested GetFileSecurity
+ Easier debugging of winacl
+ Nothing changes for most things


zanderz:
ouch, looks painful, thanks for doing ACLs. LGTM :thumbsup:
Is this why drag and drop from explorer is having troubles?

taruti:
Most of the explorer bug reports from recent days and copying files are just write throttling being ineffective and us hitting timeouts and returning a write error.

PR #542, opened 2016-11-29
folder_block_manager: use LastGCRevision instead of scanning

strib:
Instead of scanning backwards until we find the latest GC op, we
instead use a field embedded in the private metadata if it's
available.

Note that we still have to scan to find the most recent old-enough
revision, but we now limit that scan.  And that's much less of a
problem, since it's guaranteed that as soon as the client stops
writing for 5 minutes it will be able to find an old-enough revision.
And it /really/ shouldn't be a problem since clients won't even do QR
if the head is too recent anyway.

Issue: KBFS-1734


aalness:
:+: nice

PR #549, opened 2016-12-02
[KBFS-1740] Fix common block retrieval panic

jzila:
Added fix and regression test for retrieving the same block pointer into different block types. The test panics without the fix (with the same error message as Patrick's log), and passes with it.

### DIFF
--- a/libkbfs/block_retrieval_worker_test.go
+++ b/libkbfs/block_retrieval_worker_test.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/keybase/kbfs/kbfscodec"
 	"github.com/stretchr/testify/require"
@@ -19,8 +18,9 @@ import (
 // blockReturner contains a block value to copy into requested blocks, and a
 // channel to synchronize on with the worker.
 type blockReturner struct {
-	block Block
-	ch    chan struct{}
+	block      Block
+	continueCh chan struct{}
+	startCh    chan struct{}
 }
 
 // fakeBlockGetter allows specifying and obtaining fake blocks.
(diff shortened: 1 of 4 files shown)


libkbfs/block_retrieval_worker_test.go
```diff
@@ -251,3 +258,46 @@ func TestBlockRetrievalWorkerShutdown(t *testing.T) {
...
+	err = codec.Decode(bytes, testCommonBlock)
+	require.NoError(t, err)
+
+	t.Log("Make a retrieval for the same block twice, but with a different target block type.")
+	testBlock1 := &FileBlock{}
+	testBlock2 := &CommonBlock{}
+	req1Ch := q.Request(context.Background(), 1, nil, ptr1, testBlock1)
+	req2Ch := q.Request(context.Background(), 1, nil, ptr1, testBlock2)
+	// Ensure the worker picks up the request
+	time.Sleep(50 * time.Millisecond)
```
strib: Oof, I'd rather not have sleeps in tests.  (I guess I missed the one in `TestBlockRetrievalWorkerWithQueue` on a previous review.)

Seems like you can get around this pretty easily by having `fakeBlockGetter.getBlock` notify an optional unbuffered channel that it's received a request.

libkbfs/block_retrieval_worker_test.go
```diff
@@ -251,3 +258,46 @@ func TestBlockRetrievalWorkerShutdown(t *testing.T) {
...
+	w1 := newBlockRetrievalWorker(bg, q)
+	require.NotNil(t, w1)
+	defer w1.Shutdown()
+
+	t.Log("Setup source blocks")
+	ptr1 := makeFakeBlockPointer(t)
+	block1 := makeFakeFileBlock(t)
+	ch1 := bg.setBlockToReturn(ptr1, block1)
+	testCommonBlock := &CommonBlock{}
+	bytes, err := codec.Encode(block1)
```
strib: Again you can probably use `kbfscodec.Update` here instead of two separate calls to Encode/Decode.

jzila:
@strib: ready for another look

libkbfs/block_retrieval_worker_test.go
```diff
@@ -61,20 +62,15 @@ func (bg *fakeBlockGetter) getBlock(ctx context.Context, kmd KeyMetadata, blockP
...
-		bytes, err := bg.codec.Encode(source.block)
-		if err != nil {
-			return err
-		}
-		err = bg.codec.Decode(bytes, block)
-		if err != nil {
-			return err
+	for {
+		select {
+		case source.startCh <- struct{}{}:
```
strib: I'm confused as to why this works.  Don't we need to do the `startCh` (if possible), and then wait for the `continueCh`?  With this construction, it seems like we'd do the start without the continue, or vice versa.

jzila: I wanted to make the `startCh` wait optional.

jzila: This way, you can wait on `startCh` as many times as you want (although just waiting once will tell you what you need to know), and then when you tell the `continueCh` to continue, the request will finish.

strib: Ah I didn't notice the addition of the `for` loop around the `select`.  Seems fine.

jzila:
Appveyor install flake, merging.

PR #552, opened 2016-12-02
vendor update github.com/keybase/client/go/libkb for zanderz/CORE-3904

zanderz:
Depends on the matching core PR. This is for moving the target install/data directory from roaming to local 
@maxtaco @taruti


maxtaco:
👍

PR #553, opened 2016-12-03
file_data: use by final libkbfs stragglers

strib:
* `StateChecker` should use it for fetching blocks and block sizes.
* `RootMetadata` should use it when loading unembedded block changes into the reference list.

Note that there is one final direct use of `FileBlock.IPtrs` outside of `libkbfs`, in `kbfstool.md_check`, but it will already work with multiple levels of indirection and I don't really want to export `fileData` at the moment, so I'm going to leave it alone for now.

Issue: KBFS-35

### DIFF
--- a/libkbfs/root_metadata.go
+++ b/libkbfs/root_metadata.go
@@ -412,8 +412,18 @@ func (md *RootMetadata) loadCachedBlockChanges(
 		return
 	}
 
+	if len(md.data.cachedChanges.Ops) == 0 {
+		panic("MD with no ops passed to loadCachedBlockChanges")
+	}
+
 	md.data.Changes, md.data.cachedChanges =
 		md.data.cachedChanges, md.data.Changes
+
+	// We always add the ref blocks to the first operation in the MD
+	// update.  Most MD updates will only have one op anyway, and for
+	// those that have more (like conflict resolution), it is
+	// arbitrary which one lists them as references, so putting them
+	// in the first op is the easiest thing to do.
 	md.data.Changes.Ops[0].
 		AddRefBlock(md.data.cachedChanges.Info.BlockPointer)
 	// Find the block and ref any children, if any.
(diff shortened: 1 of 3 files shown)


libkbfs/root_metadata.go
```diff
@@ -406,31 +406,60 @@ func (md *RootMetadata) updateFromTlfHandle(newHandle *TlfHandle) error {
...
-				md.data.Changes.Ops[0].AddRefBlock(iptr.BlockPointer)
+	}
+
+	// uid, crypto, bsplitter and log aren't used for simply getting the
+	// indirect pointers, so set them to nil.
+	var uid keybase1.UID
+	file := path{FolderBranch{md.TlfID(), MasterBranch},
+		[]pathNode{{
+			md.data.cachedChanges.Info.BlockPointer,
+			fmt.Sprintf("<MD with revision %d>", md.Revision())}}}
```
jzila: Could you unwrap the spacing here to make it easier to read?
```go
file := path{
            FolderBranch{...},
            []pathNode{{
                ...
            }},
        }
```

libkbfs/root_metadata.go
```diff
@@ -406,31 +406,60 @@ func (md *RootMetadata) updateFromTlfHandle(newHandle *TlfHandle) error {
...
+
+	infos, err := fd.getIndirectFileBlockInfos(ctx)
+	if err != nil {
+		panic(fmt.Sprintf(
+			"Couldn't find all unembedded change blocks for %v: %v",
+			md.data.cachedChanges.Info.BlockPointer, err))
+	}
+
+	for _, info := range infos {
+		md.data.Changes.Ops[0].AddRefBlock(info.BlockPointer)
```
jzila: I'm a bit confused why we're guaranteed that only `Ops[0]` needs to modified here. Is there a comment somewhere that explains what's happening?

strib: It's arbitrary, and 0 is the easiest one to code/read.  I added a comment:

```go
	// We always add the ref blocks to the first operation in the MD
	// update.  Most MD updates will only have one op anyway, and for
	// those that have more (like conflict resolution), it is
	// arbitrary which one lists them as references, so putting them
	// in the first op is the easiest thing to do.
```

PR #558, opened 2016-12-08
kbfsfuse: support testing with mdv3

aalness:


kbfsfuse/Dockerfile
```diff
@@ -6,6 +6,7 @@ ARG KEYBASE_TEST_ROOT_CERT_PEM_B64
 
 ENV KEYBASE_TEST_ROOT_CERT_PEM=$KEYBASE_TEST_ROOT_CERT_PEM \
     KEYBASE_TEST_ROOT_CERT_PEM_B64=$KEYBASE_TEST_ROOT_CERT_PEM_B64 \
+    KBFS_METADATA_VERSION=$(KBFS_METADATA_VERSION) \
```
jzila: I think you need this to not have parens. Won't that try to execute a command?

aalness: Huh, weird. `kbfsfuse.sh` definitely spits out the expected version.

PR #565, opened 2016-12-09
Add a mock Trash on macOS and some special dir/file handling

songgao:
This adds a mock Trash that provide a symlink from what macOS uses for Trash into a directory within the private TLF of the user. This won't work for other TLFs, but should be helpful for testing macOS applications.

The mock Trash, `local` mount option, as well as other special dir/file handling, are only enabled when the `-local-experimental` CLI flag is provided.


songgao:
@strib Please hold off for review. On macOS the `keybase/kbfs/test` tests (with `-tags fuse`) run successful but leave a bunch of "KBFUSE Volume" (umounted) in Devices sections of the left sidebar in Finder. Might be a side-effect from this PR. I'll look into this and let you know!

songgao:
Turns out it's the `libfuse` tests that are leaving stuff in Finder sidebar. Looks like issues with unmounting volume from `kbfsfuse` process. I've disabled the `local` flag in tests. So this is ready for review again. Thanks!

libfuse/alias.go
```diff
@@ -12,10 +12,12 @@ import (
 	"golang.org/x/net/context"
 )
 
-// Alias is a top-level folder accessed through its non-canonical name.
+// Alias represents an alias. A use case for it is a top-level folder accessed
+// through its non-canonical name.
 type Alias struct {
-	// canonical name for this folder
-	canon string
+	// the real path this alias points to. In case of TLF alias, this is the
```
strib: Since you added periods, please start the comment with a capital "T".

libfuse/fs.go
```diff
@@ -300,7 +303,7 @@ func (r *Root) ReadDirAll(ctx context.Context) (res []fuse.Dirent, err error) {
 			Type: fuse.DT_Dir,
 			Name: PublicName,
 		},
-	}
+	}, platformRootDirs...)
```
strib: We shouldn't show these unless `r.platformParams.UseLocal` is set, right?  I don't think we want to affect the current behavior yet.

libfuse/fs_darwin.go
```diff
@@ -43,3 +106,39 @@ func bundleResourcePath(path string) (string, error) {
 	}
 	return filepath.Join(execPath, "..", "..", "..", "Resources", path), nil
 }
+
+// Trash is a mock .Trashes directory
```
strib: Please explain that for now it just links to a directory under the user's private home directory, and a TODO for how we might fix this in the future.

libfuse/mounter_osx.go
```diff
@@ -51,6 +54,7 @@ func GetPlatformSpecificMountOptionsForTest() []fuse.MountOption {
 	return []fuse.MountOption{
 		fuse.OSXFUSELocations(kbfusePath, fuse.OSXFUSELocationV3),
 		fuse.ExclCreate(),
+		// fuse.LocalVolume(),
```
strib: Please add a comment as to why this is disabled, and a TODO for getting it fixed.  Without the tests passing with this option enabled, I don't think we should consider KBFS-1751 finished.

libfuse/platform_flags_osx.go
```diff
@@ -27,5 +28,8 @@ func AddPlatformFlags(flags *flag.FlagSet) *PlatformParams {
 	var params PlatformParams
 	flags.BoolVar(&params.UseSystemFuse, "use-system-fuse", false,
 		"Use the system OSXFUSE instead of keybase's OSXFUSE")
+	flags.BoolVar(&params.UseLocal, "local-experimental", false,
+		"Use 'local' mount option"+
+			" and enable other hacky stuff for testing macOS apps.")
```
strib: Maybe explain that for now, with this flag you can only work within your private home dir?

libfuse/fs.go
songgao: Yes; good point!

libfuse/mounter_osx.go
songgao: Tests are passing; they just leave behind many items under Devices in the Finder sidebar. But I'll try to fix that

songgao:
@strib I did a couple more tests. The left-overs are not specific to one particular test, but kind of flaky. For example, `./libfuse.test -test.run=TestKbfsFileInfo` normally doesn't leave anything behind in Finder, but `for i in $(seq 1 20); do ./libfuse.test -test.run=TestKbfsFileInfo; done` normally do. It also doesn't seem to be correlated to the error message `unmount error: unmount /var/folders/ny/cd2ss6w14_zgjr3gyyl2hmbw0000gn/T/fusetest646361530: resource busy`, i.e., they appear independently.

Generally, the left-overs only show up under Devices in the Finder's sidebar, but not anywhere else (e.g. output of `mount`, in Finder's "root" device (showing as "Song's MacBook Pro" and normally would include all mounts). I said generally because I did see two test mounts showing up in output of `mount` command, but couldn't reproduce it.

This feels like a bug of Finder itself to me. What do you think?

songgao:
I just saw this happen again, but without the `local` flag. After the `kbfsfuse` process exits, there's a leftover "mount point" in "Song's Macbook Pro - Keybase", along with my other normal mounts. Clicking into it shows private/public, and the TLF inside. But it doesn't properly show the content. It looks like it has something to do with Finder caching the directories.

But anyway it looks like an issue unrelated to the `local` flag, but how Finder handles mounts.

songgao:
hmm actually this time the mount shows up in `mount` output as well. Although no kbfsfuse process was running

PR #580, opened 2016-12-14
cr: fetch ref and unref blocks in parallel

strib:
If conflict resolution has to fetch a lot of blocks to do block accounting, it's best to fetch them in parallel to make it as fast as possible.

Note that once #579 goes in, this won't be as big an issue when journaling is enabled since all of referenced blocks should be locally available on disk.  However, the unref'd blocks still may be getting fetched from the remote server.

+@jzila as a reviewer since it involves block fetching.

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -15,6 +15,7 @@ import (
 	"github.com/keybase/kbfs/kbfscodec"
 	"github.com/keybase/kbfs/tlf"
 	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
 )
 
 type overallBlockState int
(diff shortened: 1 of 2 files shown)


libkbfs/folder_block_ops.go
```diff
@@ -396,6 +397,75 @@ func (fbo *folderBlockOps) GetBlockForReading(ctx context.Context,
...
+
+	type resp struct {
+		ptr   BlockPointer
+		block Block
+	}
+
+	respChans := make([]<-chan resp, 0, len(ptrs))
+	eg, groupCtx := errgroup.WithContext(ctx)
+	for _, ptr := range ptrs {
+		ptr := ptr
```
jzila: I understand what you're doing here, but I don't love shadowing names because of the ambiguity it creates. Can you use a different name for the iteration variable?

libkbfs/folder_block_ops.go
```diff
@@ -396,6 +397,75 @@ func (fbo *folderBlockOps) GetBlockForReading(ctx context.Context,
...
+			respCh <- resp{ptr, block}
+			return nil
+		})
+	}
+
+	if err := eg.Wait(); err != nil {
+		return nil, err
+	}
+
+	blocks := make(map[BlockPointer]Block)
```
jzila: `make(map[BlockPointer]Block, len(respChans))`

libkbfs/folder_block_ops.go
```diff
@@ -396,6 +397,75 @@ func (fbo *folderBlockOps) GetBlockForReading(ctx context.Context,
...
+		respCh := make(chan resp, 1)
+		respChans = append(respChans, respCh)
+		eg.Go(func() error {
+			block, err := fbo.getBlockHelperLocked(groupCtx, nil, kmd, ptr,
+				branch, NewCommonBlock, false, path{}, blockReadParallel)
+			// TODO: we might be able to recover the size of the
+			// top-most block of a removed file using the merged
+			// directory entry, the same way we do in
+			// `folderBranchOps.unrefEntry`.
+			if isRecoverableBlockErrorForRemoval(err) &&
```
jzila: shouldn't this be `isRecoverableBlockError`?

libkbfs/folder_block_ops.go
```diff
@@ -396,6 +397,75 @@ func (fbo *folderBlockOps) GetBlockForReading(ctx context.Context,
...
+	type resp struct {
+		ptr   BlockPointer
+		block Block
+	}
+
+	respChans := make([]<-chan resp, 0, len(ptrs))
+	eg, groupCtx := errgroup.WithContext(ctx)
+	for _, ptr := range ptrs {
+		ptr := ptr
+		respCh := make(chan resp, 1)
```
jzila: Since you're handling synchronization with the `errgroup`, this be a simple slice. You'd have to loop with an index variable, but then each goroutine can set the value at its location in the slice. It saves us a channel allocation per loop iteration.

libkbfs/folder_block_ops.go
strib: I thought this was the official recommended style put forth in Effective Go, no?

jzila: So it is. Carry on.

libkbfs/folder_block_ops.go
strib: Whoa, I didn't know you could initialize map size.  Thanks!

strib:
Implemented @jzila's feedback, thanks!

libkbfs/folder_block_ops.go
```diff
@@ -396,6 +397,68 @@ func (fbo *folderBlockOps) GetBlockForReading(ctx context.Context,
...
+	ignoreRecoverableForRemovalErrors map[BlockPointer]bool,
+	branch BranchName) (map[BlockPointer]Block, error) {
+	fbo.blockLock.RLock(lState)
+	defer fbo.blockLock.RUnlock(lState)
+
+	blockResults := make([]Block, len(ptrs))
+	eg, groupCtx := errgroup.WithContext(ctx)
+	for i, ptr := range ptrs {
+		index := i
+		ptr := ptr
```
jzila: I guess you can just apply this whole pattern here.
`i, ptr := i, ptr`

jzila: FWIW I tested it here to make sure it works correctly:
https://play.golang.org/p/u2qN92yLOh

PR #582, opened 2016-12-15
libdokan: Log SID mismatch

taruti:
As discussed on Slack.


taruti:
Related to https://github.com/keybase/client/issues/4970 (should make it more visible in logs).

zanderz:
LGTM :thumbsup:

PR #584, opened 2016-12-16
fetch unmerged first in getMDByHandle so GetTLFCryptKeys works after restart on unmerged branch

songgao:
@strib I removed the second `EnableJournaling`/`EnableAuto` calls, and tested to make sure the test is still effective. Thanks @strib for your help!


songgao:
Looks like my local test went wrong; removing that chunk results in inconsistent state. So I reverted the last commit.

libkbfs/kbfs_cr_test.go
```diff
@@ -197,6 +199,114 @@ func TestMultipleMDUpdatesUnembedChanges(t *testing.T) {
...
+	config1B := ConfigAsUser(config1, userName1)
+	defer CheckConfigAndShutdown(t, config1B)
+	config1B.EnableJournaling(tempdir, TLFJournalBackgroundWorkEnabled)
+	jServer, err = GetJournalServer(config1B)
+	if err != nil {
+		t.Fatalf("error in GetJournalServer: %v\n", err)
+	}
+	jServer.onBranchChange = nil
+	jServer.onMDFlush = nil
+	jServer.EnableAuto(ctx)
```
strib: I think you can remove this line, but not the `EnableJournaling` call above.

songgao: ah you are right!

PR #587, opened 2016-12-16
bare_root_metadata_v2: readers shall not upconvert to v3

strib:
Since copying writer signatures blindly between versions doesn't work.

Issue: KBFS-1788


strib:
@aalness @akalin-keybase please take another look.  Note that I needed to change the existing private upconversion test a little since the MD didn't look readable before (because there was no blockpointer).

PR #606, opened 2016-12-27
kbfs_ops: process TLF handle changes

strib:
This fixes up the by-favorite ops map when a handle changes due to resolved assertions, conflicts, or finalization.  One specific fix is that it makes sure `GetTLFCryptKeys` will lookup the correct TLF ID for the new folder, after a previous TLF is finalized.

I tried to come up with a good test for this, but since `MDServerLocal` doesn't support handle changes, and there's no good way for the existing kbfsdocker tests to check this, so I'm punting on a test for now.

Issue: KBFS-1802

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -253,10 +253,20 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
 
 	fs.opsLock.Lock()
 	defer fs.opsLock.Unlock()
+	fav := handle.ToFavorite()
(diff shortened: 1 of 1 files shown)


strib:
@songgao: finalization happens when a user reset their account.  SBS handle updates happen when someone proves an assertion -- though this specific PR only matters in the case where they prove an assertion and it causes a TLF-level conflict, changing the name of the TLF to have a conflict prefix.  I think those are the only two cases.

songgao:
I see. Thanks!

PR #607, opened 2016-12-27
fix UID and perms

songgao:
This changes owners from `root` to the user that executes the `kbfsfuse` proces on Linux/macOS. It also changes permissions to more reasonable values. This fixes the error when saving a file in Photoshop.

The ownership and permissions might be wrong for a TLF itself if it's not initialized yet. But after the TLF is accessed it's correct.

### DIFF
--- a/libfuse/common.go
+++ b/libfuse/common.go
@@ -4,13 +4,6 @@
 
 package libfuse
 
-import (
-	"time"
-
-	"bazil.org/fuse"
-	"github.com/keybase/kbfs/libkbfs"
-)
-
 const (
 	// PublicName is the name of the parent of all public top-level folders.
 	PublicName = "public"
@@ -29,13 +22,3 @@ const (
 	// CtxIDKey is the type of the tag for unique operation IDs.
 	CtxIDKey CtxTagKey = iota
 )
-
-// fillAttr sets attributes based on the entry info. It only handles fields
-// common to all entryinfo types.
-func fillAttr(ei *libkbfs.EntryInfo, a *fuse.Attr) {
-	a.Valid = 1 * time.Minute
-
-	a.Size = ei.Size
-	a.Mtime = time.Unix(0, ei.Mtime)
-	a.Ctime = time.Unix(0, ei.Ctime)
-}
--- a/libfuse/common.go
+++ b/libfuse/common.go
@@ -33,7 +33,7 @@ const (
 
 // fillAttr sets attributes based on the entry info. It only handles fields
 // common to all entryinfo types.
-func fillAttr(ei *libkbfs.EntryInfo, a *fuse.Attr) {
+func fillAttr(ei *libkbfs.EntryInfoExtended, a *fuse.Attr) {
 	a.Valid = 1 * time.Minute
 
 	a.Size = ei.Size
--- a/libfuse/common.go
+++ b/libfuse/common.go
@@ -33,7 +33,7 @@ const (
 
 // fillAttr sets attributes based on the entry info. It only handles fields
 // common to all entryinfo types.
-func fillAttr(ei *libkbfs.EntryInfoExtended, a *fuse.Attr) {
+func fillAttr(ei *libkbfs.EntryInfo, a *fuse.Attr) {
 	a.Valid = 1 * time.Minute
 
 	a.Size = ei.Size
--- a/libfuse/common.go
+++ b/libfuse/common.go
@@ -5,7 +5,6 @@
 package libfuse
 
 import (
-	"os"
 	"time"
 
 	"bazil.org/fuse"
@@ -39,12 +38,4 @@ func fillAttr(ei *libkbfs.EntryInfo, a *fuse.Attr) {
 	a.Size = ei.Size
 	a.Mtime = time.Unix(0, ei.Mtime)
 	a.Ctime = time.Unix(0, ei.Ctime)
-	fillAttrUIDAndWritePerm(a, ei.Writable)
-}
-
-func fillAttrUIDAndWritePerm(a *fuse.Attr, writable bool) {
-	a.Uid = uint32(os.Getuid())
-	if writable {
-		a.Mode |= 0200
-	}
 }
--- a/libfuse/common.go
+++ b/libfuse/common.go
@@ -5,6 +5,7 @@
 package libfuse
 
 import (
+	"os"
 	"time"
 
 	"bazil.org/fuse"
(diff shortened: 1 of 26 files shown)


libkbfs/data_types.go
```diff
@@ -565,6 +565,14 @@ type EntryInfo struct {
 	Ctime int64
 }
 
+// EntryInfoExtended extends the EntryInfo with some locally computed
+// attributes.
+type EntryInfoExtended struct {
+	EntryInfo
+	// Writable is true iif the user has write permission on the entry.
```
strib: "iif" -> "if" (or maybe "iff" if you're trying to be really exact).

libfuse/common.go
```diff
@@ -32,10 +33,18 @@ const (
...
+func fillAttr(ei *libkbfs.EntryInfoExtended, a *fuse.Attr) {
 	a.Valid = 1 * time.Minute
 
 	a.Size = ei.Size
 	a.Mtime = time.Unix(0, ei.Mtime)
 	a.Ctime = time.Unix(0, ei.Ctime)
+	fillAttrUIDAndWritePerm(a, ei.Writable)
+}
+
+func fillAttrUIDAndWritePerm(a *fuse.Attr, writable bool) {
```
strib: Please move this above `fillAttr`.

libkbfs/data_types.go
```diff
@@ -565,6 +565,14 @@ type EntryInfo struct {
 	Ctime int64
 }
 
+// EntryInfoExtended extends the EntryInfo with some locally computed
+// attributes.
+type EntryInfoExtended struct {
```
strib: This struct shouldn't be necessary -- everywhere in `libfuse` that calls `fillAttr` should already have access to a `libfuse.Folder` object, which has a copy of the handle already (`h`).  So I don't think it's worth complicating the `KBFSOps` interface with a new struct type.  Of course you'd have to get the UID of the logged-in user (probably with `libkbfs.GetCurrentUsernameIfPossible`) from those places in `libfuse`, but that should be easy enough with a wrapper function.  Maybe `fillAttr` should be moved to be a member function of `libfuse.Folder` or something.

libfuse/fs.go
```diff
@@ -224,7 +224,7 @@ var _ fs.Node = (*Root)(nil)
 
 // Attr implements the fs.Node interface for Root.
 func (*Root) Attr(ctx context.Context, a *fuse.Attr) error {
-	a.Mode = os.ModeDir | 0755
+	a.Mode = os.ModeDir | 0555
```
strib: `FolderList.Attr` should be fixed in the same way, I think, right?

libkbfs/folder_branch_ops.go
```diff
@@ -1444,8 +1444,27 @@ func (fbo *folderBranchOps) execMDReadNoIdentifyThenMDWrite(
...
+		ret[k] = fbo.makeEntryInfoExtended(ctx, v)
+	}
+	return ret
+}
+
+func (fbo *folderBranchOps) makeEntryInfoExtended(ctx context.Context,
+	ei EntryInfo) EntryInfoExtended {
+	eie := EntryInfoExtended{EntryInfo: ei}
+	_, uid, err := fbo.config.KBPKI().GetCurrentUserInfo(ctx)
+	if err == nil { // writable is only possible if user is logged in
```
strib: This hides any other legit errors, better to use `GetCurrentUsernameIfPossible`.

libkbfs/data_types.go
songgao: I was thinking the writable permission is platform-indepedent, and putting it in libkbfs means we can reuse the same code on other platforms as well. But I'm OK with moving these to libfuse. Will update; thanks!

songgao:
@strib re: invalidation, I need to look closer on the topic to understand this better, but is there a reason we can't use https://godoc.org/bazil.org/fuse#Conn.InvalidateEntry on all directories when we have `*Folder` available?

strib:
@songgao: To use `InvalidateEntry` you need to know the name of the thing you're reverting, and the `libfuse.Dir` objects don't keep that around.  You might be able to figure it out from the `libkbfs.Node` objects if you have direct access to the `NodeCache`, but right now I don't think that's not easily done from `libfuse`.

songgao:
@strib I see; can we instead set expiration to 1 second if `*Folder` doesn't exist yet?

songgao:
@strib ready for review again!

strib:
> I see; can we instead set expiration to 1 second if `*Folder` doesn't exist yet?

I don't understand what `Folder`s have to do with anything.  For regular directory entries inside the TLF, the `Folder` will always exist.  And a user logging out or changing their username doesn't affect whether a `Folder` exists or not.

songgao:
@strib sorry, not `*Folder`, the `*Dir`. I'm referring to this: https://github.com/keybase/kbfs/pull/607/files#diff-274b505919d6e0fff53543ee07eca9a4R137

libfuse/dir.go
```diff
@@ -334,6 +342,33 @@ func (f *Folder) tlfHandleChangeInvalidate(ctx context.Context,
...
+
+	a.Uid = uint32(os.Getuid())
+
+	a.Mode &^= os.FileMode(0222) // clear write perm bits
+	_, uid, err := libkbfs.GetCurrentUserInfoIfPossible(
+		ctx, f.fs.config.KBPKI(), f.list.public)
+	// Here we get an error, but there is little that can be done.
+	// cuser will be empty in the error case in which case we will default to the
+	// canonical format.
+	if err != nil {
```
strib: We shouldn't ignore the error, because it's only non-`nil` if a real problem happened.  We should return it to the caller.  If the user is logged out, we will get an empty username and uid, with a `nil` error.

strib:
> sorry, not *Folder, the *Dir. I'm referring to this: https://github.com/keybase/kbfs/pull/607/files#diff-274b505919d6e0fff53543ee07eca9a4R137

Gotcha, but this amounts to the same thing -- it is a `Dir` representing just the top-level directory, and is always present if there are any cached kernel entries within this TLF.  Note there is nothing that unsets that when the user logs out.

songgao:
@strib hmm.. I see what you are saying. The problem is not when user first accesses the TLF, but when user logs out. Thanks for explanation! Should I create a ticket for it or maybe not worry about it right now?

libfuse/dir.go
songgao: Good point; will update!

strib:
> Should I create a ticket for it or maybe not worry about it right now?

@songgao: please create a ticket so we don't forget, but otherwise let's not work on it right now.

songgao:
Sounds good! https://keybase.atlassian.net/browse/KBFS-1808

PR #612, opened 2016-12-28
test_dynamodb_runner: use the ioutil wrapped version of IsNotExist

strib:
Because ioutil.Stat returns a wrapped error.


strib:
Yep.

strib:
Ironically the kbfs-server failures are because this fix isn't vendored there yet.  Merging.

PR #617, opened 2016-12-29
libfuse: implement Access for File

songgao:


libfuse/file.go
```diff
@@ -73,6 +73,28 @@ func (f *File) fillAttrWithMode(
...
 }
 
+// Access implements the fs.NodeAccesser interface for File. This is necessary
+// for macOS to correctly identify plaintext file as plaintext. If not
+// implemented, bazil-fuse returns a nil error for every thing, so when macOS
+// checks for executable bit using Access (instead of Attr!), it gets a
+// success, which makes it think the file is executable, yielding a "Unix
+// executable" UTI.
+func (f *File) Access(ctx context.Context, r *fuse.AccessRequest) error {
+	if r.Mask&0111 != 0 {
```
strib: If you invert this, you can save some indentation; i.e.:

```go
if r.Mask&0111 == 0 {
   return nil
}
...
```

strib: Also, we might as well also check for write permissions at the same time, right?  Who knows what other ACCESS bugs are lurking out there.  If the current user is not a writer on the folder and the mask matches `0x222`, we can also return `EPERM`.

libfuse/file.go
```diff
@@ -73,6 +73,28 @@ func (f *File) fillAttrWithMode(
 	return nil
 }
 
+// Access implements the fs.NodeAccesser interface for File. This is necessary
```
strib: Please add this above:

```go
var _ fs.NodeAccesser = (*File)(nil)
```

libfuse/file.go
```diff
@@ -73,6 +73,28 @@ func (f *File) fillAttrWithMode(
...
+// for macOS to correctly identify plaintext file as plaintext. If not
```
strib: "file" -> "files"

libfuse/file.go
```diff
@@ -73,6 +73,28 @@ func (f *File) fillAttrWithMode(
...
+// implemented, bazil-fuse returns a nil error for every thing, so when macOS
```
strib: "thing" -> "call"

libfuse/file.go
songgao: Good point. I'll just change it to do a full `fillAttWithMode` and compare against the Mask. (I was hesitant to do that in fear of performance loss)

strib: I'm not sure that's needed, since it will result in a `Stat` call even if no executable bits are set in `Mask`.  I think you can just add a case for `&0222`.

songgao: But there's also `0444`. Private entries don't have `0044` set right?

strib: And actually, shouldn't we be returning an error for any permissions that aren't on the user bit?  I.e., even if it's an executable file, we should return an error for a mask like `0001`, right?

songgao: For public TLFs, it's readable right?

songgao: Ah OK just the dirs. Although shall we clear all permission bits on g and o for public TLFs and dirs as well?

strib: > Ah OK just the dirs. Although shall we clear all permission bits on g and o for public TLFs and dirs as well?

Ah yeah, I guess we probably should, since FUSE already keeps those other groups from reading the mount.

songgao: Sounds good!

songgao: It seems `r.Mask` isn't meant to be pased as `-rwxrwxrwx`. It's just one `rwx`. I'll update the PR

songgao:
@strib would you mind giving this another look? It became bigger than the original commit. Thanks!

libfuse/dir.go
```diff
@@ -342,11 +335,41 @@ func (f *Folder) tlfHandleChangeInvalidate(ctx context.Context,
 	}
 }
 
+func (f *Folder) isWriter(
+	ctx context.Context, ei *libkbfs.EntryInfo) (bool, error) {
```
strib: Why does this function need `ei`?  It looks unused.

libfuse/dir.go
```diff
@@ -342,11 +335,41 @@ func (f *Folder) tlfHandleChangeInvalidate(ctx context.Context,
...
+	if err != nil {
+		return false, err
+	}
+	f.handleMu.RLock()
+	defer f.handleMu.RUnlock()
+	return f.h.IsWriter(uid), nil
+}
+
+func (f *Folder) writePermMode(ctx context.Context,
+	ei *libkbfs.EntryInfo, original os.FileMode) (os.FileMode, error) {
```
strib: Same here if you remove it from `isWriter()`.

libfuse/file.go
```diff
@@ -109,6 +110,48 @@ func (f *File) attr(ctx context.Context, a *fuse.Attr) (err error) {
...
+		// This is in case we enable AllowOther in the future.
+		return fuse.EPERM
+	}
+
+	if r.Mask&03 == 0 {
+		// Since we only check for w and x bits, we can return nil early here.
+		return nil
+	}
+
+	ei, err := f.folder.fs.config.KBFSOps().Stat(ctx, f.node)
```
strib: You only need to do this in the `r.Mask&01 != 0` case.

libfuse/folderlist.go
```diff
@@ -31,11 +31,24 @@ type FolderList struct {
...
+
+// Access implements fs.NodeAccesser interface for *FolderList.
+func (*FolderList) Access(ctx context.Context, r *fuse.AccessRequest) error {
+	if int(r.Uid) != os.Getuid() {
+		// short path: not accessible by anybody other than the logged in user.
+		// This is in case we enable AllowOther in the future.
+		return fuse.EPERM
+	}
+
+	return nil
```
strib: You should return an error for anything asking about write permissions in the folderlist.

libfuse/fs.go
```diff
@@ -220,11 +220,24 @@ type Root struct {
...
+
+// Access implements fs.NodeAccesser interface for *Root.
+func (*Root) Access(ctx context.Context, r *fuse.AccessRequest) error {
+	if int(r.Uid) != os.Getuid() {
+		// short path: not accessible by anybody other than the logged in user.
+		// This is in case we enable AllowOther in the future.
+		return fuse.EPERM
+	}
+
+	return nil
```
strib: You should return an error for anything asking about write permissions in the root.

libfuse/dir.go
```diff
@@ -408,6 +422,38 @@ func newDir(folder *Folder, node libkbfs.Node) *Dir {
...
+		return fuse.EPERM
+	}
+
+	if r.Mask&02 == 0 {
+		// For directory, we only check for the w bit, we can return nil early
+		// here.
+		return nil
+	}
+
+	ei, err := d.folder.fs.config.KBFSOps().Stat(ctx, d.node)
```
strib: No need to `Stat` for a directory.

libfuse/tlf.go
```diff
@@ -131,6 +131,17 @@ func (tlf *TLF) loadDirAllowNonexistent(ctx context.Context) (
 	return tlf.loadDirHelper(ctx, libkbfs.ReadMode, true)
 }
 
+// Access implements the fs.NodeAccesser interface for *TLF.
+func (tlf *TLF) Access(ctx context.Context, r *fuse.AccessRequest) error {
```
strib: The logic here should be the same as `Dir.Access()`, in that it should check write permission using the handle.  Unfortunately, `tlf.dir` might be `nil`, so you can either duplicate `Dir.Access()`, or move it up to the `Folder` level so both can share it.

libfuse/dir.go
```diff
@@ -408,6 +422,38 @@ func newDir(folder *Folder, node libkbfs.Node) *Dir {
...
+	if err != nil {
+		if isNoSuchNameError(err) {
+			return fuse.ESTALE
+		}
+		return err
+	}
+	wp, err := d.folder.writePermMode(ctx, &ei, 0)
```
strib: Why not call `d.folder.isWriter()` instead?  You just need a boolean, not the full mask, right?  (Same question in the other `Access()` methods.

libfuse/dir.go
songgao: ah you are right! Thanks!

PR #621, opened 2017-01-01
file_data: support writes and truncates for >1 indirection levels

strib:
[Sorry about the size of this PR, I can't think of a good way to shrink it and still keep the tests passing.  If you have suggestions, let me know.]

Note that this doesn't yet support syncing those multiple levels of indirection (or CR), so it is turned off for everything but the new fileData tests right now.

We view the data file as a tree of indirect blocks that ultimately point to a set of direct leaf blocks at the bottom level.  Whenever the existing leaf blocks can't accomodate a new write, we append a leaf block.  If the indirect parent can't accomodate a new block (as decided by the BlockSplitter), we add a new indirect block at that level, etc, all the way up to the root.  If the root can't handle a new block, we add a new level of indirection.

If we are writing into a "hole" in the file that was created by a previous `truncateExtend` call, the new right block will then have to be shifted into place by moving around indirect pointers in the interior of the tree.

When shrinking a file via `truncateShrink`, we truncate indirect pointer slices and unreference blocks whenever those slices no longer have any pointers. Note that for now we do not remove levels of indirection, so the left-most edge of the tree never shrinks.

Issue: KBFS-35

### DIFF
--- a/libkbfs/file_data.go
+++ b/libkbfs/file_data.go
@@ -707,7 +707,7 @@ func (fd *fileData) shiftBlocksToFillHole(
 	}
 
 	// Swap left as needed.
-	for true {
+	for {
 		var leftOff int64
 		var newParents []parentBlockAndChildIndex
 		if currIndex > 0 {
@@ -814,9 +814,7 @@ func (fd *fileData) shiftBlocksToFillHole(
 		currIndex = newCurrIndex
 		parents = newParents
 	}
-
-	return nil, fmt.Errorf("Couldn't find where to move the hole for offset %d",
-		newHoleStartOff)
+	// The loop above must exit via one of the returns.
 }
 
 // markParentsDirty caches all the blocks in `parentBlocks` as dirty,
@@ -1080,7 +1078,7 @@ func (fd *fileData) truncateExtend(ctx context.Context, size uint64,
 
 // truncateShrink shrinks the file to the given size. Return params:
 // * newDe: a new directory entry with the EncodedSize cleared if the file
-//   was extended.
+//   shrunk.
 // * dirtyPtrs: a slice of the BlockPointers that have been dirtied during
 //   the truncate.  This includes any interior indirect blocks that may not
 //   have been changed yet, but which will need to change as part of the
@@ -1140,15 +1138,19 @@ func (fd *fileData) truncateShrink(ctx context.Context, size uint64,
 			return DirEntry{}, nil, nil, 0, err
 		}
 
-		cachedPtrs := make(map[BlockPointer][]IndirectFilePtr)
+		// A map from a pointer of an indirect block to that block's
+		// original set of block pointers, before they are truncated
+		// in the loop below.  It also tracks which pointed-to blocks
+		// have already been processed.
+		savedChildPtrs := make(map[BlockPointer][]IndirectFilePtr)
 		for _, path := range pfr {
-			// Fake an indirect pointer for the top block.
-			iptr := IndirectFilePtr{BlockInfo: BlockInfo{
-				BlockPointer: fd.rootBlockPointer(),
-			}}
+			// parentInfo points to pb.pblock in the loop below.  The
+			// initial case is the top block, for which we need to
+			// fake a block info using just the root block pointer.
+			parentInfo := BlockInfo{BlockPointer: fd.rootBlockPointer()}
 			leftMost := true
 			for i, pb := range path {
-				ptrs := cachedPtrs[iptr.BlockPointer]
+				ptrs := savedChildPtrs[parentInfo.BlockPointer]
 				if ptrs == nil {
 					// Process each block exactly once, removing all
 					// now-unnecessary indirect pointers (but caching
@@ -1156,14 +1158,11 @@ func (fd *fileData) truncateShrink(ctx context.Context, size uint64,
 					// next iterations).
 					pblock := pb.pblock
 					ptrs = pblock.IPtrs
-					cachedPtrs[iptr.BlockPointer] = ptrs
+					savedChildPtrs[parentInfo.BlockPointer] = ptrs
 
 					// Remove the first child iptr and everything
 					// following it if all the child indices below
-					// this level are 0. If we remove iptr 0, this
-					// block can be unreferenced (unless it's on the
-					// left-most edge of the tree, in which case we
-					// keep it around for now -- see above TODO).
+					// this level are 0.
 					removeStartingFromIndex := pb.childIndex
 					for j := i + 1; j < len(path); j++ {
 						if path[j].childIndex > 0 {
@@ -1172,26 +1171,39 @@ func (fd *fileData) truncateShrink(ctx context.Context, size uint64,
 						}
 					}
 
+					// If we remove iptr 0, this block can be
+					// unreferenced (unless it's on the left-most edge
+					// of the tree, in which case we keep it around
+					// for now -- see above TODO).
 					if pb.childIndex == 0 && !leftMost {
-						if iptr.EncodedSize != 0 {
-							unrefs = append(unrefs, iptr.BlockInfo)
+						if parentInfo.EncodedSize != 0 {
+							unrefs = append(unrefs, parentInfo)
 						}
 					} else if removeStartingFromIndex < len(pblock.IPtrs) {
-						pblock.IPtrs = pblock.IPtrs[:removeStartingFromIndex]
-						err := fd.cacher(iptr.BlockPointer, pblock)
+						// Make sure we're modifying a copy of the
+						// block by fetching it again with blockWrite.
+						pblock, _, err = fd.getter(
+							ctx, fd.kmd, parentInfo.BlockPointer, fd.file,
+							blockWrite)
 						if err != nil {
 							return DirEntry{}, nil, nil,
 								newlyDirtiedChildBytes, err
 						}
-						dirtyMap[iptr.BlockPointer] = true
+						pblock.IPtrs = pblock.IPtrs[:removeStartingFromIndex]
+						err = fd.cacher(parentInfo.BlockPointer, pblock)
+						if err != nil {
+							return DirEntry{}, nil, nil,
+								newlyDirtiedChildBytes, err
+						}
+						dirtyMap[parentInfo.BlockPointer] = true
 					}
 				}
 
 				// Down to the next level.  If we've hit the leaf
 				// level, unreference the block.
-				iptr = ptrs[pb.childIndex]
-				if i == len(path)-1 && iptr.EncodedSize != 0 {
-					unrefs = append(unrefs, iptr.BlockInfo)
+				parentInfo = ptrs[pb.childIndex].BlockInfo
+				if i == len(path)-1 && parentInfo.EncodedSize != 0 {
+					unrefs = append(unrefs, parentInfo)
 				} else if pb.childIndex > 0 {
 					leftMost = false
 				}
--- a/libkbfs/file_data.go
+++ b/libkbfs/file_data.go
@@ -886,14 +886,21 @@ func (fd *fileData) write(ctx context.Context, data []byte, off int64,
 			needExtendFile := nextBlockOff < 0
 			needFillHole := off+nCopied < nextBlockOff
 			newBlockOff := startOff + int64(len(block.Contents))
-			if newBlockOff < off {
-				// We are writing somewhere inside a hole, not right
-				// at the start of it; all we have done so far it
-				// reached the end of an existing block (possibly
-				// zero-filling it out to its capacity).  Make sure
-				// the next block starts right at the offset we care
-				// about.
-				newBlockOff = off
+			if nCopied == 0 {
+				if newBlockOff < off {
+					// We are writing somewhere inside a hole, not right
+					// at the start of it; all we have done so far it
+					// reached the end of an existing block (possibly
+					// zero-filling it out to its capacity).  Make sure
+					// the next block starts right at the offset we care
+					// about.
+					newBlockOff = off
+				}
+			} else if newBlockOff != off+nCopied {
+				return newDe, nil, unrefs, newlyDirtiedChildBytes, 0,
+					fmt.Errorf("Copied %d bytes, but newBlockOff=%d does not "+
+						"match off=%d plus new bytes",
+						nCopied, newBlockOff, off)
 			}
 			var rightParents []parentBlockAndChildIndex
 			if needExtendFile || needFillHole {
--- a/libkbfs/file_data.go
+++ b/libkbfs/file_data.go
@@ -546,12 +546,12 @@ func (fd *fileData) createIndirectBlock(
 	return fblock, nil
 }
 
-// newRightBlock creates space for a new rightmost block,
-// creating parent blocks and a new level of indirection in the tree
-// as needed.  If there's no new level of indirection, it modifies the
-// blocks in `parentBlocks` to include the new right-most pointers.
-// If there is a new level of indirection, it returns the new top
-// block.  It also returns any newly-dirtied block pointers.
+// newRightBlock creates space for a new rightmost block, creating
+// parent blocks and a new level of indirection in the tree as needed.
+// If there's no new level of indirection, it modifies the blocks in
+// `parentBlocks` to include the new right-most pointers.  It also
+// returns the new top block (whether or not it is new), and also
+// returns any newly-dirtied block pointers.
 func (fd *fileData) newRightBlock(
 	ctx context.Context, parentBlocks []parentBlockAndChildIndex, off int64,
 	df *dirtyFile, dver DataVer) (*FileBlock, []BlockPointer, error) {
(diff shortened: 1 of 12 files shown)


jzila:
Because of the extend and complexity of this PR, I'd like @akalin-keybase to also take a look before merging.

libkbfs/file_data.go
```diff
@@ -501,44 +527,291 @@ func (fd *fileData) createIndirectBlock(
...
+
+		newDirtyPtrs = append(newDirtyPtrs, newPtr)
 	}
-	err = fd.cacher(ptr, pblock)
-	if err != nil {
-		return err
+
+	// All parents up to and including the lowest ancestor with room
+	// will have to change, so mark them as dirty.
+	ptr := fd.file.tailPointer()
```
strib: I'll make a `fileData.rootBlockPointer()` func that just returns `fd.file.tailPointer()`, for code clarity.

libkbfs/file_data.go
```diff
@@ -585,52 +863,66 @@ func (fd *fileData) write(ctx context.Context, data []byte, off int64,
...
-				return newDe, nil, unrefs, newlyDirtiedChildBytes, 0, err
+				for _, p := range newDirtyPtrs {
+					dirtyMap[p] = true
+				}
+				if newTopBlock != nil {
+					if !topBlock.IsInd {
+						// The whole block needs to be re-uploaded as
+						// a child block with a new block pointer, so
+						// below we'll need to track those dirty bytes
+						// and cache the block as dirty.
```
strib: Yes, but the dirty bcache only tracks direct block bytes right now, not the bytes of indirect blocks, so in the case you mention nothing additional needs to be tracked.

Now that we can have more indirect blocks, we might think about changing that policy, but as of now indirect block bytes don't need tracking.

strib: (I'll clarify the comment.)

strib:
Thanks for the great feedback @jzila, I think I got it all.  Ready for @akalin-keybase review!

akalin-keybase:
I'll look at this tomorrow!

libkbfs/file_data.go
```diff
@@ -750,54 +1060,120 @@ func (fd *fileData) truncateExtend(ctx context.Context, size uint64,
...
+					// keep it around for now -- see above TODO).
+					removeStartingFromIndex := pb.childIndex
+					for j := i + 1; j < len(path); j++ {
+						if path[j].childIndex > 0 {
+							removeStartingFromIndex++
+							break
+						}
+					}
+
+					if pb.childIndex == 0 && !leftMost {
```
akalin-keybase: I'm a bit confused by this case -- is the comment above 'If we remove iptr 0...' referring to this? If so, may be better to move it down here.

libkbfs/file_data.go
```diff
@@ -750,54 +1060,120 @@ func (fd *fileData) truncateExtend(ctx context.Context, size uint64,
...
+						if iptr.EncodedSize != 0 {
+							unrefs = append(unrefs, iptr.BlockInfo)
+						}
+					} else if removeStartingFromIndex < len(pblock.IPtrs) {
+						pblock.IPtrs = pblock.IPtrs[:removeStartingFromIndex]
```
akalin-keybase: I'm worried about this, because it seems to me this will modify a read-only block -- pblock is a pointer, and it's retrieved with `getLeafBlocksForOffsetRange`, which uses `blockReadParallel`, which doesn't make a copy.

libkbfs/file_data.go
```diff
@@ -750,54 +1060,120 @@ func (fd *fileData) truncateExtend(ctx context.Context, size uint64,
...
 		}
-		parentBlock.IPtrs = parentBlock.IPtrs[:indexInParent+1]
-		// always make the parent block dirty, so we will sync it
-		if err = fd.cacher(fd.file.tailPointer(), parentBlock); err != nil {
-			return DirEntry{}, nil, newlyDirtiedChildBytes, err
+
+		cachedPtrs := make(map[BlockPointer][]IndirectFilePtr)
+		for _, path := range pfr {
+			// Fake an indirect pointer for the top block.
+			iptr := IndirectFilePtr{BlockInfo: BlockInfo{
```
akalin-keybase: also probably less confusing to name this `parentInfo`, and have a comment saying that it's a loop invariant that `parentInfo.BlockPointer` points to `pb.pblock` in the loop below.

libkbfs/file_data.go
```diff
@@ -65,13 +65,21 @@ type parentBlockAndChildIndex struct {
 	childIndex int
 }
 
+func (pbci parentBlockAndChildIndex) childIPtr() IndirectFilePtr {
+	return pbci.pblock.IPtrs[pbci.childIndex]
+}
+
+func (fd *fileData) rootBlockPointer() BlockPointer {
```
akalin-keybase: mind splitting this (and the above) off into a separate PR? It would make subsequent passes easier.

strib: @jzila specifically requested this as part of this PR.  `childIPtr()` in particular makes a lot of sense as part of this PR, I think it would make the PR harder to review since it simplifies a lot of the new code you'd look at in the new passes.  I don't care as much about `rootBlockPointer()`, but it seems like it should be fairly easy to review the existing lines that only have that call changed.  Do you feel strongly about it?  How would splitting it off help exactly?

strib: (By the way, I should mention you could review the separate commits as well, if that helps, they are well separated.)

strib: I just realized you might be talking about making a new PR that is merged _before_ this one (I was assuming it would be after, or independent).  I think that could make sense.

strib: Ok done (#641).

libkbfs/file_data.go
strib: Great catch!  Will fetch the block again using `blockWrite`.

strib:
Thanks @akalin-keybase! I addressed all your  feedback, ready for another pass.

libkbfs/file_data.go
```diff
@@ -509,44 +535,298 @@ func (fd *fileData) createIndirectBlock(
...
+			Context: kbfsblock.Context{
+				Creator:  fd.uid,
+				RefNonce: kbfsblock.ZeroRefNonce,
+			},
+		}
+
+		fd.log.CDebugf(ctx, "New right block for file %d, level %d, ptr %v",
+			fd.rootBlockPointer(), i, newPtr)
+
+		pblock.IPtrs = append(pblock.IPtrs, IndirectFilePtr{
```
akalin-keybase: There's another aliasing issue here for the `i = lowestAncestorWithRoom` iteration, since that comes from `parentBlocks`, which are blocks-for-reading, right?

libkbfs/file_data.go
```diff
@@ -509,44 +535,298 @@ func (fd *fileData) createIndirectBlock(
...
+// returns the new top block (whether or not it is new), and also
+// returns any newly-dirtied block pointers.
 func (fd *fileData) newRightBlock(
-	ctx context.Context, ptr BlockPointer, pblock *FileBlock,
-	off int64) error {
-	newRID, err := fd.crypto.MakeTemporaryBlockID()
-	if err != nil {
-		return err
+	ctx context.Context, parentBlocks []parentBlockAndChildIndex, off int64,
+	df *dirtyFile, dver DataVer) (*FileBlock, []BlockPointer, error) {
```
akalin-keybase: I think this function should, instead of modifying `parentBlocks`, return a new `[]parentBlockAndChildIndex` with the new pointers. It's less memory-efficient, but somewhat clearer, and we're already returning the first element of that array anyway (i.e., the new top block).

libkbfs/file_data.go
```diff
@@ -509,44 +535,298 @@ func (fd *fileData) createIndirectBlock(
 	return fblock, nil
 }
 
-// newRightBlock appends a new block (with a temporary ID) to the end
-// of this file.
+// newRightBlock creates space for a new rightmost block, creating
```
akalin-keybase: So this function doesn't append to the end of the file anymore, right? I guess the name can still remain the same, (I can't think of a better one) but I think the comment can be a bit clearer about that. Perhaps "creates space for a new block to the right of (but not necessarily adjacent to) the leaf block pointed to by `parentBlocks`..."

libkbfs/file_data.go
```diff
@@ -509,44 +535,298 @@ func (fd *fileData) createIndirectBlock(
...
+	// left until its offset would be lower than that child's offset.
+	// If there are no children to the left, continue on with the
+	// children in the cousin block to the left.  If we swap a child
+	// between cousin blocks, we must update the offset in the right
+	// cousin's parent block.  If *that* updated pointer is the
+	// leftmost pointer in its parent block, update that one as well,
+	// up to the root.
+	var parents []parentBlockAndChildIndex
+	currBlock := newTopBlock
+	for currBlock.IsInd {
```
akalin-keybase: Ah, actually, I think there's a bug -- `newRightBlock` doesn't append to the end of the file anymore, so the right-most leaf of the tree isn't necessarily the correct one anymore, right?

libkbfs/file_data.go
strib: I don't think it claims anywhere to append to the end of the file, and this PR doesn't change that.  It's just making a rightmost leaf block under the given offset, and it's the callers responsibility to move it into the correct place in the tree.  I'll try to clarify the comment.

jzila:
The failed data race should be fixed by the new master

libkbfs/file_data.go
strib: `parentBlocks` comes from `getFileBlockOffset(..., blockWrite)` in both `write` and `truncateExtend`.  I'll clarify this requirement in the function comment.

libkbfs/file_data.go
strib: Explained above.  I don't think there's a bug, but I did change it to pass the rightmost parent chain in here.

libkbfs/file_data.go
```diff
@@ -593,52 +878,65 @@ func (fd *fileData) write(ctx context.Context, data []byte, off int64,
...
-		if nCopied < n && nextBlockOff < 0 {
-			// If the block doesn't already have a parent block, make one.
-			if ptr == fd.rootBlockPointer() {
-				topBlock, err = fd.createIndirectBlock(
-					df, DefaultNewBlockDataVersion(false))
+		if nCopied < n {
+			needExtendFile := nextBlockOff < 0
+			needFillHole := off+nCopied < nextBlockOff
+			newBlockOff := startOff + int64(len(block.Contents))
+			if newBlockOff < off+nCopied {
```
strib: Consider a file with data `data` with leaf blocks split up like so `data[0:2], data[2:4], data[4], hole(10)`.  Then a write comes in for two bytes at index 8.  (This is `TestFileDataWriteEndOfHole` in `file_data_test.go`.)

The first thing it tries to do is write to the block starting at `startOff=4`.  But since the write doesn't start until offset 8, `CopyUntilSplit` just fills out that block to capacity, but returns `0` for `nCopied`.  `off=8`, so `6 < 8+0` still, and so we need to make the new right block start at `8`.

Perhaps this only ever happens if `nCopied` is 0, so that part of the formula is useless?  Would deleting that, plus adding a better comment, make it more clear?

libkbfs/file_data.go
```diff
@@ -768,44 +1067,124 @@ func (fd *fileData) truncateExtend(ctx context.Context, size uint64,
...
+					// of the tree, in which case we keep it around
+					// for now -- see above TODO).
+					if pb.childIndex == 0 && !leftMost {
+						if parentInfo.EncodedSize != 0 {
+							unrefs = append(unrefs, parentInfo)
+						}
+					} else if removeStartingFromIndex < len(pblock.IPtrs) {
+						// Make sure we're modifying a copy of the
+						// block by fetching it again with blockWrite.
+						pblock, _, err = fd.getter(
```
strib: Because `pblock` might already be a modified copy, from another iteration of the loop.  Using `getter` guarantees we're fetching that copy if it already exists and has been placed in the dirty cache.

strib:
Thanks @akalin-keybase, I addressed all the feedback and answered questions.  Ready for another round.

libkbfs/file_data.go
akalin-keybase: The old comment did say "newRightBlock appends a new block (with a temporary ID) to the end of this file."

libkbfs/file_data.go
```diff
@@ -520,44 +543,294 @@ func (fd *fileData) createIndirectBlock(
...
+// indirect pointers in the file as needed.  It returns any block
+// pointers that were dirtied in the process.
+func (fd *fileData) shiftBlocksToFillHole(
+	ctx context.Context, newTopBlock *FileBlock, newHoleStartOff int64,
+	parents []parentBlockAndChildIndex) (
+	newDirtyPtrs []BlockPointer, err error) {
+	fd.log.CDebugf(ctx, "Shifting block with offset %d for file %v into "+
+		"position", newHoleStartOff, fd.rootBlockPointer())
+
+	// Walk down the right side of the tree to find the new rightmost
```
akalin-keybase: The first part of this comment needs to be updated, since we're not walking down the right side anymore.

libkbfs/file_data.go
```diff
@@ -520,44 +543,294 @@ func (fd *fileData) createIndirectBlock(
...
+	// cousin's parent block.  If *that* updated pointer is the
+	// leftmost pointer in its parent block, update that one as well,
+	// up to the root.
+	//
+	// We are guaranteed at least one level of indirection because
+	// `newRightBlock` should have been called before
+	// `shiftBlocksToFillHole`.
+	immedParent := parents[len(parents)-1]
+	currIndex := immedParent.childIndex
+	if off := immedParent.pblock.IPtrs[currIndex].Off; off != newHoleStartOff {
```
akalin-keybase: Any reason to pass in `newHoleStartOff` now? We can just use `off` as calculated here, right?

akalin-keybase: Also, `newHoleStartOff` is a bit confusing, since it seems to mean the offset of where the new hole starts. But the new hole will start at `newHoleStartOff + length(this block)`, right? Maybe `newBlockStartOff`?

libkbfs/file_data.go
akalin-keybase: Ah, yeah, that makes sense. I think it's worth a comment.

libkbfs/file_data.go
akalin-keybase: Ah, ok. I think it's worth a comment.

keybase-travis:
GitHub is definitely in some "eventually consistent" state right now.

libkbfs/file_data.go
```diff
@@ -604,52 +882,71 @@ func (fd *fileData) write(ctx context.Context, data []byte, off int64,
...
-		if nCopied < n && nextBlockOff < 0 {
-			// If the block doesn't already have a parent block, make one.
-			if ptr == fd.rootBlockPointer() {
-				topBlock, err = fd.createIndirectBlock(
-					df, DefaultNewBlockDataVersion(false))
+		if nCopied < n {
+			needExtendFile := nextBlockOff < 0
+			needFillHole := off+nCopied < nextBlockOff
+			newBlockOff := startOff + int64(len(block.Contents))
+			if newBlockOff < off {
```
akalin-keybase: I think it's worth baking in the assumption that `nCopied == 0`. Perhaps:

```
if nCopied == 0 {
  if newBlockOff < off {
    // We are ... 
    newBlockOff = off
  }
} else {
  assert(newBlockOff == off + nCopied)
}
```

strib:
Thanks @akalin-keybase!  Addressed those and rebased.

libkbfs/file_data_test.go
```diff
@@ -0,0 +1,739 @@
...
+		halfCapacity = capacity - 1
+	}
+	// Fills half the leaf level.
+	testFileDataWriteExtendEmptyFile(t, 2, 2, int64(halfCapacity))
+	// Fills whole leaf level.
+	testFileDataWriteExtendEmptyFile(t, 2, 2, int64(capacity))
+
+}
+
+func TestFileDataWriteNewOneLevel(t *testing.T) {
```
akalin-keybase: May want to use the new go subtests mechanism to do something like:

```
func TestFileDataWriteNew(t *testing.T) {
  for _, level := range([]int{1, 2, 3, 10}) {
    // capture range variable.
    level := level
    t.Run(fmt.Sprintf("%d levels", level), func(t *testing.T) {
      testFileDataWriteNewLevel(t, level)
    })
  }
}
```

libkbfs/file_data_test.go
```diff
@@ -0,0 +1,739 @@
...
+func testFileDataExtendExistingLevels(t *testing.T, levels float64) {
+	capacity := math.Pow(2, levels)
+	halfCapacity := capacity / 2
+	// Starts with one lower level and adds a level.
+	testFileDataWriteExtendExistingFile(
+		t, 2, 2, int64(halfCapacity), int64(capacity))
+}
+
+func TestFileDataExtendExistingOneLevel(t *testing.T) {
+	testFileDataExtendExistingLevels(t, 1)
```
akalin-keybase: here too, everywhere else.

strib:
Ignoring yet another osx test timeout.  Merging.

PR #626, opened 2017-01-04
folder_block_manager_test: reduce incremental test size

strib:
With the defaults, this test can take more than 12 seconds on a slow
OSX machine when run with a disk-based bserver.

Issue: KBFS-1774


strib:
Unrelated chat failure in kbfs-server.  Merging.

PR #635, opened 2017-01-05
return TLF breaks even if the channel is already closed

mmaxim:
As it is now, if you call `getTlfBreakAndClose` more than once you don't get any break results back on the subsequent calls. The `GetTLFCryptKeys` codepath right now calls this guy twice, the second of which fails to load any breaks into the result. With this change, that behavior is fixed, and we check to see if the `tlfBreaks` field has any data before returning an empty result.

cc @maxtaco @chrisnojima


mmaxim:
That works for me, I'm not amazingly familiar with this stuff, but that change seems to still have the right behavior from my original perspective, so I am fine with it.

PR #637, opened 2017-01-05
updating circle to use yarn and new project layout

chrisnojima:


chrisnojima:
@keybase/react-hackers @strib circle now passes

PR #638, opened 2017-01-05
Finder likes to use UID 0 so let\'s let it?

songgao:
This fixes those damaged/corrupted error from Finder on macOS

### DIFF
--- a/libfuse/dir.go
+++ b/libfuse/dir.go
@@ -385,9 +385,9 @@ func (f *Folder) fillAttrWithUIDAndWritePerm(
 }
 
 func (f *Folder) access(ctx context.Context, r *fuse.AccessRequest) error {
-	if int(r.Uid) != os.Getuid() {
-		// short path: not accessible by anybody other than the logged in user.
-		// This is in case we enable AllowOther in the future.
+	if int(r.Uid) != os.Getuid() && int(r.Uid) != 0 {
+		// short path: not accessible by anybody other than root or the user who
+		// executed the kbfsfuse process.
 		return fuse.EPERM
 	}
 
--- a/libfuse/dir.go
+++ b/libfuse/dir.go
@@ -385,7 +385,11 @@ func (f *Folder) fillAttrWithUIDAndWritePerm(
 }
 
 func (f *Folder) access(ctx context.Context, r *fuse.AccessRequest) error {
-	if int(r.Uid) != os.Getuid() && int(r.Uid) != 0 {
+	if int(r.Uid) != os.Getuid() &&
+		// Finder likes to use UID 0 for some operations. osxfuse already allows
+		// ACCESS and GETXATTR requests from root to go through. This allows root
+		// in ACCESS handler. See KBFS-1733 for more details.
+		int(r.Uid) != 0 {
 		// short path: not accessible by anybody other than root or the user who
 		// executed the kbfsfuse process.
 		return fuse.EPERM
--- a/libfuse/file.go
+++ b/libfuse/file.go
@@ -119,7 +119,11 @@ var _ fs.NodeAccesser = (*File)(nil)
 // success, which makes it think the file is executable, yielding a "Unix
 // executable" UTI.
 func (f *File) Access(ctx context.Context, r *fuse.AccessRequest) error {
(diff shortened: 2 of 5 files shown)


songgao:
p.s. I think we should also use the `auto_xattr` option, which saves us one syscall rtt in all xattr operations. Bazil doesn't have that yet so I'll open a PR there for it.

strib:
Nice job tracking this down.  I believe one of the reasons we don't allow root is to keep Spotlight from indexing private files, so this makes me a little nervous.  Let's discuss more in the JIRA.

strib:
The osxfuse documentation says:

> You need to be a member of the osxfuse Admin group to use allow_root.

Are we sure all our users will have that?  cc @gabriel.

keybase-travis:
It used to always be a problem for nfs to allow nfs files to show up as
root and have setuid possible.  It would allow a malicious nfs server to
give a local unprivileged attacker root escalation

songgao:
We dont show files as owned by root anymore. Also we didn't enable the `AllowSUID` option. Would that be fine for this exploit?

songgao:
I don't have the "osxfuse Admin" group on my macOS. I created it manually and `allow_root` still worked. Could it be deprecated already?

songgao:
Ah I see; it's actually the `admin` group: https://github.com/thejoelpatrol/fusehfs#status-7192014

gabriel:
I don't think all users on a macos computer will be guaranteed to be an administrator (in admin group)... For example a test user on my account doesn't:

```sh
> groups test
staff everyone localaccounts com.apple.sharepoint.group.1 com.apple.sharepoint.group.2 com.apple.sharepoint.group.3 _lpoperator
```

We can change the admin group for `allow_other` to `everybody` group using (in the installer):
```
sudo sysctl -w vfs.generic.kbfuse.tunables.admin_group=12
```

songgao:
@strib do you think we should filter PID as well? If we do so, we'd need to do it in every fuse op.

songgao:
Gabriel brought up a good point that we don't have to worry about spotlight since it needs to create a directory at mount root which we don't allow any way.

strib:
@songgao: no, I don't think we should special case PIDs unless we absolutely have to.  Seems difficult to get right.

This PR should probably wait until we make the `sysctl` change to the macOS installer/updater that @gabriel pointed out above.

songgao:
Sounds good; thanks!

gabriel:
The sysctl call is privileged so we'd have to update the helper tool. Because this is a helper tool update, we'd have to ask for root password to update it.

We also have the option of only setting it for new installs only (don't upgrade people). But this is only ok, if this setting is optional. What happens if this value isn't set?

(And then another option is we could alter our Fuse bundle to change the default value in code and do a fuse upgrade, which doesn't require asking for root password since we already have that privilege.)

PR for this in progress at: https://github.com/keybase/client/pull/5310

gabriel:
cc @malgorithms ^

gabriel:
Maybe we don't need to alter the vfs.generic.kbfuse.tunables.admin_group setting if we are doing allow_root (and not allow_other)?

maxtaco:
might this setting get nuked after an upgrade?

songgao:
If we don't set the admin group to something else, a user without administrator privilege would fail to mount KBFS. So this only affects the situation where KBFS is installed on a system (by an admin apparently), and another user (which is not an admin) wants to use it. Since we are not really supporting multiple users using KBFS at the same time for now, I think this would be a pretty rare case. We might just wait until next bigger update that requires the privileged updater to update the admin group?

gabriel:
I can add a check which looks at current users groups and if not an admin user and have an old helper will show an error dialog, telling them to re-install. I think this is the best of all worlds?

songgao:
@strib said he'll email Ben to ask about why `allow_root` required `admin` in the first place. Slack discussion here: https://keybase.slack.com/archives/kbfs/p1484005036003735

songgao:
I removed the `allow_root` option for now since simply allowing `uid=0` in ACCESS worked for me. And per Ben suggested osxfuse already allows ACCESS and GETXATTR requests from root.

Note that this likely doesn't fix the ticket itself, which was created before we had ACCESS handler. Still waiting for it to come up again ...

PR #652, opened 2017-01-13
put the service in debug mode

mmaxim:
It's almost impossible to figure out chat failures without having debug mode on the service.

cc @patrickxb


mmaxim:
It's fine with me, but I guess we can see if it makes it to hard to work with

PR #660, opened 2017-01-13
[KBFS-1842] Initial Fixes

jzila:
TODO

- [ ] Need to correctly handle `Put`s, in that currently the prefetcher doesn't notice when a new directory is the same as the old one.
- [ ] UpdatePointer is called both when we write a new block and when an update comes from remote. We need to figure out how to tell them apart. Only the remote one should trigger a prefetch.
- [ ] Capture these prefetch explosion scenarios in prefetcher tests.

### DIFF
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -15,6 +15,11 @@ import (
 	"github.com/keybase/kbfs/tlf"
 )
 
+type blockContainer struct {
+	block         Block
+	hasPrefetched bool
+}
+
 type idCacheKey struct {
 	tlf           tlf.ID
 	plaintextHash kbfshash.RawDefaultHash
@@ -71,11 +76,11 @@ func NewBlockCacheStandard(transientCapacity int,
 func (b *BlockCacheStandard) GetWithPrefetch(ptr BlockPointer) (Block, bool, error) {
 	if b.cleanTransient != nil {
 		if tmp, ok := b.cleanTransient.Get(ptr.ID); ok {
-			block, ok := tmp.(Block)
+			bc, ok := tmp.(blockContainer)
 			if !ok {
 				return nil, false, BadDataError{ptr.ID}
 			}
-			return block, true, nil
+			return bc.block, bc.hasPrefetched, nil
 		}
 	}
 
@@ -113,10 +118,11 @@ func getCachedBlockSize(block Block) uint32 {
 }
 
 func (b *BlockCacheStandard) onEvict(key interface{}, value interface{}) {
-	block, ok := value.(Block)
+	bc, ok := value.(blockContainer)
 	if !ok {
 		return
 	}
+	block := bc.block
 
 	b.bytesLock.Lock()
 	defer b.bytesLock.Unlock()
@@ -231,7 +237,7 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 		// Cache it later, once we know there's room
 		defer func() {
 			if madeRoom {
-				b.cleanTransient.Add(ptr.ID, block)
+				b.cleanTransient.Add(ptr.ID, blockContainer{block, hasPrefetched})
 			}
 		}()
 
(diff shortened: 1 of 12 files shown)


libkbfs/folder_block_ops.go
```diff
@@ -2588,6 +2586,7 @@ func (fbo *folderBlockOps) getDeferredWriteCountForTest(lState *lockState) int {
 }
 
 func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, newPtr BlockPointer) {
+	fbo.log.CDebugf(context.Background(), "Updating reference for pointer %s to %s", oldPtr.ID, newPtr.ID)
```
strib: This is going to be a pretty intensely frequent log message, and only useful in certain conditions.  But there have been times when I've want something like it.  If this is intentional, please print out the full `BlockPointer`, instead of just the ID, since the refnonces could be useful as well.

libkbfs/prefetcher.go
```diff
@@ -122,32 +128,35 @@ func (p *blockPrefetcher) request(priority int, kmd KeyMetadata, ptr BlockPointe
...
 }
 
-func (p *blockPrefetcher) prefetchIndirectDirBlock(b *DirBlock, kmd KeyMetadata, priority int) {
+func (p *blockPrefetcher) prefetchIndirectDirBlock(b *DirBlock, kmd KeyMetadata) {
 	// Prefetch the first <n> indirect block pointers.
 	numIPtrs := len(b.IPtrs)
 	if numIPtrs > defaultIndirectPointerPrefetchCount {
 		numIPtrs = defaultIndirectPointerPrefetchCount
 	}
+	p.log.CDebugf(context.Background(), "Prefetching pointers for indirect dir block. Num pointers to prefetch: %d", numIPtrs)
```
strib: Please use `context.TODO()` here and below.  We should probably make a new log debug tag to identify background prefetches, to make debugging easier.  Can be in a followup PR though.

libkbfs/bcache.go
```diff
@@ -85,10 +90,16 @@ func (b *BlockCacheStandard) Get(ptr BlockPointer) (Block, error) {
 		return b.cleanPermanent[ptr.ID]
 	}()
 	if block != nil {
-		return block, nil
+		return block, true, nil
```
jzila: K. From my perspective, it's because a permanent entry can only be created if this client is performing a write. Since the client is writing, it knows what goes into it, including any potential directory entries or indirect blocks.

libkbfs/block_retrieval_queue.go
```diff
@@ -223,11 +235,26 @@ func (brq *blockRetrievalQueue) WorkOnRequest() <-chan *blockRetrieval {
 	return ch
 }
 
+func (brq *blockRetrievalQueue) triggerPrefetchAfterBlockRetrieved(
+	block Block, kmd KeyMetadata, priority int, hasPrefetched bool) {
+	if hasPrefetched {
```
jzila: I did it this way because the alternative was to perform the check before each time this function is called. I could do that too, but it seemed less DRY.

libkbfs/folder_block_ops.go
jzila: I'm going to leave it for now because of debugging, but I'll add a `TODO` to remove it.

libkbfs/bcache.go
```diff
@@ -67,15 +72,15 @@ func NewBlockCacheStandard(transientCapacity int,
...
 	if b.cleanTransient != nil {
 		if tmp, ok := b.cleanTransient.Get(ptr.ID); ok {
-			block, ok := tmp.(Block)
+			bc, ok := tmp.(blockContainer)
 			if !ok {
-				return nil, BadDataError{ptr.ID}
+				return nil, false, BadDataError{ptr.ID}
 			}
-			return block, nil
+			return bc.block, bc.hasPrefetched, nil
```
jzila: I'm about to add some code that will use the `false` version of `PutWithPrefetch`.

jzila:
I'll implement the feedback in a subsequent PR.

libkbfs/block_retrieval_queue.go
jzila: I removed this function entirely and fixed it up so the deadlock won't happen.

PR #664, opened 2017-01-16
[KBFS-1842] More fixes

jzila:
In this PR:
* The Prefetcher correctly caches the prefetched state of a block.
* `Put`s to the block cache don't trigger prefetches unless they come through via the `Get` path, and they didn't already trigger prefetches before.
* `UpdatePointers` prefetches now only happen when a change comes from the server rather than locally.

### DIFF
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -238,7 +238,10 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 		if b.cleanTransient == nil {
 			return nil
 		}
-		wasInCache = b.cleanTransient.Contains(ptr.ID)
+		// We could use `cleanTransient.Contains()`, but that wouldn't update
+		// the LRU time. By using `Get`, we make it less likely that another
+		// goroutine will evict this block before we can `Put` it again.
+		_, wasInCache = b.cleanTransient.Get(ptr.ID)
 		// Cache it later, once we know there's room
 
 	case PermanentEntry:
@@ -256,6 +259,10 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 	transientCacheHasRoom := true
 	// We must make room whether the cache is transient or permanent, but only
 	// if it wasn't already in the cache.
+	// TODO: This is racy, where another goroutine can evict or add this block
+	// between our check above and our attempt to make room. If the other
+	// goroutine evicts this block, we under-count its size as 0. If the other
+	// goroutine inserts this block, we double-count it.
 	if !wasInCache {
 		size := uint64(getCachedBlockSize(block))
 		transientCacheHasRoom = b.makeRoomForSize(size)
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -216,6 +216,11 @@ func (b *BlockCacheStandard) makeRoomForSize(size uint64) bool {
 }
 
 // PutWithPrefetch implements the BlockCache interface for BlockCacheStandard.
+// This method is idempotent for a given ptr, but that invariant is not
+// currently goroutine-safe, and it does not hold if a block size changes
+// between Puts. That is, we assume that a cached block associated with a given
+// pointer will never change its size, even when it gets Put into the cache
+// again.
 func (b *BlockCacheStandard) PutWithPrefetch(
 	ptr BlockPointer, tlf tlf.ID, block Block, lifetime BlockCacheLifetime,
 	hasPrefetched bool) (err error) {
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -220,6 +220,8 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 	ptr BlockPointer, tlf tlf.ID, block Block, lifetime BlockCacheLifetime,
 	hasPrefetched bool) (err error) {
 
+	_, cacheGetError := b.Get(ptr)
+	wasInCache := cacheGetError == nil
 	madeRoom := false
 
 	switch lifetime {
@@ -240,7 +242,7 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 		}
 		// Cache it later, once we know there's room
 		defer func() {
-			if madeRoom {
+			if madeRoom || wasInCache {
 				b.cleanTransient.Add(ptr.ID, blockContainer{block, hasPrefetched})
 			} else {
 				err = cachePutCacheFullError{ptr}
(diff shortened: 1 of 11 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -179,8 +179,8 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context, priority int, kmd K
 				brq.config.BlockCache().GetWithPrefetch(ptr)
 			if err == nil && cachedBlock != nil {
 				block.Set(cachedBlock, brq.config.codec())
-				brq.triggerPrefetchAfterBlockRetrieved(
-					cachedBlock, kmd, priority, hasPrefetched)
+				brq.Prefetcher().PrefetchAfterBlockRetrieved(
```
strib: I see that CI found a couple test hangs, maybe your `TogglePrefetcher` change below wasn't enough and this still needs to be a `go` invocation for some reason?

libkbfs/folder_block_ops.go
```diff
@@ -2657,7 +2662,7 @@ func (fbo *folderBlockOps) fastForwardDirAndChildrenLocked(ctx context.Context,
 		fbo.log.CDebugf(ctx, "Fast-forwarding %v -> %v",
 			child.BlockPointer, entry.BlockPointer)
 		fbo.updatePointer(kmd, child.BlockPointer,
-			entry.BlockPointer)
+			entry.BlockPointer, false)
```
strib: Ooo, I forgot about the fast-forward cases.  These should also be `shouldPrefetch = true`.

libkbfs/block_retrieval_queue.go
jzila: I don't think this can be a `go` invocation, because we need the cache to have the entry when the channel returns, otherwise we could potentially race against future cache `Get`s.

jzila:
@strib: OK I think I got all the races and deadlocks. Ready for another look. I'll make sure CI passes, although I don't anticipate that any of the current failures are due to this PR.

libkbfs/bcache_test.go
```diff
@@ -340,8 +340,8 @@ func TestBcacheEvictIncludesPermanentSize(t *testing.T) {
 	block.SetEncodedSize(7)
 	id := kbfsblock.FakeID(8)
 	ptr = BlockPointer{ID: id}
-	if err := bcache.Put(ptr, tlf, block, TransientEntry); err != nil {
-		t.Errorf("Got error on Put for block %s: %v", id, err)
+	if err, isCacheFullError := bcache.Put(ptr, tlf, block, TransientEntry).(cachePutCacheFullError); !isCacheFullError {
+		t.Errorf("Got wrong error on Put for block %s: %v", id, err)
```
strib: Maybe "wrong" -> "unexpected"?

libkbfs/interfaces.go
```diff
@@ -1023,8 +1023,9 @@ type Prefetcher interface {
 	// PrefetchAfterBlockRetrieved allows the prefetcher to trigger prefetches
 	// after a block has been retrieved. Whichever component is responsible for
 	// retrieving blocks will call this method once it's done retrieving a
-	// block.
-	PrefetchAfterBlockRetrieved(b Block, kmd KeyMetadata, priority int,
+	// block. It caches if it has triggered a prefetch and returns that.
```
strib: "and returns that" -- it looks like this method has no return value.

libkbfs/bcache.go
```diff
@@ -238,6 +242,8 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 		defer func() {
 			if madeRoom {
 				b.cleanTransient.Add(ptr.ID, blockContainer{block, hasPrefetched})
+			} else {
+				err = cachePutCacheFullError{ptr}
```
strib: Oh, I just realized that calling `Put` and `PutWithPrefetch` for the same block will double-count the size of the block in `cleanTotalBytes` (in `makeRoomForSize`)!  So if the block is already in the cache, we should probably subtract its size (and delete it from the cache?) before making room for it again.

jzila: So `Put` isn't idempotent?

jzila: Good catch. I'll make it idempotent.

strib: Thanks!  It definitely should be.  Looks like a pre-existing bug that was made much more likely by `PutWithPrefetch`.

jzila:
@strib: ready

libkbfs/bcache.go
```diff
@@ -214,8 +218,10 @@ func (b *BlockCacheStandard) makeRoomForSize(size uint64) bool {
 // PutWithPrefetch implements the BlockCache interface for BlockCacheStandard.
 func (b *BlockCacheStandard) PutWithPrefetch(
 	ptr BlockPointer, tlf tlf.ID, block Block, lifetime BlockCacheLifetime,
-	hasPrefetched bool) error {
+	hasPrefetched bool) (err error) {
 
+	_, cacheGetError := b.Get(ptr)
```
strib: Would we ever call this twice on a permanent block?  Maybe this check should go below in each of the lifetime cases, checking the respective cache for that lifetime.

Also it's inherently racy since there's no lock taken across the get and the put.  That's probably not worth fixing, but could use a TODO here.

jzila:
@strib: ready for final review.
https://github.com/keybase/kbfs/pull/664/files/29e6e6ca9dc32e91569e18cc12956521d3909913..65616ad1502d8c3820f039614702b2dcc1919c01

libkbfs/bcache_test.go
```diff
@@ -10,10 +10,11 @@ import (
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/kbfshash"
 	"github.com/keybase/kbfs/tlf"
+	"github.com/stretchr/testify/require"
 	"golang.org/x/net/context"
 )
 
-func blockCacheTestInit(t *testing.T, capacity int,
+func blockCacheTestBlockCacheInit(t *testing.T, capacity int,
```
strib: Why the name change here?  Isn't it redundant having "blockCache" twice?

libkbfs/bcache_test.go
```diff
@@ -387,17 +341,38 @@ func TestPutNoHashCalculation(t *testing.T) {
...
+	testBcachePutWithBlock(t, id1, cache, TransientEntry, block)
+	require.Equal(t, bytes, cache.cleanTotalBytes)
+
+	t.Log("Put the same block into the cache. Check that the cache's byte usage hasn't changed.")
+	testBcachePutWithBlock(t, id1, cache, TransientEntry, block)
+	require.Equal(t, bytes, cache.cleanTotalBytes)
+
+	t.Log("Put a new block into the cache, evicting the first one. Check that the byte usage is updated correctly.")
+	block = NewFileBlock().(*FileBlock)
+	bytes = uint64(len(block.Contents))
```
strib: Isn't this always 0? Seems like the test would be a little stronger if it has a non-zero and non-16 number of bytes.

libkbfs/bcache_test.go
jzila: regex replace fail

PR #666, opened 2017-01-17
cr_chains: collapse syncOps into one when creating the chain

strib:
This will reduce log spam with for big files with consecutive writes
that end up in CR (i.e., because of journal coalescing), and should
also reduce the size of CR MD updates a bit.

Issue: KBFS-1536


strib:
osx timeout is par for the course these days. Merging.

PR #675, opened 2017-01-20
parallelize RekeyQueueStandard.processRekeys

songgao:


libkbfs/rekey_queue.go
```diff
@@ -57,7 +61,9 @@ func (rkq *RekeyQueueStandard) Enqueue(id tlf.ID) <-chan error {
 			// spawn goroutine if needed
 			var ctx context.Context
 			ctx, rkq.cancel = context.WithCancel(context.Background())
-			go rkq.processRekeys(ctx, rkq.hasWorkCh)
+			for i := 0; i < numRekeyWorkers; i++ {
```
strib: `hasWorkCh` only has a buffer of size 1, and when `Enqueue` sends on it, there's a default case that does nothing.  I think there can be races where multiple Rekeys get enqueued, but only one worker thread is woken up to process them, because the other signals are dropped.  We should probably bump the channel size to `numRetryWorkers`, or change the queue into a proper channel itself.

libkbfs/rekey_queue.go
```diff
@@ -14,6 +14,10 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	numRekeyWorkers = 8
```
strib: Any particular reason for 8?  I feel like we could probably handle more...

songgao: No particular reason ... Does 64 sound good? Does it need to be in Config?

libkbfs/rekey_queue.go
songgao: Good point; Thanks! I'm not sure we can simply replace the queue with `chan` here, since there's a `IsRekeyPending` method which needs to iterate through the queue. I suppose we could use a separate `map` to track pending ones, which should also make the code more readable. Will try to refactor this a little bit

songgao:
@strib I refactored it a little bit; tests are also passing. Could you please take another look? Thanks!

libkbfs/rekey_queue.go
```diff
@@ -37,148 +60,114 @@ var _ RekeyQueue = (*RekeyQueueStandard)(nil)
...
-			go rkq.processRekeys(ctx, rkq.hasWorkCh)
-		}
-		rkq.queue = append(rkq.queue, rekeyQueueEntry{id, c})
-	}()
-	// poke the channel
-	select {
-	case rkq.hasWorkCh <- struct{}{}:
-	default:
+
+	if ch := rkq.GetRekeyChannel(id); ch != nil {
```
strib: There's a race here since this isn't done under the same lock as the one adding it to `pendings`.  The race already existed, and is not a big deal, but maybe worth a comment?

libkbfs/rekey_queue.go
```diff
@@ -37,148 +60,114 @@ var _ RekeyQueue = (*RekeyQueueStandard)(nil)
...
+
+	rkq.wg.Add(1)
+
+	rkq.lock.Lock()
+	defer rkq.lock.Unlock()
+
+	rkq.ensureRunningLocked()
+
+	ch := make(chan error, 1)
+	rkq.queue <- rekeyQueueEntry{id: id, ch: ch}
```
strib: Appending to the queue while holding the lock is probably not a good idea, since it can block.

In fact, `folderBranchOps` calls `Enqueue` while holding an important lock as well, so if this blocks it might cause issues.

I understand why it's attractive to put the ids in a channel, but maybe we should do something simple like this to be safe:

```go
select {
  case rkq.queue <- rekeyQueueEntry{id: id, ch: ch}:
  default:
    // The queue is full; avoid blocking by spawning a goroutine.
    rkq.log.CDebugf(context.TODO(), "Rekey queue is full; enqueuing %s in the background", id)
    go func() { rkq.queue <- rekeyQueueEntry{id: id, ch: ch} }
}

```

libkbfs/rekey_queue.go
```diff
@@ -37,148 +60,114 @@ var _ RekeyQueue = (*RekeyQueueStandard)(nil)
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case entry := <-rkq.queue:
+			newCtx := ctxWithRandomIDReplayable(ctx, CtxRekeyIDKey,
+				CtxRekeyOpID, nil)
+			rkq.log.CDebugf(newCtx, "Processing rekey for %s", entry.id)
+			err := rkq.config.KBFSOps().Rekey(newCtx, entry.id)
+			rkq.done(entry, err)
```
strib: Maybe this case should be wrapped in a `func()`, with this `done` call being a `defer` so that `wg` still gets decremented even on a panic?

libkbfs/rekey_queue.go
```diff
@@ -37,148 +60,114 @@ var _ RekeyQueue = (*RekeyQueueStandard)(nil)
...
-
-// CtxRekeyOpID is the display name for the unique operation
-// enqueued rekey ID tag.
-const CtxRekeyOpID = "REKEYID"
-
-// Dedicated goroutine to process the rekey queue.
-func (rkq *RekeyQueueStandard) processRekeys(ctx context.Context, hasWorkCh chan struct{}) {
-	for {
+	rkq.cancel()
+	for more := true; more; { // drain rkq.queue
```
strib: Can do `for e := range rkq.queue` instead.

libkbfs/rekey_queue.go
songgao: Good idea; will do!

libkbfs/rekey_queue.go
songgao: Ah that's why the `done` call was deferred in previous version. Thanks!

libkbfs/rekey_queue.go
songgao: hmm... doesn't it block when `rkq.queue` is empty but not closed? This approach would drain it and move on when it's empty

strib: Oh, I forgot you weren't closing the channel. Ok then, fine as is.

libkbfs/rekey_queue.go
songgao: good point! I'll just do another check below the exclusive lock then

PR #681, opened 2017-01-22
bcache: use same lifetime for PutWithPrefetch, and delete transient block along with pointer

strib:
This fixes a few things needed to make #663 pass tests.

In block_util, delete cached block when deleting known pointer.
Otherwise dedup code might still try to retrieve the block from the
cache, and the subsequent PutWithPrefetch with re-add the hash to
bcache.ids.

And in bcache, return lifetime from GetWithPrefetch. So the subsequent PutWithPrefetch can use the same lifetime (in the case of PermanentEntry blocks).

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -169,11 +169,15 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context, priority int, kmd K
 	for {
 		br, exists := brq.ptrs[bpLookup]
 		if !exists {
-			// Attempt to retrieve the block from the cache. This might be a
-			// specific type where the request blocks are CommonBlocks, but
-			// that direction can Set correctly. The cache will never have
-			// CommonBlocks.
-			cachedBlock, hasPrefetched, err :=
+			// Attempt to retrieve the block from the cache. This
+			// might be a specific type where the request blocks are
+			// CommonBlocks, but that direction can Set correctly. The
+			// cache will never have CommonBlocks.  TODO: verify that
+			// the returned lifetime here matches `lifetime` (which
+			// should always be TransientEntry, since a PermanentEntry
+			// would have been served directly from the cache
+			// elsewhere)?
+			cachedBlock, hasPrefetched, _, err :=
 				brq.config.BlockCache().GetWithPrefetch(ptr)
 			if err == nil && cachedBlock != nil {
 				block.Set(cachedBlock, brq.config.Codec())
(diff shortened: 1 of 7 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -173,7 +173,7 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context, priority int, kmd K
 			// specific type where the request blocks are CommonBlocks, but
 			// that direction can Set correctly. The cache will never have
 			// CommonBlocks.
-			cachedBlock, hasPrefetched, err :=
```
jzila: You should probably do the same thing here. Although we'll have a race for Permanent -> Transient changes.

When we `DeletePermanent`, should we automatically put the block into `Transient` under lock?

jzila: Or maybe we need a `PutPermanentToTransient` which atomically removes an entry from the Permanent cache and adds it to the Transient cache.

strib: Since `lifetime` is passed into `Request()`, I didn't know what to do if `bcache.Get()` returns something different.  I _think_ it should be the case that all requests that end up hitting the block retrieval queue will be transient entries anyway, since permanent entries by definition are guaranteed to be in the cache and would have been returned earlier by `folderBlockOps.getBlockHelperLocked()`.

What race are you worried about exactly?  Something getting re-put as Permanent after it's transitioned to Transient?  Even if the above logic is wrong, I think we can fix that by not calling `PrefetchAfterBlockRetrieved` when `hasPrefetched == true` in the cache already, since there's no point.

Anyway, I'll add a TODO here and merge as is, but let me know if you want me to fix something in a follow up PR.

jzila: In `FinishSync` we `DeletePermanent` for each pointer in `syncState.newIndirectFileBlockPtrs`, but I don't see where we insert those items into the transient cache. Is that done somewhere else?

The race I'm worried about is what happens if `GetWithPrefetch` returns a `PermanentEntry`, but it's deleted before you can `Put` it back.

strib: The blocks re put in the transient cache in `folderBranchOps.finalizeBlocks`, after the MD is successfully put to the mdserver.

And I don't think that race can happen here, since `Request()` should never see permanent entries.  I think it also can't happen in `getBlockHelperLocked` either, since `blockLock` is taken across both the `GetWithPrefetch` and `PrefetchAfterBlockRetrieved` calls, as well as the `DeletePermanent` call.

PR #682, opened 2017-01-23
[KBFS-1874] Prefetcher Regression Tests

jzila:
Also fixed a bug with the block cache where byte accounting broke if
permanent entries overfilled the cache.

### DIFF
--- a/libkbfs/prefetcher_test.go
+++ b/libkbfs/prefetcher_test.go
@@ -55,7 +55,7 @@ func makeFakeDirBlock(t *testing.T, name string) *DirBlock {
 	}}
 }
 
-func initPrefetcherTest(t *testing.T) (*blockRetrievalQueue, *blockRetrievalWorker, *fakeBlockGetter, func() BlockCache) {
+func initPrefetcherTest(t *testing.T) (*blockRetrievalQueue, *blockRetrievalWorker, *fakeBlockGetter, *testBlockRetrievalConfig) {
 	config := newTestBlockRetrievalConfig(t)
 	q := newBlockRetrievalQueue(1, config)
 	require.NotNil(t, q)
@@ -64,7 +64,7 @@ func initPrefetcherTest(t *testing.T) (*blockRetrievalQueue, *blockRetrievalWork
 	w := newBlockRetrievalWorker(bg, q)
 	require.NotNil(t, w)
 
-	return q, w, bg, config.BlockCache
+	return q, w, bg, config
 }
 
 func shutdownPrefetcherTest(q *blockRetrievalQueue, w *blockRetrievalWorker) {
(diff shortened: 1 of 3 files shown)


libkbfs/prefetcher_test.go
```diff
@@ -272,7 +276,106 @@ func TestPrefetcherDirectDirBlockAlreadyCached(t *testing.T) {
...
 	<-q.Prefetcher().Shutdown()
 
-	block, err = cacheFunc().Get(dir2.Children["b"].BlockPointer)
+	block, err = config.BlockCache().Get(dir2.Children["b"].BlockPointer)
 	require.NoError(t, err)
 	require.Equal(t, file1, block)
 }
+
+func TestPrefetcherNoPrefetchWhileCacheFull(t *testing.T) {
+	t.Log("Test that prefetches aren't triggered when the cache is full.")
```
strib: "is full with permanent entries", right?  The cache can be "full" of transient entries, but prefetching will still happen because the older ones get evicted I think.

libkbfs/prefetcher_test.go
```diff
@@ -272,7 +276,106 @@ func TestPrefetcherDirectDirBlockAlreadyCached(t *testing.T) {
...
+	block = &DirBlock{}
+	ch = q.Request(context.Background(), defaultOnDemandRequestPriority, makeKMD(), ptr2, block, TransientEntry)
+	continueCh2 <- nil
+	err = <-ch
+	require.NoError(t, err)
+	require.Equal(t, dir1, block)
+
+	t.Log("Shutdown the prefetcher and wait until it's done prefetching." +
+		" This shouldn't block, indicating that no prefetches were triggered.")
+	<-q.Prefetcher().Shutdown()
```
strib: If this shouldn't block, should we add a `default` case with an error?

jzila: We still need it to wait. The `<-` here is for synchronization.

libkbfs/prefetcher_test.go
jzila: Indeed. I can make a test for that scenario too, if you like.

libkbfs/prefetcher_test.go
strib: The comment above says "This shouldn't block".  So shouldn't a check like this succeed?

```go
select {
  case <-q.Prefetcher().Shutdown():
  default:
    t.Fatalf("Shutdown blocked")
}
```

Or am I misunderstanding the log message?

jzila: It should really be "this shouldn't hang". Your suggestion would be racy.

libkbfs/prefetcher_test.go
strib: No need, just making sure I understand.  I think just clarifying in the log message is good enough for me, but more tests are always welcome of course.

libkbfs/prefetcher_test.go
strib: Oh ok.  Yeah, then I guess please clarify the log message.

PR #691, opened 2017-01-24
[KBFS-1874] Prefetcher Fixes

jzila:


libkbfs/folder_block_ops.go
```diff
@@ -2679,29 +2679,32 @@ func (fbo *folderBlockOps) getDeferredWriteCountForTest(lState *lockState) int {
...
-	// TODO KBFS-1850: Eventually we should use the codec library's ability to
-	// decode into a nil interface to no longer need to pre-initialize the
-	// correct type.
-	block, err := fbo.config.BlockCache().Get(oldPtr)
-	if err != nil {
-		return
-	}
+	// Only prefetch if the updated pointer is a new block ID.
+	if oldPtr.ID != newPtr.ID {
+		fbo.log.CDebugf(context.TODO(), "Updated reference for pointer %s to %s.", oldPtr.ID, newPtr.ID)
```
jzila: I tried `DeleteTransient` here, but looks like there are still requests for blocks even after the pointer has been updated.

libkbfs/folder_block_ops.go
```diff
@@ -2679,29 +2679,32 @@ func (fbo *folderBlockOps) getDeferredWriteCountForTest(lState *lockState) int {
 func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, newPtr BlockPointer, shouldPrefetch bool) {
 	// TODO: Remove this comment when we're done debugging because it'll be
```
strib: Move this comment along with the debug msg.

libkbfs/folder_block_ops.go
strib: In tests, or in real life?  Seems likely to happen during CR, but otherwise not sure how it would happen in real life.  Tests might be weird though.

libkbfs/prefetcher_test.go
```diff
@@ -289,12 +291,30 @@ func TestPrefetcherDirectDirBlockAlreadyCached(t *testing.T) {
...
+
+	t.Log("Remove the prefetched file block from the cache.")
+	cache.DeleteTransient(dir2.Children["b"].BlockPointer, kmd.TlfID())
+	_, err = cache.Get(dir2.Children["b"].BlockPointer)
+	require.EqualError(t, err, NoSuchBlockError{dir2.Children["b"].BlockPointer.ID}.Error())
+
+	t.Log("Restart the prefetcher.")
+	q.TogglePrefetcher(context.Background(), true)
+
+	t.Log("Request the second-level directory block again. No prefetches should be triggered.")
```
strib: Is there a way to verify this?  Maybe try to get `dir2.Children["b"].BlockPointer` from the cache again after shutting down the prefetcher?

libkbfs/folder_block_ops.go
jzila: Real Life.
Real Life

jzila: FWIW it looks like it's happening during the journaling-related CR.

libkbfs/prefetcher_test.go
jzila: It is verified below where the prefetcher doesn't hang. If it had requested its child block, `<-q.Prefetcher().Shutdown()` would hang on `continueCh3`.

PR #692, opened 2017-01-24
libkbfs: don't let block changes blocks explode in test

strib:
And re-enable flaky test that was timing out.

The size of block pointers increased since I added a direct type to them, and that caused some tests to exceed the default `BlockSplitterSimple.blockChangeEmbedMaxSize` when they had `BlockSplitterSimple.maxSize = 5`.  That caused a HUGE explosion in the number of blocks being written by the test.

So instead bump up the embed max size, so we don't have to waste test time on unembedded block change blocks.


strib:
(A review from either of you is good enough for this one.)

PR #697, opened 2017-01-26
GetTLFCryptKeys: don't early out in identifyOnce for chat CORE-4465

mmaxim:
We don't want to use the KBFS cache on identifies in chat, the hope is the caching in the service will be able to handle the influx of calls. The particular symptom is that if an identify breaks and the service knows it (probably from the background identifier), then we want the chat system to be consistent with that knowledge. 

cc @maxtaco

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -837,7 +837,9 @@ func (fbo *folderBranchOps) identifyOnce(
 	ctx context.Context, md ReadOnlyRootMetadata) error {
 	fbo.identifyLock.Lock()
 	defer fbo.identifyLock.Unlock()
-	if fbo.identifyDone {
+
+	ei := getExtendedIdentify(ctx)
+	if fbo.identifyDone && !ei.behavior.AlwaysRunIdentify() {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


mmaxim:
Also FWIW, maybe we should think about a story here where the background identifier can push to KBFS and this cache can be cleared? Otherwise I can be in a state where `keybase id` will show a failure but I can still roll into the folder with that person.

maxtaco:
The solution might be a simple cache bust call over to kbfs

maxtaco:
Good point!!

strib:
> The solution might be a simple cache bust call over to kbfs

Agreed that would be a much better solution.  Is that possible within whatever time frame you need this done?

mmaxim:
We would have to track down all the cache busting points in the service and make sure they call into KBFS with the busting call. I think that should be doable within a reasonable amount of time. Could we merge this in the meantime though?

keybase-travis:
Mike's original proposal seems much easier, since the background identifier
can easily know old vs new identify status

mmaxim:
Note I don't just need the background identifier to trigger the behavior above. I could just have easily typed `keybase id -s candrencil`.

libkbfs/folder_branch_ops.go
```diff
@@ -837,7 +837,9 @@ func (fbo *folderBranchOps) identifyOnce(
 	ctx context.Context, md ReadOnlyRootMetadata) error {
 	fbo.identifyLock.Lock()
 	defer fbo.identifyLock.Unlock()
-	if fbo.identifyDone {
+
+	ei := getExtendedIdentify(ctx)
+	if fbo.identifyDone && !ei.behavior.AlwaysRunIdentify() {
```
strib: Please add a comment/TODO that this is temporary until the service sends cache-busting notices.

mmaxim: Will do!

PR #711, opened 2017-01-30
don't mark node as dirty in Write() if it's unlinked

songgao:
... so that a running process writing into an already-holding handle doesn't cause the file to show up again, or result in a stubborn entry in `DirtyNodes`.

### DIFF
--- a/libfuse/file.go
+++ b/libfuse/file.go
@@ -217,6 +217,7 @@ func (f *File) Write(ctx context.Context, req *fuse.WriteRequest,
 	if f.node.GetID().ParentID() == nil {
 		f.folder.fs.log.CDebugf(ctx,
 			"Ignoring write on unlinked file.", len(req.Data))
+		resp.Size = len(req.Data)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


songgao:
I think this fixes my particularly experiment with `cat` as described in the ticket. But the fact that it only happens when it's in a sub-dir, and not when it's the TLF root, is weird and concerning. Any ideas on why that's the case? @strib

songgao:
hmm the same test passes on my machine under a `repeat 100 ./libfuse.test -test.run=TestRemoveTLF`. Will take a look later to see why it fails on CI ...

strib:
This seems like the wrong fix to me.  You're allowed to write to a file handle even if it's been unlinked, and subsequent reads on that same handle should be able to see the data that was written.

I'd guess the real fix is something deeper relating to showing unlinked but dirty files in a `GetDirChildren`.

There's some trickiness in the root dir vs subdirs, since the root pointer is kept in the metadata, and I wouldn't be surprised if that leads to different behavior in the dirty case.  Would be good to have test cases for both so we can figure out what's different.

songgao:
Ah good point; Thanks! I totally missed the case where the same handle could be used for reading. Will work on a different fix.

PR #713, opened 2017-01-31
cr_chains: used cached changes when getting ops, if needed

strib:
The journal can't easily re-embed block changes (since it involves reading and decrypting blocks from disk), so we need to look in cached changes too.

Included is a regression test, but it won't actually test the regression until KBFS-1912 is fixed.  I'll be doing that in a separate PR.

Issue: KBFS-1901

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1862,8 +1862,8 @@ func (fbo *folderBranchOps) unembedBlockChanges(
 
 	md.AddRefBytes(uint64(info.EncodedSize))
 	md.AddDiskUsage(uint64(info.EncodedSize))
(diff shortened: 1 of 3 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -1862,8 +1862,8 @@ func (fbo *folderBranchOps) unembedBlockChanges(
 
 	md.AddRefBytes(uint64(info.EncodedSize))
 	md.AddDiskUsage(uint64(info.EncodedSize))
-	changes.Info = info
 	md.data.cachedChanges = *changes
+	changes.Info = info
```
strib: Having it above kept the block info in the cached changes, which could end up confusing things quite a bit.

PR #714, opened 2017-02-01
test: set block sizes for all users, and fix resulting bugs

strib:
It turns out the DSL tests had a problem where the block sizes weren't being properly set for all users after the first one.  Once I fixed it, there was some other fallout that needed fixing:

* When syncOps were collapsed together during CR, the unref pointers from the now-deleted sync ops did not make it to the final resolution.  This would leak quota for the user, and leave defunct blocks unarchived.
* When unflushed blocks were deleted from the journal after CR, we were still leaving them in the resulting `resolutionOp`.  This is harmless in practice except for making the `resolutionOp` larger than necessary.  But in tests, it confused the `StateChecker`.
* Fixing the test issue caused the libkbfs DSL tests to take 3x as long as before, due to all the new blocks.  This was mostly caused by tests where `blockSize=20` and `blockChangeSize=5`, because the size of the unembedded block changes would grow exponentially, creating a ton more work for the test.  Instead I increased the block size for those tests a bit, which makes a huge difference and doesn't sacrifice coverage as far as I can tell.  On my laptop, both libkbfs and FUSE-based DSL tests only each increase by ~10s.

Issue: KBFS-1912

### DIFF
--- a/test/cr_multi_blocks_test.go
+++ b/test/cr_multi_blocks_test.go
@@ -11,7 +11,7 @@ import "testing"
 // bob writes a multi-block file while unmerged, no conflicts
 func TestCrUnmergedWriteMultiblockFile(t *testing.T) {
 	test(t,
-		blockSize(20), users("alice", "bob"),
+		blockSize(20), blockChangeSize(20*1024), users("alice", "bob"),
 		as(alice,
 			mkdir("a"),
 		),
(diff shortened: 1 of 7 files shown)


strib:
Ugh: https://ci.keyba.se/job/kbfs/job/PR-714/1/execution/node/344/log/.  Looks like I'll have to tweak some more parameters, but it shouldn't be a big change.  Still worth reviewing.

test/cr_multi_blocks_test.go
```diff
@@ -11,7 +11,7 @@ import "testing"
 // bob writes a multi-block file while unmerged, no conflicts
 func TestCrUnmergedWriteMultiblockFile(t *testing.T) {
 	test(t,
-		blockSize(20), users("alice", "bob"),
+		blockSize(20), blockChangeSize(20*1024), users("alice", "bob"),
```
strib: These two tests were causing us to use too many goroutines with `-race` using the default block change size, I think due to prefetching.

PR #717, opened 2017-02-01
folder_branch_ops: cancel registration when clearing folder MD

strib:
If we don't, then the next time the head is set we'll have two update goroutines, and will end up panicking in MDServerRemote.

Issue: KBFS-1918

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4954,6 +4954,12 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 		expBackoff.MaxElapsedTime = 0
 		// Register and wait in a loop unless we hit an unrecoverable error
 		fbo.cancelUpdatesLock.Lock()
+		if fbo.cancelUpdates != nil {
+			// It should be impossible to get here without having
+			// already called the cancel function, but just in case
+			// call it here again.
+			fbo.cancelUpdates()
+		}
 		ctx, fbo.cancelUpdates = context.WithCancel(ctx)
 		fbo.cancelUpdatesLock.Unlock()
 		for {
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -280,6 +280,10 @@ type folderBranchOps struct {
 	// Can be used to turn off notifications for a while (e.g., for testing)
 	updatePauseChan chan (<-chan struct{})
 
+	cancelUpdatesLock sync.Mutex
+	// Cancels the goroutine currently waiting on TLF MD updates.
+	cancelUpdates context.CancelFunc
+
 	// After a shutdown, this channel will be closed when the register
 	// goroutine completes.
 	updateDoneChan chan struct{}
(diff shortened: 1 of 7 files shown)


libkbfs/mdserver_remote.go
```diff
@@ -385,10 +388,30 @@ func (md *MDServerRemote) cancelObservers() {
...
+		// not registered
+		return
+	}
+
+	// signal that we've seen the update
+	md.signalObserverLocked(
+		observerChan, id, errors.New("Registration canceled"))
+	// Setting nil here indicates that the server thinks we're still
+	// registered, though locally no one is listening.
+	md.observers[id] = nil
```
songgao: How do we know if the server thinks we are still registered? The comment in the interface says "let the MDServer know that we are no longer interested ...". Does the server refer to a different context?

strib: Ah, yeah that's confusing.  The actual server in the backend still thinks we are registered, because we registered previously and never told the server to stop sending us updates (there is no RPC for that, besides killing our connection).  The comment is referring to the local `MDServerRemote` instance.  I'll fix up that interface comment to be more clear.

libkbfs/folder_branch_ops.go
```diff
@@ -4976,7 +4982,9 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 					fbo.log.CDebugf(ctx, "Abandoning updates since we can't "+
 						"read the newest metadata: %+v", err)
 					fbo.status.setPermErr(err)
-					cancel()
+					// No need to lock here, since `cancelUpdates` is
+					// only set within this same goroutine.
```
jzila: I don't understand this comment. What if a `ClearPrivateFolderMD` call happens in the meantime?

libkbfs/mdserver_remote.go
songgao: I see; thanks for explaining!

libkbfs/folder_branch_ops.go
```diff
@@ -4949,7 +4953,9 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 		// Never give up hope until we shut down
 		expBackoff.MaxElapsedTime = 0
 		// Register and wait in a loop unless we hit an unrecoverable error
-		ctx, cancel := context.WithCancel(ctx)
+		fbo.cancelUpdatesLock.Lock()
+		ctx, fbo.cancelUpdates = context.WithCancel(ctx)
```
jzila: What if `fbo.cancelUpdates` was already set?

strib: Since we're enforcing there can only be one update goroutine at a time, we can safely overwrite any existing `cancelUpdates` here.  Whatever the old one was, it's useless now.

libkbfs/folder_branch_ops.go
strib: `ClearPrivateFolderMD` doesn't write to the `fbo.cancelUpdates` field, so whatever's there is safe to read from this goroutine without taking the lock.  Furthermore, `ClearPrivateFolderMD` waits until the goroutine completes before releasing any of its locks, so no new goroutine can be spawned to update it either.  Any suggestions how I can make that more clear?

libkbfs/folder_branch_ops.go
jzila: Shouldn't we call the old one if it's non-nil, just in case?

strib: I can do that, and will, but I think it's impossible to get here a second time without it having been called once already.

libkbfs/folder_branch_ops.go
jzila: Ah I see. Fair enough. Carry on!

PR #726, opened 2017-02-04
file_data: treat unknown children as direct blocks

strib:
Since we didn't have multiple levels of indirection before introducing
the direct type.  Without this, a new client removing an old file
needs to download all child direct data blocks, which can be slow.

Also fix temporary block pointers and the fileData test to set direct
types on block pointers, since otherwise we'd violate the above
assumption about multiple levels of indirection.

Issue: KBFS-1924

### DIFF
--- a/libkbfs/file_data_test.go
+++ b/libkbfs/file_data_test.go
@@ -24,9 +24,10 @@ func setupFileDataTest(t *testing.T, maxBlockSize int64,
 	maxPtrsPerBlock int) (*fileData, BlockCache, DirtyBlockCache, *dirtyFile) {
 	// Make a fake file.
 	ptr := BlockPointer{
-		ID:      kbfsblock.FakeID(42),
-		KeyGen:  1,
-		DataVer: 1,
+		ID:         kbfsblock.FakeID(42),
+		KeyGen:     1,
+		DataVer:    1,
+		DirectType: DirectBlock,
 	}
 	id := tlf.FakeID(1, false)
 	file := path{FolderBranch{Tlf: id}, []pathNode{{ptr, "file"}}}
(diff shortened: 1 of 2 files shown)


libkbfs/file_data_test.go
```diff
@@ -24,9 +24,10 @@ func setupFileDataTest(t *testing.T, maxBlockSize int64,
 	maxPtrsPerBlock int) (*fileData, BlockCache, DirtyBlockCache, *dirtyFile) {
 	// Make a fake file.
 	ptr := BlockPointer{
-		ID:      kbfsblock.FakeID(42),
-		KeyGen:  1,
-		DataVer: 1,
+		ID:         kbfsblock.FakeID(42),
+		KeyGen:     1,
+		DataVer:    1,
```
jzila: Shouldn't this be `AtLeastTwoLevelsOfChildrenDataVer`?

libkbfs/file_data_test.go
```diff
@@ -378,13 +379,18 @@ func testFileDataLevelExistingBlocks(t *testing.T, fd *fileData,
...
 			for j, child := range children {
 				id, err := crypto.MakeTemporaryBlockID()
 				require.NoError(t, err)
 				ptr := BlockPointer{
-					ID:      id,
-					KeyGen:  1,
-					DataVer: 1,
+					ID:         id,
+					KeyGen:     1,
+					DataVer:    1,
```
jzila: Same here, `AtLeastTwoLevelsOfChildrenDataVer`

libkbfs/file_data_test.go
```diff
@@ -405,6 +411,11 @@ func testFileDataLevelExistingBlocks(t *testing.T, fd *fileData,
 		prevChildren = level
 		numLevels++
 	}
+
+	if numLevels > 1 {
+		fd.file.path[len(fd.file.path)-1].DirectType = IndirectBlock
```
jzila: So you're setting the last pointer in the path to `IndirectBlock`. Shouldn't the last one be `DirectBlock`?

strib: The `path` represents the set of directories leading up the file; the tail pointer of the file is actually the root block.  (Same as `fs.rootBlockPointer()`.)  So if there's more than one level in the file, this sets the root block to be indirect.

libkbfs/file_data_test.go
strib: Well it depends on the size of the tree I guess.  But this test doesn't actually use this field at all, so I'm just going to delete it in both places so it's not confusing.

strib:
Ready for further review.

strib:
Not sure why appveyor failed or why jenkins didn't run, but it previously passed tests and this is just a rebase (and it builds/passes libkbfs tests locally), so I'm merging.

PR #730, opened 2017-02-06
[KBFS-1895] Refactor before Prefetching Workers

jzila:
Let's get this refactor in no matter what, because it's better separation of concerns, and will make it easier to add prefetch workers if we still want to.

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -21,19 +21,19 @@ const (
 	defaultOnDemandRequestPriority       int = 100
 )
 
-type blockRetrievalParentConfig interface {
+type blockRetrievalPartialConfig interface {
 	dataVersioner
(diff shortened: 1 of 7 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -16,16 +16,31 @@ import (
 
 const (
 	defaultBlockRetrievalWorkerQueueSize int = 100
+	defaultNumPrefetchWorkers            int = 1
 	testBlockRetrievalWorkerQueueSize    int = 5
 	defaultOnDemandRequestPriority       int = 100
 )
 
-type blockRetrievalConfig interface {
+type blockRetrievalParentConfig interface {
```
strib: The word "parent" is throwing me off a bit for some reason.  Maybe "partial" or "sub" instead?

PR #738, opened 2017-02-08
send notification for EXDEV so frontend can do popups if moving across TLFs is from an unsupported app e.g. Finder

songgao:


sysutils/pid_path_darwin.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
```
strib: 2017 in the copyrights, please.

sysutils/pid_path_darwin.go
```diff
@@ -0,0 +1,31 @@
...
+const (
+	procpidpathinfo     = 11
+	procpidpathinfosize = 1024
+	proccallnumpidinfo  = 2
+)
+
+// GetExecPathFromPID returns the process's executable path for given PID.
+func GetExecPathFromPID(pid uint32) (string, error) {
+	buf := make([]byte, procpidpathinfosize)
+	_, _, errno := syscall.Syscall6(syscall.SYS_PROC_INFO, proccallnumpidinfo, uintptr(pid), procpidpathinfo, 0, uintptr(unsafe.Pointer(&buf[0])), procpidpathinfosize)
```
strib: Please break up this line a bit.

sysutils/pid_path_darwin.go
```diff
@@ -0,0 +1,31 @@
...
+	if errno != 0 {
+		return "", errno
+	}
+	nonZero := bytes.IndexByte(buf, 0)
```
strib: Isn't this the first instance of the 0 byte?  `nonZero` seems like a confusing name.  `firstZero` seems better.

sysutils/pid_path_others.go
```diff
@@ -0,0 +1,20 @@
...
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+// +build !darwin,!linux
+
+package sysutils
+
+// NotImplementedError is returned on platforms where GetExecPathFromPID is not
+// implemented.
+type NotImplementedError struct{}
```
strib: Should this be defined in another file, so that other code can check for it specifically without needing `!darwin,!linux`?  Either that or it probably shouldn't be exported.

libfuse/dir.go
```diff
@@ -681,35 +682,35 @@ func (d *Dir) Rename(ctx context.Context, req *fuse.RenameRequest,
...
 
-	return nil
+	err = d.folder.fs.config.KBFSOps().Rename(ctx,
+		d.node, req.OldName, realNewDir.node, req.NewName)
+
+	switch err.(type) {
+	case nil:
+		return nil
+	case libkbfs.RenameAcrossDirsError:
+		epath, er := sysutils.GetExecPathFromPID(req.Pid)
```
strib: I think `execPathErr` would be more clear than `er`.

libfuse/dir.go
```diff
@@ -681,35 +682,35 @@ func (d *Dir) Rename(ctx context.Context, req *fuse.RenameRequest,
...
+		if er != nil {
+			d.folder.fs.log.CDebugf(ctx,
+				"Dir Rename: getting exec path for PID %d error: %v", req.Pid, er)
+			return err
+		}
+		return libkbfs.ExdevForUnsupportedApplicationError{ExecPath: epath}
```
strib: The comment for `ExdevForUnsupportedApplicationError` says that the error should only be used for programs that are known not to respect EXDEV.  But this code doesn't check anything about the program.  Is the idea that the desktop app will check?  If so, I guess that comment should be updated.  If not, we need a list of unsupported applications to check against here.

songgao: I talked to Gabriel and decided to have that list in frontend. I'll update the comment; thanks!

songgao:
@strib I ended up removing `ExdevForUnsupportedApplicationError` and added a field to `RenameAcrossDirsError`. Please take another look thanks!

PR #745, opened 2017-02-09
key_manager: mark rekey identifies with the KBFS_REKEY behavior

strib:
This lets the service suppress popups unless something is wrong.

Issue: KBFS-1928


maxtaco:
Thanks.  I think we're going to be suppressing tracker popups in all cases, even if there is an error. I think the issue is that it's just too confusing otherwise.

CC: @malgorithms  on what's best here.

strib:
cc: @maxtaco: do I need to wait until the client-side support for this identify behavior is ready, or will the client continue to perform the default behavior if it doesn't recognize the behavior?

strib:
> Thanks. I think we're going to be suppressing tracker popups in all cases, even if there is an error

Wow, that seems super duper dangerous.  Rekey is a pretty sensitive operation.

cc: @malgorithms

strib:
@maxtaco: oh maybe you mean the rekey will still fail, there just won't be a popup?

jzila:
Maybe instead of tracker popups, we should be introducing a tracker notification pane on the main app, along with high priority un-toggleable toasts.

maxtaco:
I meant the identify will fail, which will block the rekey, but the user won't know. 

Asking @oconnor663 what happens when we get a new enum we don't know.  I think we're ok.

maxtaco:
@jzila yes that's the plan on the "people tab" when we get around to it.

strib:
@maxtaco: thanks make sense!  I misinterpreted.

maxtaco:
@strib, yes....  presumably we can just return an error to `Identify2` but suppress all UI, and it hits all of the requirements?

oconnor663:
Here are the important functions for the identify behaviors: https://github.com/keybase/client/blob/master/go/protocol/keybase1/extras.go#L751-L764

The WarningInsteadOfErrorOnBrokenTracks function in particular is conservative, and it won't suppress errors for a new behavior that it doesn't recognize.

maxtaco:
@songgao yeah @oconnor663 is on it!

maxtaco:
Nice work here by @oconnor663 to make this addition so easy on the Core side btw...

oconnor663:
Pretty sure I was just following @songgao's pattern :)

PR #747, opened 2017-02-10
fix Ctrl-C on kbfsfuse

songgao:


songgao:
Please hold off the review. There still seems to be problems

songgao:
I looked at osxfuse's example filesystems' signal handlers. It seems unmounting the filesystem is enough and we shouldn't be closing the device file descriptor (i.e. `c.Close()` is wrong).

It looks like `umount` or `diskutil unmount force` only works after `fs.Serve` is called. We could drain `c.ReadRequest()` manually in a separate routine but that has dead lock problem again. So I moved `libkbfs.Init()` before the mount, and added a layer of indirection to the `interruptFn`.

Also bazil uses `syscall.Unmount` on darwin. The syscall has a flag argument but bazil uses constant `0`. Just to be safe, I changed our `Unmount` implementations to use `mount` command on darwin, which matches osxfuse's [way](https://github.com/osxfuse/fuse/blob/b2f7f794bf0e23cfe99b52e1e474510230d6645b/lib/fuse_darwin.c#L330-L353) of [handling signals](https://github.com/osxfuse/fuse/blob/b2f7f794bf0e23cfe99b52e1e474510230d6645b/lib/fuse_signals.c#L25-L69)

So this is open for review again @strib Thanks!

libkbfs/init.go
```diff
@@ -365,9 +366,25 @@ func Init(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn, onI
 			onInterruptFn()
 		}
 
-		os.Exit(1)
+		close(done)
 	}()
 
+	errCh := make(chan error)
+	go func() {
```
strib: Can you please add some comments here about why we need to call `doInit` in another goroutine?

libfuse/start.go
```diff
@@ -37,74 +37,51 @@ func Start(mounter Mounter, options StartOptions, kbCtx libkbfs.Context) *libfs.
...
-
-	config, err := libkbfs.Init(kbCtx, options.KbfsParams, nil, onInterruptFn, log)
-	if err != nil {
-		return libfs.InitError(err.Error())
-	}
-
-	defer libkbfs.Shutdown()
-
 	if c != nil {
+		<-c.Ready
```
strib: Previously we were only calling `<-done` (same as `<-c.Ready`) _after_ calling `fs.Serve()`, but now we're calling it before.  I assume you've tested it (on Linux too?) and it must be fine, but I just wanted to call it out and make sure you thought through this change.

songgao: `<-c.Ready` is closed when the mount is ready or has failed ([godoc](https://godoc.org/bazil.org/fuse#Conn)). So I thought it makes sense to do it before the Serve. I'll test on Linux to make sure before merging.

libkbfs/init.go
songgao: This is so that we can select on `done` and `errCh` below, in case SIGINT happens in the middle of the actual init. With this, we can (and I did) remove the `os.Exit(1)` from the SIGINT handler so everything finishes naturally. I'll add a comment. Thanks for pointing it out!

PR #758, opened 2017-02-13
update vendor to include timezone in timestamp in logger

songgao:
Sister PR in client: https://github.com/keybase/client/pull/5837


songgao:
Ah right!

PR #766, opened 2017-02-15
Simplefs implementation 1/N

taruti:
@zanderz

### DIFF
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -241,7 +241,7 @@ func (k *KeybaseDaemonRPC) OnConnect(ctx context.Context,
 		keybase1.NotifyPaperKeyProtocol(k),
 		keybase1.NotifyFSRequestProtocol(k),
 		keybase1.TlfKeysProtocol(k),
-//		keybase1.SimpleFSProtocol(&simplefs.SimpleFS{}),
+		//		keybase1.SimpleFSProtocol(&simplefs.SimpleFS{}),
 	}
 
 	if k.protocols != nil {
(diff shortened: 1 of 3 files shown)


libkbfs/keybase_daemon_rpc.go
```diff
@@ -241,7 +241,7 @@ func (k *KeybaseDaemonRPC) OnConnect(ctx context.Context,
 		keybase1.NotifyPaperKeyProtocol(k),
 		keybase1.NotifyFSRequestProtocol(k),
 		keybase1.TlfKeysProtocol(k),
-		keybase1.SimpleFSProtocol(&simplefs.SimpleFS{}),
+//		keybase1.SimpleFSProtocol(&simplefs.SimpleFS{}),
```
zanderz: What's the issue with leaving this in? I can't try it otherwise.

simplefs/simplefs.go
```diff
@@ -117,3 +232,238 @@ func (k *SimpleFS) SimpleFSGetOps(_ context.Context) ([]keybase1.OpDescription,
...
+func remotePath(path keybase1.Path) (ps []string, public bool, err error) {
+	pt, err := path.PathType()
+	if err != nil {
+		return nil, false, err
+	}
+	if pt != keybase1.PathType_KBFS {
+		return nil, false, errOnlyRemotePathSupported
+	}
+	raw := path.Kbfs()
+	if raw != `` && raw[0] == '/' {
```
zanderz: I think the way I have CLI now is that on Windows, the incoming KBFS path can start with the mounted drive letter instead of "keybase" - but we could normalize that to always start with "/keybase" if you want. Also, speaking of slashes, shouldn't we use os.PathSeparator instead of slash? or does that always have to be normalized to slashes too?

libkbfs/keybase_daemon_rpc.go
taruti: That would create an import cycle. libkbfs importing simplefs and simplefs importing libkbfs.

simplefs/simplefs.go
taruti: Pass for Windows `K:/private/foo` and Linux `/keybase/private/foo` just `/private/foo`. Just using slashed on all platforms makes sense.

PR #769, opened 2017-02-16
simplefs protocol tweak

zanderz:
Vendored from a client branch - still unclear whether I should be waiting for the client branch to be merged first before vendoring into here?


strib:
Please merge client first so that when you vendor, you have the real commit ID in vendor.json.

PR #777, opened 2017-02-20
make du/df work

songgao:
The investigation on what block size we should use turns out to be meaningless after all. FUSE seems to always assume and report 512-byte block size regardless what's provided by user-space implementation. This is verified on both Linux and macOS.

There still seems to be problem with very small files on macOS. It reports minimum 8 blocks. Consequently, a 1-byte file shows up as 4.0KB from `du`. In Linux, however, this works fine (report 1 block, and `du` shows 512 byte).

I also changed `crypto_common.go` so that file content with exactly power-of-two bytes of content doesn't get padded to the next power-of-two. I don't think, but am not sure if this breaks anything.


songgao:
hmm df doesn't work on macOS; will fix soon

songgao:
@strib It turns out on macOS we need to populate `FrSize` as well. I've verified on  Linux and macOS that `du` and `df` both work properly. So this is open for review again. Thanks!

songgao:
Not sure how, but it seems now with 7bb856f, CI consistently fails TestJournalCrManyFiles. I see there's a "flake" [ticket](https://keybase.atlassian.net/browse/KBFS-1826) for this test. Does it seem like the same case?

strib:
I haven't looked at the failure yet, but if it's failing because of a timeout when going through a FS, maybe it's causing the OS to write in chunks that are too small, slowing down the test so much that it hits the timeout? Just speculating, since that test writes a ton of files and any slowdown will affect it more than other tests.

songgao:
Ah interesting. Thanks for the theory! I'll try to confirm tomorrow

songgao:
With my Keybase app's KBFS, (i.e. admin build), I couldn't reproduce what `TestJournalCrManyFiles` tries to check. What I saw was exactly what I see when running the test on *this branch* (newly-created-then-removed files show up in unflushed paths when journal is paused). Running the same test on master branch, however, doesn't reproduce the same thing. I'm confused!

songgao:
Ready for review! (I'll fix all those tests in a different PR; there seems to be a few of them)

libfuse/block_size.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2016 Keybase Inc. All rights reserved.
```
strib: 2017

libfuse/fs.go
```diff
@@ -199,18 +199,22 @@ func (f *FS) Root() (fs.Node, error) {
...
-		Bavail:  ^uint64(0) / uint64(bsize),
-		Files:   0,
-		Ffree:   0,
-		Bsize:   bsize,
+		Bsize:   fuseBlockSize,
 		Namelen: ^uint32(0),
-		Frsize:  0,
+		Frsize:  fuseBlockSize,
 	}
+	status, _, err := f.config.KBFSOps().Status(ctx)
```
strib: I have some concerns about making an RPC to the bserver every time this is called, because it seems to be called a lot on OSX.  What do you think about caching the usage/quota info somewhere for (say for ~1 min, or until there's a local write operation / the journal flushes).  I don't have a great plan for how to do that, just brainstorming...

libkbfs/crypto_common.go
```diff
@@ -294,15 +294,15 @@ func (c CryptoCommon) DecryptPrivateMetadata(
 
 const minBlockSize = 256
 
-// nextPowerOfTwo returns next power of 2 greater than the input n.
+// thisPowerOfTwo returns next power of 2 greater than or equal to the input n.
 // https://en.wikipedia.org/wiki/Power_of_two#Algorithm_to_round_up_to_power_of_two
-func nextPowerOfTwo(n uint32) uint32 {
-	if n < minBlockSize {
+func thisPowerOfTwo(n int) int {
```
strib: @patrickxb: do you remember why you originally implemented this function as `nextPowerOfTwo` (called when doing KBFS block padding)?  I can't remember if we had a good reason for it or not.

test/journal_test.go
```diff
@@ -111,14 +111,19 @@ func TestJournalCrSimple(t *testing.T) {
 	)
 }
 
-// bob creates many conflicting files while running the journal.
-func TestJournalCrManyFiles(t *testing.T) {
-	var busyWork []fileOp
-	iters := 20
+func giveMeHiWork(iters int) (busyWork []fileOp) {
```
strib: I think `makeBusyWork` would be a more clear name.

songgao: It was `giveMeWork`. I added Hi to indicate it was all about the file "hi". Perhaps `makeBusyWorkForHi`?

libfuse/fs.go
songgao: Would it work if we use cached version until a md update comes in?

libfuse/block_size.go
songgao: Oops thx!

libfuse/fs.go
strib: This is a per-user thing, not a per-TLF thing, and the current device might not be listening for updates on all its TLFs being written to by another device.

test/journal_test.go
strib: If the caller needs to know the fliename, maybe just have them pass it in?  It seems awkward in the function name.

libfuse/fs.go
songgao: Ah that makes sense. I guess we'd have to do time based cache then. Or perhaps on top of that, if pid points to a df binary, make sure to hit the bserver?

test/journal_test.go
songgao: Sounds good!

libkbfs/crypto_common.go
patrickxb: Sorry, I don't remember...

libfuse/fs.go
strib: We can do a time-based cache, and just bust it early whenever there's a merged MD update on any TLF we have an FBO for?

Maybe we can add the `df` thing later, but I'm worried about other things like Finder (and similar things on other platforms) that show the total free space.

libkbfs/crypto_common.go
strib: Ok, I guess I'm fine with this, then.  But @songgao I find the name a little confusing, maybe just `powerOfTwo`, or `powerOfTwoEqualOrGreater`.  Also, looks like the wikipedia link has changed slightly: https://en.wikipedia.org/wiki/Power_of_two#Algorithm_to_find_a_power_of_two_greater_than_or_equal_to_a_number

It seem like we might not need the `if` case to check if it's already a power of two.

strib:
I'll approve after a rebase and nit fixes.  (I'm also going to test it out on my account.)

strib:
Works great!

libkbfs/crypto_common.go
songgao: Yes it's idempotent if it's already power of two; I thought that was there to save a couple instructions? New name sounds good!

libfuse/fs.go
songgao: Ah, good point on Finder might show total free space. 

> whenever there's a merged MD update on any TLF we have an FBO for
Doesn't this suffer from the same problem where we might not be tracking all TLFs updates? Or did you mean this is still better than pure time-based cache busting?

What expiration time do you think is a good choice? I'm thinking more of a 5s rather than 1m

strib:
(Actually, shouldn't have said I'll approve after nits -- this should probably have at least basic caching too...)

songgao:
Yup! I'll make the updates and tag you for review again. Thanks!

libfuse/fs.go
strib: > Or did you mean this is still better than pure time-based cache busting?

Yeah, we'd just be using it as a hint since we know that the usage will have changed.  I guess it can be an optimization for a later PR though, if it's too hard.

> What expiration time do you think is a good choice? I'm thinking more of a 5s rather than 1m

Can you test on macOS how often it calls `statfs` both when you're idle and when you're using KBFS?  Hard to guess at a value without knowing the numbers.

songgao: >  I guess it can be an optimization for a later PR though

OK! I'll add the time-based cache first and see how complicated the hint part is ...

>  how often it calls `statfs`

IIRC, it's fairly often. I'll make sure. But doesn't expiring at 5s mean it's at-most 1 per-5s? Is that bad?

strib: It seems pretty bad for every osx client to be making a non-trivial RPC every 5 seconds when otherwise idle.

songgao: Yeah it's fairly often when user is doing anything. Pretty much every operation in Finder triggers at least 1 Statfs call. CLI is less excessive, where it seems only doing things in mount root triggers it.

It doesn't seem like there's any Statfs calls trigerred in idle state though.

So if user is reading/writing stuff from/into KBFS, they are already making RPCs a lot right? If idle is not a problem, does 5s still seems too short?

strib: Ok, good that there are none in an idle state.  Though now I'm worried that we'll be blocking foreground operations on an unnecessary server call once every 5 seconds.  Once journaling is on, this will be pretty noticeable.  (Not sure if those `statfs` calls actually block the other Finder operation or if they're off the critical path.)

Maybe we can do something where we trigger a background RPC to update the cache, but still serve it from the cache if it's less than 2*timeout seconds old?

songgao: Hmm but if user hasn't touched KBFS for e.g. 1h, and suddenly decide to do a `df`, this would give user data from 1h ago. Same thing applies to looking at storage in Finder.

But background RPC is a good idea I think. If the concern if "block ever 5s", perhaps we can do it in background, but only if the last available one was retrieved within 10s? That is, if the data is older than 10s, a RPC would be made. This should cover the case when user is doing a lot of stuff in Finder, I think

strib: > perhaps we can do it in background, but only if the last available one was retrieved within 10s?

That's exactly what I said above: "still serve it from the cache if it's less than 2*timeout seconds old".

songgao: Oops; so sorry I missed that part! sounds good then. Thanks Jeremy

songgao:
Ready for another round of review. Thanks @strib!

libfuse/constants.go
```diff
@@ -0,0 +1,12 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libfuse
+
+import "time"
+
+// QuotaUsageStaleTolerance is the lifespan of stale usage data that libfuse
+// accetps in the Statfs handler. In other words, this causes libkbfs to issue
```
strib: accetps -> accepts

libfuse/fs.go
```diff
@@ -199,18 +202,22 @@ func (f *FS) Root() (fs.Node, error) {
...
-		Ffree:   0,
-		Bsize:   bsize,
+		Bsize:   fuseBlockSize,
 		Namelen: ^uint32(0),
-		Frsize:  0,
+		Frsize:  fuseBlockSize,
+	}
+	usageBytes, limitBytes, err := f.quotaUsage.Get(ctx, QuotaUsageStaleTolerance)
+	if err != nil {
+		return fuse.EIO
```
strib: Why not `return err`?  It will automatically get an EIO error code if the error doesn't have a specific code attached to it already.

strib: Also, since we don't have a `reportErr` call for this function, we should probably log any errors here.

libkbfs/kbfs_ops.go
```diff
@@ -667,15 +668,9 @@ func (fs *KBFSOpsStandard) Status(ctx context.Context) (
...
-			limitBytes = quotaInfo.Limit
-			if quotaInfo.Total != nil {
-				usageBytes = quotaInfo.Total.Bytes[kbfsblock.UsageWrite]
-			} else {
-				usageBytes = 0
-			}
-		}
+		// The error is ignored here so that other fields can still be populated
+		// even if this fails.
+		usageBytes, limitBytes, _ = fs.quotaUsage.Get(ctx, 0)
```
strib: I think it's still worth logging the error.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	"golang.org/x/net/context"
+
+	"github.com/keybase/client/go/logger"
+	"github.com/keybase/kbfs/kbfsblock"
+)
+
+//ECQUCtxTagKey is the type for unique ECQU background opertaion IDs.
+type ECQUCtxTagKey struct{}
+
+// ECQUID == "ECQU"
```
strib: I think a sentence explanation of what this is used for would be more useful.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/keybase/client/go/logger"
+	"github.com/keybase/kbfs/kbfsblock"
+)
+
+//ECQUCtxTagKey is the type for unique ECQU background opertaion IDs.
```
strib: Need space after "//".

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+
+package libkbfs
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"golang.org/x/net/context"
+
```
strib: Most of the time we don't have separate import blocks here, but I don't feel too strongly about it.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	cached cachedQuotaUsage
+}
+
+// NewEventuallyConsistentQuotaUsage creates a new
+// EventuallyConsistentQuotaUsage object.
+func NewEventuallyConsistentQuotaUsage(
+	config Config) *EventuallyConsistentQuotaUsage {
+	return &EventuallyConsistentQuotaUsage{
+		config: config,
+		log:    config.MakeLogger(ECQUID),
```
strib: It's more useful to have an instance-specific logger tag, rather than a struct-specific one, since we can already tell that it comes from the struct using the file name.  I would either get rid of the tag, or pass one in so we can distinguish between libfuse and KBFSOps requests.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (q *EventuallyConsistentQuotaUsage) getAndCache(
+	ctx context.Context) (usage cachedQuotaUsage, err error) {
+	defer func() {
+		q.log.CDebugf(ctx, "getAndCache: error=%v", err)
+	}()
+	quotaInfo, err := q.config.BlockServer().GetUserQuotaInfo(ctx)
+	if err != nil {
+		return usage, err
```
strib: Explicitly doing `return cachedQuotaUsage{}, err` would be a little safer.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	usage.limitBytes = quotaInfo.Limit
+	if quotaInfo.Total != nil {
+		usage.usageBytes = quotaInfo.Total.Bytes[kbfsblock.UsageWrite]
+	} else {
+		usage.usageBytes = 0
+	}
+	usage.timestamp = time.Now()
```
strib: Please use `q.config.Clock().Now()`.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+// RPC is issued and the function only returns after RPC finishes, with the
+// newest data from RPC.
+func (q *EventuallyConsistentQuotaUsage) Get(ctx context.Context,
+	tolerance time.Duration) (usageBytes, limitBytes int64, err error) {
+	c := func() cachedQuotaUsage {
+		q.mu.RLock()
+		defer q.mu.RUnlock()
+		return q.cached
+	}()
+	past := time.Since(c.timestamp)
```
strib: If you use `q.config.Clock()` above, you'll have to change this to be an explicit `Sub()` call, sorry.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	q.cached = usage
+
+	return usage, nil
+}
+
+// Get returns KBFS bytes used and limit for user. To help avoid having too
+// frequent calls into bserver, caller can provide a positive tolerance, to
+// accept stale LimitBytes and UsageByts data.
+//
+// 1) If the cached (stale) data is older for less than half of tolerance, the
```
strib: Grammar nit: "If the age of the cached data is less than half..."  Please fix the other ones below in a similar way.

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+	c := func() cachedQuotaUsage {
+		q.mu.RLock()
+		defer q.mu.RUnlock()
+		return q.cached
+	}()
+	past := time.Since(c.timestamp)
+	switch {
+	case past > tolerance:
+		q.log.CDebugf(ctx, "Blocking on getAndCache. Cached data is %s old.", past)
+		c, err = q.getAndCache(ctx)
```
strib: We could optimize here to make sure there's only ever one RPC outstanding at any point in time, though it's probably not worth doing for this PR.  Maybe add a TODO comment?

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+
+	q.mu.Lock()
+	defer q.mu.Unlock()
+	q.cached = usage
+
+	return usage, nil
+}
+
+// Get returns KBFS bytes used and limit for user. To help avoid having too
+// frequent calls into bserver, caller can provide a positive tolerance, to
```
strib: You mention it should be a positive tolerance, but don't check for that.  Return an error in that case?

libkbfs/quota_usage.go
```diff
@@ -0,0 +1,134 @@
...
+				logTags := make(logger.CtxLogTags)
+				logTags[ECQUCtxTagKey{}] = ECQUID
+				bgCtx := logger.NewContextWithLogTags(context.Background(), logTags)
+				bgCtx = context.WithValue(bgCtx, ECQUCtxTagKey{}, id)
+				// Make sure a timeout is on the context, in case the RPC blocks
+				// forever somehow, where we'd end up with never resetting
+				// backgroundInProcess flag again.
+				bgCtx, cancel := context.WithTimeout(bgCtx, 10*time.Second)
+				defer cancel()
+				q.getAndCache(bgCtx)
```
strib: Should log any errors here.

libfuse/constants.go
```diff
@@ -0,0 +1,12 @@
...
+// a fresh RPC call if cached usage data is older than 10s.
+const QuotaUsageStaleTolerance = 10 * time.Second
```
strib: I think you could put this in `fs.go` rather than have a whole file for it (it also doesn't need to be exported).

libfuse/fs.go
songgao: >  will automatically get an EIO error code

Ah right!

>  don't have a `reportErr` call for this function

Would you prefer to call `reportErr` here or is it overkill?

libkbfs/quota_usage.go
songgao: That was `goimports` when I changed `context` to `golang.org/x/net/context` I think. I removed space here and it happily rearranged it to have only two blocks. Will update!

libkbfs/quota_usage.go
songgao: That makes sense. I'll add a suffix to `ECQUID` to differentiate the two

libkbfs/quota_usage.go
songgao: If it's not positive, it's still valid; we'd just always make a RPC to ensure it's latest. I'll update comment to reflect it

libfuse/fs.go
strib: > Would you prefer to call reportErr here or is it overkill?

Nah, we removed statfs logging for the happy path since it caused the dsl fuse tests to log too much.

libkbfs/quota_usage.go
songgao: `getAndCache()` already logs the error in a deferred block. Is there any added information to log here?

songgao:
Thanks for the detailed review!

libfuse/fs.go
songgao: Got you; sounds good!

libkbfs/quota_usage.go
strib: Oh, I missed that. That's good. Though some linters might get upset that we're not checking the error here. Not sure what to do about that though. Maybe assigning to _ explicitly will tell linters that we're ignoring it on purpose?

songgao: Sounds good; will also add comment.

PR #779, opened 2017-02-21
bare_root_metadata_v3: don't error if MDDiskUsage looks bad

strib:
As long as the MDRefBytes fields are the same, it probably means that
an old client just copied the value as an unknown field from an MD
written by a new client.


strib:
Self-approving as this needs to go in ASAP.

PR #783, opened 2017-02-22
mdserver_remote: reset auth status on disconnect

strib:
If the device suspends for a while and comes back (without any new
keybase login events), the old mserver connection will disconnect
(resetting the background check-rekey timer to 60min) and it will
immediately try to reconnect.  But that reconnect won't call
`getFoldersForRekey` when it starts up, because `md.isAuthenticated`
is still `true`, even though we aren't really authenticated anymore.
This means that any rekey requests that came into the server while the
device was suspended won't get processed by the device for 60 minutes.

Instead, set `md.isAuthenticated` to false on a disconnect, so that
the reconnect will trigger a `getFoldersForRekey()` call.

Issue: keybase/client#5972

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -351,6 +351,10 @@ func (md *MDServerRemote) OnDisconnected(ctx context.Context,
 		md.serverOffset = 0
 	}()
 
+	md.authenticatedMtx.Lock()
+	md.isAuthenticated = false
+	md.authenticatedMtx.Unlock()
+
 	md.cancelObservers()
 	md.resetPingTicker(0)
 	if md.authToken != nil {


strib:
> Do we need to do the same thing here too?

I don't think so, since `OnConnectError` shouldn't get called if `OnConnect` is called.

I'll wait for @jzila to take a look too, since he wrote the original `isAuthenticated` stuff, I think.

libkbfs/mdserver_remote.go
```diff
@@ -351,6 +351,10 @@ func (md *MDServerRemote) OnDisconnected(ctx context.Context,
 		md.serverOffset = 0
 	}()
 
+	md.authenticatedMtx.Lock()
```
jzila: Don't you usually do the following approach to insure against panics? Or do you only do that when there's a reasonable chance that the line can panic?
```go
func() {
    md.authenticatedMtx.Lock()
    defer md.authenticatedMtx.Unlock()
    md.isAuthenticated = false
}()
```

strib: I was following your lead above in `resetAuth` to keep it consistent -- that approach is fine with me when the line has no chance of panicking.

jzila: Sounds good. Ship it!

PR #788, opened 2017-02-22
remove fsapi

zanderz:
I guess someone updated the vendoring after I took this out of the client side; should have done this sooner.


zanderz:
bigger mess than I thought, I will just bring back fs.avdl for now

PR #799, opened 2017-02-24
Taruti/simplefs8

taruti:
Async things as per email.


zanderz:
LGTM when CI passes :thumbsup:

PR #804, opened 2017-02-27
block_ops: if a block is in the local journal, fetch it directly

strib:
Otherwise, the block could get stuck behind a full queue of prefetching requests, for no reason.

@jzila: it strikes me that the disk cache will have exactly the same problem -- might want to move that logic out of `BlockServerRemote.Get()` and put it in `BlockOps.Get()` instead.

Issue: KBFS-1981

### DIFF
--- a/libkbfs/block_getter.go
+++ b/libkbfs/block_getter.go
@@ -8,8 +8,6 @@ import (
 	"fmt"
 
 	"github.com/keybase/kbfs/kbfsblock"
-	"github.com/keybase/kbfs/kbfscrypto"
-
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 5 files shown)


jzila:
@strib: I'm not sure I follow. The disk cache will respond quickly and free up workers immediately, if a block is present.

strib:
Yes, so does the journal.  The problem is that if the queue is already full with prefetching requests, the request for a local block will get stuck for no reason.  And once it returns, another prefetch request will take over the worker, and the next local block request will be stuck as well.  This is what was happening with journal requests.

It's less crucial for the disk cache case, because CR in particular makes some assumptions about how fast it can read journal blocks and no one should do that for the cache.  But still, it might be a good optimization to avoid useless waits.

jzila:
K thanks, I'll give it some thought.

strib:
@jzila: I agree this is hacky.  The only better thing I can think of is limit the number of workers dedicated to prefetching.

I figured the `realBlockGetter` thing is fine, since it's being instantiated in `BlockOps` already, so it already knows what kind of getter it's using.  But I could refactor `assembleBlock()` into a freestanding util function that takes a `blockOpsConfig` if you prefer.

What type of clarification do you have in mind?  Are you asking for more comments somewhere?

jzila:
I like your suggestion of a freestanding util function for `assembleBlock()`, except I'd explicitly pass in the `config.pureCrypto()`, `config.Codec()`, and `config.keyGetter()`, since configs are meant for actual components.

Regarding clarification, sorry, I realize that I wasn't very clear in my original comment. I wasn't implying any work for you in this PR, more that we should create a ticket to think about how the components relate. This ticket's story would be to more deliberately design how these components relate, in contrast to the current evolved relationship.

strib:
@jzila: done in 613cca2.  Will merge when CI passes, but let me know if you have any other comments.

PR #805, opened 2017-02-27
refactor rekey scheduling and add a 30s recheck

songgao:


libkbfs/folder_branch_ops.go
```diff
@@ -4678,10 +4681,30 @@ func (fbo *folderBranchOps) rekeyLocked(ctx context.Context,
...
+	if rekeyDone && !stillNeedsRekey && ttl > 0 {
+		// We did do the rekey, and there isn't another rekey scheduled. We enqueue
+		// the rekey here again, in case we missed a device due to a race
+		// condition. This is specifically for the situation where user provisions
+		// two devices in a row, and the key update for the 2nd device only comes
+		// in after rekey for a TLF is done, which didn't include the second
+		// device. At this point, there wouldn't be a new MD with rekey bit set
+		// since it's already set. As a result, the TLF won't get rekeyed for the
+		// second device until the next 1-hour timer triggers another scan.
+		time.AfterFunc(rekeyRecheckInterval, func() {
```
strib: It would be nice to have a log message before this stating that another rekey check has been scheduled for 30s.

strib: Also, will this leave around a bunch of goroutines in tests, after a successful rekey test finishes?  We should probably save the timer and stop it on `Shutdown()`.  That could also help us avoid scheduling multiple retry attempts.

strib: We could also have any new rekey request stop an old timer.

songgao:
Thanks for the review!

Re: putting it back to rekey queue: I went with that approach initially (see commits), but turned to this approach because otherwise we'd have to plumb through the TTL to the rekey queue as well. We could have a rate limiter on each TLF in rekey queue, but that seems less ideal than a TTL.

libkbfs/folder_branch_ops.go
songgao: good point!

songgao:
Or did you mean letting `fbo.Rekey()` report whether rekey did happen, and have the RekeyQueue manage the TTL itself?

strib:
> Or did you mean letting fbo.Rekey() report whether rekey did happen, and have the RekeyQueue manage the TTL itself?

Yes, that's what I was thinking, though I didn't completely think it through.

songgao:
Hmm.. OK I think that might work. Let me try that approach

songgao:
@strib ready for review again. Sorry for the huge PR! The meat's in `rekey_fsm.go`

libkbfs/rekey_fsm_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+package libkbfs
+
+import (
+	"context"
+
+	"github.com/keybase/kbfs/tlf"
+)
+
+func requestRekeyWithContextAndWaitForOneFinishEvent(
```
strib: Seems like you wouldn't need this function (or file) if you follow my suggestion above.

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+	m.muListenersForTest.Lock()
+	defer m.muListenersForTest.Unlock()
+	m.listenersForTest[event] = append(m.listenersForTest[event], listenerForTest{
+		body:       callback,
+		repeatedly: repeatedly,
+	})
+}
+
+func getRekeyFSMForTest(ops KBFSOps, tlfID tlf.ID) RekeyFSM {
+	switch o := ops.(type) {
```
strib: `default` case with a `panic`?

libkbfs/rekey_queue.go
```diff
@@ -36,188 +35,92 @@ import (
...
-				rkq.log.CDebugf(newCtx, "Processing rekey for %s", entry.id)
-				err = rkq.config.KBFSOps().Rekey(newCtx, entry.id)
-			}()
+// start spawns a goroutine that dispatches rekey requests to correct folder
+// branch ops while conforming to the rater limiter.
+func (rkq *RekeyQueueStandard) start(ctx context.Context) {
+	go func() {
+		for id := range rkq.queue {
+			if err := rkq.limiter.Wait(ctx); err != nil {
+				rkq.log.Debug("Waiting on rate limiter for tlf=%v error: %v", id, err)
```
strib: Did you intend to continue after the error?  Seems like this should be a panic or a return?

libkbfs/rekey_queue.go
```diff
@@ -36,188 +35,92 @@ import (
...
-		}
-	}
 	rkq.cancel = nil
 }
 
-// Wait implements the RekeyQueue interface for RekeyQueueStandard.
-func (rkq *RekeyQueueStandard) Wait(ctx context.Context) error {
-	return rkq.wg.Wait(ctx)
+// New implements the RekeyQueue interface for RekeyQueueStandard.
+func (rkq *RekeyQueueStandard) New() RekeyQueue {
```
strib: It seems confusing to have this instance construct its successor -- why not just have the caller do these two things?

libkbfs/kbfs_ops.go
```diff
@@ -482,7 +482,8 @@ func (fs *KBFSOpsStandard) getMaybeCreateRootNode(
 		if ops, ok := fs.ops[fb]; ok {
 			fs.log.CDebugf(ctx, "Triggering a paper prompt rekey on folder "+
 				"access due to unreadable MD for %s", h.GetCanonicalPath())
-			go ops.rekeyWithPrompt()
+			// TODO: do we really need explicit paper prompt here?
```
strib: Yes!  I think we do need it.  If someone tries to access the folder explicitly and a paper key is needed, we should show a prompt right away, rather than 10 minutes later.

libkbfs/kbfs_ops.go
```diff
@@ -698,12 +699,12 @@ func (fs *KBFSOpsStandard) UnstageForTesting(
 	return ops.UnstageForTesting(ctx, folderBranch)
 }
 
-// Rekey implements the KBFSOps interface for KBFSOpsStandard
-func (fs *KBFSOpsStandard) Rekey(ctx context.Context, id tlf.ID) error {
+// RequestRekey implements the KBFSOps interface for KBFSOpsStandard
+func (fs *KBFSOpsStandard) RequestRekey(id tlf.ID) {
```
strib: I think `ctx` needs to be passed in, since `getOps` uses it.

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+
+// rekeyTask describes a rekey task.
+type rekeyTask struct {
+	// timeout, if non-nil, causes rekey to fail if it takes more than this
+	// duration since it enters rekeyStateStarted.
+	timeout     *time.Duration
+	ttl         int
+	promptPaper bool
+
+	ctx context.Context
```
songgao: The `ctx` was added here mostly for logging, so that we'd have same REKEYID through out different retries. If that's not a desired behavior, I've no problem moving it to function parameters!

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+		return newRekeyStateStarted(r.fsm, r.task)
+	case rekeyRequestEvent:
+		task := r.task
+		task.promptPaper = task.promptPaper || event.request.promptPaper
+		if task.timeout == nil {
+			task.timeout = event.request.timeout
+		}
+		task.ttl = event.request.ttl
+		r.fsm.log.CDebugf(task.ctx, "Replacing context")
+		task.ctx = event.request.ctx
```
songgao: This is specifically to handle the stallers, which put staller key in the context. This is actually another reason why the context needs to be persisted through different states.

libkbfs/rekey_fsm.go
songgao: Sorry; actually stallers is another reason; see comments below

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+		r.fsm.log.CDebugf(task.ctx, "Context replaced")
+		if !r.deadline.After(time.Now().Add(event.request.delay)) {
+			r.fsm.log.CDebugf(r.task.ctx, "Reusing existing timer")
```
songgao: ah right. Thanks!

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+			r.fsm.log.CDebugf(r.task.ctx,
+				"Not scheduling new rekey because TTL expired.")
+			return newRekeyStateIdle(r.fsm)
+		}
+
+		switch event.finished.err {
+		case nil:
+		default:
+			r.fsm.log.CDebugf(r.task.ctx, "Rekey errored. Scheduling new rekey.")
+			return newRekeyStateScheduled(r.fsm, 0, rekeyTask{
```
songgao: I thought it wouldn't hurt now that there's a TTL. It could fail on other transient errors such as networking errors, etc. But I'm OK with checking errors explicitly as well

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+	current rekeyState
+
+	muListenersForTest sync.Mutex
+	listenersForTest   map[rekeyEventType][]listenerForTest
+}
+
+func (m *rekeyFSM) loop() {
+	for e := range m.reqs {
+		if e.eventType == rekeyShutdownEvent {
+			close(m.reqs)
```
songgao: It's only closed during shutdown. hmm I suppose I should add a guard in `Event()`?

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,488 @@
...
+		m.log.Debug("RekeyFSM transition: %T + %s -> %T",
+			m.current, e, next)
+		m.current = next
+
+		m.triggerCallbacksForTest(e)
+	}
+}
+
+// NewRekeyFSM creates a new rekey FSM.
+func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
```
songgao: Sure; although I was trying to make functions only call functions above them (I thought that was the convention?)

libdokan/rekey_file.go
```diff
@@ -24,7 +24,9 @@ func (f *RekeyFile) WriteFile(ctx context.Context, fi *dokan.FileInfo, bs []byte
 	if len(bs) == 0 {
 		return 0, nil
 	}
-	err = f.folder.fs.config.KBFSOps().Rekey(ctx, f.folder.getFolderBranch().Tlf)
+	// TODO: get rid of this and use RequestRekey instead
```
songgao: The difference between this and `RequestRekey` is, this is sync, and `RequestRekey` is async. We really only wait for rekey to finish here other than in tests. I was thinking we could get rid of "waiting for rekeys" in non-tests altogether, since a return from `rekeyLocked` doesn't necessarily mean the rekey is finished. It can spawn paper rekey prompts, or regular 30s rekey rechecks, which are all caused by the same rekey request.

libkbfs/rekey_fsm.go
strib: We definitely want a `ctx` threaded throughout the rekey, I just don't want it saved to a struct if possible.  But now that I've looked through all the code I don't see a great way to do that.  I suppose adding a `rekeyState.getContext` functor might be a little better (that is passed in explicitly to `reactToEvent`), since then at least it's clear that the context is not supposed to be modified by the FSM.

libkbfs/folder_branch_ops.go
```diff
@@ -4710,27 +4694,22 @@ func (fbo *folderBranchOps) rekeyLocked(ctx context.Context,
...
 
 	default:
-		if err == context.DeadlineExceeded {
-			fbo.log.CDebugf(ctx, "Paper key prompt timed out")
-			// Reschedule the prompt in the timeout case.
-			stillNeedsRekey = true
-		} else {
-			return err
-		}
+		return RekeyResult{}, err
```
songgao: > since other error types might be fatal
hmmm it seems the question here is if we want to stop when there is an error that could be fatal, or keep trying when there is an error that could be non-fatal. We have a TTL now so I was leaning towards the second option, thus this change.

libkbfs/rekey_fsm.go
strib: The problem is that some state cause goroutines, which can themselves enqueue events, and nothing shuts down those goroutines on `Shutdown` (which might be another issue -- leaking goroutines).

Yes a guard in `Event()` would be good.

libkbfs/rekey_fsm.go
strib: Ah, yeah, I guess there are two competing conventions.  In my code I usually put the constructor first no matter what, so it's easy to find.  But I guess it's not fully consistent.

libkbfs/rekey_fsm.go
songgao: > that the context is not supposed to be modified by the FSM

Did you mean by RekeyFSM, or by any of the states? The RekeyFSM doesn't have a reference to it, so I'm assuming the latter. I'll see if I can get a workaround in test so that we don't have to replace ctx.

libdokan/rekey_file.go
strib: I think this needs to be synchronous, and it should return after the first rekey attempt either succeeds or fails.

libkbfs/rekey_fsm.go
songgao: ah you are right; Thanks!

libdokan/rekey_file.go
songgao: > this needs to be synchronous

Did you mean just the Rekey file or somewhere else as well?

Also, I suppose we are adding back `KBFSOps().Rekey()`, which is a "synchronous version" that calls "RequestRekey" and wait for first finish event?

libkbfs/folder_branch_ops.go
strib: Even so, the FSM uses a 0 delay in the error case, which is going to be annoying to someone who purposely ignored the popup.  (Before it retried after the 10 minute wait.)

I guess I can see the logic in retrying no matter what, I'm just worried that's going to make the logs harder to debug and put unnecessary strain on other parts of the system.  I still lean towards explicitly checking codes, but if you prefer this way, I'm ok with trying it.  But I do think specifically in the timeout case, we shouldn't retry immediately.

libkbfs/rekey_fsm.go
strib: Ah true.  The states shouldn't be able to modify it either, in my opinion.

libkbfs/folder_branch_ops.go
songgao: Sounds good! Thanks for the example. I'll add the check for timeout then. Does that cover the case where user purposely ignores the popup too?

libdokan/rekey_file.go
strib: The `Rekey` file and all tests want the same thing, so there only needs to be one non-test-specific way of doing that.

libkbfs/folder_branch_ops.go
strib: Yes, that's exactly the case I'm talking about.

songgao:
Thanks for the review!

libkbfs/rekey_queue.go
songgao: I think I intended to do that because we had `Clear()`. Now that it's removed, we can return here. Thanks!

libkbfs/rekey_fsm.go
songgao: Hmm I don't see an easy way to not store the context in the states. Do you have any ideas?

Arguably state structs are a way of breaking up function calls, so rekeyTask is like function arguments :)

songgao:
@strib I think I've implemented all feedback from your first pass except for storing the context. Please take another look when you get a chance; thanks!

libkbfs/rekey_fsm.go
strib: As I mentioned above, my best idea is to turn it into a function instead, like `getCtx() context.Context`, so at least it's not mutable.  But if stallers somehow make that impossible, oh well.

songgao: By not mutable, do you mean the states should not *replace* it? Yeah I think the staller means we need to replace the context with the one in the new rekey request for stalling to be effective

strib: Yeah that's what I mean.  Ok.  Please put a comment here explaining why it's ok for us to store a context in this case.

songgao: When taking a shower this morning I thought of a way to isolate this better; will push the change soon!

songgao: pushed; Thanks!

libdokan/rekey_file.go
strib: Can you get rid of the TODO now?

libkbfs/key_manager_test.go
```diff
@@ -2141,6 +2160,26 @@ func testKeyManagerRekeyAddDeviceWithPromptViaFolderAccess(t *testing.T, ver Met
...
 
+// maybeReplaceContext, defined on *protectedContext, enables replacing context
+// stored in protectedContext.
+//
+// This is importantly defined in a _test.go file since we should not use it
+// anywhere other than in tests.
+//
+// For now, the only application for this is to cause states in rekey_fsm
+// to replace existing context when new request comes in. This is needed
+// because stallers stores staller key in context, and context needs to be
```
strib: "stores" -> "store"

libkbfs/protected_context.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/keybase/client/go/logger"
+	"golang.org/x/net/context"
+)
+
+type defaultContextReplacer struct{}
+
+func (defaultContextReplacer) maybeReplaceContext(context.Context) {}
+
+// protectedContexte is a construct that helps avoid unwanted change of context
```
strib: "protectedContexte" -> "protectedContext"

libkbfs/protected_context.go
```diff
@@ -0,0 +1,46 @@
...
+// when the context needs to be stored.
+type protectedContext struct {
+	mu  sync.Mutex
```
strib: Maybe `RWMutex`, since `context()` will be called way more than the other methods and one caller doesn't need to block the other.

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,526 @@
...
+		r.fsm.log.CDebugf(r.task.ctx.context(),
+			"Rekey finished, ttl: %d -> %d", r.task.ttl, ttl)
+
+		if ttl <= 0 {
+			r.fsm.log.CDebugf(r.task.ctx.context(),
+				"Not scheduling new rekey because TTL expired")
+			return newRekeyStateIdle(r.fsm)
+		}
+
+		switch event.finished.err {
```
strib: I think you mentioned you'd be adding a explicit check for `context.DeadlineExceeded` (and maybe add a 10 min delay instead of 0), but this still seems to retry no matter what the error.

libkbfs/rekey_fsm.go
```diff
@@ -0,0 +1,526 @@
...
+			}
+		}
+	}()
+	for _, cb := range cbs {
+		cb.onEvent(e)
+	}
+}
+
+// listenOnEventForTest implements RekeyFSM interface for rekeyFSM.
+func (m *rekeyFSM) listenOnEventForTest(
```
strib: I think you can get rid of "ForTest" everywhere.

libdokan/rekey_file.go
songgao: oops yes thanks!

libkbfs/rekey_fsm.go
songgao: This was added back to fbo [here](https://github.com/keybase/kbfs/pull/805/commits/ca9507e9ff792b7e17514ccd1c0220edadcacb2b). It doesn't return the error in case of `DeadlineExceeded` any more. Although thinking about it more, I realized we should also add a delay for other errors (as mentioned in your "maybe" clause); will do!

strib: Ah thanks, I missed that.

songgao:
@strib Thanks for the effort you put in in reviewing this PR!

PR #807, opened 2017-02-28
don't add TLF to favorite in GetTLFCryptKeys

songgao:
It seems using a context key for this makes code easier to understand and maintain, than passing a parameter around.

### DIFF
--- a/libkbfs/keybase_service_base.go
+++ b/libkbfs/keybase_service_base.go
@@ -649,8 +649,6 @@ func (k *KeybaseServiceBase) GetTLFCryptKeys(ctx context.Context,
 		return keybase1.GetTLFCryptKeysRes{}, err
 	}
 
-	ctx = context.WithValue(ctx, CtxNoAddToFavoriteKey{}, true)
-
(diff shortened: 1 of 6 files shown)


libkbfs/keybase_service_base.go
```diff
@@ -649,6 +649,8 @@ func (k *KeybaseServiceBase) GetTLFCryptKeys(ctx context.Context,
 		return keybase1.GetTLFCryptKeysRes{}, err
 	}
 
+	ctx = context.WithValue(ctx, CtxNoAddToFavoriteKey{}, true)
```
strib: `true` seems weird, since setting it to `false` doesn't do the opposite.  Maybe just set it to `struct{}{}`?

songgao:
Hmm I think you convinced me using context key here is a bad idea -- It can be easily ignored. It's not really a lot of plumbing, but I did want to have a more formalized/unified way to specify whether an operation should be added to favorites.

songgao:
@strib I added a `FavoritesOp` for this. PTAL Thanks!

songgao:
@rjkmurray It's a ticket ID in our internal issue tracker

PR #809, opened 2017-02-28
conflict_resolver: don't unref the original half of a resOp update

strib:
If we unref the original, then a subsequent CR over the resulting `resolutionOp` will think the chain has been deleted, and the ops related to that chain won't get included in the resolution.

Issue: KBFS-1979


strib:
Unrelated CORE failure, merging.

maxtaco:
Which test is failing? We really should not have anybody flakes. They must
be new.

strib:
@maxtaco: this was the session expiration kbfs-server tests that have been failing for the last day,

keybase-travis:
See the #ci channel. We believe it is fixed.

strib:
Great.  The last run of this PR was hours ago though.  If master fails now I'll revisit.

maxtaco:
Super

PR #810, opened 2017-02-28
tlf_journal_test: fix race in TestTLFJournal/*WhileFlushing

strib:
In rare cases, `testTLFJournalPauseBlocksAndConvertBranch` can return
before the journal is converted to a branch.  In that case wait
explicitly for the branch change before moving onto the next part of
the test.

cc: @songgao 

Issue: KBFS-1985

### DIFF
--- a/libkbfs/tlf_journal_test.go
+++ b/libkbfs/tlf_journal_test.go
@@ -1088,13 +1088,24 @@ func testTLFJournalFlushInterleaving(t *testing.T, ver MetadataVer) {
 	}
 	require.NotZero(t, md1Slot)
 	require.NotZero(t, md2Slot)
-
 }
+
+type testBranchChangeListener struct {
+	c chan<- struct{}
+}
+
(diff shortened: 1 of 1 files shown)


libkbfs/tlf_journal_test.go
```diff
@@ -1148,6 +1159,13 @@ func testTLFJournalPauseBlocksAndConvertBranch(t *testing.T,
 		markers++
 	}
 
+	// Wait for the local squash branch to appear.
+	select {
+	case <-branchCh:
+	case <-ctx.Done():
```
songgao: I can't find a deadline or timeout on this ctx. If we rely on `-test.timeout`, maybe this isn't necessary?

strib: Look in `setupTLFJournalTest()`.

songgao: ah I see. Thanks!

strib:
circleci actually succeeded.

PR #811, opened 2017-02-28
folder_branch_ops: don't panic if no sync progress is made

strib:
This can happen naturally if the journal is really full and flushing
slowly, and we don't want KBFS to completely break in that case.
Instead, print a warning and keep trying, though we should still wait
for the ticker so we don't end up in a tight loop for too long.

(It's been a while since this panic found anything useful, and I really don't want KBFS to crash completely when the journal is full.)


strib:
Unrelated CORE failure, merging.

PR #812, opened 2017-02-28
config_local: increase the min head age for QR on other devices

strib:
This increases the time for another device to start QRing a TLF, after
one device writes to it.  The problem is that someone could write to a
journal, and then go offline for a while with items still unflushed.
If another device runs QR in the meantime, it will cause an
unnecessary conflict.

(Obviously CR still needs to work well in that case, in case the other
device does a _real_ write instead of just a QR write, but there's no
need to impose that extra burden if we don't have to.)

As long as the writing device is still online, it will be the one
doing QR on the TLF.

Issue: KBFS-1986


strib:
Failure is an unrelated CORE-related kbfs-server failure.  Merging.

PR #819, opened 2017-03-03
simplefs: Client and libkbfs differ in time resolution

taruti:


taruti:
Trivial one-liner, merging.

PR #832, opened 2017-03-08
libkbfs: Pass username+uid to identify instead of only uid

taruti:
Pass the username to the identify routines rather than using only uid and getting the username based on that.


strib:
@taruti: was this reviewed?

taruti:
@strib, Mis-clicked it in while merging another PR on wed. Discussed with akalin and reverted and made a new PR ( https://github.com/keybase/kbfs/pull/835 ).

PR #836, opened 2017-03-09
[KBFS-1920] Disk Cache Limiter

jzila:
This PR adds the ability for the journal's `diskLimiter` to also work with the `DiskBlockCacheStandard`. The limit function is as follows:

* min(20GB, 0.1 * free space on drive)

In practice the disk cache will only hit 20GB when the user has >=200GB of available space. We can ship future builds where this will be a tunable parameter.

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -190,9 +190,17 @@ func (bt *backpressureTracker) onBlocksDelete(blockResources int64) {
 	bt.updateSemaphoreMax()
 }
 
-func (bt *backpressureTracker) beforeDiskBlockCachePut(ctx context.Context,
-	blockResources int64) (availableResources int64, err error) {
-	return bt.semaphore.ForceAcquire(blockResources), nil
+func (bt *backpressureTracker) beforeDiskBlockCachePut(blockResources int64) (
+	availableResources int64) {
+	defer func() {
+		if availableResources < 0 {
+			// We must roll back the acquisition of resources. We should still
+			// return the negative number, however, so the disk block cache
+			// knows how much to evict.
+			bt.afterBlockPut(blockResources, false)
+		}
+	}()
+	return bt.semaphore.ForceAcquire(blockResources)
 }
 
 type backpressureTrackerStatus struct {
@@ -243,9 +251,9 @@ type backpressureDiskLimiter struct {
 	// lock protects everything in the trackers, including the
 	// (implicit) maximum values of the semaphores, but not the
 	// actual semaphore itself.
-	lock                     sync.RWMutex
-	byteTracker, fileTracker *backpressureTracker
-	diskCacheByteTracker     *backpressureTracker
+	lock                                   sync.RWMutex
+	journalByteTracker, journalFileTracker *backpressureTracker
+	diskCacheByteTracker                   *backpressureTracker
 }
 
 var _ DiskLimiter = (*backpressureDiskLimiter)(nil)
@@ -348,12 +356,12 @@ func (bdl *backpressureDiskLimiter) getSnapshotsForTest() (
 	byteSnapshot, fileSnapshot bdlSnapshot) {
 	bdl.lock.RLock()
 	defer bdl.lock.RUnlock()
-	return bdlSnapshot{bdl.byteTracker.used, bdl.byteTracker.free,
-			bdl.byteTracker.semaphoreMax,
-			bdl.byteTracker.semaphore.Count()},
-		bdlSnapshot{bdl.fileTracker.used, bdl.fileTracker.free,
-			bdl.fileTracker.semaphoreMax,
-			bdl.fileTracker.semaphore.Count()}
+	return bdlSnapshot{bdl.journalByteTracker.used, bdl.journalByteTracker.free,
+			bdl.journalByteTracker.semaphoreMax,
+			bdl.journalByteTracker.semaphore.Count()},
+		bdlSnapshot{bdl.journalFileTracker.used, bdl.journalFileTracker.free,
+			bdl.journalFileTracker.semaphoreMax,
+			bdl.journalFileTracker.semaphore.Count()}
 }
 
 func (bdl *backpressureDiskLimiter) onJournalEnable(
@@ -361,8 +369,8 @@ func (bdl *backpressureDiskLimiter) onJournalEnable(
 	availableBytes, availableFiles int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	availableBytes = bdl.byteTracker.onJournalEnable(journalBytes)
-	availableFiles = bdl.fileTracker.onJournalEnable(journalFiles)
+	availableBytes = bdl.journalByteTracker.onJournalEnable(journalBytes)
+	availableFiles = bdl.journalFileTracker.onJournalEnable(journalFiles)
 	return availableBytes, availableFiles
 }
 
@@ -370,14 +378,14 @@ func (bdl *backpressureDiskLimiter) onJournalDisable(
 	ctx context.Context, journalBytes, journalFiles int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	bdl.byteTracker.onJournalDisable(journalBytes)
-	bdl.fileTracker.onJournalDisable(journalFiles)
+	bdl.journalByteTracker.onJournalDisable(journalBytes)
+	bdl.journalFileTracker.onJournalDisable(journalFiles)
 }
 
 func (bdl *backpressureDiskLimiter) getDelayLocked(
 	ctx context.Context, now time.Time) time.Duration {
-	byteDelayScale := bdl.byteTracker.delayScale()
-	fileDelayScale := bdl.fileTracker.delayScale()
+	byteDelayScale := bdl.journalByteTracker.delayScale()
+	fileDelayScale := bdl.journalFileTracker.delayScale()
 	delayScale := math.Max(byteDelayScale, fileDelayScale)
 
 	// Set maxDelay to min(bdl.maxDelay, time until deadline - 1s).
@@ -398,14 +406,14 @@ func (bdl *backpressureDiskLimiter) beforeBlockPut(
 	availableBytes, availableFiles int64, err error) {
 	if blockBytes == 0 {
 		// Better to return an error than to panic in Acquire.
-		return bdl.byteTracker.semaphore.Count(),
-			bdl.fileTracker.semaphore.Count(), errors.New(
+		return bdl.journalByteTracker.semaphore.Count(),
+			bdl.journalFileTracker.semaphore.Count(), errors.New(
 				"backpressureDiskLimiter.beforeBlockPut called with 0 blockBytes")
 	}
 	if blockFiles == 0 {
 		// Better to return an error than to panic in Acquire.
-		return bdl.byteTracker.semaphore.Count(),
-			bdl.fileTracker.semaphore.Count(), errors.New(
+		return bdl.journalByteTracker.semaphore.Count(),
+			bdl.journalFileTracker.semaphore.Count(), errors.New(
 				"backpressureDiskLimiter.beforeBlockPut called with 0 blockFiles")
 	}
 
@@ -421,8 +429,8 @@ func (bdl *backpressureDiskLimiter) beforeBlockPut(
 			return 0, err
 		}
 
-		bdl.byteTracker.updateFree(freeBytes)
-		bdl.fileTracker.updateFree(freeFiles)
+		bdl.journalByteTracker.updateFree(freeBytes)
+		bdl.journalFileTracker.updateFree(freeFiles)
 
 		delay := bdl.getDelayLocked(ctx, time.Now())
 		if delay > 0 {
@@ -430,37 +438,37 @@ func (bdl *backpressureDiskLimiter) beforeBlockPut(
 				"journalBytes=%d, freeBytes=%d, "+
 				"journalFiles=%d, freeFiles=%d)",
 				blockBytes, blockFiles, delay.Seconds(),
-				bdl.byteTracker.used, freeBytes,
-				bdl.fileTracker.used, freeFiles)
+				bdl.journalByteTracker.used, freeBytes,
+				bdl.journalFileTracker.used, freeFiles)
 		}
 
 		return delay, nil
 	}()
 	if err != nil {
-		return bdl.byteTracker.semaphore.Count(),
-			bdl.fileTracker.semaphore.Count(), err
+		return bdl.journalByteTracker.semaphore.Count(),
+			bdl.journalFileTracker.semaphore.Count(), err
 	}
 
 	// TODO: Update delay if any variables change (i.e., we
 	// suddenly free up a lot of space).
 	err = bdl.delayFn(ctx, delay)
 	if err != nil {
-		return bdl.byteTracker.semaphore.Count(),
-			bdl.fileTracker.semaphore.Count(), err
+		return bdl.journalByteTracker.semaphore.Count(),
+			bdl.journalFileTracker.semaphore.Count(), err
 	}
 
-	availableBytes, err = bdl.byteTracker.beforeBlockPut(ctx, blockBytes)
+	availableBytes, err = bdl.journalByteTracker.beforeBlockPut(ctx, blockBytes)
 	if err != nil {
-		return availableFiles, bdl.fileTracker.semaphore.Count(), err
+		return availableFiles, bdl.journalFileTracker.semaphore.Count(), err
 	}
 	defer func() {
 		if err != nil {
-			bdl.byteTracker.afterBlockPut(blockBytes, false)
-			availableBytes = bdl.byteTracker.semaphore.Count()
+			bdl.journalByteTracker.afterBlockPut(blockBytes, false)
+			availableBytes = bdl.journalByteTracker.semaphore.Count()
 		}
 	}()
 
-	availableFiles, err = bdl.fileTracker.beforeBlockPut(ctx, blockFiles)
+	availableFiles, err = bdl.journalFileTracker.beforeBlockPut(ctx, blockFiles)
 	return availableBytes, availableFiles, err
 }
 
@@ -468,16 +476,16 @@ func (bdl *backpressureDiskLimiter) afterBlockPut(
 	ctx context.Context, blockBytes, blockFiles int64, putData bool) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	bdl.byteTracker.afterBlockPut(blockBytes, putData)
-	bdl.fileTracker.afterBlockPut(blockFiles, putData)
+	bdl.journalByteTracker.afterBlockPut(blockBytes, putData)
+	bdl.journalFileTracker.afterBlockPut(blockFiles, putData)
 }
 
 func (bdl *backpressureDiskLimiter) onBlocksDelete(
 	ctx context.Context, blockBytes, blockFiles int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	bdl.byteTracker.onBlocksDelete(blockBytes)
-	bdl.fileTracker.onBlocksDelete(blockFiles)
+	bdl.journalByteTracker.onBlocksDelete(blockBytes)
+	bdl.journalFileTracker.onBlocksDelete(blockFiles)
 }
 
 func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
@@ -508,15 +516,7 @@ func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
 	bt := bdl.diskCacheByteTracker
 	bt.updateFree(freeBytes)
 
-	defer func() {
-		if err != nil || availableBytes < 0 {
-			// We must roll back the acquisition of resources. We should still
-			// return the negative number, however, so the disk block cache
-			// knows how much to evict.
-			bt.afterBlockPut(blockBytes, false)
-		}
-	}()
-	return bt.beforeDiskBlockCachePut(ctx, blockBytes)
+	return bt.beforeDiskBlockCachePut(blockBytes), nil
 }
 
 func (bdl *backpressureDiskLimiter) afterDiskBlockCachePut(
@@ -547,7 +547,7 @@ func (bdl *backpressureDiskLimiter) getStatus() interface{} {
 
 		CurrentDelaySec: currentDelay.Seconds(),
 
-		ByteTrackerStatus: bdl.byteTracker.getStatus(),
-		FileTrackerStatus: bdl.fileTracker.getStatus(),
+		ByteTrackerStatus: bdl.journalByteTracker.getStatus(),
+		FileTrackerStatus: bdl.journalFileTracker.getStatus(),
 	}
 }
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -15,6 +15,10 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	diskBlockCacheFrac float64 = 0.5
+)
+
 // backpressureTracker keeps track of the variables used to calculate
 // backpressure. It keeps track of a generic resource (which can be
 // either bytes or files).
@@ -188,6 +192,11 @@ func (bt *backpressureTracker) onBlocksDelete(blockResources int64) {
 
 	bt.used -= blockResources
 	bt.updateSemaphoreMax()
+}
+
+func (bt *backpressureTracker) beforeDiskBlockCachePut(ctx context.Context,
+	blockResources int64) (availableResources int64, err error) {
+	return bt.semaphore.ForceAcquire(blockResources), nil
 }
 
 type backpressureTrackerStatus struct {
@@ -480,7 +489,8 @@ func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
 }
 
 func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
-	ctx context.Context, blockBytes int64) (availableBytes int64, err error) {
+	ctx context.Context, blockBytes, diskBlockCacheBytes int64) (
+	availableBytes int64, err error) {
 	if blockBytes == 0 {
 		// Better to return an error than to panic in Acquire.
 		return bdl.byteTracker.semaphore.Count(),
@@ -489,8 +499,20 @@ func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
 	}
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	// TODO: finish
-	return 0, nil
+	diskBlockCacheLimit := int64(bdl.byteTracker.currLimit() *
+		bdl.byteTracker.minThreshold * diskBlockCacheFrac)
+	if diskBlockCacheLimit < diskBlockCacheBytes+blockBytes {
+		// The overall disk block cache limit has been exceeded.  Return a
+		// negative result so the block cache knows how much it must evict
+		// before retrying.
+		return diskBlockCacheLimit - (diskBlockCacheBytes + blockBytes),
+			errors.New("backpressureDiskLimiter needs more space")
+	}
+	// The disk block cache is using up less than diskBlockCacheLimit bytes, so
+	// we can add more, even if it makes backpressure worse. This can return a
+	// negative number with a nil error, in which case the disk block cache
+	// must Put the block. It can choose to evict anyway afterwards.
+	return bdl.byteTracker.beforeDiskBlockCachePut(ctx, blockBytes)
 }
 
 type backpressureDiskLimiterStatus struct {
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -15,10 +15,6 @@ import (
 	"golang.org/x/net/context"
 )
 
-const (
-	diskBlockCacheFrac float64 = 0.5
-)
-
 // backpressureTracker keeps track of the variables used to calculate
 // backpressure. It keeps track of a generic resource (which can be
 // either bytes or files).
@@ -491,11 +487,11 @@ func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
 	}
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	bdl.byteTracker.onBlocksDelete(blockBytes)
+	bdl.diskCacheByteTracker.onBlocksDelete(blockBytes)
 }
 
 func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
-	ctx context.Context, blockBytes, diskBlockCacheBytes int64) (
+	ctx context.Context, blockBytes int64) (
 	availableBytes int64, err error) {
 	if blockBytes == 0 {
 		// Better to return an error than to panic in ForceAcquire.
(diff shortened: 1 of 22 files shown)


jzila:
Need to factor the `DiskLimiter` into `Config`, because the disk cache can use it even if the journal isn't active. Also need to factor `JournalRoot` and `DiskCacheRoot` into one `StorageRoot`.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -185,6 +189,11 @@ func (bt *backpressureTracker) onBlocksDelete(blockResources int64) {
 	bt.updateSemaphoreMax()
 }
 
+func (bt *backpressureTracker) beforeDiskBlockCachePut(ctx context.Context,
```
akalin-keybase: you can just omit the unused `ctx` param and unused `err` return value

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
 	bdl.fileTracker.onBlocksDelete(blockFiles)
 }
 
+func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
+	ctx context.Context, blockBytes int64) {
```
akalin-keybase: no `ctx` needed here, either.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	bdl.lock.Lock()
+	defer bdl.lock.Unlock()
+	bdl.byteTracker.onBlocksDelete(blockBytes)
+}
+
+func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
+	ctx context.Context, blockBytes, diskBlockCacheBytes int64) (
+	availableBytes int64, err error) {
+	if blockBytes == 0 {
+		// Better to return an error than to panic in Acquire.
```
akalin-keybase: Acquire -> ForceAcquire

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+		return 0, errors.New("backpressureDiskLimiter.beforeDiskBlockCachePut" +
+			" called with 0 blockBytes")
+	}
+	bdl.lock.Lock()
+	defer bdl.lock.Unlock()
+	freeBytes, freeFiles, err := bdl.freeBytesAndFilesFn()
+	if err != nil {
+		return
```
akalin-keybase: `return 0, err` to be consistent

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	diskBlockCacheLimit := int64(bdl.byteTracker.currLimit() *
+		bdl.byteTracker.minThreshold * diskBlockCacheFrac)
+	if diskBlockCacheLimit < diskBlockCacheBytes+blockBytes {
+		// The overall disk block cache limit has been exceeded.  Return a
+		// negative result so the block cache knows how much it must evict
+		// before retrying.
+		return diskBlockCacheLimit - (diskBlockCacheBytes + blockBytes),
+			errors.New("backpressureDiskLimiter needs more space")
+	}
+	// The disk block cache is using up less than diskBlockCacheLimit bytes, so
```
akalin-keybase: Hmm...I think we can avoid the possibility of making backpressure worse by having the semaphore have a `TryAcquire` function. Add a TODO perhaps, since it's not urgent. I had thought that having the diskBlockCacheLimit be some fraction of currLimit() was enough to guarantee that the semaphore couldn't go negative, but I'd have to think more about that...

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
 	bdl.fileTracker.onBlocksDelete(blockFiles)
 }
 
+func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
```
akalin-keybase: perhaps `Evict` is a better verb to use

libkbfs/semaphore_disk_limiter.go
```diff
@@ -103,6 +103,23 @@ func (sdl semaphoreDiskLimiter) onBlocksDelete(
...
+	blockBytes int64) {
+	sdl.onBlocksDelete(ctx, blockBytes, 0)
+}
+
+func (sdl semaphoreDiskLimiter) beforeDiskBlockCachePut(ctx context.Context,
+	blockBytes, diskBlockCacheBytes int64) (bytesAcquired int64, err error) {
+	if blockBytes == 0 {
+		return 0, nil
+	}
+	availableBytes, err := sdl.byteSemaphore.Acquire(ctx, blockBytes)
```
akalin-keybase: this should be `ForceAcquire` to match the backpressure behavior, right?

libkbfs/semaphore_disk_limiter.go
```diff
@@ -103,6 +103,23 @@ func (sdl semaphoreDiskLimiter) onBlocksDelete(
 	}
 }
 
+func (sdl semaphoreDiskLimiter) onDiskBlockCacheDelete(ctx context.Context,
+	blockBytes int64) {
+	sdl.onBlocksDelete(ctx, blockBytes, 0)
+}
+
+func (sdl semaphoreDiskLimiter) beforeDiskBlockCachePut(ctx context.Context,
+	blockBytes, diskBlockCacheBytes int64) (bytesAcquired int64, err error) {
```
akalin-keybase: `bytesAcquired` -> `availableBytes`

libkbfs/backpressure_disk_limiter.go
jzila: Not quite, this gets called both when blocks are evicted, and when they're deleted for another reason, like when they get garbage collected.

libkbfs/backpressure_disk_limiter.go
jzila: IMO this is in keeping with the purpose of the disk limiter. Do you think the possibility of going negative can cause issues elsewhere?

libkbfs/semaphore_disk_limiter.go
jzila: Sure, although this doesn't actually get used anywhere.

libkbfs/backpressure_disk_limiter.go
jzila: `onBlocksDelete` takes a `ctx` though.

libkbfs/backpressure_disk_limiter.go
jzila: The `err` return value is used in `DiskBlockCache.Put`, and I think it makes sense to pass through the context just like all the other methods.

libkbfs/backpressure_disk_limiter.go
akalin-keybase: Going negative is currently an edge case (e.g. if a journal gets enabled while we're putting lots of blocks) so I'm just a bit troubled by introducing another way for it to happen.

What do you mean by this "keeping with the purpose of the disk limiter"?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	if blockBytes == 0 {
+		return
+	}
+	bdl.lock.Lock()
+	defer bdl.lock.Unlock()
+	bdl.byteTracker.onBlocksDelete(blockBytes)
+}
+
+func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
+	ctx context.Context, blockBytes, diskBlockCacheBytes int64) (
```
akalin-keybase: I think instead of passing in the total disk block cache size here, we need an `onDIskBlockCacheEnable` function. Otherwise, the disk limiter doesn't know about the bytes consumed by the disk cache except in this function.

That is, the bytes `backpressureTracker` should be keeping track of both the disk block cache and the journal bytes, i.e. the U in `U <= min(k(U+F), L)` in the comment for `backpressureTracker` would also count the disk block cache bytes.

Also, can you change the comment "Let U be the (approximate) resource usage of the journal..." to add "(and maybe the disk block cache)" near the top of the file?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	availableBytes int64, err error) {
```
akalin-keybase: I think we also need an `afterDiskBlockCachePut`, since otherwise if the cache put fails somehow, we'd 'leak' those bytes. And if it succeeds we want to do something similar to `afterBlockPut` (i.e., update `bt.used`).

libkbfs/backpressure_disk_limiter.go
jzila: Sorry I meant that the behavior here seems correct: as long as the disk cache is meeting its contract of <=25% of `currLimit()`, it should be allowed to `Put`, no matter what that does to backpressure.

libkbfs/backpressure_disk_limiter.go
akalin-keybase: Ah, I guess this means we have to keep around a `diskBlockCacheBytes` member variable in bdl, too, to do the calculations below.

libkbfs/disk_block_cache.go
```diff
@@ -411,8 +404,22 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
 		return err
 	}
 	if !hasKey {
-		for cache.currBytes+uint64(encodedLen) > cache.maxBytes {
-			cache.evictLocked(ctx, defaultNumBlocksToEvict)
+		for {
```
akalin-keybase: I think you want to limit the iteration count here, to avoid infinite loops if somehow there's a bug in the limiter, or when we don't want to put the block in the cache, even if we evict everything from the cache first. (see comment above for example)

libkbfs/backpressure_disk_limiter.go
```diff
@@ -470,6 +479,49 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+		// negative result so the block cache knows how much it must evict
+		// before retrying.
+		return diskBlockCacheLimit - (diskBlockCacheBytes + blockBytes),
+			errors.New("backpressureDiskLimiter needs more space")
+	}
+	// The disk block cache is using up less than diskBlockCacheLimit bytes, so
+	// we can add more, even if it makes backpressure worse. This can return a
+	// negative number with a nil error, in which case the disk block cache
+	// must Put the block. It can choose to evict anyway afterwards.
+	return bdl.byteTracker.beforeDiskBlockCachePut(ctx, blockBytes)
```
akalin-keybase: Okay, I think I've convinced myself that we want to avoid going negative here, and instead we need a `TryAcquire` function (which I'm happy to do). Here's an example:

Say we have a ~2GB journal, a nearly empty cache (~0GB), and 6 GB of free space. Then since 2 + 0 <= 25% * (2 + 0 + 6), we're right at the limit, and the semaphore has count 0. However, the condition above still passes, since the limit is 1/16 * (2 + 0 + 6) = 0.5 GB. In this case, I think we want to tell the caller to not put the block in the cache at all, rather than go over the limit.

jzila: Right, I think my point was that we want to go over here. Otherwise the journal can starve the disk cache, which we don't want to allow.

libkbfs/backpressure_disk_limiter.go
jzila: We may need to rethink how the disk limiter is designed then. Right now it is tightly coupled into its uses, which isn't ideal. And we're talking about exacerbating that further.

libkbfs/backpressure_disk_limiter.go
jzila: I guess in this picture, the "true" limit is 25% + (25% of 25%).

libkbfs/backpressure_disk_limiter.go
jzila: In my ideal world the disk limiter would be an oracle that tells me how much storage Keybase is using, and how much is left available. Each component tracks how much storage _it_ is using, and could then make decisions about its future storage based on those 3 parameters.

akalin-keybase: Okay, let's discuss this on Monday, then...also, https://github.com/keybase/kbfs/pull/837 might help things somewhat, since it imposes a limit that is more stringent than anything we have now, e.g. if the quota is 10 GB, then the journal will most likely not go very much past that, and we can let the disk cache eat up everything else.

libkbfs/backpressure_disk_limiter.go
akalin-keybase: It seems there's a tension between starving the disk cache, and making journal backpressure worse, or hitting the disk limits. Not sure what the right answer is...will think about it.

jzila: Consider that the common use case for this scenario is blocks being flushed from the journal, thus being `Put` into the disk cache. It should be a zero-sum operation (within reason).

libkbfs/backpressure_disk_limiter.go
jzila: Coding against a moving target

moving target

jzila:
@akalin-keybase I think this is ready for another review.

libkbfs/disk_block_cache.go
```diff
@@ -411,29 +406,59 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
...
+				cache.log.CWarningf(ctx, "Error obtaining space for the disk"+
+					" block cache: %+v", err)
+				return err
+			}
+			if bytesAvailable >= 0 {
+				break
+			}
+			_, _, err = cache.evictLocked(ctx, defaultNumBlocksToEvict)
+			if err != nil {
+				return err
```
akalin-keybase: Need an `afterDiskBlockCachePut` call here too. Probably worth decomping out this whole block at this point, so you can use a `defer`.

libkbfs/backpressure_disk_limiter.go
akalin-keybase: Yeah, but the `err` return is only used by comparing to `nil`, so if you make this not return an error, you can omit that case in the caller below.

`ctx` is passed through all methods of `backpressureDiskLimiter` (for logging), but not `backpressureTracker` (only in `beforeBlockPut`, where it's needed.)

libkbfs/backpressure_disk_limiter.go
```diff
@@ -240,39 +245,46 @@ type backpressureDiskLimiter struct {
...
+	// that the journal should consume.
+	journalByteLimit := int64((float64(byteLimit) * journalFrac) + 0.5)
 	byteTracker, err := newBackpressureTracker(
 		backpressureMinThreshold, backpressureMaxThreshold,
-		limitFrac, byteLimit, freeBytes)
+		journalFrac, journalByteLimit, freeBytes)
 	if err != nil {
 		return nil, err
 	}
+	journalFileLimit := int64((float64(fileLimit) * journalFrac) + 0.5)
```
akalin-keybase: I don't think it makes sense to apply `journalFrac` to the file limit, since currently only the journal is using it. Maybe rename `journalFrac` and `diskCacheFrac` to `journalByteFrac`, etc. and apply those only to the byte trackers.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -469,6 +480,52 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	freeBytes, _, err := bdl.freeBytesAndFilesFn()
+	if err != nil {
+		return 0, err
+	}
+
+	bt := bdl.diskCacheByteTracker
+	bt.updateFree(freeBytes)
+
+	defer func() {
+		if err != nil || availableBytes < 0 {
```
akalin-keybase: if you make `bt.beforeDiskBlockCachePut` not return an error, then this can just be `if availableBytes < 0`. In fact, this logic probably belongs in `bt.beforeDiskBlockCachePut`, anyway, as ideally it would do `TryAcquire`, and we're just faking it now with `ForceAcquire` + rollback.

libkbfs/config_local.go
```diff
@@ -935,26 +978,16 @@ func (c *ConfigLocal) EnableJournaling(
...
-	const journalByteLimit int64 = 50 * 1024 * 1024 * 1024
-	// Set the absolute journal file limit to 1.5 million for now.
-	const journalFileLimit int64 = 1500000
-	bdl, err := newBackpressureDiskLimiter(
-		log, backpressureMinThreshold, backpressureMaxThreshold,
-		journalByteLimitFrac, journalByteLimit, journalFileLimit,
-		defaultDiskLimitMaxDelay, journalRoot)
-	if err != nil {
-		return err
+	if limiter == nil {
```
akalin-keybase: Can you instead make `Config` expose `c.GetDiskLimiter()` and use that here instead of passing it in? Then you'd only have to create the limiter once when constructing the config.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -469,6 +480,52 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	}
+	bdl.lock.Lock()
+	defer bdl.lock.Unlock()
+	freeBytes, _, err := bdl.freeBytesAndFilesFn()
+	if err != nil {
+		return 0, err
+	}
+
+	bt := bdl.diskCacheByteTracker
+	bt.updateFree(freeBytes)
```
akalin-keybase: So all byte trackers need to keep track of the space used by both the journal and the cache. Otherwise, if each tracker independently thinks it's the only one using the free space, then they'll affect each other implicitly.

Example: Say you have a 100 GB limit and 25% limits for both the journal and cache, both which start off empty. Both trackers see the usage fraction as 0/(0 + 100), so let's say both fill up fairly quickly to 25GB each. But then, on the next journal block put, the journal sees 50 GB free and knows it's using 25 GB, so it calculates its usage fraction as 25/(25 + 50) = 33%, which is over the limit, and thus starts applying backpressure, which is undesirable, since by having separate limits we wanted to avoid having to deal with caching affecting journaling and vice versa.

The cleanest solution I can think of is to treat 'used' journal bytes as 'free' in the cache byte tracker, and to treat 'used' cache bytes as 'free' in the journal byte tracker. For example, here, it would look like:

```
bt.updateFree(freeBytes + bdl.journalByteTracker.used)
```

I think it suffices to update the free bytes only when we'd normally do it, and not every time the `used` field changes.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -469,6 +480,52 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
...
+	ctx context.Context, blockBytes int64) {
+	if blockBytes == 0 {
+		return
+	}
+	bdl.lock.Lock()
+	defer bdl.lock.Unlock()
+	bdl.diskCacheByteTracker.onBlocksDelete(blockBytes)
+}
+
+func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
```
akalin-keybase: I don't see a new `onDiskBlockCacheEnable(cacheBytes int64)` function. I think that's still necessary, as otherwise every time we restart we'd think the cache is empty.

libkbfs/backpressure_disk_limiter.go
jzila: I agree regarding the context, but I'm going to leave the error because it allows us to explicitly check the preconditions (e.g. `blockBytes != 0`).

libkbfs/disk_block_cache.go
jzila: It is called in the limiter as a defer, but you're right, we shouldn't be splitting up that behavior.

libkbfs/backpressure_disk_limiter.go
jzila: I think that's fine, because the first time we try to `beforeDiskBlockCachePut` or `onDelete` we'll call `updateFree`.

libkbfs/backpressure_disk_limiter.go
jzila: All right so I'll move it into the tracker, but then we still have the issue that if `availableBytes > 0`, we'll need to call `afterDiskBlockCachePut`, but otherwise we won't. So we call `afterDiskBlockCachePut` if and only if we have tried to `Put`. Does that sound right?

libkbfs/backpressure_disk_limiter.go
jzila: Ugh, you're right. I'm not a fan of that solution though. I'd really like to keep the trackers concerns separate, preferably by abstracting out overall usage to an overall tracker.

libkbfs/disk_block_cache.go
jzila: Actually per your feedback I've moved the functionality to the tracker. Thus, this is actually correct.

libkbfs/backpressure_disk_limiter.go
jzila: This creates an inconsistency in how the files vs bytes are handled. Are you sure that's what we want to do?

libkbfs/backpressure_disk_limiter.go
jzila: OK you're right; now that `updateFree` is calibrated to both caches' usage, we have to consider them together.

libkbfs/backpressure_disk_limiter.go
akalin-keybase: I don't think it's an inconsistency per se -- in general, we'd have a `journalByteFrac` and `diskCacheByteFrac` that add up to 1, and `journalFileFrac` and `diskCacheFileFrac` that add up to 1. But since the disk cache uses up no files we can just set `journalFileFrac` to 1. If we ever end up doing individual files for the disk cache, then we'd have to revisit.

So we can be explicit about it by passing in `*FileFrac`, but I don't know if that buys us anything at this point.

jzila: Actually the current `journalByteFrac` and `diskCacheByteFrac` don't add up to 1: they add up to whatever total percentage of the disk they are allowed to take up.

akalin-keybase: ah, sorry, right.

jzila:
All right @akalin-keybase: I think I've implemented all your feedback except for the file limit fraction. I can also do that if you strongly prefer it.

[Here](https://github.com/keybase/kbfs/pull/836/files/a97f2f786ac902561bcdbca4295ce511ed18a05d..b7b2465fb176f26ece1270244b3cfabc991e2017) are the commits from the final round.

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -487,13 +487,13 @@ func testBackpressureDiskLimiterSmallDiskDelay(
...
-		return diskBytes - bdl.byteTracker.used,
-			diskFiles - bdl.fileTracker.used, nil
+		return diskBytes - bdl.journalByteTracker.used,
+			diskFiles - bdl.journalFileTracker.used, nil
 	}
 
 	log := logger.NewTestLogger(t)
 	bdl, err := newBackpressureDiskLimiterWithFunctions(
-		log, 0.1, 0.9, 0.25, math.MaxInt64, math.MaxInt64,
+		log, 0.1, 0.9, 0.25, 0.1, math.MaxInt64, math.MaxInt64,
```
akalin-keybase: Doesn't have to be as comprehensive as the existing tests, but it would be nice to have at least one test for the interaction between the cache and journal limits.

Could be something like the scenario I described in the previous review:

1) Set up journal and cache with 25% each of 100 GB.
2) Fill them both up.
3) Test that the journal isn't in backpressure, and that the cache doesn't need to evict.

libkbfs/config_local.go
```diff
@@ -935,26 +978,16 @@ func (c *ConfigLocal) EnableJournaling(
...
-	const journalByteLimit int64 = 50 * 1024 * 1024 * 1024
-	// Set the absolute journal file limit to 1.5 million for now.
-	const journalFileLimit int64 = 1500000
-	bdl, err := newBackpressureDiskLimiter(
-		log, backpressureMinThreshold, backpressureMaxThreshold,
-		journalByteLimitFrac, journalByteLimit, journalFileLimit,
-		defaultDiskLimitMaxDelay, journalRoot)
-	if err != nil {
-		return err
+	if c.DiskLimiter() == nil {
```
akalin-keybase: Can we just do `MakeDiskLimiter` in whatever initializes `ConfigLocal`? Then we wouldn't have to do this. Note that this function can technically be called from multiple goroutines, anyway, so this would be racy.

libkbfs/disk_block_cache.go
akalin-keybase: I don't understand -- it sounds like you're referring to the error returned by the `beforeDiskBlockCachePut()` call above. But if that succeeds, and the call to `evictLocked` fails, you need an `afterDiskBlockCachePut` call, right?

libkbfs/disk_block_cache.go
```diff
@@ -411,29 +400,66 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
 		return err
 	}
 	if !hasKey {
-		for cache.currBytes+uint64(encodedLen) > cache.maxBytes {
-			cache.evictLocked(ctx, defaultNumBlocksToEvict)
+		for {
+			select {
```
akalin-keybase: Sorry, I forgot to mention this last review, but I'm a bit nervous that we can get into a state (e.g. due to a bug) where the limiter tells us to evict, but we can't evict any more, and so we'd burn the CPU until the context times out. Perhaps check whether `evictLocked` didn't actually do anything, and treat that case as an error?

libkbfs/config_local.go
akalin-keybase: Ah, looks like we already do that. So I don't think this block is needed at all.

libkbfs/backpressure_disk_limiter.go
jzila: I've quadrupled it to compensate.

libkbfs/disk_block_cache.go
jzila: `bytesAvailable < 0` will have reverted already, thus requiring the next iteration of the loop.

akalin-keybase: D'oh, got it!

libkbfs/backpressure_disk_limiter.go
```diff
@@ -337,36 +355,54 @@ func (bdl *backpressureDiskLimiter) getSnapshotsForTest() (
...
 func (bdl *backpressureDiskLimiter) onJournalEnable(
 	ctx context.Context, journalBytes, journalFiles int64) (
 	availableBytes, availableFiles int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	availableBytes = bdl.byteTracker.onJournalEnable(journalBytes)
-	availableFiles = bdl.fileTracker.onJournalEnable(journalFiles)
+	availableBytes = bdl.journalByteTracker.onEnable(journalBytes)
+	availableFiles = bdl.journalFileTracker.onEnable(journalFiles)
+	bdl.updateFreeLocked()
```
jzila: this `updateFreeLocked` isn't strictly necessary; the next time either the journal or disk cache do a respective `Put`, they will call it again, correctly, and handle errors correctly.

akalin-keybase: ah, yeah, you're right. Yeah, that sounds fine to me!

akalin-keybase: those are the two points where the numbers matter, anyway.

akalin-keybase: What I mean is that you can just omit those calls, then?

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -433,6 +433,144 @@ func testBackpressureDiskLimiterLargeDiskDelay(
 	}, fileSnapshot)
 }
 
+// testBackpressureDiskLimiterLargeDiskDelay checks the delays when
```
akalin-keybase: comment has wrong name

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -433,6 +433,144 @@ func testBackpressureDiskLimiterLargeDiskDelay(
...
+// pretending to have a large disk.
+func TestBackpressureDiskLimiterJournalAndDiskCache(
+	t *testing.T) {
```
akalin-keybase: can't fit on prev line?

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -433,6 +433,144 @@ func testBackpressureDiskLimiterLargeDiskDelay(
...
+		lastDelay = delay
+		return nil
+	}
+
+	const blockBytes int64 = 100
+	// Big number, but no risk of overflow
+	maxFreeBytes := int64(1 << 30)
+
+	// Set the bottleneck; i.e. set parameters so that semaphoreMax for the
+	// bottleneck always has value 10 * blockX when called in beforeBlockPut,
```
akalin-keybase: blockX -> blockBytes

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -433,6 +433,144 @@ func testBackpressureDiskLimiterLargeDiskDelay(
...
+	// and every block put beyond the min threshold leads to an increase in
+	// timeout of 1 second up to the max.
+	byteLimit := 10 * blockBytes
+	// arbitrarily large number
+	fileLimit := int64(100000000)
```
akalin-keybase: maybe `math.MaxInt64` would be a better constant to use if we don't care about it (and remove the *4 below)

libkbfs/backpressure_disk_limiter_test.go
```diff
@@ -433,6 +433,144 @@ func testBackpressureDiskLimiterLargeDiskDelay(
...
+
+	log := logger.NewTestLogger(t)
+	bdl, err := newBackpressureDiskLimiterWithFunctions(
+		log, 0.1, 0.9, 0.25, 0.25, byteLimit*4, fileLimit*4,
```
akalin-keybase: why the `*4` here?

libkbfs/disk_block_cache_test.go
```diff
@@ -53,15 +53,16 @@ func newDiskBlockCacheStandardForTest(config *testDiskBlockCacheConfig,
...
-	// Sooooo hacky :(
+	if limiter == nil {
+		limiter, err = newBackpressureDiskLimiterWithFunctions(
+			config.MakeLogger(""), 0.5, 0.95, 0.25, 0.25,
+			testDiskBlockCacheMaxBytes, maxFiles, time.Second,
+			defaultDoDelay, func() (int64, int64, error) {
+				// hackity hackeroni
+				freeBytes := maxBytes - int64(cache.currBytes)
+				return freeBytes, maxFiles, nil
+			})
```
akalin-keybase: `require.NoError(t, err)`?

libkbfs/disk_block_cache_test.go
```diff
@@ -53,15 +53,16 @@ func newDiskBlockCacheStandardForTest(config *testDiskBlockCacheConfig,
...
-			freeBytes := maxBytes - int64(cache.currBytes)
-			return freeBytes, maxFiles, nil
-		})
-	// Sooooo hacky :(
+	if limiter == nil {
+		limiter, err = newBackpressureDiskLimiterWithFunctions(
+			config.MakeLogger(""), 0.5, 0.95, 0.25, 0.25,
+			testDiskBlockCacheMaxBytes, maxFiles, time.Second,
+			defaultDoDelay, func() (int64, int64, error) {
+				// hackity hackeroni
```
akalin-keybase: Maybe a slightly more detailed comment? Something like: `Hack: simulate the disk cache taking up space.`

libkbfs/backpressure_disk_limiter_test.go
jzila: Same reason it's in the test I adapted. `byteLimit` represents the byte limit for the journal cache (and disk cache), and so `byteLimit * 4` is the overall byte limit.

akalin-keybase: Ah, yes! Serves me right for not making that a constant 😅

PR #838, opened 2017-03-09
simplefs: Move out simplefs to separate package

taruti:
As discussed on Slack.

The instance is passed through InitParams because the connection to service is created before a separate hypothetical SetSimpleFSImplementation could be created. With the current way the simplefs implementation can be present when the connection is opened.


taruti:
Review from either @zanderz or @strib.

taruti:
API change for mobile in https://github.com/keybase/client/pull/6200

PR #840, opened 2017-03-10
bserver_remote: add pings and disconnect on timeouts

strib:
Also add timeout disconnects for mdserver as well, factoring out a bit of common code.

Since bserver actually has two connections, this required a bit of refactoring of bserver connection handling as well.

This helps the case where a (Linux?) device loses its network (e.g., due to a suspend).  Normally it would take a long time (15-20 mins) for the TCP connections to figure out that they need to reconnect.  This brings that down to ~40s.

One downside is that `connection.go` now spams the log every 2 seconds while the device is offline, and before it wouldn't do that until the TCP connection noticed the disconnect.  We might want to suppress some of that logging in a future PR.

Issue: KBFS-1982

### DIFF
--- a/libkbfs/bserver_remote.go
+++ b/libkbfs/bserver_remote.go
@@ -95,8 +95,9 @@ func (b *blockServerRemoteClientHandler) initNewConnection() {
 	}
 
 	b.conn = rpc.NewTLSConnection(
-		b.srvAddr, kbfscrypto.GetRootCerts(b.srvAddr), MDServerErrorUnwrapper{},
-		b, b.rpcLogFactory, b.log, b.connOpts)
+		b.srvAddr, kbfscrypto.GetRootCerts(b.srvAddr),
+		kbfsblock.BServerErrorUnwrapper{}, b, b.rpcLogFactory, b.log,
+		b.connOpts)
 	b.client = keybase1.BlockClient{Cli: b.conn.GetClient()}
 }
 
--- a/libkbfs/bserver_remote.go
+++ b/libkbfs/bserver_remote.go
@@ -114,7 +114,7 @@ func (b *blockServerRemoteClientHandler) shutdown() {
 	}
 
 	// cancel the ping ticker
-	b.pinger.resetTicker(0)
+	b.pinger.cancelTicker()
 }
 
 func (b *blockServerRemoteClientHandler) getConn() *rpc.Connection {
@@ -197,7 +197,7 @@ func (b *blockServerRemoteClientHandler) OnConnectError(err error, wait time.Dur
 	if b.authToken != nil {
 		b.authToken.Shutdown()
 	}
-	b.pinger.resetTicker(0)
+	b.pinger.cancelTicker()
 	// TODO: it might make sense to show something to the user if this is
 	// due to authentication, for example.
 }
@@ -216,7 +216,7 @@ func (b *blockServerRemoteClientHandler) OnDisconnected(ctx context.Context,
 	if b.authToken != nil {
 		b.authToken.Shutdown()
 	}
-	b.pinger.resetTicker(0)
+	b.pinger.cancelTicker()
 }
 
 // ShouldRetry implements the ConnectionHandler interface.
--- a/libkbfs/bserver_remote.go
+++ b/libkbfs/bserver_remote.go
@@ -25,6 +25,12 @@ const (
 	BServerTokenServer = "kbfs_block"
 	// BServerTokenExpireIn is the TTL to use when constructing an authentication token.
 	BServerTokenExpireIn = 2 * 60 * 60 // 2 hours
+	// BServerDefaultPingIntervalSeconds is the default interval on which the
+	// client should contact the block server.
+	BServerDefaultPingIntervalSeconds = 10
+	// BServerPingTimeout is how long to wait for a ping response
+	// before breaking the connection and trying to reconnect.
+	BServerPingTimeout = 30 * time.Second
 )
 
 // blockServerRemoteAuthTokenRefresher is a helper struct for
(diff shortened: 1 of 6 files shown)


strib:
Rebased and ready for review again.

strib:
Rebased again, ready for review.

Should I say . . . _ping_?

libkbfs/connection_status.go
```diff
@@ -50,7 +50,13 @@ func (kcs *kbfsCurrentStatus) PushConnectionStatusChange(service string, err err
 	defer kcs.lock.Unlock()
 
 	if err != nil {
+		// Exit early if the service is already failed, to avoid an
+		// invalidation.
+		_, errExisted := kcs.failingServices[service]
```
jzila: To be clear, this is because the previous error would have already invalidated, correct? Have we seen any panics due to the potential double close below?

libkbfs/mdserver_remote.go
```diff
@@ -32,21 +32,29 @@ const (
...
 // MDServerRemote is an implementation of the MDServer interface.
 type MDServerRemote struct {
-	config       Config
-	conn         *rpc.Connection
-	client       keybase1.MetadataClient
-	log          logger.Logger
-	mdSrvAddr    string
-	authToken    *kbfscrypto.AuthToken
-	squelchRekey bool
+	config        Config
```
jzila: It might be good to refactor this to an `mdServerRemoteConfig`, that only includes the `Config` methods we need.

libkbfs/mdserver_remote.go
```diff
@@ -79,25 +87,31 @@ var _ rpc.ConnectionHandler = (*MDServerRemote)(nil)
...
+		observers:     make(map[tlf.ID]chan<- error),
+		log:           config.MakeLogger(""),
+		mdSrvAddr:     srvAddr,
+		rpcLogFactory: rpcLogFactory,
+		rekeyTimer:    time.NewTimer(MdServerBackgroundRekeyPeriod),
+	}
+
+	mdServer.pinger = pinger{
+		name:    "MDServerRemote",
+		doPing:  mdServer.pingOnce,
```
jzila: I'm not sure I love this design (bidirectional dependency), but I don't immediately see a cleaner way to do it.

libkbfs/mdserver_remote.go
```diff
@@ -643,6 +638,18 @@ func (md *MDServerRemote) FolderNeedsRekey(ctx context.Context,
 	return nil
 }
 
+func (md *MDServerRemote) getConn() *rpc.Connection {
+	md.connMu.RLock()
+	defer md.connMu.RUnlock()
+	return md.conn
+}
+
+func (md *MDServerRemote) getClient() keybase1.MetadataClient {
```
jzila: Please move this above its first use.

libkbfs/mdserver_remote.go
```diff
@@ -788,12 +795,15 @@ func (md *MDServerRemote) getFoldersForRekey(ctx context.Context,
...
+	md.connMu.Lock()
+	defer md.connMu.Unlock()
+
 	// close the connection
 	md.conn.Shutdown()
 	// cancel pending observers
 	md.cancelObservers()
 	// cancel the ping ticker
-	md.resetPingTicker(0)
+	md.pinger.resetTicker(0)
```
jzila: Now that we've abstracted out the ticker, maybe instead of special-casing a 0-valued `resetTicker`, we can have an `md.pinger.cancelTicker()`. IMO it makes more sense anyway for `pinger.resetTicker(0)` to still ping once, but then not repeat.

libkbfs/bserver_remote.go
```diff
@@ -141,11 +248,45 @@ func (b *blockServerRemoteClientHandler) ShouldRetryOnConnect(err error) bool {
 
 var _ rpc.ConnectionHandler = (*blockServerRemoteClientHandler)(nil)
 
+func (b *blockServerRemoteClientHandler) pingOnce(ctx context.Context) {
+	_, err := b.getClient().BlockPing(ctx)
```
jzila: What about the `putClient()`? It's conceivable that the `getClient` will stay refreshed but the `putClient` will time out.

libkbfs/bserver_remote.go
```diff
@@ -25,41 +26,147 @@ const (
 	BServerTokenServer = "kbfs_block"
 	// BServerTokenExpireIn is the TTL to use when constructing an authentication token.
 	BServerTokenExpireIn = 2 * 60 * 60 // 2 hours
+	// BServerDefaultPingIntervalSeconds is the default interval on which the
+	// client should contact the block server.
+	BServerDefaultPingIntervalSeconds = 10
+	// BServerPingTimeout is how long to wait for a ping response
+	// before breaking the connection and trying to reconnect.
+	BServerPingTimeout = 30 * time.Second
```
strib: The pings are launched from a single goroutine that blocks on each ping, so there should only ever be one outstanding as far as I can tell.  I was worried if I made it too small, I'd end up missing pings when uploading lots of data, or when the server is busy, and in that case re-initializing the connection would just make the problem worse.

Not sure what the right balance is.  Any thoughts?

libkbfs/bserver_remote.go
strib: Oh, this is somewhat confusing naming!  The `getClient()` here is actually an accessor function for the underlying rpc client.  `BServerRemote.getConn/putConn` is at a higher level.  So the pings do happen for both connections.

libkbfs/connection_status.go
strib: Yes, the previous error has already invalidated it, but there's no risk of double close because after the invalidation, it immediately makes a new channel.  The reason to avoid the second invalidation (with the new channel) is to avoid the log spam that comes with it (and all subsequent, periodic invalidations).

libkbfs/mdserver_remote.go
strib: That might be nice, but that's not a goal of this PR, so I'll skip it for now.

libkbfs/bserver_remote.go
jzila: Sounds reasonable.

PR #841, opened 2017-03-10
kbfsfuse: turn off journaling as part of the kbfsfuse command

strib:
So we don't open the tests up to any races.


strib:
@jzila: I thought you were seeing journaling-related flakes in kbfs-server CI.  This was trying to address that.  So it's as urgent as those flakes are, if they exist.

PR #845, opened 2017-03-13
more ctrl-c fixes

songgao:


libfuse/mounter.go
```diff
@@ -87,17 +87,27 @@ func (m ForceMounter) Unmount() (err error) {
...
+			} else {
+				_, err = exec.Command("fusermount", "-u", dir).Output()
+			}
 		}
-	case "linux":
-		if force {
-			_, err = exec.Command("fusermount", "-ul", dir).Output()
-		} else {
-			_, err = exec.Command("fusermount", "-u", dir).Output()
+		if err != nil {
```
strib: Do you really need a double switch here?  Why not just move this `if` block down below the `switch`?  Seems like it would work just fine in the default case.

libkbfs/init.go
```diff
@@ -354,11 +354,18 @@ func Init(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn, onI
...
 		_ = <-interruptChan
 
+		close(done)
+
 		if onInterruptFn != nil {
 			onInterruptFn()
+
+			// Keep listening on the signal channel in case unmount fails the first
+			// time and user continues pressing Ctrl-C.
+			for range interruptChan {
```
strib: Why isn't this an infinite loop?  I don't see anything that closes `interruptChan`.  Is it a requirement that `onInterruptFn` kills the whole process?  Could use some more comments here.

songgao: `interruptChan` is the channel that `os/signal` sends signals into, so we need to keep it open until program exits, i.e. not closing. I'll add comment in the code.

PR #850, opened 2017-03-14
Fix removing open file in sub dir

songgao:
@strib thanks for your help figuring it out!

resolves #711

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -3091,12 +3091,6 @@ func (fbo *folderBranchOps) unrefEntry(ctx context.Context,
 func (fbo *folderBranchOps) removeEntryLocked(ctx context.Context,
 	lState *lockState, md *RootMetadata, dir path, name string) error {
 	fbo.mdWriterLock.AssertLocked(lState)
-	fmt.Printf("SONGGAO: removeEntryLocked: md.data.Dir.BlockPointer.ID:  %s\n", md.data.Dir.BlockPointer.ID)
-	fmt.Printf("SONGGAO: removeEntryLocked: dir.path[0].BlockPointer.ID:  %s\n", dir.path[0].BlockPointer.ID)
-	defer func() {
-		fmt.Printf("SONGGAO: defer removeEntryLocked: md.data.Dir.BlockPointer.ID:  %s\n", md.data.Dir.BlockPointer.ID)
-		fmt.Printf("SONGGAO: defer removeEntryLocked: dir.path[0].BlockPointer.ID:  %s\n", dir.path[0].BlockPointer.ID)
-	}()
 
 	pblock, err := fbo.blocks.GetDir(
 		ctx, lState, md.ReadOnly(), dir, blockWrite)
@@ -3196,9 +3190,7 @@ func (fbo *folderBranchOps) RemoveDir(
 func (fbo *folderBranchOps) RemoveEntry(ctx context.Context, dir Node,
 	name string) (err error) {
 	fbo.log.CDebugf(ctx, "RemoveEntry %s %s", getNodeIDStr(dir), name)
-	fmt.Printf("SONGGAO: RemoveEntry dir: %#v\n", dir)
 	defer func() {
-		fmt.Printf("SONGGAO: defer RemoveEntry dir: %#v\n", dir)
 		fbo.deferLog.CDebugf(ctx, "RemoveEntry %s %s done: %+v",
 			getNodeIDStr(dir), name, err)
 	}()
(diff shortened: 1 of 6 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -3960,6 +3950,36 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 	afterUpdateFn func() error) error {
 	fbo.headLock.AssertLocked(lState)
 
+	// We need to get unlinkPath before calling UpdatePointers so that
+	// nodeCache.Unlink can properly update cachedPath.
+	unlinkPath, toUnlink := func() (path, bool) {
```
strib: Might as well put this in a named function, it might be a little cleaner.  Looks like it only needs to take `ctx` and `op`.

libkbfs/folder_branch_ops.go
```diff
@@ -3960,6 +3950,36 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
...
+		var (
```
strib: Is there a benefit to using the `()` construct?  Seems like it just uses two extra lines for no reason.

libkbfs/folder_branch_ops.go
```diff
@@ -3960,6 +3950,36 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
...
+			childName = realOp.NewName
+		}
+		if node == nil {
+			return path{}, false
+		}
+
+		p, err := fbo.pathFromNodeForRead(node)
+		if err != nil {
+			fbo.log.CErrorf(ctx,
+				"Calling pathFromNodeForRead error: %v", err)
```
strib: Seems like an `err` here should be reported back to the caller.

libkbfs/folder_branch_ops.go
```diff
@@ -3960,6 +3950,36 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
...
+			node      Node
+			childName string
+		)
+
+		switch realOp := op.(type) {
+		case *rmOp:
+			node = fbo.nodeCache.Get(realOp.Dir.Unref.Ref())
+			childName = realOp.OldName
+		case *renameOp:
+			node = fbo.nodeCache.Get(realOp.NewDir.Unref.Ref())
```
strib: When a file is renamed within the same directory, `realOp.NewDir` is set to the empty value.  In that case, you need to use `realOp.OldDir` here.  See the existing rename code you're (partially) removing.

libkbfs/folder_branch_ops.go
songgao: No benefit performance wise; I just thought it looked clearer to group them together. But I've no problem with not using it as well; will change

libkbfs/folder_branch_ops.go
songgao: Should it be returned all the way to the caller of notifyOnOpLocked? It seem `UpdatePointers` and `batchChanges` should still be called in this case

strib: The current code returns it all the way to the caller, yes.  I don't see any reason to change it.  It would be a pretty unexpected error.  (The only way to get the error is if the node has an invalid path.)

songgao: OK sounds good. I guess it wouldn't matter, but just to make sure, should `UpdatePointers` still be called in case of the error? The existing code would have called `UpdatePointers` in case of that error.

songgao:
@strib I implemented your feedback, including adding two tests for rename, one of which fails without the new commit. PTAL thanks!

libkbfs/folder_branch_ops.go
strib: Ah right, I see.  I think it's ok not to call it.

libkbfs/folder_branch_ops.go
```diff
@@ -3955,12 +3945,52 @@ func (fbo *folderBranchOps) unlinkFromCache(op op, oldDir BlockPointer,
...
 	afterUpdateFn func() error) error {
 	fbo.headLock.AssertLocked(lState)
 
-	err := fbo.blocks.UpdatePointers(
+	// We need to get unlinkPath before calling UpdatePointers so that
+	// nodeCache.Unlink can properly update cachedPath.
+	unlinkPath, toUnlink, err := fbo.getUnlinkPathBeforeUpdatingPointers(ctx, op)
+	if err != nil {
+		return err
+		fbo.log.CErrorf(ctx,
```
strib: Unreachable (and unneeded) code.

songgao: oops thanks!

PR #852, opened 2017-03-15
SimpleFSCheckOp, more testing

zanderz:
This satisfies KBFS-1945


simplefs/simplefs.go
```diff
@@ -519,8 +519,14 @@ func (k *SimpleFS) SimpleFSClose(ctx context.Context, opid keybase1.OpID) (err e
...
-	// TODO
-	return 0, simpleFSError{"Not implemented"}
+	k.lock.RLock()
+	defer k.lock.RUnlock()
+	if _, ok := k.inProgress[opid]; ok {
+		return 0, nil
+	} else if _, ok := k.handles[opid]; ok {
+		return 0, nil
+	}
+	return 0, errNoResult
```
songgao: What does the first returned variable do? Also perhaps this method could use some comment regarding to what it actually checks?

PR #862, opened 2017-03-20
init: Fix crash when mode isn't explicitly specified

jzila:


jzila:
Why not both?

strib:
I just feel better about making the caller specify it explicitly.  Seems like returning an error is better on an empty string.

jzila:
Anyway feel free to close this PR when you submit a fix as you wish.

strib:
Closed via b6f2053f.

PR #863, opened 2017-03-20
init: add default mode to DefaultInitParam

strib:


jzila:
@strib: I'm not sure this will work: in `AddFlags`, we already specify the default to be `InitDefaultString`, so I'm not sure how adding it to the `defaultParams` will help, unless we reference that in `AddFlags`.

strib:
@jzila: The client calls this function directly.  `AddFlags` isn't involved: https://github.com/keybase/client/blob/0f8ce2b1a334529fdda57efce63e27e017909c05/go/bind/keybase.go#L92

PR #864, opened 2017-03-20
vendor: update libkb to get NoSessionError exported error

strib:
cc: @patrickxb


vendor/vendor.json
```diff
@@ -153,10 +159,10 @@
 			"revisionTime": "2017-02-13T21:07:17Z"
 		},
 		{
-			"checksumSHA1": "iCQM0dov+82oscLvOAhMcNRjbvM=",
+			"checksumSHA1": "izAKnxnWAKBhfisFveXwWSd1Gz4=",
 			"path": "github.com/keybase/client/go/libkb",
-			"revision": "19e0d7ac1a781b4d23a9f386f8c2cd0057bab914",
-			"revisionTime": "2017-02-13T21:07:17Z"
+			"revision": "66b4b1333325587c1161b18c2b462b3d1781696e",
```
jzila: This is a different revision than the one in kbfs-server (b9cdd24a571538ecb27ae338deff7b25fdbc3c34). Is that ok?

strib: Yeah good question, I should have mentioned that.  I'm vendoring the branch of keybase/client#6299, since otherwise that PR will fail the kbfs-server tests.

PR #865, opened 2017-03-20
simplefs: Cancelling v1

taruti:
As discussed on Slack.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -30,8 +30,9 @@ type SimpleFS struct {
 }
 
 type inprogress struct {
-	desc keybase1.OpDescription
-	done chan error
+	desc   keybase1.OpDescription
+	cancel context.CancelFunc
+	done   chan error
 }
 
 type handle struct {
@@ -219,10 +220,28 @@ func (k *SimpleFS) doCopy(ctx context.Context, srcPath, destPath keybase1.Path) 
 	defer dst.Close()
 
 	if src.Type() == keybase1.DirentType_FILE || src.Type() == keybase1.DirentType_EXEC {
-		_, err = io.Copy(dst, src)
+		err = copyWithCancellation(ctx, dst, src)
+		if err != nil {
+			return err
+		}
 	}
 
-	return err
+	return nil
+}
+
+func copyWithCancellation(ctx context.Context, dst io.Writer, src io.Reader) error {
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		_, err := io.CopyN(dst, src, 64*1024)
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+	}
 }
 
 type pathPair struct {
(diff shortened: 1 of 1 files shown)


simplefs/simplefs.go
```diff
@@ -517,6 +542,20 @@ func (k *SimpleFS) SimpleFSClose(ctx context.Context, opid keybase1.OpID) (err e
...
+	k.lock.Lock()
+	defer k.lock.Unlock()
+	w, ok := k.inProgress[opid]
+	if !ok {
+		return
+	}
+	delete(k.inProgress, opid)
+	delete(k.handles, opid)
+	w.cancel()
+}
```
oconnor663: Will this work for `keybase fs write`, for example? Or would we still need to call `Close` there?

taruti: k.handles is what Close clears, so after a Cancel no Close or Wait needed.

oconnor663: Cool. Is there nothing behind the `Node` in the handle that needs to get closed / cleaned up? Just the `Node` itself?

simplefs/simplefs.go
```diff
@@ -219,10 +220,28 @@ func (k *SimpleFS) doCopy(ctx context.Context, srcPath, destPath keybase1.Path)
...
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		_, err := io.CopyN(dst, src, 64*1024)
+		if err == io.EOF {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
```
zanderz: `err == nil` always means there is more to copy?

oconnor663: I think the implementation of `Reader` is allowed to choose whether to return `EOF` with the last non-zero-length read, or to return it on the first zero-length read. But either way, I think it's correct for callers to loop until `EOF` like this.

simplefs/simplefs.go
```diff
@@ -517,6 +542,20 @@ func (k *SimpleFS) SimpleFSClose(ctx context.Context, opid keybase1.OpID) (err e
 	return err
 }
 
+// SimpleFSCancel starts to cancel op with the given opid.
+// Also remove any pending references of opid everywhere.
+func (k *SimpleFS) SimpleFSCancel(_ context.Context, opid keybase1.OpID) {
```
oconnor663: The generated protocol is going to expect a method that returns `error` I think.

oconnor663:
LGTM with the error change above. It looks like this is working on my example branches, https://github.com/keybase/client/tree/jack/CORE-4684/cancel_scratch and https://github.com/keybase/kbfs/tree/jack/CORE-4684/cancel_scratch.

simplefs/simplefs.go
taruti: Node does not need closing.

simplefs/simplefs.go
taruti: This uses CopyN for which ```On return, written == n if and only if err == nil.```.

However on reflection using CopyBuffer is better for the given types.

simplefs/simplefs.go
taruti: Will do.

PR #867, opened 2017-03-20
mdserver_remote: don't treat random errors as NoCurrentSessionErrors

strib:
If the service gives us some other error message when getting the session (as opposed to `libkb.NoSessionError`), we should retry the connection after a delay using the rpc connection's retry logic.

This also fixes up logging in `MDServerRemote` to be more consistent with the rest of the package.

Also, it re-vendors client to get the most up-to-date content hashes (previously, we had vendored from an unmerged branch).

Issue: KBFS-2031

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -129,7 +129,7 @@ func (md *MDServerRemote) OnConnect(ctx context.Context,
 		}
 	}()
 
-	md.log.Debug("MDServerRemote: OnConnect called with a new connection")
+	md.log.CDebugf(ctx, "OnConnect called with a new connection")
 
 	// we'll get replies asynchronously as to not block the connection
 	// for doing other active work for the user. they will be sent to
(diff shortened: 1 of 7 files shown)


libkbfs/mdserver_remote.go
```diff
@@ -171,34 +171,41 @@ func (md *MDServerRemote) resetAuth(ctx context.Context, c keybase1.MetadataClie
...
-		return MdServerDefaultPingIntervalSeconds, NoCurrentSessionError{}
+	if _, ok := err.(NoCurrentSessionError); ok {
+		md.log.CDebugf(ctx, "User logged out (%+v), skipping resetAuth", err)
+		return MdServerDefaultPingIntervalSeconds, err
+	} else if err != nil {
+		// We can't rely on getting a `LoggedIn` error from the
+		// service in this case, so let the RPC connection fail and
+		// retry a bit later.
+		md.log.CDebugf(ctx, "Unexpected error when getting session: %+v", err)
+		return MdServerDefaultPingIntervalSeconds, err
```
jzila: We're returning the same thing in both cases, so why split them up?

strib: Logging different messages.  I could dedup the `return` statement with an extra `if err != nil` around the whole thing, if you prefer.

jzila: It's fine. I'm just not sure whether the log message should be here or in the caller that actually branches on this functionality. It seems like the comment and log message are prescribing behavior for an upstream caller.

jzila: That wouldn't be bad if we were returning something different in the two cases, but it's confusing if we're just sending back the same error no matter what.

strib:
Cool, I just made the caller do the logging instead, and added a comment
about it.

PR #868, opened 2017-03-21
journal_md_ops: avoid remote mdserver calls for squash branches

strib:
We know for sure that all data is available locally.

On my work machine, this reduced the time to copy the Go source into KBFS from 4m50s to 4m25s, though the effect is probably better on slower internet connections.

Issue: KBFS-2038

### DIFF
--- a/libkbfs/journal_md_ops.go
+++ b/libkbfs/journal_md_ops.go
@@ -310,15 +310,24 @@ func (j journalMDOps) getRange(
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


jzila:
Nice! Have you compared the performance of this with existing performance, say on extracting the Go source?

strib:
> Nice! Have you compared the performance of this with existing performance, say on extracting the Go source?

@jzila: Yep!  See the description of this PR.

PR #870, opened 2017-03-21
vendor: update to new rpc.NewClient call

strib:
Issue: KBFS-2035


strib:
Note the `github.com/keybase/client` hashes won't match the client repo until I push keybase/client#6321 (and then I'll update this PR with the new hashes before merging).

PR #874, opened 2017-03-22
revendor SimpleFS protocol changes

oconnor663:
TODO: Get the hashes right after these changes land in client.

r? @zanderz @taruti


vendor/github.com/keybase/client/go/protocol/keybase1/simple_fs.go
```diff
@@ -548,6 +552,8 @@ type SimpleFSInterface interface {
 	// Close OpID, cancels any pending operation.
 	// Must be called after list/copy/remove
```
zanderz: I don't understand why separate close and cancel should be required, but the comments on Close should at least be changed. The "must be called after" part I don't think is true any more. @taruti ? 

Also, Taru said we should make the size argument a long instead of int for simpleFSRead and ReadArgs - can we get thjis in too?

In case the .avdl change was in a different PR: never mind!

oconnor663: It is in a different PR. Will be up in a sec.

oconnor663: https://github.com/keybase/client/pull/6328

oconnor663:
Upstream client PR landed. Revendoring and rebasing this one, and waiting for CI.

PR #875, opened 2017-03-23
[KBFS-2001] Skip Retrieval Workers for Items in the Disk Cache

jzila:
* [x] Export `blockRetriever` to `BlockRetriever`, and allow accessing it from `BlockOps`.
* [x] Factor out the `PrefetchAfterBlockRetrieved` responsibility to `BlockRetriever` as `CacheAndPrefetch`.
* [x] `blockRetrievalQueue` is now responsible for handling caching, and `Prefetcher` is only responsible for prefetching.
* [x] Move `DiskBlockCache.Get` calls from `BlockServerRemote` to `blockRetrievalQueue`, and call them before a request is placed in the request queue.
* [x] Check memory and disk caches before locking the `blockRetrievalQueue` mutex, so we don't lock the whole retrieval subsystem based on I/O to the disk. `DiskBlockCache.Get` and `DiskBlockCache.UpdateMetadata` both only take a read lock on the disk cache, so other goroutines attempting `Get`s won't block unless there's a `DiskBlockCache.Put` or `DiskBlockCache.Delete`.

### DIFF
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -13,6 +13,7 @@ import (
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/kbfshash"
 	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
 )
 
 type blockContainer struct {
@@ -242,6 +243,18 @@ func (b *BlockCacheStandard) makeRoomForSize(size uint64, lifetime BlockCacheLif
 func (b *BlockCacheStandard) PutWithPrefetch(
 	ptr BlockPointer, tlf tlf.ID, block Block, lifetime BlockCacheLifetime,
 	hasPrefetched bool) (err error) {
+	// Just in case we tried to cache a block type that shouldn't be cached,
+	// return an error. This is an insurance check. That said, this got rid of
+	// a flake in TestSBSConflicts, so we should still look for the underlying
+	// error.
+	switch block.(type) {
+	case *DirBlock:
+	case *FileBlock:
+	case *CommonBlock:
+		return errors.New("attempted to Put a common block")
+	default:
+		return errors.Errorf("attempted to Put an unknown block type %T", block)
+	}
 
 	var wasInCache bool
 
@@ -251,7 +264,8 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 
 	case TransientEntry:
 		// If it's the right type of block, store the hash -> ID mapping.
-		if fBlock, isFileBlock := block.(*FileBlock); b.ids != nil && isFileBlock && !fBlock.IsInd {
+		if fBlock, isFileBlock := block.(*FileBlock); b.ids != nil &&
+			isFileBlock && !fBlock.IsInd {
 
 			key := idCacheKey{tlf, fBlock.GetHash()}
 			// zero out the refnonce, it doesn't matter
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -264,10 +264,10 @@ func (b *BlockCacheStandard) PutWithPrefetch(
 		// We could use `cleanTransient.Contains()`, but that wouldn't update
 		// the LRU time. By using `Get`, we make it less likely that another
 		// goroutine will evict this block before we can `Put` it again.
-		var block interface{}
-		block, wasInCache = b.cleanTransient.Get(ptr.ID)
+		var bc interface{}
+		bc, wasInCache = b.cleanTransient.Get(ptr.ID)
 		if wasInCache {
-			hasPrefetched = (hasPrefetched || block.(blockContainer).hasPrefetched)
+			hasPrefetched = (hasPrefetched || bc.(blockContainer).hasPrefetched)
 		}
 		// Cache it later, once we know there's room
 
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -183,6 +183,12 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 			}
 		}()
 	}
+	defer func() {
+		if err != nil {
+			brq.log.CWarningf(ctx, "Error Putting into the block cache: %+v",
+				err)
+		}
+	}()
 	if hasPrefetched {
 		return brq.config.BlockCache().PutWithPrefetch(ptr, kmd.TlfID(), block,
 			lifetime, true)
@@ -205,8 +211,9 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 			"is full")
 		return err
 	default:
-		brq.log.CWarningf(ctx, "Error Putting into the block cache: %+v",
-			err)
+		// We should return the error here because otherwise we could thrash
+		// the prefetcher.
+		return err
 	}
 	// This must be called in a goroutine to prevent deadlock in case this
 	// CacheAndPrefetch call was triggered by the prefetcher itself.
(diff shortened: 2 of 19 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -11,6 +11,8 @@ import (
 	"reflect"
 	"sync"
 
+	"github.com/keybase/client/go/logger"
+
```
strib: Blank line not needed.

libkbfs/block_retrieval_queue.go
```diff
@@ -117,14 +121,16 @@ type blockRetrievalQueue struct {
...
 
 // newBlockRetrievalQueue creates a new block retrieval queue. The numWorkers
 // parameter determines how many workers can concurrently call Work (more than
 // numWorkers will block).
-func newBlockRetrievalQueue(numWorkers int, config blockRetrievalConfig) *blockRetrievalQueue {
+func newBlockRetrievalQueue(numWorkers int,
+	config blockRetrievalConfig) *blockRetrievalQueue {
 	q := &blockRetrievalQueue{
 		config:      config,
+		log:         config.MakeLogger("BRQ"),
```
strib: You don't need a prefix for the logger if there's only one instance of the queue.  The printed filename is enough in my opinion.

libkbfs/block_retrieval_queue.go
```diff
@@ -163,8 +169,84 @@ func (brq *blockRetrievalQueue) notifyWorker() {
...
+	// 1) To prevent any other Gets from prefetching.
+	// 2) To prevent prefetching if a cache Put fails, since prefetching if
+	//	  only useful when combined with the cache.
+	err = brq.config.BlockCache().PutWithPrefetch(ptr, kmd.TlfID(), block,
+		lifetime, true)
+	if _, isCacheFullError := err.(cachePutCacheFullError); isCacheFullError {
+		brq.log.CDebugf(context.TODO(), "Skipping prefetch because the cache "+
+			"is full")
+		return err
+	}
```
strib: Probably good to add an `else if err != nil` clause that prints the other errors so they don't get completely lost.

libkbfs/block_retrieval_queue.go
```diff
@@ -163,8 +169,84 @@ func (brq *blockRetrievalQueue) notifyWorker() {
 	}
 }
 
+func (brq *blockRetrievalQueue) CacheAndPrefetch(ptr BlockPointer, block Block,
```
strib: Function comment please (especially noting that the disk cache LRU time is updated, regardless of whether prefetching happened or not).

libkbfs/block_retrieval_queue.go
```diff
@@ -163,8 +169,84 @@ func (brq *blockRetrievalQueue) notifyWorker() {
...
+		return nil
+	}
+
+	// Check the disk cache.
+	dbc := brq.config.DiskBlockCache()
+	if dbc == nil {
+		return NoSuchBlockError{ptr.ID}
+	}
+	blockBuf, serverHalf, err := dbc.Get(ctx, kmd.TlfID(), ptr.ID)
+	if err != nil || len(blockBuf) == 0 {
```
strib: Do you intend to translate every possible leveldb or disk error into a `NoSuchBlockError`?  I guess that could make sense since the disk cache isn't critical, but I think we should have a comment here about that and log unexpected error types.

libkbfs/block_retrieval_worker_test.go
```diff
@@ -85,6 +86,12 @@ func (bg *fakeBlockGetter) getBlock(ctx context.Context, kmd KeyMetadata, blockP
 	}
 }
 
+func (bg *fakeBlockGetter) assembleBlock(ctx context.Context,
```
strib: You could save some characters by using `_` as the param names, but whatever you prefer is fine with me.

libkbfs/folder_branch_ops.go
```diff
@@ -1947,8 +1947,8 @@ func (fbo *folderBranchOps) unembedBlockChanges(
 		}
 		fblock, ok := block.(*FileBlock)
 		if !ok {
-			return nil, false,
-				errors.Errorf("Block for %s is not a file block", ptr)
+			return nil, false, errors.Errorf(
+				"Block for %s is not a file block. Block type: %T", ptr, block)
```
strib: I'd prefer avoiding sentences in these, maybe: `Block for %s is not a file block, block type: %T` instead?

libkbfs/interfaces.go
```diff
@@ -840,8 +840,12 @@ type DiskBlockCache interface {
 	// Put puts a block to the disk cache.
 	Put(ctx context.Context, tlfID tlf.ID, blockID kbfsblock.ID, buf []byte,
 		serverHalf kbfscrypto.BlockCryptKeyServerHalf) error
-	// DeleteByTLF deletes some blocks from the disk cache.
-	DeleteByTLF(ctx context.Context, tlfID tlf.ID, blockIDs []kbfsblock.ID) (numRemoved int, sizeRemoved int64, err error)
+	// Delete deletes some blocks from the disk cache.
+	Delete(ctx context.Context, blockIDs []kbfsblock.ID) (numRemoved int,
+		sizeRemoved int64, err error)
+	// UpdateMetadata updates the metadata for the given block, including
```
strib: "including"?  Since it doesn't take any parameters, I'm not sure how it can update anything else...

libkbfs/block_retrieval_queue.go
```diff
@@ -163,8 +169,84 @@ func (brq *blockRetrievalQueue) notifyWorker() {
 	}
 }
 
+func (brq *blockRetrievalQueue) CacheAndPrefetch(ptr BlockPointer, block Block,
+	kmd KeyMetadata, priority int, lifetime BlockCacheLifetime,
+	hasPrefetched bool) (err error) {
+	dbc := brq.config.DiskBlockCache()
+	if dbc != nil {
+		go dbc.UpdateMetadata(context.TODO(), ptr.ID)
```
strib: A linter might complain about ignoring the error here.  I'd prefer something that prints out unexpected errors, or at least assigns it to `_` so a linter would be satisfied.  But I won't fight too hard for it.

libkbfs/block_retrieval_queue.go
strib: I just realized this is already commented in `interfaces.go`.  Maybe you could add a note about the LRU time there?  It's just not obvious from the name that that happens.

libkbfs/block_retrieval_queue.go
jzila: You're right, I don't even use the `NoSuchBlockError` anyway; might as well just pass through the regular error.

jzila:
@strib there's some new fixes in here:
https://github.com/keybase/kbfs/pull/875/files/65d5d831af1be9fa1a8aa604cd468b7534aa6bfb..53d663f86e0dc4f937b2010963ffe53d03235b65

libkbfs/bcache.go
```diff
@@ -242,6 +243,18 @@ func (b *BlockCacheStandard) makeRoomForSize(size uint64, lifetime BlockCacheLif
...
 	hasPrefetched bool) (err error) {
+	// Just in case we tried to cache a block type that shouldn't be cached,
+	// return an error. This is an insurance check. That said, this got rid of
+	// a flake in TestSBSConflicts, so we should still look for the underlying
+	// error.
+	switch block.(type) {
+	case *DirBlock:
+	case *FileBlock:
+	case *CommonBlock:
+		return errors.New("attempted to Put a common block")
```
strib: Should we only return this error when `lifetime == NoCacheEntry`?  Or should we get rid of `NoCacheEntry` altogether if we want to force the caller into policing this?

jzila: I think #2 is probably correct. `NoCacheEntry` seems pretty brittle, since something was clearly slipping through.

jzila:
You're right in that we haven't actually found the place where we're trying to cache a `CommonBlock` (although it's pretty clear that this is happening). You're also right that `PutWithPrefetch` errors are being swallowed. I don't think that's incorrect behavior though: caching isn't mission-critical, so errors with it probably shouldn't percolate up.

That said, this definitely fixes the EIO flake in https://keybase.atlassian.net/browse/KBFS-2050, since there's no longer any way for the cache to return something that isn't a `{File,Dir}Block`.

libkbfs/bcache.go
strib: (Oops, above I meant `lifetime != NoCacheEntry`.)

strib:
Ok thanks.  I'm cool with it, I'm just worried that if we hide cache errors too much, we won't find out when something is going horribly wrong.  But hopefully performance would be bad enough that we'd be alerted to it in other ways.

jzila:
We're still logging them including stacks, so it'll leave a trail of breadcrumbs when something does go wrong.

PR #876, opened 2017-03-23
timeout mdserver ping much faster

mmaxim:
The advantage of this quicker timeout is that KBFS is more responsive to changes in the network, and the client's connectivity to it. The hope is we hit this timeout, disconnect the connection, and then move this whole deal into the connection retry mechanic. For context, we use the same timing for Gregor (10 second loop, 5 second timeout) in the client.

If we can't even ping the MD server in 5 seconds, KBFS is likely majorly slow anyway.

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -34,7 +34,7 @@ const (
 	MdServerDefaultPingIntervalSeconds = 10
 	// MdServerPingTimeout is how long to wait for a ping response
 	// before breaking the connection and trying to reconnect.
-	MdServerPingTimeout = 30 * time.Second
+	MdServerPingTimeout = 5 * time.Second
 )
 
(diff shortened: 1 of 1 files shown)


strib:
I'm not sure about this.  I chose 30s deliberately to avoid false timeouts when ping times were super slow or the connection was super congested or the servers were way overloaded.  In those cases, closing the connection and reconnecting is just about the worst thing you can do.

> If we can't even ping the MD server in 5 seconds, KBFS is likely majorly slow anyway.

As I said above, I consider this a pretty weak argument, because if KBFS is slow, this will only make things much worse.

I could be convinced, but at the moment I feel like the right thing to do is hook into OS constructs for network up/down if we want to really respond faster to local changes.  A timeout isn't the right thing for that . . .

mmaxim:
Ok sounds good, I will have another patch soon that I think will be better. I'll close this one.

PR #878, opened 2017-03-23
Allow identify skip

patrickxb:
This adds a new keybase1.IdentifyBehavior, CHAT_SKIP.  If present, KBFS will skip identify calls.  Service only sets this in the case of GetTLFCryptKeys where it does the identifies.

cc @mmaxim

### DIFF
--- a/libkbfs/identify_util.go
+++ b/libkbfs/identify_util.go
@@ -156,6 +156,13 @@ func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
 func identifyUser(ctx context.Context, nug normalizedUsernameGetter,
 	identifier identifier, username libkb.NormalizedUsername,
 	uid keybase1.UID, isPublic bool) error {
+
+	// check to see if identify should be skipped altogether
+	ei := getExtendedIdentify(ctx)
+	if ei.behavior == keybase1.TLFIdentifyBehavior_CHAT_SKIP {
+		return nil
+	}
+
 	var reason string
 	if isPublic {
 		reason = "You accessed a public folder."
--- a/libkbfs/identify_util.go
+++ b/libkbfs/identify_util.go
@@ -156,8 +156,7 @@ func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
 func identifyUser(ctx context.Context, nug normalizedUsernameGetter,
 	identifier identifier, username libkb.NormalizedUsername,
 	uid keybase1.UID, isPublic bool) error {
-
-	// check to see if identify should be skipped altogether
+	// Check to see if identify should be skipped altogether.
 	ei := getExtendedIdentify(ctx)
 	if ei.behavior == keybase1.TLFIdentifyBehavior_CHAT_SKIP {
 		return nil
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -964,6 +964,8 @@ func (fbo *folderBranchOps) identifyOnce(
 		len(ei.getTlfBreakAndClose().Breaks) > 0 {
 		fbo.log.CDebugf(ctx,
 			"Identify finished with no error but broken proof warnings")
+	} else if ei.behavior == keybase1.TLFIdentifyBehavior_CHAT_SKIP {
+		fbo.log.CDebugf(ctx, "Identify skipped")
 	} else {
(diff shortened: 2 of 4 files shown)


strib:
Oh wait, scratch that, misunderstood.  One second...

mmaxim:
I think @patrickxb wants the opposite behavior here.

libkbfs/identify_util.go
```diff
@@ -156,6 +156,13 @@ func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
 func identifyUser(ctx context.Context, nug normalizedUsernameGetter,
 	identifier identifier, username libkb.NormalizedUsername,
 	uid keybase1.UID, isPublic bool) error {
+
```
strib: We usually avoid blank lines at the start of functions.

patrickxb:
Yes, I want to kbfs to skip identify calls.

libkbfs/identify_util.go
```diff
@@ -156,6 +156,13 @@ func identifyUID(ctx context.Context, nug normalizedUsernameGetter,
...
+	// check to see if identify should be skipped altogether
```
strib: Please capitalize comments and end them in punctuation.

patrickxb:
In the case of `GetTLFCryptKeys`, the service will now be responsible for running identify.

I don't see `GetTLFCryptKeys` calls going through folderBranchOps.

    GetTLFCryptKeys
    getHandleFromFolderName
    ParseTLFHandle
    parseTLFHandleLoose
    identifyHandle
    identifyUsersForTLF
    identifyUsers
    identifyUser
    identifier.Identify
    identifyClient.Identify2

Am I missing something?

strib:
Yes, you're missing:

```
KBFSOpsStandard.GetTLFCryptKeys
KBFSOpsStandard.getMDByHandle
folderBranchOps.getMDForReadNeedIdentifyOnMaybeFirstAccess
... etc
```

If the service's definition of a successful identify differs from KBFS's, we don't want to cache the result.

mmaxim:
It seems like this might be as easy as adding an additional check in `identifyOnce` about whether or not to cache the result:

```
if ei.behavior.WarningInsteadOfErrorOnBrokenTracks() &&
		len(ei.getTlfBreakAndClose().Breaks) > 0 {
		fbo.log.CDebugf(ctx,
			"Identify finished with no error but broken proof warnings")
	} else {
		fbo.log.CDebugf(ctx, "Identify finished successfully")
		fbo.identifyDone = true
		fbo.identifyTime = fbo.config.Clock().Now()
	}
```

Just add an `else if` in there that checks for the `SKIP` or `VOUCH` and don't set the cache.

strib:
Also, `parseTlfHandleLoose` only runs identify when the folder name is non-canonical (social assertions, etc).

patrickxb:
What I'm saying is that the first path I showed could call Identify2 without regard to folderBranchOps.  We want it to not call Identify2.

strib:
> What I'm saying is that the first path I showed could call Identify2 without regard to folderBranchOps. We want it to not call Identify2.

Understood.  And I'm saying it could _also_ call it under `folderBranchOps` (and would do so in most cases, when the service gives us a canonical TLF name), and end up messing up the cache there.  So the right thing is to keep the code you have here, AND do what @mmaxim suggests.  In that case, I guess CHAT_SKIP is a ok name.

patrickxb:
ok, PTAL

mmaxim:
@patrickxb It might be also worth decluttering all the rest of the identify mode stuff scattered through KBFS as part of this change, since we don't need it anymore. Alternatively, we could save that for a different ticket, but I know @strib is not super excited about how the former code is in there in a couple places.

mmaxim:
LIke this for example:

```
if fbo.identifyDone && !ei.behavior.AlwaysRunIdentify() {
		// TODO: provide a way for the service to break this cache when identify
		// state changes on a TLF. For now, we do it this way to make chat work.
		return nil
	}
```

mmaxim:
Maybe better to save for another ticket actuall now that I think about it more, it doesn't make sense to make this more complicated than it needs to be.

patrickxb:
I'll leave that for another ticket.

strib:
Yeah I assumed it would be pushed simultaneously with the service side.  (Usually it's good to put a pre-emptive note in #breaking-ci or #ci to tell people they need to rebase to get circleci to pass.)

PR #879, opened 2017-03-23
Add read/write to GetOps (fs ps)

zanderz:
Not sure if this is the right approach but I tried it and it gave a result when a second client process did `fs ps` while another did `fs read`

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -420,16 +420,25 @@ func (k *SimpleFS) SimpleFSSetStat(ctx context.Context, arg keybase1.SimpleFSSet
 	return nil
 }
 
-func (k *SimpleFS) addReadWriteDesc(opid keybase1.OpID, desc keybase1.OpDescription) {
+func (k *SimpleFS) startReadWriteOp(ctx context.Context, opid keybase1.OpID, desc keybase1.OpDescription) (context.Context, error) {
+	ctx, err := k.startSyncOp(ctx, desc.AsyncOp__.String(), desc)
+	if err != nil {
+		return nil, err
+	}
 	k.lock.RLock()
 	k.inProgress[opid] = &inprogress{desc, func() {}, make(chan error, 1)}
 	k.lock.RUnlock()
+	return ctx, err
 }
 
-func (k *SimpleFS) clearReadWriteDesc(opID keybase1.OpID) {
+func (k *SimpleFS) doneReadWriteOp(ctx context.Context, opID keybase1.OpID, err error) {
 	k.lock.RLock()
 	delete(k.inProgress, opID)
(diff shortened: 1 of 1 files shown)


zanderz:
It occurs to me that we might want to list open file handles in ps, even if no active operation is happening?

maxtaco:
Took a quick look but I defer to @taruti on the review!

taruti:
@zanderz Thinking of this there are a few issues:

1) Is this needed?
2) Multiple read/writes may be in progress with the same file [thus sharing the same opid]
3) If we limit to one op why not use inProgress instead of adding a new struct member?

zanderz:
1. The author of CORE-4740, @maxtaco, says we need this. 
2. I don't think we need to necessarily support multi read/writes on one opid. Closing it would still end all the ops, wouldn't it? Or does Close synchronously wait until the read/write is finished?
3. I thought inProgress was for `cancel`, not `close`? And therefore needs a different type of context. etc.? Are you saying I should just make read/write async like the other inProgress ops?

maxtaco:
what if it's a really big or stalled operation?  might still be useful.  i wasn't dead set on a change, I was wondering what the implications would be is all.

maxtaco:
btw, my suggestion was made without regard to any implementation complexities, or the costs of the change, so obviously is open to discussion...

taruti:
So we have the following ops at the moment:
1) Async, cancelable, visible in ps, requiring Wait and Close (and ReadList): List, ListRecursive
2) Async, cancellable, visible in ps, requiring Wait: Copy, CopyRecursive, Move, Remove
3) Sync special: ReadList, Close [after Open, List, or ListRecursive], Cancel, Check, GetOps, Wait
4) Sync normal: Rename, SetStat, Stat
5) Sync requires Close: Open
6) Sync requires handle opened by Open: Read, Write

Currently only async operations can be cancelled and are visible through GetOps and they have to be waited. Certainly exposing Read and Write also to ps even through they are sync can work out.

zanderz:
@taruti ping

taruti:
Why not startReadWriteOp + doneReadWriteOp?

i.e.
```
func (k *SimpleFS) startSyncOp(ctx context.Context, opid keybase1.OpID, opname string, desc keybase1.OpDescription) (context.Context, error) {
    ctx, err := k.startSyncOp(ctx, "Read", opDesc)
    if err!=nil { return nil, err }
    k.lock.RLock()
    k.inProgress[opid] = &inprogress{desc, func() {}, make(chan error, 1)}
    k.lock.RUnlock()
    return ctx, err
}
```
And the same for clearReadWriteDesc -> doneReadWriteDesc. Would make the actual code calling those functions simpler.

Otherwise LGTM.

PR #881, opened 2017-03-27
fix routine leak in rekey_queue

songgao:
I tested locally using `gops` to make sure there wasn't any left-over routines after `test()` returned. Thanks for discovering this!


libkbfs/rekey_queue.go
```diff
@@ -74,17 +74,23 @@ func NewRekeyQueueStandard(config Config) (rkq *RekeyQueueStandard) {
...
+					return
+				}
+				rkq.config.KBFSOps().RequestRekey(context.Background(), id)
+				func(id tlf.ID) {
+					rkq.mu.Lock()
+					defer rkq.mu.Unlock()
+					delete(rkq.pendings, id)
+				}(id)
+			case err := <-ctx.Done():
+				rkq.log.Debug("%v", err)
```
jzila: More detailed log message here please.

PR #883, opened 2017-03-28
simplefs: add more comments

taruti:


taruti:
Only comments.

simplefs/simplefs.go
```diff
@@ -520,8 +528,7 @@ func (k *SimpleFS) SimpleFSMakeOpid(_ context.Context) (keybase1.OpID, error) {
 	return opid, err
 }
 
-// SimpleFSClose - Close OpID, cancels any pending operation.
-// Must be called after list/copy/remove
```
zanderz: These comments are from the avdl and its which generated output, which we should update too. I'll make a ticket.

simplefs/simplefs.go
```diff
@@ -22,11 +22,19 @@ import (
...
+	// config for the fs - constant, does not need locking.
+	config libkbfs.Config
+	// lock protects handles and inProgress
+	lock sync.RWMutex
+	// handles contains handles opened by SimpleFSOpen,
+	// closed by SimpleFSClose (or SimpleFSCancel) and used
+	// by listing, reading and writing.
+	handles map[keybase1.OpID]*handle
+	// inProgress is for keeping state of operations in progress,
+	// values are removed by SimpleFSWait (or SimpleFSCancel).
```
zanderz: These changes aren't part of this PR are they? They're already in?

taruti: This documents how it is working at the moment in master.

simplefs/simplefs.go
taruti: Ok.

PR #885, opened 2017-03-29
[KBFS-2058] Compact disk block cache asynchronously

jzila:


libkbfs/disk_block_cache.go
```diff
@@ -271,12 +276,27 @@ func (cache *DiskBlockCacheStandard) syncBlockCountsFromDb() error {
...
+	// Execute these in a goroutine so we don't block reads or the completion
+	// of the delete that called this. Use a sentinel channel to make sure only
+	// one compaction can happen at once.
+	select {
+	case <-cache.compactCh:
+		blockDb := cache.blockDb
+		metaDb := cache.metaDb
+		tlfDb := cache.tlfDb
+		go func() {
+			metaDb.CompactRange(util.Range{})
```
strib: Can you add logging here before and after the compactions (maybe before and after each one?) so we can tell how long the compactions take, and when they start/stop, so that we can line that up with CPU graphs?

jzila:
@strib: LevelDB definitely doesn't lock out reads, but because of I/O bandwidth use they'll be slower.

PR #886, opened 2017-03-29
[KBFS-2057] Initialize the disk block cache asynchronously

jzila:


jzila:
Good idea strib. I'll do that!

PR #889, opened 2017-03-29
journal_server: parallelize enabling TLF journals

strib:
On my spinning disk, with 86 journals, this saves 15ms out of a total of 95ms. On my SSD it's basically negligible (2ms out of 24ms).  That said I don't see anywhere near the delays that were seen in KBFS-2057, so I'm not sure what this would be like on that device.  Still this doesn't seem like it's worse and doesn't make the code much more complicated, so it seems like it's worth merging.

Issue: KBFS-2057

### DIFF
--- a/libkbfs/journal_server.go
+++ b/libkbfs/journal_server.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"sync"
 
@@ -16,6 +17,7 @@ import (
 	"github.com/keybase/kbfs/tlf"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
 )
 
 type journalServerConfig struct {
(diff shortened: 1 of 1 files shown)


libkbfs/journal_server.go
```diff
@@ -311,55 +313,86 @@ func (j *JournalServer) EnableExistingJournals(
...
-			j.log.CDebugf(
-				ctx, "Skipping dir %q due to mismatched UID %s",
-				name, uid)
-			continue
+	eg, groupCtx := errgroup.WithContext(ctx)
+
+	fileCh := make(chan os.FileInfo, len(fileInfos))
+	worker := func() error {
+		for fi := range fileCh {
+			name := fi.Name()
```
strib: The contents of this loop are unchanged from the previous code, except it uses the `groupCtx` instead of `ctx` everywhere.

jzila: Cool thx. For future reference you can link to https://github.com/keybase/kbfs/pull/889/files?w=1 (note the w=1) to see the whitespace-agnostic diff (but we have to comment on the regular diff).

strib: Ah nice, didn't know that.

jzila: #secretfeatures

PR #891, opened 2017-03-30
journal_server: don't write to tlfJournals in enableLocked

strib:
Since now it might be called in parallel by EnableExistingJournals,
this would cause a concurrent read/write map access failure.  Instead,
return the tlfJournal and have the caller add it to the map under the
lock.

Issue: KBFS-2065

### DIFF
--- a/libkbfs/journal_server.go
+++ b/libkbfs/journal_server.go
@@ -316,6 +316,11 @@ func (j *JournalServer) EnableExistingJournals(
 	eg, groupCtx := errgroup.WithContext(ctx)
 
 	fileCh := make(chan os.FileInfo, len(fileInfos))
+	type journalRet struct {
+		id      tlf.ID
+		journal *tlfJournal
+	}
+	journalCh := make(chan journalRet, len(fileInfos))
 	worker := func() error {
 		for fi := range fileCh {
 			name := fi.Name()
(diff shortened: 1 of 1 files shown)


libkbfs/journal_server.go
```diff
@@ -316,6 +316,11 @@ func (j *JournalServer) EnableExistingJournals(
 	eg, groupCtx := errgroup.WithContext(ctx)
 
 	fileCh := make(chan os.FileInfo, len(fileInfos))
+	type journalRet struct {
+		id      tlf.ID
+		journal *tlfJournal
+	}
+	journalCh := make(chan journalRet, len(fileInfos))
```
strib: Well it's the same size as `fileInfos` and `fileCh` themselves right?  So it doesn't add much more overhead compared to what's already there, as far as I can tell.  Plus, this is the same size as the eventual `tlfJournals`, which has to reside in memory.  So I don't think there's anything too scary about this.

libkbfs/journal_server.go
```diff
@@ -390,16 +396,24 @@ func (j *JournalServer) EnableExistingJournals(
 		// happen...
 		return err
 	}
+	close(journalCh)
+
+	for r := range journalCh {
```
strib: Well, I could read from it, but I couldn't update `j.tlfJournals` until waiting on `eg`, because that would trigger the same race.

jzila: Understood, thanks.

PR #892, opened 2017-03-30
[KBFS-2050] bcache: Fix attempt to cache nil block

jzila:


libkbfs/block_types.go
```diff
@@ -250,6 +250,9 @@ func (fb *FileBlock) Set(other Block) {
 	fb.Contents = fbCopy.Contents
 	fb.IPtrs = fbCopy.IPtrs
 	fb.ToCommonBlock().Set(fbCopy.ToCommonBlock())
+	if otherFb.hash != nil {
```
jzila: This was necessary to avoid having to special-case hashing in the tests.

libkbfs/block_retrieval_worker_test.go
```diff
@@ -89,7 +89,12 @@ func (bg *fakeBlockGetter) getBlock(ctx context.Context, kmd KeyMetadata, blockP
 func (bg *fakeBlockGetter) assembleBlock(ctx context.Context,
 	kmd KeyMetadata, ptr BlockPointer, block Block, buf []byte,
 	serverHalf kbfscrypto.BlockCryptKeyServerHalf) error {
-	return errors.New("Not implemented")
+	source, ok := bg.blockMap[ptr]
```
jzila: Somewhat cheating to avoid having to pass test crypto stuff to the fake block getter.

libkbfs/block_retrieval_queue.go
```diff
@@ -176,12 +176,9 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 	lifetime BlockCacheLifetime, hasPrefetched bool) (err error) {
 	dbc := brq.config.DiskBlockCache()
 	if dbc != nil {
-		go func() {
-			err := dbc.UpdateLRUTime(ctx, ptr.ID)
-			if err != nil {
-				brq.log.CWarningf(ctx, "Error updating metadata: %+v", err)
-			}
-		}()
+		if err := dbc.UpdateLRUTime(ctx, ptr.ID); err != nil {
```
jzila: With this in a goroutine, the new test would flake because it'd try to call `dbc.UpdateLRUTime` after the LevelDB was closed.

jzila:
@strib: Added a test. The test hangs without the fix.

libkbfs/block_retrieval_queue.go
strib: That seems like something you should fix inside `dbc.UpdateLRUTime` by turning it into a no-op if Shutdown has already been called, right?  No need to penalize happy-path performance.

libkbfs/block_types.go
strib: Please put this in a comment here so we don't forget why we're doing it.  Though maybe expand it a little because I don't totally follow this comment.

libkbfs/disk_block_cache_test.go
```diff
@@ -447,12 +455,56 @@ func TestDiskBlockCacheDynamicLimit(t *testing.T) {
...
+	require.NotNil(t, q)
+	defer q.Shutdown()
+
+	ctx := context.Background()
+	kmd := makeKMD()
+	ptr1, block1, block1Encoded, serverHalf1 := setupBlockForDiskCache(
+		t, dbcConfig)
+	err := dbc.Put(ctx, kmd.TlfID(), ptr1.ID, block1Encoded, serverHalf1)
+	_, _ = bg.setBlockToReturn(ptr1, block1)
+	require.NoError(t, err)
```
strib: Would be nice to swap this with the line above so the err checking happens right away.

libkbfs/disk_block_cache_test.go
```diff
@@ -447,12 +455,56 @@ func TestDiskBlockCacheDynamicLimit(t *testing.T) {
...
+	ch := q.Request(ctx, 1, kmd, ptr1, block, TransientEntry)
+	err = <-ch
+	require.NoError(t, err)
+	require.Equal(t, block1, block)
+
+	t.Log("Remove the block from the disk cache to rule it out for " +
+		"the next step.")
+	numRemoved, _, err := dbc.Delete(ctx, []kbfsblock.ID{ptr1.ID})
+	require.Equal(t, 1, numRemoved)
+	require.NoError(t, err)
```
strib: Swap this line too.

PR #896, opened 2017-04-04
tlf_handle_resolve: don't trust Resolve for compound assertions

strib:
Resolve2 only checks the first component of a compound assertion, so
as not to rely on server trust.  KBFS was relying on Resolve2 for
compound assertions, and thus was broken.  Instead, call Identify2 for
compound assertions for now, until we can get a better fix with
KBFS-2022.

Issue: KBFS-2068


strib:
cc: @maxtaco @patrickxb @malgorithms

maxtaco:
LGTM!

maxtaco:
Thanks for rushing out the fix

PR #897, opened 2017-04-05
folder_update_prepper: refactor code that helps get MD updates ready for sending to servers

strib:
In preparation for batching updates in `folderBranchOps`, this moves some code out of both `folderBranchOps` and `conflictResolver` into a new helper struct, `folderUpdatePrepper`.  This doesn't change any of the code materially, it just does code renames, moves, and re-formatting.

The idea is that `folderBranchOps` will be able to call `folderUpdatePrepper.prepUpdateForPaths` with a whole batch of updates, rather than calling `folderUpdatePrepper.prepUpdateForPath` once for every FUSE operation, and thus save a bunch of encode/encrypt/journal time for most FUSE operations.

Issue: KBFS-2064


strib:
cc: @songgao @akalin-keybase, in case you guys are also interested in reviewing the series of PRs that will eventually make up KBFS-2064.  Signal your desire by reviewing this one to start, and I'll tag you as a reviewer on future PRs in this series as well.

libkbfs/folder_update_prepper.go
```diff
@@ -0,0 +1,1223 @@
...
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+// folderUpdatePrepper is a helper struct for preparing blocks and MD
+// updates before they get synced to the backend servers.  It can be
+// used for a single update or for a batch of updates (e.g. conflict
+// resolution).
+type folderUpdatePrepper struct {
+	config       Config
```
jzila: Could you create a custom `folderUpdatePrepperConfig` type to represent which config elements this needs?
From my accounting, that is:
```go
type folderUpdatePrepperConfig interface {
    clockGetter
    blockCacher
    blockOpsGetter // new
    cryptoPureGetter // you'll have to change the .Crypto() calls below to .cryptoPure()
    codecGetter
    dataVersioner
    blockSplitterGetter // new
    blockServerGetter // new
    currentSessionGetter
}
```

strib: Oh god, that looks horrific to me.  I know you guys are into it, but I can't see any benefit to it for this type. Can you explain why this is useful for this type?

jzila: 1. It makes explicit the dependencies this component takes.
1. It makes components more testable: you only have to implement test versions of the methods actually used (or pass in stock test implementations of each interface).

If the names bother you, you can explicitly name the methods:
```go
type folderUpdatePrepperConfig interface {
    Clock() Clock
    Codec() kbfscodec.Codec
    BlockCache() BlockCache
    ...etc
}
```
The `cryptoPure` change below should happen anyway though. All the methods that use `.Crypto()` take a `cryptoPure`, so there's no need to pass the full `Crypto`.

strib: > It makes components more testable: you only have to implement test versions of the methods actually used (or pass in stock test implementations of each interface).

How does the `Config` interface not give this?  Isn't exactly the same amount of work to populate certain fields in a `ConfigLocal` as to do it in this new one, except it makes the test code harder to understand because the types change from test to test?

I can understand the explicit dependency argument, but to me that's only useful if we ever split things out into more packages, and it's over-engineering to do it now, and in my opinion makes the code harder to follow for no reason.

jzila: Yeah I suppose instead of `testFooConfig` we could have
`newTestConfigLocalForFoo`, which populates only the fields that are needed.

I do like the explicit dependency specification, but I understand if you don't like the slight code overhead it adds per component.

strib: I made the `cryptoPure` change, but will leave the full `Config` for now.  It's not just the code overhead in this file I'm worried about, but future flexibility: if this struct ends up needing to initialize some other struct that needs something else, then I need to plumb the new thing down through all the configs.

If we ever move this into a separate package, I volunteer to break up the config at that point.

PR #903, opened 2017-04-10
kbfsdokan: don't error out on non-force fs mount failure

zanderz:
At least people will be able to chat and use simplefs if dokan goes all wonky.


libdokan/mounter.go
```diff
@@ -64,9 +64,13 @@ func (m *DefaultMounter) Mount(cfg *dokan.Config, log logger.Logger) error {
 	}
 
 	if err != nil {
-		return err
+		if m.force {
+			return err
+		}
+		log.Info("continuing after filesystem mount failure %s", err)
```
strib: Please use `%+v` for error printing.

PR #912, opened 2017-04-17
process reachability notifications from the service, and plug MDServer into them

mmaxim:
The patch does the following:

1.) Implements the `keybase1.ReachabilityProtocol` for the MDServer.
2.) When receiving a `ReachabilityChanged` notification from the service, then we do a basic test (`DialWithTimeout`), to see if we can still reach the MDServer at the network level.

`ReachabilityChanged` is triggered from an OS level update indicating the network state has changed in some way. The service currently uses it to be aggressive about dropping the Gregor connection in response to these updates, and this patch extends that same behavior to MDServer.

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -391,14 +391,13 @@ func (md *MDServerRemote) ShouldRetryOnConnect(err error) bool {
 // CheckReachability implements the MDServer interface.
 func (md *MDServerRemote) CheckReachability(ctx context.Context) {
 	conn, err := net.DialTimeout("tcp", md.mdSrvAddr, MdServerPingTimeout)
-	if conn != nil {
-		conn.Close()
-		return
-	}
 	if err != nil {
 		md.log.CDebugf(ctx,
 			"MDServerRemote: CheckReachability(): failed to connect, reconnecting: %s", err.Error())
 		md.initNewConnection()
+	}
+	if conn != nil {
+		conn.Close()
 	}
 }
 
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"fmt"
+	"net"
 	"sync"
 	"time"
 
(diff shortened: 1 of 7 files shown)


libkbfs/mdserver_remote.go
```diff
@@ -387,6 +388,20 @@ func (md *MDServerRemote) ShouldRetryOnConnect(err error) bool {
...
 }
 
+// CheckReachability implements the MDServer interface.
+func (md *MDServerRemote) CheckReachability(ctx context.Context) {
+	conn, err := net.DialTimeout("tcp", md.mdSrvAddr, MdServerPingTimeout)
+	if conn != nil {
+		conn.Close()
+		return
+	}
+	if err != nil {
```
strib: Can we re-order these cases so the err check is first?  Seems more natural.  Also, if `err == nil`, will `conn` ever be `nil`?

libkbfs/mdserver_remote.go
```diff
@@ -387,6 +388,20 @@ func (md *MDServerRemote) ShouldRetryOnConnect(err error) bool {
 	return !inputCanceled
 }
 
+// CheckReachability implements the MDServer interface.
+func (md *MDServerRemote) CheckReachability(ctx context.Context) {
```
strib: It would be less code, would avoid opening a new connection, and would test the actual connection in use, if you call `md.pinger.pingOnce()` instead of using `net.DialTimeout`.  I don't really see a downside, unless the reachability protocol is explicitly trying to measure _new_ connectivity to the server, rather than existing connectivity.

libkbfs/mdserver_remote.go
mmaxim: Sure.

libkbfs/mdserver_remote.go
mmaxim: I've found that using `DialTimeout` has been a better way to determine the connection is down after getting one of these "reachability" notifications than running a ping on the current connection. As an example, if you pull the network cable, any additional pings on an existing connection will just hang (at least on macOS). Whereas, if we try to `Dial` we hit an error immediately and kill the connection. 

We want to kill the connection because there is another annoying problem on macOS where if you pull the network cable, and then plug it back in, you can get a weird error about trying to bind to the local port that is already in use. The best way I have found to counter this problem on the CORE side is to just kill the connection and try to reconnect (in the presence of this reachability notification). Here is an example of what that error looks like:

2017-04-18T10:07:20.667529-04:00 ▶ [DEBU kbfs mdserver_remote.go:291] 1682 MDServerRemote: ping error write tcp 192.168.1.7:60118->52.73.199.240
:443: write: can't assign requested address

strib: Cool thanks, understood.  I guess my point is the opposite -- if the current connection is bad, but for some reason a _new_ connection will succeed easily, this won't actually fix KBFS's connectivity.  But I guess in that case we can just rely on the normal background pings.

mmaxim: Yeah, you are right this is _only_ testing the ability to create a new connection. If the new one somehow succeeds but the old one is bad, the hope is the ping loop will detect it.

PR #913, opened 2017-04-17
[KBFS-2000] Store whether blocks have been prefetched in the disk cache

jzila:


libkbfs/block_retrieval_queue.go
```diff
@@ -174,35 +174,38 @@ func (brq *blockRetrievalQueue) notifyWorker() {
...
 		if err != nil {
 			brq.log.CWarningf(ctx, "Error Putting into the block cache: %+v",
 				err)
 		}
+		dbc := brq.config.DiskBlockCache()
+		if dbc != nil {
+			go func() {
+				err := dbc.UpdateMetadata(ctx, ptr.ID, hasPrefetched)
+				if err != nil {
+					brq.log.CWarningf(ctx, "Error updating metadata: %+v", err)
```
strib: Can we take this opportunity to avoid printing a warning in the "block not found" case? Seems pretty common when I look through logs.

libkbfs/disk_block_cache.go
```diff
@@ -409,14 +411,20 @@ func (cache *DiskBlockCacheStandard) Get(ctx context.Context, tlfID tlf.ID,
...
 			NoSuchBlockError{blockID}
 	}
-	err = cache.updateMetadataLocked(ctx, tlfID, blockKey, len(entry))
+	md, err := cache.getMetadata(blockID)
 	if err != nil {
-		return nil, kbfscrypto.BlockCryptKeyServerHalf{}, err
+		return nil, kbfscrypto.BlockCryptKeyServerHalf{}, false, err
+	}
+	err = cache.updateMetadataLocked(ctx, tlfID, blockKey, len(entry),
+		hasPrefetched)
```
strib: Should this be `md.HasPrefetched`?  Otherwise you'll just overwrite it with `false` on every Get.

PR #917, opened 2017-04-19
don't ask for quota in statfs in logged out and other small fixes

songgao:
I verified that without ignoring quota error in statfs, macOS gets permission denied error. This PR fixes that


libfuse/fs.go
```diff
@@ -363,7 +363,9 @@ func (f *FS) Statfs(ctx context.Context, req *fuse.StatfsRequest, resp *fuse.Sta
 	usageBytes, limitBytes, err := f.quotaUsage.Get(ctx, quotaUsageStaleTolerance)
 	if err != nil {
 		f.log.CDebugf(ctx, "Getting quota usage error: %v", err)
-		return err
+		// Ignore the error here so Statfs() can succeeded. Otherwise reading a
+		// public TLF while logged out fails on macOS.
+		return nil
```
strib: This is a good start, but what do you think about just using `libkbfs.GetCurrentSessionIfPossible()` to avoid calling this altogether if there's no session?  That way we can avoid continuous RTTs to the server when logged out.

songgao: sounds good!

songgao:
ready for review again!

libfuse/fs.go
```diff
@@ -354,6 +354,12 @@ func (f *FS) Statfs(ctx context.Context, req *fuse.StatfsRequest, resp *fuse.Sta
 		Namelen: ^uint32(0),
 		Frsize:  fuseBlockSize,
 	}
+	if _, err := libkbfs.GetCurrentSessionIfPossible(
+		ctx, f.config.KBPKI(), false); err != nil {
```
strib: Did you test this?  I think it's the wrong usage of `GetCurrentSessionIfPossible`.  It returns a nil error if you're logged in, but with a zero-value session.  That way you can still return unexpected errors.

songgao: I did test this but let me test again to make sure

strib: Oops, I meant to say "if you're *not* logged in".  Will edit.

strib: https://github.com/keybase/kbfs/blob/ad3b9231b631c41a62729cf6a10520e61c8829d0/libkbfs/kbpki_client.go#L228-L231

songgao: It does work while logged out. Weird. I'll try to figure out why

songgao: Ah I remember now. We are passing in `false` here. So it's this case: https://github.com/keybase/kbfs/blob/ad3b9231b631c41a62729cf6a10520e61c8829d0/libkbfs/kbpki_client.go#L223

strib: Ah.  I think you should use `true` instead, to keep from suppressing unexpected errors.

songgao:
ready for review :D

songgao:
Ah nice; thanks Jeremy!

PR #918, opened 2017-04-19
libfuse: support directory syncing

strib:
By simply calling SyncAll() for that TLF.

(Depends on #909.)

Issue: KBFS-2070


libfuse/tlf.go
```diff
@@ -243,6 +243,15 @@ func (tlf *TLF) Setattr(ctx context.Context, req *fuse.SetattrRequest, resp *fus
 	return dir.Setattr(ctx, req, resp)
 }
 
+// Fsync implements the fs.NodeFsyncer interface for TLF.
+func (tlf *TLF) Fsync(ctx context.Context, req *fuse.FsyncRequest) (err error) {
+	dir, err := tlf.loadDir(ctx)
```
songgao: If we don't have a `dir` yet, we wouldn't need to do a sync right? If so, perhaps we can just `getStoredDir()` like `Attr()`?

strib: Good idea, done in 5156abc.

PR #919, opened 2017-04-19
[KBFS-2095] Enable disk cache for everyone

jzila:


jzila:
Yep, leaving it around for the future, since I had to dig through previous commits last time I wanted to restore it.

PR #935, opened 2017-05-02
ops: small truncate fixes

strib:
* When collapsing writes, the later truncate always wins.
* When inverting ops, make copies of gcOps and resolutionOps, otherwise we're setting ourselves up for trouble.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/ops.go
+++ b/libkbfs/ops.go
@@ -1248,9 +1248,9 @@ func invertOpForLocalNotifications(oldOp op) (newOp op, err error) {
 			return nil, err
 		}
 	case *GCOp:
-		newOp = op
+		newOp = newGCOp(op.LatestRev)
 	case *resolutionOp:
-		newOp = op
+		newOp = newResolutionOp()
 	}
 
 	// Now reverse all the block updates.  Don't bother with bare Refs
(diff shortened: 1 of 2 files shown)


libkbfs/ops.go
```diff
@@ -867,10 +867,7 @@ func addToCollapsedWriteRange(writes []WriteRange,
 			// Truncate past the last write.
 			return append(head, wNew)
 		} else if mid[0].isTruncate() {
-			// Min truncate wins
-			if mid[0].Off < wNew.Off {
-				return append(head, mid[0])
-			}
+			// Later truncates always win.
```
songgao: After this change, if a 3MB file is truncated to 1MB, then truncated to 2MB, does it retain the data from 1MB to 2MB or fill with `\0`?

strib: Good question, it is filled with `\0` in that case.  These write ranges aren't used for the actual data, that still lives in the block.  The write ranges are just describing the data so that things like CR can figure out if a file with two different sets of operations on them need conflict resolution, or if they are still essentially the same file.

You're right though that a more accurate description in that case would be to add a `WriteRange{Len=1000, off=1000}` to reflect the fact that those bytes changed.  I'll update this and add a test when I get a chance.

strib: @songgao: Done in a58fd5a.

songgao: Ah nice! Thanks for the explanation and the amendment

PR #938, opened 2017-05-02
cr: merge new, truncated files

strib:
A future commit will immediately truncate every newly-created file, in
order to mark it dirty.  To avoid unnecessary conflicts, we should
treat the same file name that's only been truncated as mergeable
everywhere in the CR code, not just during action computation.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/conflict_resolver.go
+++ b/libkbfs/conflict_resolver.go
@@ -442,28 +442,39 @@ func (sp crSortedPaths) Swap(i, j int) {
 	sp[j], sp[i] = sp[i], sp[j]
 }
 
-func fileWithConflictingWrite(unmergedChains, mergedChains *crChains,
+func createdFileWithConflictingWrite(unmergedChains, mergedChains *crChains,
 	unmergedOriginal, mergedOriginal BlockPointer) bool {
 	mergedChain := mergedChains.byOriginal[mergedOriginal]
 	unmergedChain := unmergedChains.byOriginal[unmergedOriginal]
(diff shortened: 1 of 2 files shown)


libkbfs/conflict_resolver.go
```diff
@@ -442,28 +442,39 @@ func (sp crSortedPaths) Swap(i, j int) {
...
+	// If the unmerged one is just a truncate, we can safely ignore
+	// the unmerged chain.
+	if len(unmergedWriteRange) == 1 && unmergedWriteRange[0].isTruncate() &&
+		unmergedWriteRange[0].Off == 0 {
+		unmergedChain.removeSyncOps()
+		return false
+	}
+
+	// If the merged one is just a truncate, we can safely ignore
+	// the unmerged chain.
```
jzila: merged chain

jzila: Also what if the truncates are of different sizes?

strib: Truncates are defined as a `WriteRange` where `Len == 0`.  So they don't really have "sizes" per se.  If you mean what happens if the unmerged branch truncates to a different place, or even does some writes -- that's ok.  Here's we're just ignoring the merged truncate to the beginning of the file.  This function is only called for _created_ files, so there's no danger that the merged truncate to 0 actually changed the data in the file.

strib:
I filed https://keybase.atlassian.net/browse/KBFS-2145 for the previous test run failure., I'm pretty sure it's a random, unrelated flake.

PR #939, opened 2017-05-02
cr: don't unref block IDs that have already been GC'd

strib:
If blocks have already been completely deleted on the merged branch, then CR over unmerged branch operations that also delete those same blocks don't need to bother unreferencing them.

This will be tested in a future PR for KBFS-2076.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/folder_update_prepper.go
+++ b/libkbfs/folder_update_prepper.go
@@ -572,7 +572,18 @@ func (fup folderUpdatePrepper) updateResolutionUsage(ctx context.Context,
 // addUnrefToFinalResOp makes a resolutionOp at the end of opsList if
 // one doesn't exist yet, and then adds the given pointer as an unref
 // block to it.
-func addUnrefToFinalResOp(ops opsList, ptr BlockPointer) opsList {
+func addUnrefToFinalResOp(ops opsList, ptr BlockPointer,
+	doNotUnref map[BlockPointer]bool) opsList {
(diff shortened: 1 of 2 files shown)


libkbfs/folder_update_prepper.go
```diff
@@ -572,7 +572,18 @@ func (fup folderUpdatePrepper) updateResolutionUsage(ctx context.Context,
...
 // one doesn't exist yet, and then adds the given pointer as an unref
 // block to it.
-func addUnrefToFinalResOp(ops opsList, ptr BlockPointer) opsList {
+func addUnrefToFinalResOp(ops opsList, ptr BlockPointer,
+	doNotUnref map[BlockPointer]bool) opsList {
+	// Make sure the block ID we want to unref isn't in the "do not
+	// unref" list -- it could mean that block has already been GC'd
+	// by the merged branch.  We can't compare pointers directly
+	// because GC'd pointers contain no block context.
+	for noUnref := range doNotUnref {
```
songgao: Is there a reason not to do something like `if _, ok := doNotUnref[ptr.ID]; ok { return ops }` here?

strib: Yeah, the key of `doNotUnref` is a `BlockPointer`, not a `kbfsblock.ID`, so that won't work.  And it needs to stay a `BlockPointer` for other reasons elsewhere in the code unfortunately.

songgao: Ah I see. Thanks!

PR #940, opened 2017-05-02
folder_branch_ops: don't ref/unref pointers within same MD

strib:
This is a no-op until a later PR cached directory ops in `fbo.dirOps`.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -2886,14 +2886,16 @@ func (fbo *folderBranchOps) CreateLink(
 
 // unrefEntry modifies md to unreference all relevant blocks for the
 // given entry.
-func (fbo *folderBranchOps) unrefEntry(ctx context.Context,
+func (fbo *folderBranchOps) unrefEntryLocked(ctx context.Context,
 	lState *lockState, md *RootMetadata, dir path, de DirEntry,
 	name string) error {
+	fbo.mdWriterLock.AssertLocked(lState)
 	if de.Type == Sym {
 		return nil
 	}
 
-	md.AddUnrefBlock(de.BlockInfo)
+	unrefsToAdd := make(map[BlockPointer]BlockInfo)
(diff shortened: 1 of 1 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -2908,9 +2914,26 @@ func (fbo *folderBranchOps) unrefEntry(ctx context.Context,
 			return err
 		}
 		for _, blockInfo := range blockInfos {
-			md.AddUnrefBlock(blockInfo)
+			unrefsToAdd[blockInfo.BlockPointer] = blockInfo
 		}
 	}
+
+	// Any referenced blocks that were unreferenced since the last
+	// sync can just be forgotten about.  Note that any updated
```
songgao: Double spaces

libkbfs/folder_branch_ops.go
```diff
@@ -3081,18 +3104,19 @@ func (fbo *folderBranchOps) renameLocked(
 	}
 
 	// does name exist?
-	if de, ok := newPBlock.Children[newName]; ok {
+	replacedDe, ok := newPBlock.Children[newName]
+	if ok {
```
songgao: `replacedDe` and `ok` are only used with in the `if` block, so it seems it's better to have the assignment within `if` line like before, so the two variables are not left around?

libkbfs/folder_branch_ops.go
strib: Intentional.  That's how I roll!

libkbfs/folder_branch_ops.go
strib: Ah, that's because of a future PR (#941) coming later that uses `replacedDe` after that block.  If you feel strongly that we should keep master clean of stuff like this, I'd be happy to fix it here and change it later in that PR, if you want.

songgao: Ah no need to change if it's used later. Thanks for the context!

PR #941, opened 2017-05-02
libkbfs: fix deCache to update timestamps better, and store unlinked DirEntrys in the NodeCache

strib:
This PR does two somewhat related things that are needed for the rest of KBFS-2076:

1. It updates the parent mtime/ctime on creates and removes, and stores those changes in the `deCache`.  These changes need to be applied when looking up the directory in the future.  But we might not want to replace the entire `DirEntry` because our cached copy might not have a proper `BlockPointer` or other important data, so update carefully.
2. When a file is unlinked, there might still be a handle lying around for it in some application.  That handle still has a right to the node's DirEntry.  Previously we were storing it in the `deCache`, so a) it would only survive until the next `SyncAll`, even though the handle might live past that, and b) when processing directory adds, we might end up using the new DirEntry for an entry that was renamed over the one we had the handle for.  Instead, this stores the DirEntry alongside the `Node` object itself that's associated with the handle, so we can't lose track of it before the handle itself goes away.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -131,9 +131,9 @@ type deCacheEntry struct {
 	// been added to the DirBlock for the BlockPointer that maps to
 	// this struct.
 	adds map[string]BlockPointer
-	// dels are the DirEntrys that have been removed from the DirBlock
-	// for the BlockPointer that maps to this struct, index by name.
-	dels map[string]DirEntry
+	// dels are the name that have been removed from the DirBlock
+	// for the BlockPointer that maps to this struct.
+	dels map[string]bool
 }
 
 type deferredState struct {
@@ -868,9 +868,9 @@ func (fbo *folderBlockOps) removeDirEntryInCacheLocked(lState *lockState,
 	fbo.blockLock.AssertLocked(lState)
 	cacheEntry := fbo.deCache[dir.tailPointer().Ref()]
 	if cacheEntry.dels == nil {
-		cacheEntry.dels = make(map[string]DirEntry)
+		cacheEntry.dels = make(map[string]bool)
 	}
-	cacheEntry.dels[oldName] = oldDe
+	cacheEntry.dels[oldName] = true
 	// In case it was added in the cache but not flushed yet.
 	delete(cacheEntry.adds, oldName)
 
@@ -890,6 +890,8 @@ func (fbo *folderBlockOps) RemoveDirEntryInCache(lState *lockState, dir path,
 	fbo.blockLock.Lock(lState)
 	defer fbo.blockLock.Unlock(lState)
 	fbo.removeDirEntryInCacheLocked(lState, dir, oldName, oldDe)
+	_ = fbo.nodeCache.Unlink(
+		oldDe.Ref(), dir.ChildPath(oldName, oldDe.BlockPointer), oldDe)
 }
 
 // RenameDirEntryInCache updates the entries of both the old and new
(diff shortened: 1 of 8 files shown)


strib:
Ok, this is ready to review.  I didn't end up changing it; instead I made #944 to fix the issues that this PR exposed.

libkbfs/folder_block_ops.go
```diff
@@ -855,7 +864,7 @@ func (fbo *folderBlockOps) AddDirEntryInCache(lState *lockState, dir path,
 }
 
 func (fbo *folderBlockOps) removeDirEntryInCacheLocked(lState *lockState,
-	dir path, oldName string) {
+	dir path, oldName string, oldDe DirEntry) {
```
strib: Oh sorry, it's not used until #951.  Sorry, bad job of splitting things up.  Let me know if you want me to move its introduction until then.

libkbfs/folder_block_ops.go
```diff
@@ -1001,6 +1027,37 @@ func (fbo *folderBlockOps) ClearCachedDirEntry(lState *lockState, dir path) {
...
+		refDe.dirEntry.Ctime == 0 {
+		return false, de
+	}
+
+	if refDe.dirEntry.IsInitialized() {
+		return true, refDe.dirEntry
+	}
+
+	// Just update the times.
+	if refDe.dirEntry.Mtime > 0 {
```
strib: In `addDirEntryInCacheLocked()`, we update the `Ctime` and `Mtime` of the parent directory, but we don't necessarily have the full `DirEntry` for it (which would need to be fetched from the grandparent block).

libkbfs/folder_branch_ops.go
```diff
@@ -4074,53 +4069,92 @@ func (fbo *folderBranchOps) searchForNode(ctx context.Context,
...
-		// It's ok to modify this path, since we break as soon as the
-		// node cache takes a reference to it.
-		unlinkPath.path[len(unlinkPath.path)-1].BlockPointer = ptr
-		found := fbo.nodeCache.Unlink(ptr.Ref(), unlinkPath)
-		if found {
-			break
-		}
-	}
+func (fbo *folderBranchOps) unlinkFromCache(
+	op op, unlinkPath path, unlinkDe DirEntry) error {
```
strib: Oops, I think I do that in a later PR, but it's easy enough to do it here, so will do.

libkbfs/folder_branch_ops.go
```diff
@@ -4074,53 +4069,92 @@ func (fbo *folderBranchOps) searchForNode(ctx context.Context,
...
 	if err != nil {
-		return path{}, false, err
+		return path{}, DirEntry{}, false, err
 	}
 
-	return p.ChildPathNoPtr(childName), true, nil
+	// If the first op in this MD update is a resolutionOp, we need to
+	// inspect it to look for the *real* original pointer for this
+	// node.
+	if resOp, ok := md.data.Changes.Ops[0].(*resolutionOp); ok {
```
strib: Yeah, we can assume that `op` came from `md`, so we know at least there's one.  But I can add a defensive check to make sure.

libkbfs/folder_branch_ops.go
```diff
@@ -4074,53 +4069,92 @@ func (fbo *folderBranchOps) searchForNode(ctx context.Context,
...
+		fbo.log.CDebugf(ctx,
+			"Ignoring unlink when resolutionOp never fixed up %v",
+			p.tailPointer())
+		return path{}, DirEntry{}, false, nil
+	}
+
+	// If the original (clean) parent block is already GC'd from the
+	// server, this might not work, but hopefully we'd be
+	// fast-forwarding in that case anyway.
+	dblock, err := fbo.blocks.GetDir(ctx, lState, md, p, blockRead)
```
strib: `folderBlockOps.GetDir()` checks the memory cache before going out to the disk cache / server.  But we don't have it anywhere locally in the function yet, if that's what you're asking.

libkbfs/folder_block_ops.go
```diff
@@ -1001,6 +1027,37 @@ func (fbo *folderBlockOps) ClearCachedDirEntry(lState *lockState, dir path) {
 	// in there as well.)
 }
 
+func (fbo *folderBlockOps) updateDirtyEntryLocked(
```
strib: Done.

libkbfs/folder_block_ops.go
jzila: Nope that's fine, it can stay here. I was just confused.

jzila:
Thanks for the responses strib! Looks good.

PR #942, opened 2017-05-03
tests: add `SyncAll` calls to a bunch of tests that modify directories

strib:
This will be needed once we turn on the feature in a future PR.  Note that this PR doesn't update *every* test, especially the complicated ones in `kbfs_cr_test.go` and `kbfs_ops_concur_test.go`, because those will require a bit more work than just a simple call to `SyncAll`.


strib:
The test failure here is due to its parent PR #941.  This PR is still worth reviewing as is.

PR #947, opened 2017-05-04
node_cache: add a way to undo unlinks and moves

strib:
Issue: KBFS-2076

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -1689,7 +1689,8 @@ type NodeCache interface {
 	// called to undo the effect of the move (or `nil` if nothing
 	// needs to be done); if newParent cannot be found, it returns an
 	// error.
-	Move(ref BlockRef, newParent Node, newName string) (func(), error)
+	Move(ref BlockRef, newParent Node, newName string) (
+		undoFn func(), err error)
 	// Unlink set the corresponding node's parent to nil and caches
 	// the provided path in case the node is still open. NodeCache
 	// ignores the call when ptr is not cached.  The path is required
@@ -1697,7 +1698,7 @@ type NodeCache interface {
 	// already that shouldn't be reflected in the cached path.  It
 	// returns a function that can be called to undo the effect of the
 	// unlink (or `nil` if nothing needs to be done).
-	Unlink(ref BlockRef, oldPath path, oldDe DirEntry) func()
+	Unlink(ref BlockRef, oldPath path, oldDe DirEntry) (undoFn func())
 	// IsUnlinked returns whether `Unlink` has been called for the
 	// reference behind this node.
 	IsUnlinked(node Node) bool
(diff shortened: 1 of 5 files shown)


libkbfs/interfaces.go
```diff
@@ -1683,17 +1683,20 @@ type NodeCache interface {
...
+	// found, it returns an error.
+	Move(ref BlockRef, newParent Node, newName string) (func(), error)
 	// Unlink set the corresponding node's parent to nil and caches
 	// the provided path in case the node is still open. NodeCache
 	// ignores the call when ptr is not cached.  The path is required
 	// because the caller may have made changes to the parent nodes
 	// already that shouldn't be reflected in the cached path.
-	// Returns whether a node was actually updated.
-	Unlink(ref BlockRef, oldPath path, oldDe DirEntry) bool
+	// Returns whether a node was actually updated.  It returns a
```
jzila: No longer true, it only returns a function.

libkbfs/node_cache.go
```diff
@@ -193,31 +193,38 @@ func (ncs *nodeCacheStandard) Move(
...
+	oldName := entry.core.pathNode.Name
+
 	entry.core.parent = newParentNS
 	entry.core.pathNode.Name = newName
-	return nil
+
+	return func() {
+		entry.core.parent = oldParent
+		entry.core.pathNode.Name = oldName
+	}, nil
```
jzila: This is a pretty clever approach. Nice!

libkbfs/node_cache_test.go
```diff
@@ -229,6 +229,12 @@ func TestNodeCacheMoveSuccess(t *testing.T) {
 	if ncs.nodes[childPtr2.Ref()].refCount != 1 {
 		t.Errorf("Child1 has wrong refcount: %d", ncs.nodes[childPtr2.Ref()].refCount)
 	}
+
+	undoMove()
+	if childNode2.GetBasename() != "child2" {
```
jzila: I guess it's fine not to check the parent?

strib: I was just being lazy.  Added a check.

libkbfs/interfaces.go
```diff
@@ -1683,17 +1683,21 @@ type NodeCache interface {
...
-	// If newParent is nil, it treats the ptr's corresponding node as
-	// being unlinked from the old parent completely.
-	Move(ref BlockRef, newParent Node, newName string) error
+	// is not cached.  If newParent is nil, it treats the ptr's
+	// corresponding node as being unlinked from the old parent
+	// completely. If successful, it returns a function that can be
+	// called to undo the effect of the move (or `nil` if nothing
+	// needs to be done); if newParent cannot be found, it returns an
+	// error.
+	Move(ref BlockRef, newParent Node, newName string) (func(), error)
```
songgao: Would you mind write out the name of returned variable e.g. `(undo func(), err error)`? It'd make autocompletion plugins show the meaning of them. Same for `Unlink` below, and their implementations if it's not too much to change :)

strib: @songgao: sure.  Everywhere, or just here in the interface definition?

songgao: Everywhere would be great! The reason is sometimes it's called using the actual type, and the completer would pull the concrete type's method signature

PR #948, opened 2017-05-04
init: feature flag directory op batching

strib:
Until the admins test it out a bit.

Issue: KBFS-2076

### DIFF
--- a/libkbfs/init.go
+++ b/libkbfs/init.go
@@ -645,10 +645,7 @@ func doInit(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn,
 	}
 
 	session, err := config.KBPKI().GetCurrentSession(context.Background())
-	if err != nil {
-		return nil, err
-	}
-	if adminFeatureList[session.UID] {
+	if err == nil && adminFeatureList[session.UID] {
 		log.Debug("Enabling a dir op batch size of %d",
 			params.BGFlushDirOpBatchSize)
 		config.SetBGFlushDirOpBatchSize(params.BGFlushDirOpBatchSize)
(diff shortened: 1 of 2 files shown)


libkbfs/keybase_service_util.go
```diff
@@ -62,6 +62,19 @@ func serviceLoggedIn(ctx context.Context, config Config, name string,
 		}
 	}
 
+	if adminFeatureList[session.UID] {
+		log.CDebugf(ctx, "Enabling a dir op batch size of %d",
+			bgFlushDirOpBatchSizeDefault)
+		// NOTE: This overrides any command-line parameter.  It only
+		// matters until we un-feature-flag this, I think it's ok for
+		// now.
+		config.SetBGFlushDirOpBatchSize(bgFlushDirOpBatchSizeDefault)
```
songgao: Does this mean all admins get the default batch size (which is 20) no matter the value of `-sync-batch-size`? It seems the command line flag doesn't do anything then does it?

strib: Only when an admin logs in _after_ KBFS initializes.  I know that's not ideal, but it's temporary until this is default for everyone.  Open to other ideas -- I just didn't want to plumb an extra config field here for the original value of the command line flag.

songgao: Ah OK! that makes sense. Thanks

libkbfs/init.go
```diff
@@ -235,7 +235,8 @@ func AddFlags(flags *flag.FlagSet, ctx Context) *InitParams {
 	flags.IntVar((*int)(&params.BGFlushDirOpBatchSize), "sync-batch-size",
 		int(defaultParams.BGFlushDirOpBatchSize),
 		"The number of unflushed directory operations in a TLF that will "+
-			"trigger an immediate data sync.")
+			"trigger an immediate data sync.  EXPERIMENTAL, only "+
+			"available for Keybase admins")
```
songgao: Perhaps mentioned here in the help doc that the parameter gets reverted to default if logged out and logged back in again?

strib: Done, thanks for the suggestion!

PR #949, opened 2017-05-04
folder_branch_ops: save old block pointers in the blockPutState

strib:
Track the pre-ready block pointers in the blockPutState, so that in a
future commit SyncAll can figure out how to call `UpdatePointers` for
blocks that haven't yet been synced (and were given random
BlockPointers when they were created).

Issue: KBFS-2076

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1899,6 +1899,8 @@ func (bps *blockPutState) addNewBlock(
 		blockState{blockPtr, block, readyBlockData, syncedCb, zeroPtr})
 }
 
+// saveOldPtr stores the given BlockPointer as the old (pre-readied)
+// pointer for the most recent blockState.
 func (bps *blockPutState) saveOldPtr(oldPtr BlockPointer) {
 	bps.blockStates[len(bps.blockStates)-1].oldPtr = oldPtr
 }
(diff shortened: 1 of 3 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -1891,10 +1892,15 @@ func newBlockPutState(length int) *blockPutState {
...
+func (bps *blockPutState) addNewBlock(
+	blockPtr BlockPointer, block Block,
 	readyBlockData ReadyBlockData, syncedCb func() error) {
 	bps.blockStates = append(bps.blockStates,
-		blockState{blockPtr, block, readyBlockData, syncedCb})
+		blockState{blockPtr, block, readyBlockData, syncedCb, zeroPtr})
+}
+
+func (bps *blockPutState) saveOldPtr(oldPtr BlockPointer) {
+	bps.blockStates[len(bps.blockStates)-1].oldPtr = oldPtr
```
songgao: Why is it always in the last `blockState`? Is it because in the future this will used *while* building out the `blockPutState.blockStates`?

strib: Yeah, that's right.  Sorry for the confusion, I'll add a function comment.

songgao: Thanks!

PR #951, opened 2017-05-04
folder_block_ops: fixes need for dir op batching

strib:
* Make cache/status changes undo-able
* Use the undo functions in folderBranchOps.
* Provide more control over whether to apply cached attributes when
  getting directory entries or not.

Issue: KBFS-2076

PR #952, opened 2017-05-04
folder_branch_ops: enable dir op batching

strib:
This is the final PR for KBFS-2076.  Sorry I couldn't make it any smaller.  It does everything necessary to enable dir op batching as long as the batch size > 1 (right now, that's the case only for admin users).  Note that when the batch size == 1, it's not exactly the same behavior as the current releases: it syncs uses `SyncAll`, rather than the old method of finalizing writes, which leads to slightly different lists of ops.  But at least ops aren't buffered in memory, and only one op at a time goes in the MD.

One other thing to note: this holds `mdWriterLock` throughout the entire sync process.  That means, while batching is happening, other incoming directory updates are blocked.  So if the batch sizes are big, the user might start noticing periodic pauses as blocks get written to the journal during the sync.  We should be able to unlock the next level of performance by relaxing this requirement, but it seems pretty risky and I didn't want to pack it into this work.  But if we ever need another performance, we should look into it.

Issue: KBFS-2076


strib:
I'll try to go through this and annotate some interesting things in the morning.

libkbfs/folder_branch_ops.go
```diff
@@ -2117,10 +2067,30 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
...
+			// then the next call will get an
+			// UnmergedSelfConflictError but fail to find any new
+			// updates and fail the operation, but things will get
+			// fixed up once conflict resolution finally completes.
+			//
+			// TODO: how confused will the kernel cache get if the
+			// pointers are updated but the file system operation
+			// still gets an error returned by the wrapper function
+			// that calls us (in the event of a user cancellation).
+			fbo.log.CInfof(ctx, "Ignoring a PutUnmerged error: %+v", err)
```
strib: This is a pretty nasty hack.  Hopefully the comment above explains it well.  I'm open to other suggestions but couldn't think of anything good.  I thought of checking with the server to see if the put really made it, but the most likely time for this to happen is during network connectivity issues (when journaling is off), and so checking with the server won't work...

libkbfs/folder_branch_ops.go
```diff
@@ -2440,10 +2380,29 @@ func (fbo *folderBranchOps) canonicalPath(ctx context.Context, dir Node, name st
...
+	default:
+	}
+	// A local write always means any ongoing CR should be cancel,
+	// because the set of unmerged writes has changed.
+	fbo.cr.ForceCancel()
+}
+
+func (fbo *folderBranchOps) syncDirUpdateOrSignal(
+	ctx context.Context, lState *lockState) error {
+	if fbo.config.BGFlushDirOpBatchSize() == 1 {
```
strib: This is where the feature-flag magic happens.

libkbfs/folder_branch_ops.go
```diff
@@ -2117,10 +2067,30 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
...
+			// then the next call will get an
+			// UnmergedSelfConflictError but fail to find any new
+			// updates and fail the operation, but things will get
+			// fixed up once conflict resolution finally completes.
+			//
+			// TODO: how confused will the kernel cache get if the
+			// pointers are updated but the file system operation
+			// still gets an error returned by the wrapper function
+			// that calls us (in the event of a user cancellation)?
+			fbo.log.CInfof(ctx, "Ignoring a PutUnmerged error: %+v", err)
```
strib: This is a pretty nasty hack. Hopefully the comment above explains it well. I'm open to other suggestions but couldn't think of anything good. I thought of checking with the server to see if the put really made it, but the most likely time for this to happen is during network connectivity issues (when journaling is off), and so checking with the server won't work...

libkbfs/folder_branch_ops.go
```diff
@@ -2522,13 +2483,22 @@ func (fbo *folderBranchOps) createEntryLocked(
 		Context: kbfsblock.MakeFirstContext(
 			session.UID, keybase1.BlockType_DATA),
 	}
+	co.AddRefBlock(newPtr)
+	co.AddUpdate(parentPtr, parentPtr)
```
strib: The "updates" between the same pointers are just placeholders that will be fixed up by the folder update prep process in `syncAllLocked`.  It tells that process what blocks to ready, and the final pointer gets replaced with the new ID at the end of the process.

libkbfs/folder_branch_ops.go
```diff
@@ -3787,10 +3741,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+
+	newBlocks := make(map[BlockPointer]bool)
+	fileBlocks := make(fileBlockMap)
+	parentsToAddChainsFor := make(map[BlockPointer]bool)
+	for _, dop := range fbo.dirOps {
+		// Copy the op before modifying it, in case there's an error
+		// and we have to retry with the original ops.  TODO: make a
+		// better way of copying a single op.
+		ops := make(opsList, 1)
+		err := kbfscodec.Update(
```
strib: If anyone prefers, I can add a `DeepCopy` method to the `op` interface, to make this cleaner (and probably more performant).

libkbfs/folder_branch_ops.go
```diff
@@ -3787,10 +3741,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+				}
+				lbc[newPath.tailPointer()] = dblock
+				resolvedPaths[newPath.tailPointer()] = newPath
+			}
+
+			if len(dblock.Children) > 0 {
+				continue
+			}
+
+			// If the directory is empty, we need to explicitly clean
```
strib: If the directory *isn't* empty, it's guaranteed that we'll process other createOps for that directory, and we'll set it to be cleaned up at that point.

libkbfs/folder_branch_ops.go
```diff
@@ -4027,39 +4100,14 @@ func (fbo *folderBranchOps) UnregisterFromChanges(obs Observer) error {
...
-		// resolution op with all the new pointer updates.  It's the
-		// one that gets the afterUpdateFn.
-		if md.data.Changes.Ops[0].(*resolutionOp) == nil {
-			return errors.Errorf("First update in local batch is not a "+
-				"resolutionOp: %T", md.data.Changes.Ops[0])
-		}
-		err := fbo.notifyOneOpLocked(
-			ctx, lState, md.data.Changes.Ops[0], md, false, afterUpdateFn)
+	for _, op := range md.data.Changes.Ops {
+		err := fbo.notifyOneOpLocked(ctx, lState, op, md.ReadOnly(), false)
```
strib: Since `syncAllLocked` calls `UpdatePointers` directly, without going through `notifyBatchLocked`, there's no longer any need for complex logic here around the `afterUpdateFn`.

libkbfs/kbfs_cr_test.go
```diff
@@ -867,13 +901,15 @@ func TestCRDouble(t *testing.T) {
...
-		syncCtx, config2, StallableMDPutUnmerged, 1)
+		syncCtx, config2, StallableMDAfterPutUnmerged, 1)
 
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
 		err = kbfsOps2.SyncAll(syncCtx, fileNodeC.GetFolderBranch())
-		assert.Equal(t, context.Canceled, err)
+		if err != nil {
+			assert.Equal(t, context.Canceled, err)
```
strib: Since `SyncAll` forks a goroutine and then waits on both the goroutine and `ctx.Done()`, there's a race on cancel -- sometimes it will return `nil` if the goroutine finishes first (and ignores the `PutUnmerged` error), and sometimes it will return `context.Canceled`.  Either is ok.

strib:
Numbers with the final version (debugging turned off), on my SSD laptop.  `bigfile` is a 128MB file.

With this PR:
```
$ time cp -r /tmp/go go6 && time cp /tmp/bigfile .

real	0m19.734s
user	0m0.059s
sys	0m0.386s

real	0m2.492s
user	0m0.002s
sys	0m0.070s
```

Master:
```
$ time cp -r /tmp/go go7 && time cp /tmp/bigfile .

real	1m46.190s
user	0m0.091s
sys	0m0.423s

real	0m3.001s
user	0m0.000s
sys	0m0.075s
```

libkbfs/kbfs_cr_test.go
songgao: It might be worth moving this GH comment into a code comment?

test/engine_libkbfs.go
```diff
@@ -52,6 +52,9 @@ func (k *LibKBFS) InitTest(ver libkbfs.MetadataVer,
 	config.SetMetadataVersion(ver)
 
 	setBlockSizes(k.tb, config, blockSize, blockChangeSize)
+	if batchSize > 0 {
+		config.SetBGFlushDirOpBatchSize(batchSize)
+	}
```
songgao: Do we not need to worry about `BGFlushPeriod` here because of the fake clock?

strib:
@songgao for the win!

libkbfs/folder_block_ops.go
```diff
@@ -986,8 +986,13 @@ func (fbo *folderBlockOps) RemoveDirEntryInCache(lState *lockState, dir path,
 	fbo.blockLock.Lock(lState)
 	defer fbo.blockLock.Unlock(lState)
 	undoFn := fbo.removeDirEntryInCacheLocked(lState, dir, oldName, oldDe)
+	unlinkUndoFn := fbo.nodeCache.Unlink(
+		oldDe.Ref(), dir.ChildPath(oldName, oldDe.BlockPointer), oldDe)
 	return fbo.wrapWithBlockLock(func() {
 		undoFn()
+		if unlinkUndoFn != nil {
```
strib: You're right, I fixed that up as per @jzila's comments in #951.  Will be fixed here next time I push.

libkbfs/folder_block_ops.go
```diff
@@ -1002,17 +1007,31 @@ func (fbo *folderBlockOps) RemoveDirEntryInCache(lState *lockState, dir path,
...
 	oldParent path, oldName string, newParent path, newName string,
-	newDe DirEntry, replacedDe DirEntry) dirCacheUndoFn {
+	newDe DirEntry, replacedDe DirEntry) (dirCacheUndoFn, error) {
 	fbo.blockLock.Lock(lState)
 	defer fbo.blockLock.Unlock(lState)
 	if newParent.tailPointer() == oldParent.tailPointer() &&
 		oldName == newName {
 		// Noop
-		return fbo.wrapWithBlockLock(func() {})
+		return fbo.wrapWithBlockLock(func() {}), nil
```
strib: I'll be changing this to just return `nil` in the future, and checking it at the caller.

libkbfs/folder_branch_ops.go
strib: The code in this block is executed when the error _isn't_ `UnmergedSelfConflictError` (it's in an `else` clause).  The kind of errors I'm worried about here are `context.Canceled` and `context.DeadlineExceeded`.  We'll mostly be getting those when the network is being slow or broken.

libkbfs/folder_branch_ops.go
strib: Sure, will do.

libkbfs/kbfs_cr_test.go
strib: Yeah good call.  I have a comment at the other place this happens, but I'll copy it here too.

test/engine_libkbfs.go
strib: Background flushing is disabled entirely by default in these DSL tests, so the flush period has no effect.  We do that to make it easier to have more control about when things flush.

Really the only reason to set this parameter is to go into `non-admin` mode and force a `SyncAll` during every directory operation, rather than waiting until the end of the `as` block.

songgao: Make sense; thanks!

libkbfs/folder_branch_ops.go
```diff
@@ -3179,101 +3207,24 @@ func (fbo *folderBranchOps) renameLocked(
...
 		}
 	}()
 
-	_, err = doBlockPuts(ctx, fbo.config.BlockServer(), fbo.config.BlockCache(),
-		fbo.config.Reporter(), fbo.log, md.TlfID(),
-		md.GetTlfHandle().GetCanonicalName(), *newBps)
+	// It's safe to notify before we've synced, since it is only
+	// sending invalidation notifications.  At worst the upper layer
+	// will just have to refresh its cache needlessly.
+	err = fbo.notifyOneOp(ctx, lState, ro, md.ReadOnly(), false)
```
jzila: The above defer through to this `notifyOneOp` is repeated at least 3-4 times (except for the specific undo function that's called). Is there a way we could factor it out a bit?

libkbfs/folder_branch_ops.go
```diff
@@ -3789,10 +3743,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
 		var ref BlockRef
-		switch realOp := op.dirOp.(type) {
+		switch realOp := newOp.(type) {
+		case *createOp:
+			if realOp.Type == Sym {
+				continue
+			}
+
+			// New files and directories need explicitly
+			// pointer-updating, because the sync process will turn
```
jzila: `explicitly need` or `need explicit`

libkbfs/folder_branch_ops.go
```diff
@@ -3789,10 +3743,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+			// them into simple refs and will forget about the local,
+			// temporary ID.
+			newNode := dop.nodes[1]
+			newPath := fbo.nodeCache.PathFromNode(newNode)
+			newBlocks[newPath.tailPointer()] = true
```
jzila: please make a `newPointer` as well and use that below instead of `newPath.tailPointer()`.

libkbfs/folder_branch_ops.go
jzila: Are we guaranteed about the order though? Or does the order not matter?

libkbfs/folder_branch_ops.go
```diff
@@ -3801,21 +3830,39 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+				}
+			}
 		}()
 	}
 
-	fileBlocks := make(fileBlockMap)
 	var blocksToRemove []BlockPointer
 	var afterUpdateFns []func() error
 
+	afterUpdateFns = append(afterUpdateFns, func() error {
```
jzila: This seems like it's going to do a lot of dynamic dispatch of closures. Not sure of a better way right now though. Maybe a `TODO` to revisit?

libkbfs/folder_branch_ops.go
```diff
@@ -3789,10 +3743,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+			if _, ok := newOp.(*setAttrOp); ok {
+				// For a setattr, the node is the file, but that
+				// doesn't get updated, so use the current parent
+				// node.
+				p = *p.parentPath()
+			}
+
+			for i, pn := range p.path {
+				if i == len(p.path)-1 {
+					newOp.AddUpdate(pn.BlockPointer, pn.BlockPointer)
```
jzila: We could use the previously suggested `AddSelfUpdate` here.

libkbfs/folder_branch_ops.go
```diff
@@ -3876,13 +3923,16 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+		// like updates during the prepping.
 		lastOp := md.Data().Changes.Ops[len(md.Data().Changes.Ops)-1]
-		for _, p := range file.path {
-			lastOp.AddUpdate(p.BlockPointer, p.BlockPointer)
+		for i, pn := range file.path {
+			if i == len(file.path)-1 {
+				lastOp.AddUpdate(pn.BlockPointer, pn.BlockPointer)
+			} else {
+				parentsToAddChainsFor[pn.BlockPointer] = true
+			}
```
jzila: This block is the same as https://github.com/keybase/kbfs/pull/952/files#diff-f6470cb87de71869397b1e97dd1ee43dR3776. Helper function?

libkbfs/kbfs_ops_test.go
```diff
@@ -2952,35 +2952,30 @@ func TestKBFSOpsBackgroundFlush(t *testing.T) {
...
-		t.Errorf("Got error on write: %+v", err)
-	}
+	ops := getOps(config, rootNode.GetFolderBranch().Tlf)
+	oldPtr := ops.nodeCache.PathFromNode(nodeA).tailPointer()
 
-	c := make(chan struct{})
-	observer := &testBGObserver{c}
-	config.Notifier().RegisterForChanges(
-		[]FolderBranch{rootNode.GetFolderBranch()}, observer)
+	staller := NewNaïveStaller(config)
```
songgao: I added that to experiment and the plan was keep it until somebody complains. So I'm OK with changing it to `NaiveStaller`. :-) I'll make the PR later if @strib you don't want to deal with it in this PR

libkbfs/folder_branch_ops.go
strib: You mean the order that we process createOps in?  That doesn't matter, because we're checking for emptiness by looking up the dirty block itself, which will have children filled in for any createOps that happened during this batch.  So we'll know that a createOp is coming at some point in this batch.

libkbfs/folder_branch_ops.go
strib: Added a `AddSelfUpdate` helper.

libkbfs/folder_block_ops.go
```diff
@@ -1002,17 +1007,31 @@ func (fbo *folderBlockOps) RemoveDirEntryInCache(lState *lockState, dir path,
...
 	undoAdd := fbo.addDirEntryInCacheLocked(lState, newParent, newName, newDe)
 	undoRm := fbo.removeDirEntryInCacheLocked(
 		lState, oldParent, oldName, DirEntry{})
+	undoUnlink := fbo.nodeCache.Unlink(
+		replacedDe.Ref(), newParent.ChildPath(newName, replacedDe.BlockPointer),
+		replacedDe)
+
+	newParentNode := fbo.nodeCache.Get(newParent.tailPointer().Ref())
+	undoMove, err := fbo.nodeCache.Move(newDe.Ref(), newParentNode, newName)
+	if err != nil {
```
strib: I clarified that in the interface comment for `Move`.

libkbfs/kbfs_ops_test.go
strib: I'm not going to touch that rename as part of this PR, and am neutral about whether it gets done in a future PR or not.

libkbfs/kbfs_ops_test.go
```diff
@@ -3372,347 +3367,3 @@ func TestForceFastForwardOnEmptyTLF(t *testing.T) {
...
-	defer CheckConfigAndShutdown(ctx, t, config2)
-
-	rootNodeBob := GetRootNodeOrBust(ctx, t, config2, "alice,bob", false)
-	checkChild(config2, rootNodeBob)
-
-	if len(ops.blocks.deCache) > 0 {
-		t.Fatalf("%d unexpected deCache entries leftover",
-			len(ops.blocks.deCache))
-	}
-}
```
strib: Yes, that's why they're being removed.  Coverage went up with this PR:

Master:
```
libkbfs$ go test -coverPASS
coverage: 66.4% of statements
ok  	github.com/keybase/kbfs/libkbfs	7.832s
```

This PR:
```
libkbfs$ go test -cover
PASS
coverage: 66.7% of statements
ok  	github.com/keybase/kbfs/libkbfs	7.848s
```

libkbfs/folder_branch_ops.go
```diff
@@ -3789,10 +3743,85 @@ func (fbo *folderBranchOps) syncAllLocked(
...
+	newBlocks := make(map[BlockPointer]bool)
+	fileBlocks := make(fileBlockMap)
+	parentsToAddChainsFor := make(map[BlockPointer]bool)
+	for _, dop := range fbo.dirOps {
+		// Copy the op before modifying it, in case there's an error
+		// and we have to retry with the original ops.  TODO: make a
+		// better way of copying a single op.
+		ops := make(opsList, 1)
+		err = kbfscodec.Update(
+			fbo.config.Codec(), &ops, opsList([]op{dop.dirOp}))
```
jzila: And by extension on the `op` interface as suggested above.

libkbfs/folder_branch_ops.go
```diff
@@ -2119,10 +2069,30 @@ func (fbo *folderBranchOps) finalizeMDWriteLocked(ctx context.Context,
...
+			// remain cached locally and will be re-tried
+			// (non-idempotently) on the next sync call.  This should
+			// be a very rare situation when journaling is enabled, so
+			// instead let's pretend it succeeded so that the cached
+			// data is cleared and the nodeCache is updated.  If we're
+			// wrong, and the update didn't make it to the server,
+			// then the next call will get an
+			// UnmergedSelfConflictError but fail to find any new
+			// updates and fail the operation, but things will get
+			// fixed up once conflict resolution finally completes.
```
strib: Ok, I've converted `TestUnmergedPutAfterCanceledUnmergedPut` to test this.

libkbfs/folder_branch_ops.go
strib: Done.

libkbfs/folder_branch_ops.go
strib: Ok, I added it to the `op` interface, which means we won't need it on `opsList` since only one op is being copied at a time here.

PR #953, opened 2017-05-04
libdokan: don't report GetDiskFreeSpace errors

strib:
This function is usually called by the OS/desktop, not the user, so it
doesn't make sense to show an error notification (usually a "read
timeout -- send your logs" type of message) for this.

Issue: KBFS-2103

### DIFF
--- a/libdokan/fs.go
+++ b/libdokan/fs.go
@@ -164,7 +164,6 @@ func (f *FS) GetDiskFreeSpace(ctx context.Context) (freeSpace dokan.FreeSpace, e
 	_, usageBytes, limitBytes, err := f.quotaUsage.Get(
 		ctx, quotaUsageStaleTolerance/2, quotaUsageStaleTolerance)
 	if err != nil {
-		f.log.CDebugf(ctx, "Getting quota usage error: %v", err)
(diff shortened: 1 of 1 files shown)


libdokan/fs.go
```diff
@@ -151,7 +151,16 @@ func (f *FS) GetDiskFreeSpace(ctx context.Context) (freeSpace dokan.FreeSpace, e
...
 	}
-	defer func() { f.reportErr(ctx, libkbfs.ReadMode, err) }()
+	defer func() {
+		if err == nil {
+			f.log.CDebugf(ctx, "Request complete")
+		} else {
+			// Don't report the error (perhaps resulting in a user
+			// notification) since this method is mostly called by the
+			// OS and not because of a user action.
+			f.log.CDebugf(ctx, err.Error())
```
songgao: The error is already logged [below](https://github.com/keybase/kbfs/pull/953/files#diff-3c07e38475555636a646b747704ace53R167). Perhaps remove one of them? But the previous logs twice too, so I don't feel very strong about this

strib: Ah good catch, I'll remove the one below to be more consistent with other calls.

PR #954, opened 2017-05-05
folder_branch_ops: in SyncAll, don't clean up file blocks

strib:
They're already handled by `folderBlockOps.CleanupSyncState()` (called
via the `cleanup` function returned by `startSyncLocked()`), and some
of those blocks might be re-used by subsequent retries.  So remove
them from the state passed to
`folderBlockManager.cleanUpBlockState()`.

Issue: KBFS-2145

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1907,6 +1907,34 @@ func (bps *blockPutState) mergeOtherBps(other *blockPutState) {
 	bps.blockStates = append(bps.blockStates, other.blockStates...)
 }
 
+func (bps *blockPutState) removeOtherBps(other *blockPutState) {
+	if len(other.blockStates) == 0 {
+		return
+	}
+
+	otherPtrs := make(map[BlockPointer]bool, len(other.blockStates))
+	for _, bs := range other.blockStates {
+		otherPtrs[bs.blockPtr] = true
+	}
+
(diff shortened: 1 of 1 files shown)


strib:
> Could this leave blocks orphaned?

It shouldn't -- we have several tests checking that actually.  All of the blocks removed from `bps` will be dealt with by the `folderBlockOps.CleanupSyncState()` calls in the same way they've always been, even before all the `SyncAll` stuff was added.

libkbfs/folder_branch_ops.go
```diff
@@ -1907,6 +1907,34 @@ func (bps *blockPutState) mergeOtherBps(other *blockPutState) {
...
+	otherPtrs := make(map[BlockPointer]bool, len(other.blockStates))
+	for _, bs := range other.blockStates {
+		otherPtrs[bs.blockPtr] = true
+	}
+
+	// Assume that `other` is a subset of `bps` when initializing the
+	// slice length.
+	newLen := len(bps.blockStates) - len(other.blockStates)
+	if newLen <= 0 {
+		newLen = 1
```
songgao: Although at the beginning we return 0 if `len(other.blockStates) == 0`, it's still possible to have an empty `newBlockStates` later right? If `cap` matters why not use `0` here?

songgao:
Whoa this PR's old; not sure why it just showed up in my "Review requests" tab!

PR #959, opened 2017-05-09
folder_block_ops: use Nodes for reading, not paths

strib:
Since the caller would have gotten the path without holding
`blockLock`, which means it could have changed between then and now.

Issue: KBFS-2155

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -3280,12 +3280,12 @@ func (fbo *folderBranchOps) Read(
 		return 0, err
 	}
 
-	filePath, err := fbo.pathFromNodeForRead(file)
-	if err != nil {
-		return 0, err
-	}
-
 	{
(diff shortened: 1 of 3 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -3320,8 +3320,10 @@ func (fbo *folderBranchOps) Read(
 			return err
 		}
 
+		// Read using the `file` Node, not `filePath`, since the path
+		// could change until we take `blockLock` for reading.
```
jzila: Doesn't this make the `git` special case above susceptible to the same issue?

strib: Nah -- there we're just checking the path names, not the BlockPointers.  I suppose if someone renames the .git directory to something else right in the middle of a read, then the special git behavior may or may not be enabled, but that seems ok.  As long as the right file contents is still read.

PR #961, opened 2017-05-09
folder_block_ops: can't cancel a write due to a sync context err

strib:
If a write is blocked on waiting for the dirty block cache to release
more bytes that can be dirtied, we have a mechanism where we can error
that write if something fatal and non-recoverable happens to an
existing sync that's currently draining the dirty bytes.  For example,
the user fills up their quota, and the sync fails, there's no reason
to accept the write knowing that it will just fail in the same way.

However, some tests cancels Syncs on purpose, but expect the next
Write to succeed, and there's no good way to tell for sure when the
Sync goroutine that was forked has finished.  (The actual Sync() call
could return earlier with the canceled error.)

We can safely ignore those errors, since the sync failing in that way
doesn't imply anything about syncs failing in the future.

Issue: KBFS-2164

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -1787,7 +1787,7 @@ func (fbo *folderBlockOps) maybeWaitOnDeferredWrites(
 	ctx context.Context, lState *lockState, file Node,
 	c DirtyPermChan) error {
 	var errListener chan error
-	err := func() error {
+	registerErr := func() error {
 		fbo.blockLock.Lock(lState)
 		defer fbo.blockLock.Unlock(lState)
 		filePath, err := fbo.pathFromNodeForBlockWriteLocked(lState, file)
@@ -1798,7 +1798,8 @@ func (fbo *folderBlockOps) maybeWaitOnDeferredWrites(
 		errListener = make(chan error, 1)
 		df.addErrListener(errListener)
 		return nil
-	}()
+	}
+	err := registerErr()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


libkbfs/folder_block_ops.go
```diff
@@ -1824,10 +1825,24 @@ func (fbo *folderBlockOps) maybeWaitOnDeferredWrites(
...
-			// errors from which we couldn't recover)?
+			// Context errors are safe to ignore, since they are
+			// likely to be specific to a previous sync (e.g., a user
+			// hit ctrl-c during an fsync, or a sync timed out, or a
+			// test was provoking an error specifically [KBFS-2164]).
+			if err == context.Canceled || err == context.DeadlineExceeded {
+				fbo.log.CDebugf(ctx, "Ignoring sync err: %+v", err)
+				err := registerErr()
+				if err != nil {
+					return err
```
jzila: Just to make sure a reader sees the comment below before, I'd suggest a refactor to the following:
```go
if ...
    fbo.log..CDebugf(...)
    err = registerErr()
    if err == nil {
        continue
    }
}
// Treat other errors...
return err
```

jzila:
Ooh one more thing: I wonder if this will also fix the issue I was seeing, since that one was triggered by me `<ctrl>-C`ing a large copy.

libkbfs/folder_block_ops.go
strib: Happy to do this, but not sure I follow the logic.  The "Treat other errors" comment doesn't apply to errors returned from `registerErr()`, since those are purely local to this write and aren't affected by whatever happened in the previous sync.  Can you elaborate?

jzila: Yeah on reading it again, nvmind. I'm a little confused why need to re-register the error though.

strib: Ah, because the semantics are the registrants get notified exactly once, and then the list of registrants is cleared (over in `dirtyFile.notifyErrListeners()`).  So if some other writes snuck in and blocked this write between the sync finishes and this goroutine waking up, we might have to wait for the next Sync to finish, so we register for that as well.

PR #962, opened 2017-05-10
folder_branch_ops: fix rm'd dirty dirs and rm -rf performance

strib:
This PR does two things:

1. Removes dirty nodes when the entry is removed. A removed dir entry is also removed from the deCache, which means it will never be synced.  Removed files still go through the sync process, so that their state is cleared.
2. Caches block infos during a cached `rmOp`. Otherwise the prepper will fetch all the blocks in order to find out the sizes, which could take a long time and would block writes and leave dirty paths around during the entire download process, until the SyncAll finishes.

The test makes sure extraneous blocks are fetched, and the dirty paths are all cleaned up.

Issue: KBFS-2161

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -2987,6 +2987,7 @@ func (fbo *folderBranchOps) unrefEntryLocked(ctx context.Context,
 	}
 
 	unrefsToAdd := make(map[BlockPointer]bool)
+	fbo.prepper.cacheBlockInfos([]BlockInfo{de.BlockInfo})
 	unrefsToAdd[de.BlockPointer] = true
 	// construct a path for the child so we can unlink with it.
 	childPath := dir.ChildPath(name, de.BlockPointer)
@@ -3005,6 +3006,7 @@ func (fbo *folderBranchOps) unrefEntryLocked(ctx context.Context,
 		} else if err != nil {
 			return err
 		}
+		fbo.prepper.cacheBlockInfos(blockInfos)
 		for _, blockInfo := range blockInfos {
 			unrefsToAdd[blockInfo.BlockPointer] = true
 		}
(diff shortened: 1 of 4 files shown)


libkbfs/folder_update_prepper.go
```diff
@@ -24,6 +25,9 @@ type folderUpdatePrepper struct {
 	folderBranch FolderBranch
 	blocks       *folderBlockOps
 	log          logger.Logger
+
+	cacheLock   sync.Mutex
```
jzila: As `go vet` has identified, all the methods of `folderUpdatePrepper` now need to be accessed by pointer.

libkbfs/folder_update_prepper.go
```diff
@@ -476,7 +480,7 @@ func (fup folderUpdatePrepper) prepTree(ctx context.Context, lState *lockState,
 // updateResolutionUsage figures out how many bytes are referenced and
 // unreferenced in the merged branch by this resolution.  Only needs
 // to be called for non-squash resolutions.
-func (fup folderUpdatePrepper) updateResolutionUsage(ctx context.Context,
+func (fup folderUpdatePrepper) updateResolutionUsageLocked(ctx context.Context,
```
jzila: This seems to imply that the `lState` should be locked, but it actually refers to the `cacheLock`. Should the name reflect that? Same thing below.

libkbfs/folder_branch_ops.go
```diff
@@ -3005,6 +3006,7 @@ func (fbo *folderBranchOps) unrefEntryLocked(ctx context.Context,
 		} else if err != nil {
 			return err
 		}
+		fbo.prepper.cacheBlockInfos(blockInfos)
```
jzila: No race between these two cache calls because of the `mdWriterLock`, correct?

strib: Yeah, exactly.

PR #965, opened 2017-05-10
add MakeIDRandomOrBust

songgao:


tlf/test_common.go
```diff
@@ -16,6 +16,17 @@ func FakeID(b byte, public bool) ID {
 	return ID{bytes}
 }
 
+// FakeIDRandomOrBust creates a fake public or private TLF ID from the given
+// byte, and fill the rest with empty bytes.
```
jzila: IMO this doesn't need to be in the `kbfs` repo. And I don't like setting the byte. We can always compare with the randomly generated `TlfID`.

songgao: I just didn't want to change all the signatures. I think some `libkbfs` tests relies on `FakeIDByte`. I guess we should just change them to random in the future. I'll go ahead and use `tlf.MakeRandomID` in kbfs-server tests

jzila: Right I'm not saying we need to change the `libkbfs` tests; just the server ones.

PR #978, opened 2017-05-15
kbpki_client: allow slack time when checking revoke times

strib:
The mdserver might still allow writes from an online device that's
just been revoked, until it hears about the revocation from the API
server.  One user has TLFs in this state, and our strict check here
means they can't read any data in it.

Eventually we should switch all these to Merkle roots, but that's a
much bigger project (and will also need some concept of slack anyway,
for the same reason).

Issue: keybase/client#6990


maxtaco:
Fine with me

strib:
> Changes LGTM, as long as the slack only applies to KBFS/chat activity, and not device management activity.

Yep, only affects KBFS and chat.

PR #979, opened 2017-05-15
init: enable dir op batching for everyone, and raise journal disk limit to 85%

strib:


songgao:
1. This doesn't automatically enable batching; still needs to manually pass in the flag, correct?
2. I remember there was an issue that if user gets logged out and back in, batch size gets overriden. Now that it's the same flag for everybody, should we address that as well?

strib:
> This doesn't automatically enable batching; still needs to manually pass in the flag, correct?

It does enable it, because the default for the flag is 100.

> I remember there was an issue that if user gets logged out and back in, batch size gets overriden. Now that it's the same flag for everybody, should we address that as well?

That's what the changes in `keybase_server_util` in this PR already address.

songgao:
ah OK! sorry my head is probably not working properly

PR #981, opened 2017-05-15
tlf_journal: check MD server for conflicts during flush

strib:
If we have a lot of blocks to flush from the journal, we should check
the MD server periodically to see if the next MD we're going to flush
would result in a conflict.

Specifically, if we have already once done CR locally, we might have a
huge set of blocks to flush, and only one MD to flush at the very end.
If we wait to the end to find out if there's a conflict, it could take
a long time for the remote changes (written by another device) to be
reflected in the local view of the TLF.  If instead we force CR
earlier, those changes will become visible earlier.

Issue: KBFS-2106

### DIFF
--- a/libkbfs/tlf_journal.go
+++ b/libkbfs/tlf_journal.go
@@ -89,6 +89,8 @@ const (
 	// Maximum number of blocks to delete from the local saved block
 	// journal at a time while holding the lock.
 	maxSavedBlockRemovalsAtATime = uint64(500)
+	// How often to check the server for conflicts while flushing.
+	tlfJournalServerMDCheckInterval = 1 * time.Minute
 )
 
 // TLFJournalStatus represents the status of a TLF's journal for
@@ -219,7 +221,8 @@ type tlfJournal struct {
 	backgroundShutdownCh chan struct{}
 
 	// Serializes all flushes.
-	flushLock sync.Mutex
+	flushLock         sync.Mutex
+	lastServerMDCheck time.Time // protected by `flushLock`
 
 	// Tracks background work.
 	wg kbfssync.RepeatedWaitGroup
@@ -779,6 +782,7 @@ func (j *tlfJournal) flush(ctx context.Context) (err error) {
 
 		// TODO: Flush MDs in batch.
 
+		flushedOneMD := false
 		for {
 			flushed, err := j.flushOneMDOp(ctx, maxMDRevToFlush)
 			if err != nil {
@@ -787,7 +791,16 @@ func (j *tlfJournal) flush(ctx context.Context) (err error) {
 			if !flushed {
 				break
 			}
+			flushedOneMD = true
+			j.lastServerMDCheck = j.config.Clock().Now()
 			flushedMDEntries++
+		}
+
+		if !flushedOneMD {
+			err = j.checkServerForConflicts(ctx)
+			if err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
 	return "tlfJournal is not empty"
 }
 
+func (j *tlfJournal) checkServerForConflicts(ctx context.Context) error {
+	if time.Since(j.lastServerMDCheck) < tlfJournalServerMDCheckInterval ||
+		!j.config.MDServer().IsConnected() {
```
jzila: Shouldn't this rather attempt a reconnect?

libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
 	return "tlfJournal is not empty"
 }
 
+func (j *tlfJournal) checkServerForConflicts(ctx context.Context) error {
+	if time.Since(j.lastServerMDCheck) < tlfJournalServerMDCheckInterval ||
```
jzila: This implicitly considers `time.Now()` rather than `config.Clock().Now()`. Not sure if that matters, but it makes it less testable.
```go
j.config.Clock().Now().Sub(j.lastServerMDCheck) < tlfJournalServerMDCheckInterval
```

libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
...
+
+	isConflict, err := j.isOnConflictBranch()
+	if err != nil {
+		return err
+	}
+	if isConflict {
+		return nil
+	}
+
+	nextMDToFlush, err := func() (MetadataRevision, error) {
```
jzila: You may need to replace this with `kbfsmd.Revision` if @akalin-keybase has already merged his refactor.

libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
 	return "tlfJournal is not empty"
 }
 
+func (j *tlfJournal) checkServerForConflicts(ctx context.Context) error {
```
jzila: This is locked by `flushLock`, not sure how you want to capture that in the name (important because it reads/writes `lastServerMDCheck`).

libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
...
+
+	nextMDToFlush, err := func() (MetadataRevision, error) {
+		j.journalLock.RLock()
+		defer j.journalLock.RUnlock()
+		return j.mdJournal.readEarliestRevision()
+	}()
+	if err != nil {
+		return err
+	}
+	if nextMDToFlush == MetadataRevisionUninitialized {
```
jzila: Same as above: `kbfsmd.RevisionUninitialized`.

libkbfs/tlf_journal.go
```diff
@@ -813,6 +826,56 @@ func (e errTLFJournalNotEmpty) Error() string {
...
+		return nil
+	}
+
+	j.log.CDebugf(ctx, "Checking the MD server for the latest revision; "+
+		"next MD revision in the journal is %d", nextMDToFlush)
+	// TODO: implement a lighterweight server RPC that just returns
+	// the latest revision number, so we don't have to fetch the
+	// entire MD?
```
jzila: Ticket + mention in comment please.

libkbfs/tlf_journal.go
strib: We don't have a convention for it in this file, and there are many other functions with the same issue.  I'm inclined to favor consistency with those other functions for now.

libkbfs/tlf_journal.go
strib: If we are offline, one of the flush operations will block or timeout, leading to an exponential backoff retry at a higher level.  I only put this check here to avoid blocking on a pretty useless command, but it's not strictly necessary and I don't think there's a reason to explicitly reconnect here and not do so everywhere else there's an RPC.

I think I'll remove the check, it's just confusing.

strib:
Thanks for the feedback, implemented the suggestions I didn't comment on directly.  Will merge after CI passes.

PR #982, opened 2017-05-16
libkbfs: cache MD at the same time it's put to the server

strib:
This PR fixes a nasty race:

1. folderBranchOps puts a new MD update X into the journal, but execution doesn't return immediately to FBO.
2. The journal, while flushing, notices a conflict, and converts the journal, including new update X, to a conflict branch. This updates all MDs in the cache with a new branch ID . . . except update X hasn't been put into the cache yet.
3. Execution returns to folderBranchOps, and update X gets put into the cache as a "merged" revision with a null BranchID.
4. Conflict resolution runs, and reads update X as one of the merged revisions, even though the real merged revision X was written by another device and should be fetched from the server.

To fix, I decided to put things in the cache under the same journal lock that puts things in the journal, so we can be guaranteed that branch conversion will catch all cached updates when a journal put doesn't fail with a conflict. This ended up requiring some changes to the MDOps interface so that an immutable RootMetadata could be created at the journal level, instead of in FBO.

The downside of this approach is that now we don't cache in memory MDs that have been fetched from the journal.  That seems ok, since they're present on disk anyway, and if they were written by us since restart, they should already be in the cache.

Issue: KBFS-2182

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -692,10 +692,6 @@ func (fbo *folderBranchOps) setHeadLocked(
 	}
 
 	fbo.log.CDebugf(ctx, "Setting head revision to %d", md.Revision())
-	err := fbo.config.MDCache().Put(md)
-	if err != nil {
-		return err
-	}
 
 	// If this is the first time the MD is being set, and we are
 	// operating on unmerged data, initialize the state properly and
@@ -1435,7 +1431,7 @@ func (fbo *folderBranchOps) initMDLocked(
 	// about delayed conflicts (since this is essentially a rekey, and
 	// we always bypass the journal for rekeys).  The caller will have
 	// to intelligently deal with a conflict.
-	mdID, err := fbo.config.MDOps().Put(ctx, md)
+	irmd, err := fbo.config.MDOps().Put(ctx, md, session.VerifyingKey)
 	if err != nil {
 		return err
 	}
@@ -1450,8 +1446,7 @@ func (fbo *folderBranchOps) initMDLocked(
 			md.TlfID(), fbo.head.mdID)
 	}
 
-	fbo.setNewInitialHeadLocked(ctx, lState, MakeImmutableRootMetadata(
-		md, session.VerifyingKey, mdID, fbo.config.Clock().Now()))
+	fbo.setNewInitialHeadLocked(ctx, lState, irmd)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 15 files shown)


libkbfs/tlf_journal.go
```diff
@@ -1902,19 +1902,19 @@ func (j *tlfJournal) getMDRange(
 }
 
 func (j *tlfJournal) doPutMD(ctx context.Context, rmd *RootMetadata,
-	mdInfo unflushedPathMDInfo,
-	perRevMap unflushedPathsPerRevMap) (
-	mdID kbfsmd.ID, retryPut bool, err error) {
+	mdInfo unflushedPathMDInfo, perRevMap unflushedPathsPerRevMap,
+	verifyingKey kbfscrypto.VerifyingKey) (
```
jzila: I'm not sure where to put this comment, because it's percolated everywhere. But I'm confused why `verifyingKey` needs to be passed everywhere. Wouldn't it be cleaner to add `currentSessionGetter` or make a new `verifyingKeyGetter` and add it to `tlfJournalConfig`?

jzila: Ah, I see now the `lastWriterVerifyingKey` in `folderBranchOps`. Hmm. I guess leave it then, although a comment somewhere would be good. But where?

libkbfs/folder_branch_ops.go
```diff
@@ -2239,7 +2238,20 @@ func (fbo *folderBranchOps) finalizeMDRekeyWriteLocked(ctx context.Context,
 		}
 		mdOps = jServer.delegateMDOps
 	}
-	mdID, err := mdOps.Put(ctx, md)
+
+	var key kbfscrypto.VerifyingKey
```
jzila: Maybe the comment describing why we have to pass through a verifying key needs to go here along with the ticket number.

libkbfs/tlf_journal.go
jzila: Actually...that `if` block on `md` could also go here as `rmd.IsWriterMetadataCopiedSet`.

strib: Yeah, we have to pass the verifying key all the way down from FBO because it could possibly come from the predecessor MD update, which isn't available easily here in the journal -- we'd have to re-fetch it from the cache/server.  Seems cleaner to pass it all the way down.  I guess I'll add a comment on the MDOps interface function description.

libkbfs/folder_branch_ops.go
strib: I'll just make it clear in the `MDOps.Put` description.

PR #986, opened 2017-05-18
disk_limiter: return currLimit(), rather than limit

strib:
`getDiskLimitInfo` was returning the max limit for the journal, rather
than the limit calculated based on available space.  Oops!

Turns out this is a float64, so all those types needed to change
sadly.  When reporting to the user, it formats the float with no
decimals.

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -205,8 +205,8 @@ func (bt *backpressureTracker) beforeDiskBlockCachePut(blockResources int64) (
 	return availableResources
 }
 
-func (bt *backpressureTracker) getLimitInfo() (used int64, limit int64) {
-	return bt.used, bt.limit
+func (bt *backpressureTracker) getLimitInfo() (used int64, limit float64) {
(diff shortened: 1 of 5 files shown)


libkbfs/backpressure_disk_limiter.go
```diff
@@ -205,8 +205,8 @@ func (bt *backpressureTracker) beforeDiskBlockCachePut(blockResources int64) (
 	return availableResources
 }
 
-func (bt *backpressureTracker) getLimitInfo() (used int64, limit int64) {
-	return bt.used, bt.limit
+func (bt *backpressureTracker) getLimitInfo() (used int64, limit float64) {
+	return bt.used, bt.currLimit()
```
jzila: Would it be too complicated to refactor `currLimit` to be an `int64`?

strib: @jzila: `currLimit()` is used in fraction calculations (like is `usedFrac()`), so that doesn't seem like a good solution.  I could cast it to `int64` here if you prefer.  I don't have a strong preference but seemed a little better to only lose information at the highest layer.

PR #987, opened 2017-05-18
libfuse: don't try to get quota when disconnected

strib:
Make Linux/macOS work the same way as Windows.  I saw someone complain
about this on Twitter:

https://twitter.com/bytebot/status/865182626977640449


strib:
> It seems the ExtraFilename from libfs.RemoteStatus is used to indicate user's not logged in or "human error"? How is it used?

It also indicates when the client is disconnected from the mdserver, which is the main thing this PR is targeting.

(I think the "Human" prefix in the const names just indicates that it is a file explaining the error in a way that's readable by humans.)

songgao:
OK thanks!

PR #989, opened 2017-05-19
disk_block_cache: Make disk cache errors nonfatal

jzila:
First, ensure that a failure to load a version file isn't fatal for the disk cache.

Then, even if the disk cache fails to load, start KBFS anyway.


libkbfs/disk_block_cache.go
```diff
@@ -256,27 +260,38 @@ func getVersionedPathForDiskCache(dirPath string) (versionedDirPath string,
...
-		if version < currentDiskCacheVersion {
-			return "", errors.WithStack(
-				InvalidVersionError{fmt.Sprintf("New disk cache version."+
-					" Delete the existing disk cache at path: %s", dirPath)})
-		}
-		// Existing disk cache version is newer than we expect for this client.
-		// This is an error since our client won't understand its format.
-		if version > currentDiskCacheVersion {
-			return "", errors.WithStack(OutdatedVersionError{})
+		if err != nil {
```
strib: It would be good to log something in all three of these cases.

libkbfs/init.go
```diff
@@ -616,11 +616,13 @@ func doInit(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn,
 		log.Warning("Could not enable disk limiter: %+v", err)
 		return nil, err
 	}
+	goctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
strib: Nit: why "goctx"?  It's not used in a goroutine.  Maybe `ctxWithTimeout` or `ctx10s`?

jzila: `ctx` was taken by something that isn't a `context.Context`, so I was just going for a name that differentiated it. I'll change it to `ctx10s`.

libkbfs/disk_block_cache.go
jzila: K I'll plumb the logger through.

libkbfs/init.go
strib: Ah I see, the "go" referred to "golang", not "goroutine".  Cool.  `ctx10s` still seems better to me, thanks.

PR #991, opened 2017-05-19
interfaces: change KBPKI interface to deal with UserOrTeamID

strib:
And also Identify() now calls keybase1.IdentifyLite, since none of the
callers were directly depending on any of the keys being available in
the response.  This will be needed later since teams can't be
identified through Identify2, as their keys are incompatible with that
result data structure.

Note that this depends on keybase/client#7046 and #990.

Issue: KBFS-2185

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1872,16 +1872,12 @@ func (fbo *folderBranchOps) GetNodeMetadata(ctx context.Context, node Node) (
 		return res, err
 	}
 	res.BlockInfo = de.BlockInfo
-	uid := de.Writer
-	if uid == keybase1.UserOrTeamID("") {
-		uid = de.Creator
-	}
-	asUser, err := uid.AsUser()
-	if err != nil {
-		return NodeMetadata{}, err
+	id := de.Writer
+	if id == keybase1.UserOrTeamID("") {
+		id = de.Creator
 	}
 	res.LastWriterUnverified, err =
-		fbo.config.KBPKI().GetNormalizedUsername(ctx, asUser)
+		fbo.config.KBPKI().GetNormalizedUsername(ctx, id)
 	if err != nil {
 		return res, err
 	}
@@ -5998,8 +5994,8 @@ func (fbo *folderBranchOps) GetUpdateHistory(ctx context.Context,
 	for _, rmd := range rmds {
 		writer, ok := writerNames[rmd.LastModifyingWriter()]
 		if !ok {
-			name, err := fbo.config.KBPKI().
-				GetNormalizedUsername(ctx, rmd.LastModifyingWriter())
+			name, err := fbo.config.KBPKI().GetNormalizedUsername(
+				ctx, rmd.LastModifyingWriter().AsUserOrTeam())
 			if err != nil {
 				return TLFUpdateHistory{}, err
 			}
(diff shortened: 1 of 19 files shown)


jzila:
Is the failure due to the merge sequence, or is it something that needs to be fixed?

strib:
@jzila: it's test failures in `libfuse/` I forgot to check for locally.  Will fix in the morning.  Shouldn't affect the review much though.

strib:
CI should be fixed once keybase/client#7117 goes in.

strib:
Ok, finally passes CI.  @jzila should I block on your review?

jzila:
Reviewing now

libkbfs/folder_branch_ops.go
```diff
@@ -1872,16 +1872,12 @@ func (fbo *folderBranchOps) GetNodeMetadata(ctx context.Context, node Node) (
...
-	uid := de.Writer
-	if uid == keybase1.UserOrTeamID("") {
-		uid = de.Creator
-	}
-	asUser, err := uid.AsUser()
-	if err != nil {
-		return NodeMetadata{}, err
+	id := de.Writer
+	if id == keybase1.UserOrTeamID("") {
+		id = de.Creator
```
jzila: Why remove the `asUser`? The`Writer` is still guaranteed to be a `UID`, right?

jzila: nvmind; I see that you moved this block into `GetNormalizedUsername`

libkbfs/identify_util.go
```diff
@@ -168,7 +168,7 @@ func identifyUser(ctx context.Context, nug normalizedUsernameGetter,
 	} else {
 		reason = fmt.Sprintf("You accessed a private folder with %s.", username.String())
 	}
-	userInfo, err := identifier.Identify(ctx, username.String(), reason)
+	resName, resID, err := identifier.Identify(ctx, username.String(), reason)
```
jzila: `res*` == `resource*`/`result*`?

libkbfs/keybase_daemon_rpc_test.go
```diff
@@ -92,10 +92,10 @@ func (c *fakeKeybaseClient) Call(ctx context.Context, s string, args interface{}
 			return fmt.Errorf("Could not find user info for UID %s", uid)
 		}
 
-		*res.(*keybase1.Identify2Res) = keybase1.Identify2Res{
-			Upk: keybase1.UserPlusKeys{
-				Uid:      uid,
-				Username: string(userInfo.Name),
+		*res.(*keybase1.IdentifyLiteRes) = keybase1.IdentifyLiteRes{
+			Ul: keybase1.UserOrTeamLite{
```
jzila: Is there a reason why this key is so opaquely named?

libkbfs/identify_util.go
strib: I thought "res" was pretty standard for RPC results, but I'll expand it to "result*".

libkbfs/keybase_daemon_rpc_test.go
strib: It was supposed to be symmetric with the field `Identify2Res.Upk` ("UserPlusKeys"), but it stems from a previous draft of the protocol when I called the type just "UserLite".  Sorry, would need to change it in the client protocol (and break everyone's CIs) to fix it.

PR #992, opened 2017-05-19
add random backoff to MD server reconnects

songgao:


libkbfs/mdserver_remote.go
```diff
@@ -71,6 +71,8 @@ type MDServerRemote struct {
 	serverOffsetMu    sync.RWMutex
 	serverOffsetKnown bool
 	serverOffset      time.Duration
+
+	reconnectBackoff *CancelableRandomBackoff
```
kunkun-tang: space between * and CancelableRandomBackoff

songgao: `*CancelableRandomBackoff` is a type (pointer to `CancelableRandomBackoff`) and in Go this is traditionally written with no space.

But thanks for reviewing! 😊

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+		return
+	}
+	<-o.ch
+}
+
+// CancelableRandomBackoff can be used to wait on a random backoff timer. A
+// pointer to a zero value of CancelableRandomBackoff if usable.
+type CancelableRandomBackoff struct {
+	mu sync.Mutex
+	ch *fireOnce
```
strib: Why do we need `fireOnce`?  Doesn't `time.Timer` already provide us with a cancellable timer that fires exactly once?

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (o *fireOnce) wait() {
+	if o == nil {
+		return
+	}
+	<-o.ch
+}
+
+// CancelableRandomBackoff can be used to wait on a random backoff timer. A
```
strib: Also, it's not a big deal, but in `delayed_cancellation.go`, all of the forms of "cancel" are spelled with two Ls in the middle, and this one has only one L.  Would be nice to be consistent.  (I think both spellings are technically correct, so it doesn't really matter which.)

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+func (b *CancelableRandomBackoff) get() *fireOnce {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.ch
+}
+
+// Start starts a random backoff timer. The timer is cancelable with
+// b.Cancel(). Use b.Wait() to wait for the timer.
+//
+// It's OK to call b.Start() multiple times. Any pending b.Wait() honors only
```
strib: The last sentence is confusing.  I would word it more like "Any pending b.Wait() calls will be reset, and will not return until the new timer completes."

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+func (b *CancelableRandomBackoff) Start(maxWait time.Duration) {
+	f := newFireOnce()
+	b.swap(f).fire()
+	time.AfterFunc(time.Duration(randomBackoffRand.Int63()%int64(maxWait)),
+		f.fire)
+}
+
+// Wait waits on any existing random backoff timer. If there isn't a timer
+// started, Wait() returns immediately. If b.Start() is called in the middle of
+// the wait, the new timer is honored.
```
strib: Here also "the new timer is honored" is hard to understand.  Maybe "this call will not return until the new timer completes"?

libkbfs/constants.go
```diff
@@ -15,3 +15,7 @@ const rekeyRecheckInterval = 30 * time.Second
 
 // rekeyInitialTTL is the maximum number rechecks each rekey request can trigger.
 const rekeyInitialTTL = 4
+
+// reconnectBackoffWindow is a backoff window within which we try to wait
+// randomly for before reconnecting to MD server.
+const reconnectBackoffWindow = 2 * time.Minute
```
strib: The comment says this is for the MD server -- should it be in `mdserver_remote.go` then?  Or should `MD` be in the name somewhere?

libkbfs/random_backoff.go
songgao: hmm.. good point! Name it `CancelableRandomTimer`?

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Start starts a random backoff timer. The timer is cancelable with
+// b.Cancel(). Use b.Wait() to wait for the timer.
+//
+// It's OK to call b.Start() multiple times. Any pending b.Wait() honors only
+// the last call to b.Start() until it fires.
+func (b *CancelableRandomBackoff) Start(maxWait time.Duration) {
+	f := newFireOnce()
+	b.swap(f).fire()
+	time.AfterFunc(time.Duration(randomBackoffRand.Int63()%int64(maxWait)),
```
strib: Also it might be better to use `kbfscrypto.RandRead()` (or introduce a helper there to get a 64-bit number), to avoid managing randomness in two places.  It might make it easier to unit test or have reproducible randomness for testing.

libkbfs/random_backoff.go
songgao: It also needs to be fire-able outside the timer, when the `*fireOnce` gets switched out

strib: Why?  That doesn't seem to be used, since `Wait` just switches over to the next timer anyway.

libkbfs/random_backoff.go
songgao: Sounds good. Although if the new timer completes earlier than the old one, that supersedes the old one too

libkbfs/random_backoff.go
strib: Sure (modulo the two Ls comment below).

libkbfs/mdserver_remote.go
```diff
@@ -351,6 +354,8 @@ func (md *MDServerRemote) OnDisconnected(ctx context.Context,
 		md.log.CWarningf(ctx, "MDServerRemote is disconnected")
 		md.config.Reporter().Notify(ctx,
 			connectionNotification(connectionStatusDisconnected))
+		md.reconnectBackoff.Start(reconnectBackoffWindow)
+		defer md.reconnectBackoff.Wait()
```
songgao: It's only called in the reconnect routine I think. But yeah I agree it's a good idea to add a method to the interface. I'll try to refactor this a little bit, which also has the benefit of making it more generally useful. Thanks!

songgao: That was the only approach I could think of without changing the interface. But it seems to be a good idea to just change the interface now. Hope there won't be too much shave!

libkbfs/random_backoff.go
songgao: Sounds good

libkbfs/random_backoff.go
songgao: Yes it does the switch, but only after `wait()` returns. So during the switch, the old one needs to be canceled, right?

libkbfs/random_backoff.go
```diff
@@ -0,0 +1,89 @@
...
+	ch   chan struct{}
+	once sync.Once
+}
+
+func newFireOnce() *fireOnce {
+	return &fireOnce{ch: make(chan struct{})}
+}
+
+func (o *fireOnce) fire() {
+	if o == nil {
```
songgao: hmm.. I guess we can change `fireOnce` to be a value type, and their fields to be pointer types, and check pointers in its methods (so that we can still do stuff like `b.swap(f).fire()`)

libkbfs/random_backoff.go
strib: I don't really follow.  With a `*Timer`, you can still switch it out, and then `Timer.Stop` the old timer.

songgao: Sorry for the confusion. `Timer.Stop` stops the timer, but doesn't close the channel. So it prevents it from firing at all. But at the same time, `CancelableRandomBackOff.Wait` is still waiting on `f.wait()`. We'd need a `Timer.Fire` method in order to use it.

strib: Ah ok, I see, I didn't remember that about Timer.  Oh well.  Maybe add a comment somewhere explaining why Timer isn't good enough.

songgao: sounds good!

songgao:
Thanks for the reviews! I've addressed most comments in `random_backoff.go`. Please let me know if it looks good. I'm gonna close this and move it to `go-framed-msgpack-rpc`.

PR #994, opened 2017-05-22
add random backoff to MD server reconnects

songgao:
This is a successor to https://github.com/keybase/kbfs/pull/992

Depends on https://github.com/keybase/go-framed-msgpack-rpc/pull/113

### DIFF
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -18,8 +18,6 @@ import (
 	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libfuse"
 	"github.com/keybase/kbfs/libkbfs"
-
-	_ "github.com/songgao/stacktraces/on/SIGUSR2"
 )
 
 var runtimeDir = flag.String("runtime-dir", os.Getenv("KEYBASE_RUNTIME_DIR"), "runtime directory")
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -18,6 +18,8 @@ import (
 	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libfuse"
 	"github.com/keybase/kbfs/libkbfs"
+
+	_ "github.com/songgao/stacktraces/on/SIGUSR2"
 )
 
 var runtimeDir = flag.String("runtime-dir", os.Getenv("KEYBASE_RUNTIME_DIR"), "runtime directory")
--- a/libkbfs/constants.go
+++ b/libkbfs/constants.go
@@ -15,3 +15,7 @@ const rekeyRecheckInterval = 30 * time.Second
 
 // rekeyInitialTTL is the maximum number rechecks each rekey request can trigger.
 const rekeyInitialTTL = 4
+
+// mdserverReconnectBackoffWindow is a backoff window within which we try to
+// wait randomly for before reconnecting to MD server.
+const mdserverReconnectBackoffWindow = 2 * time.Minute
--- a/libkbfs/constants.go
+++ b/libkbfs/constants.go
@@ -19,4 +19,4 @@ const rekeyInitialTTL = 4
 // mdserverReconnectBackoffWindow is a backoff window within which we try to
 // wait randomly for before reconnecting to MD server.
 // TODO: increase this to 5 min
-const mdserverReconnectBackoffWindow = 30 * time.Second
+const mdserverReconnectBackoffWindow = 5 * time.Minute
--- a/libkbfs/constants.go
+++ b/libkbfs/constants.go
@@ -18,4 +18,5 @@ const rekeyInitialTTL = 4
 
 // mdserverReconnectBackoffWindow is a backoff window within which we try to
 // wait randomly for before reconnecting to MD server.
-const mdserverReconnectBackoffWindow = 2 * time.Minute
+// TODO: increase this to 5 min
+const mdserverReconnectBackoffWindow = 30 * time.Second
--- a/libkbfs/constants.go
+++ b/libkbfs/constants.go
@@ -19,4 +19,8 @@ const rekeyInitialTTL = 4
 // mdserverReconnectBackoffWindow is a backoff window within which we try to
 // wait randomly for before reconnecting to MD server.
 // TODO: increase this to 5 min
-const mdserverReconnectBackoffWindow = 5 * time.Minute
+const mdserverReconnectBackoffWindow = 30 * time.Second
+
+// registerForUpdatesFireNowThreshold is the maximum length of time that
+// KBFS can be idle for, in order to trigger FireNow from RegisterForUpdate.
+const registerForUpdatesFireNowThreshold = 10 * time.Minute
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -569,7 +569,7 @@ func (fbo *folderBranchOps) doFavoritesOp(ctx context.Context,
 func (fbo *folderBranchOps) updateLastGetHeadTimestamp() {
 	fbo.muLastGetHead.Lock()
 	defer fbo.muLastGetHead.Unlock()
-	fbo.lastGetHead = time.Now()
+	fbo.lastGetHead = fbo.config.Clock().Now()
 }
 
 // getTrustedHead should not be called outside of folder_branch_ops.go.
@@ -599,7 +599,7 @@ func (fbo *folderBranchOps) getHead(lState *lockState) (
 	fbo.headLock.RLock(lState)
 	defer fbo.headLock.RUnlock(lState)
 
-	// See getTrustedHead for explaination.
+	// See getTrustedHead for explanation.
 	fbo.config.MDServer().FastForwardBackoff()
 	fbo.updateLastGetHeadTimestamp()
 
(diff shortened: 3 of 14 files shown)


songgao:
Sounds good. Although I'm also thinking about reducing it to either 30s or 1min for now, just to see if it works well. In other words, if I screwed up somewhere, a 30s delay in reconnection is not as bad as 2min or 5min. What do you think?

jzila:
How will you be monitoring the results?

songgao:
If nobody complains about slow reconnect, it's likely fine; also we can tell from deployments

strib:
I know you said on keybase/go-framed-msgpack-rpc#113 that the pinger doesn't matter because the ticker gets stopped during `OnDisconnect()`, but a `Ping` could in theory start before `OnDisconnect()` is called, and return after, which would invoke `initNewConnection()`, right?  Maybe that is rare enough that it won't much affect the stampeding herd that happens on mdserver redeploys, just wanted to mention it in case you think it's important enough to address in this PR.

songgao:
Thanks @strib! I don't actually know for sure how rare it would be, since the mdserver disconnection doesn't seem to be detected immediately. Perhaps this should block on disabling the pinger.

strib:
I don't think we should disable the pinger.  It would be enough for `pingOnce` to make sure that the state of the connection hasn't changed since before it sent the ping RPC.  (For example, it could check that the rekey queue is the same before and after, or maybe that `isAuthenticated == false`, or something new you add.  But the ping is important for keeping the connection alive and catching laptop resumes.

songgao:
If I understand correctly, the reachability stuff address the same set of problems (including resuming from standby), by monitoring the gregor connection. The service calls the `CheckRechability` into KBFS if it receives the NetInfo change from frontend. @mmaxim could you confirm?

strib:
@songgao: even if it solves the laptop resume issue on all platforms, it won't address keeping alive the MD server connection (which gets closed by the AWS load balancer after being idle for a certain amount of time).  So maybe it's possible to get rid of the `initNewConnection` call if we're sure `CheckReachability` will be called in all cases, but I don't think we should get rid of the pinger entirely.  It also computes time differences from the server, which is needed for other features.

songgao:
Ah I see. Good point. Thanks! I'll update this tomorrow morning

jzila:
We're still doing the thing where we need to add stuff to every RPC: `ctx.WithFireNow()` and now `fbo.config.MDServer().FastForwardBackoff()`. Why is the second one needed?

I'm afraid that requiring these is brittle, and I'd love to have a way to abstract these in the library.

songgao:
Some MD operations don't necessarily go through FBO, e.g. when a TLF is first accessed or created, and when chat needs MD

songgao:
ah sorry, that was the answer for the first one. The second one is because we rely on MD updates pushes from server to detect changes once connected. So without fast forwarding the timer, during the backoff the client would just return cached stuff

jzila:
All right. I'm still not quite happy with the sprinkled maintenance calls; they seem like a bug magnet for people forgetting to put them where they're needed in the future, but I don't have a solid suggestion for getting rid of them right now.

songgao:
@jzila I don't have a better idea either, but will keep thinking

libkbfs/constants.go
```diff
@@ -15,3 +15,8 @@ const rekeyRecheckInterval = 30 * time.Second
 
 // rekeyInitialTTL is the maximum number rechecks each rekey request can trigger.
 const rekeyInitialTTL = 4
+
+// mdserverReconnectBackoffWindow is a backoff window within which we try to
+// wait randomly for before reconnecting to MD server.
+// TODO: increase this to 5 min
```
strib: No need for this TODO anymore, right?

songgao: ah this was accidentally committed as I was testing it. Meant to leave it as 30sec for now

songgao:
@strib You are right; it would still return stale data. But since the updates are already asynchronous, I figured this would just be like really slow network connections. Having a timestamp of last call as heuristic is a good idea. Will do!

On the other hand I was thinking this morning that, if we get to offline mode in KBFS at some point in the future, mdserver connection status would be factored into that part, and the backoff could just be a brief period of offline mode where user is aware of from UI. Does this sound correct?

strib:
> But since the updates are already asynchronous, I figured this would just be like really slow network connections.

Though by this logic, we wouldn't need any changes in `folderBranchOps`!

> On the other hand I was thinking this morning that, if we get to offline mode in KBFS at some point in the future, mdserver connection status would be factored into that part, and the backoff could just be a brief period of offline mode where user is aware of from UI. Does this sound correct?

Since the MD updates are already asynchronous, the current code basically already implements offline mode (with respect to the MD, anyway).  So I don't know if anything needs to change in that part of the code really.  For MD stuff, it will be more about caching the MD head on disk so we can continue work after a restart.

songgao:
> Though by this logic, we wouldn't need any changes in folderBranchOps!

Ah that's true!

> the current code basically already implements offline mode

Make sense. In this case, would it make sense to expose the status of KBFS connectivity to user in someway?

strib:
@songgao: also maybe you can move the `fbo.config.MDServer().FastForwardBackoff()` call to `folderBranchOps.getHead` and `folderBranchOps.getTrustedHead`, and that way you wouldn't have to call it in every API method.

strib:
> Make sense. In this case, would it make sense to expose the status of KBFS connectivity to user in someway?

We already do that, by putting a special error file in `/keybase` when we're not connected to the mdserver.  I assume as part of offline mode we might also want something to happen in the GUI/icon as well.

songgao:
> I assume as part of offline mode we might also want something to happen in the GUI/icon as well.

Cool; that's what I had in mind too

songgao:
Still fixing mock tests; but please take another look? Thanks!

kbfsfuse/main.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libfuse"
 	"github.com/keybase/kbfs/libkbfs"
+
+	_ "github.com/songgao/stacktraces/on/SIGUSR2"
```
strib: Intentionally committed?  I'm not against it, we already have a way to get goroutines without killing the process: `/keybase/.kbfs_profiles/goroutine`.

libkbfs/folder_branch_ops.go
```diff
@@ -557,6 +566,12 @@ func (fbo *folderBranchOps) doFavoritesOp(ctx context.Context,
 	}
 }
 
+func (fbo *folderBranchOps) updateLastGetHeadTimestamp() {
+	fbo.muLastGetHead.Lock()
+	defer fbo.muLastGetHead.Unlock()
+	fbo.lastGetHead = time.Now()
```
strib: Please use `fbo.config.Clock().Now()`.

libkbfs/folder_branch_ops.go
```diff
@@ -574,6 +598,11 @@ func (fbo *folderBranchOps) getHead(lState *lockState) (
 	ImmutableRootMetadata, headTrustStatus) {
 	fbo.headLock.RLock(lState)
 	defer fbo.headLock.RUnlock(lState)
+
+	// See getTrustedHead for explaination.
```
strib: "explanation"

libkbfs/folder_branch_ops.go
```diff
@@ -5549,15 +5578,31 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
 	<-childDone
 }
 
+func (fbo *folderBranchOps) registerForUpdatesShouldFireNow() bool {
+	fbo.muLastGetHead.Lock()
+	defer fbo.muLastGetHead.Unlock()
+	return time.Since(fbo.lastGetHead) < registerForUpdatesFireNowThreshold
```
strib: Let's use the abstracted clock, so:

```go
fbo.config.Clock().Now().Sub(fbo.lastGetHead) < ...
```

libkbfs/folder_branch_ops.go
```diff
@@ -5549,15 +5578,31 @@ func (fbo *folderBranchOps) registerAndWaitForUpdates() {
...
+
 func (fbo *folderBranchOps) registerForUpdates(ctx context.Context) (
 	updateChan <-chan error, err error) {
 	lState := makeFBOLockState()
 	currRev := fbo.getLatestMergedRevision(lState)
-	fbo.log.CDebugf(ctx, "Registering for updates (curr rev = %d)", currRev)
+
+	fireNow := false
+	if fbo.registerForUpdatesShouldFireNow() {
+		fbo.log.CDebugf(ctx, "Registering for updates (curr rev = %d)", currRev)
```
strib: I don't think this log message is needed.

libkbfs/mdserver_remote.go
```diff
@@ -1096,6 +1128,11 @@ func (md *MDServerRemote) GetKeyBundles(ctx context.Context,
 	return wkb, rkb, nil
 }
 
+// FastForwardBackoff implements the MDServer interface for MDServerRemote.
+func (md *MDServerRemote) FastForwardBackoff() {
```
strib: This function should take `md.connMu` to avoid races with `initNewConnection()`.

libkbfs/kbfs_ops_test.go
```diff
@@ -706,6 +706,7 @@ func TestKBFSOpsGetBaseDirChildrenCacheSuccess(t *testing.T) {
 	ops := getOps(config, id)
 	n := nodeFromPath(t, ops, p)
 
+	config.mockMdserv.EXPECT().FastForwardBackoff().AnyTimes()
```
strib: You can put this in `kbfsOpsInit()` in you want, so we don't have to call it in every test.

kbfsfuse/main.go
songgao: Oops; not intentional

libkbfs/kbfs_ops_test.go
songgao: whoa that's the swiss army knife!

songgao: holy shit that one line fixed it. Thanks!

PR #995, opened 2017-05-23
kbfs: properly type TLFs, instead of using just a boolean

strib:
This ended up being a far-reaching change, not only because of
switching the TLF handles over, but because determining the proper
type of a handle for a RootMetadata object required storing the
Writers/Readers as keybase1.UserOrTeamIDs, instead of keybase1.UIDs.

This is a largely mechanical change that doesn't add much new logic.

Issue: KBFS-2185

### DIFF
--- a/fsrpc/fs.go
+++ b/fsrpc/fs.go
@@ -10,6 +10,7 @@ import (
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/kbfs/libkbfs"
+	"github.com/keybase/kbfs/tlf"
 	"golang.org/x/net/context"
 )
 
@@ -30,7 +31,7 @@ func (f fs) favorites(ctx context.Context, path Path) (keybase1.ListResult, erro
 	}
 	files := []keybase1.File{}
 	for _, fav := range favs {
-		if fav.Public == path.Public {
+		if fav.Public == (path.TLFType == tlf.Public) {
 			favPath, err := path.Join(fav.Name)
 			if err != nil {
 				return keybase1.ListResult{}, err
--- a/libkbfs/bare_root_metadata_v2.go
+++ b/libkbfs/bare_root_metadata_v2.go
@@ -30,8 +30,9 @@ type WriterMetadataV2 struct {
 	// The last KB user with writer permissions to this TLF
 	// who modified this WriterMetadata
 	LastModifyingWriter keybase1.UID
-	// For public TLFs (since those don't have any keys at all).
-	Writers []keybase1.UID `codec:",omitempty"`
+	// For public and single-team TLFs (since those don't have any
+	// keys at all).
+	Writers []keybase1.UserOrTeamID `codec:",omitempty"`
 	// For private TLFs. Writer key generations for this metadata. The
 	// most recent one is last in the array. Must be same length as
 	// BareRootMetadata.RKeys.
@@ -63,7 +64,7 @@ type WriterMetadataV2 struct {
 // WriterMetadataV3.
 func (wmdV2 *WriterMetadataV2) ToWriterMetadataV3() WriterMetadataV3 {
 	var wmdV3 WriterMetadataV3
-	wmdV3.Writers = make([]keybase1.UID, len(wmdV2.Writers))
+	wmdV3.Writers = make([]keybase1.UserOrTeamID, len(wmdV2.Writers))
 	copy(wmdV3.Writers, wmdV2.Writers)
 
 	wmdV3.UnresolvedWriters = make([]keybase1.SocialAssertion, len(wmdV2.Extra.UnresolvedWriters))
@@ -78,7 +79,7 @@ func (wmdV2 *WriterMetadataV2) ToWriterMetadataV3() WriterMetadataV3 {
 	wmdV3.UnrefBytes = wmdV2.UnrefBytes
 	wmdV3.MDRefBytes = wmdV2.MDRefBytes
 
-	if wmdV2.ID.IsPublic() {
+	if wmdV2.ID.Type() == tlf.Public {
 		wmdV3.LatestKeyGen = PublicKeyGen
 	} else {
 		wmdV3.LatestKeyGen = wmdV2.WKeys.LatestKeyGeneration()
@@ -142,20 +143,19 @@ type BareRootMetadataV2 struct {
 // must be done separately.
 func MakeInitialBareRootMetadataV2(tlfID tlf.ID, h tlf.Handle) (
 	*BareRootMetadataV2, error) {
-	if tlfID.IsPublic() != h.IsPublic() {
-		return nil, errors.New(
-			"TlfID and TlfHandle disagree on public status")
+	if tlfID.Type() != h.Type() {
+		return nil, errors.New("TlfID and TlfHandle disagree on TLF type")
 	}
 
-	var writers []keybase1.UID
+	var writers []keybase1.UserOrTeamID
 	var wKeys TLFWriterKeyGenerationsV2
 	var rKeys TLFReaderKeyGenerationsV2
-	if tlfID.IsPublic() {
-		writers = make([]keybase1.UID, len(h.Writers))
-		copy(writers, h.Writers)
-	} else {
+	if tlfID.Type() == tlf.Private {
 		wKeys = make(TLFWriterKeyGenerationsV2, 0, 1)
 		rKeys = make(TLFReaderKeyGenerationsV2, 0, 1)
+	} else {
+		writers = make([]keybase1.UserOrTeamID, len(h.Writers))
+		copy(writers, h.Writers)
 	}
 
 	var unresolvedWriters, unresolvedReaders []keybase1.SocialAssertion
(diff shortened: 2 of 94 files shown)


tlf/id.go
```diff
@@ -22,8 +23,36 @@ const (
...
 	// pubIDSuffix is the last byte of a public top-level folder ID
 	pubIDSuffix = 0x17
+	// singleTeamIDSuffix is the last byte of a single-team top-level
+	// folder ID
+	singleTeamIDSuffix = 0x26
 )
 
+type Type int
+
+const (
```
strib: Here is where the TLF types are defined.

tlf/handle.go
```diff
@@ -12,8 +12,8 @@ import (
 //
 // TODO: Have separate types for writers vs. readers.
 type Handle struct {
-	Writers           []keybase1.UID             `codec:"w,omitempty"`
-	Readers           []keybase1.UID             `codec:"r,omitempty"`
+	Writers           []keybase1.UserOrTeamID    `codec:"w,omitempty"`
```
strib: This file is worth close inspection.

strib:
This one is ready for review.  I suggest roughly this review order:

1. `tlf/id.go`
2. `tlf/handle.go`
3. `libkbfs/tlf_handle.go`
4. `libkbfs/tlf_handle_resolve.go`
5. `libkbfs/errors.go`
6. `libkbfs/path.go`
7. `libkbfs/identify*.go`
8. `libkbfs/bare_root_metadata_*.go`
9. `libkbfs/root_metadata.go`
10. `libkbfs/reporter_kbpki.go`
11. `libkbfs/key_manager.go`
12. `libdokan` and `libfuse`
13. `test/`
14. Everything else (almost everything else are pretty dumb mechanical changes)

libkbfs/errors.go
```diff
@@ -169,8 +169,12 @@ type ReadAccessError struct {
 
 // Error implements the error interface for ReadAccessError
 func (e ReadAccessError) Error() string {
+	t := tlf.Private
+	if e.Public {
+		t = tlf.Public
+	}
 	return fmt.Sprintf("%s does not have read access to directory %s",
-		e.User, buildCanonicalPathForTlfName(e.Public, e.Tlf))
+		e.User, buildCanonicalPathForTlfName(t, e.Tlf))
```
songgao: What if it's a team folder? I guess this is fine if we decide to let team folders stay in `/keybase/private`

libkbfs/path.go
```diff
@@ -46,21 +49,29 @@ func BuildCanonicalPath(pathType PathType, paths ...string) string {
...
-	pathType := PrivatePathType
-	if public {
+func buildCanonicalPathForTlfType(t tlf.Type, paths ...string) string {
+	var pathType PathType
+	switch t {
+	case tlf.Private:
+		pathType = PrivatePathType
+	case tlf.Public:
 		pathType = PublicPathType
+	case tlf.SingleTeam:
```
songgao: I can't find a way to comment there, but should there be an added case at https://github.com/keybase/kbfs/pull/995/files#diff-98b8e5fde78a630e2ca6a9b7ef015871R42 ?

tlf/id.go
```diff
@@ -22,8 +23,38 @@ const (
 	idSuffix = 0x16
 	// pubIDSuffix is the last byte of a public top-level folder ID
 	pubIDSuffix = 0x17
+	// singleTeamIDSuffix is the last byte of a single-team top-level
+	// folder ID
+	singleTeamIDSuffix = 0x26
```
strib: Yep, I updated it, thanks. (It's in the keybase repo.)

libkbfs/path.go
songgao: ugh, the link doesn't bring you to the correct place. I meant in the function `BuildCanonicalPath` in `libkbfs/path.go`

libkbfs/errors.go
strib: Team folders will be in `/keybase/teams` or something like that.  I don't see what the issue is here -- it's passing the type (which could be `tlf.SingleTeam`) into `buildCanonicalPathForTlfName`, which picks the correct path type based on that.

libkbfs/path.go
strib: Ah yes, good catch, will do.

libkbfs/errors.go
songgao: It's not; `t` is either `tlf.Private` or `tlf.Public` here, and never `tlf.SingleTeam`, right?

strib: Oh, I see what you're saying.  Yeah I didn't update those errors yet.  Was just being lazy.  I'll do that now.

songgao:
Do we need to also update vendor `tlf` and `libkbfs` in `kbfs-server` soon after this is merged? A quick look told me we aren't making the assumption that there can only be public and private TLFs, but I'm not entirely sure, so it seems to me it's best to vendor them soon. If we do vendor them, we should probably loosen those panics before doing so

strib:
> Do we need to also update vendor tlf and libkbfs in kbfs-server soon after this is merged?

Yep, my plan is to do it right away.  But I won't let the server actually make any team TLFs yet until we're ready to do the full set of server work.

libkbfs/bare_root_metadata_v3_test.go
```diff
@@ -15,12 +15,13 @@ import (
 )
 
 func TestBareRootMetadataVersionV3(t *testing.T) {
-	tlfID := tlf.FakeID(1, false)
+	tlfID := tlf.FakeID(1, tlf.Private)
```
songgao: (also in `test/`)

strib: I can't easily add any tests until the next PR (#996) which vendors in `keybase1.MakeTestTeamID()`.  I'll add a simple bare RMD test to that one.  Later PRs have more involved libkbfs tests.

As for `test/`, that will require some DSL changes.  I'll do those in a separate PR.

songgao: Sounds good. Thanks!

fsrpc/fs.go
```diff
@@ -30,7 +31,7 @@ func (f fs) favorites(ctx context.Context, path Path) (keybase1.ListResult, erro
 	}
 	files := []keybase1.File{}
 	for _, fav := range favs {
-		if fav.Public == path.Public {
+		if fav.Public == (path.TLFType == tlf.Public) {
```
jzila: Will we be properly typing favorites as well?

jzila: nvmind I see the `TODO` below.

libkbfs/bare_root_metadata_v2.go
```diff
@@ -309,9 +309,11 @@ func (md *BareRootMetadataV2) IsFinal() bool {
 // IsWriter implements the BareRootMetadata interface for BareRootMetadataV2.
 func (md *BareRootMetadataV2) IsWriter(
 	user keybase1.UID, deviceKey kbfscrypto.CryptPublicKey, _ ExtraMetadata) bool {
-	if md.ID.IsPublic() {
+	if md.ID.Type() != tlf.Private {
```
jzila: IMO this should be `md.ID.Type() == tlf.Public`, since the teams version of this won't need to iterate through the `md.Writers` list.

libkbfs/kbpki_client.go
```diff
@@ -217,17 +217,19 @@ func (k *KBPKIClient) loadUnverifiedKeys(ctx context.Context, uid keybase1.UID)
 	return k.serviceOwner.KeybaseService().LoadUnverifiedKeys(ctx, uid)
 }
 
-// GetCurrentSessionIfPossible returns the current username and UID from
-// kbpki.GetCurrentSession.
-// If isPublic is true NoCurrentSessionError is ignored and empty username
-// and uid will be returned. If it is false all errors are returned.
-func GetCurrentSessionIfPossible(ctx context.Context, kbpki KBPKI, isPublic bool) (SessionInfo, error) {
+// GetCurrentSessionIfPossible returns the current username and UID
+// from kbpki.GetCurrentSession.  If sessionNotRequired is true
```
jzila: true,

libkbfs/key_manager.go
```diff
@@ -454,14 +456,16 @@ func (km *KeyManagerStandard) Rekey(ctx context.Context, md *RootMetadata, promp
 	defer func() { km.deferLog.CDebugf(ctx, "Rekey %s done: %+v", md.TlfID(), err) }()
 
 	currKeyGen := md.LatestKeyGeneration()
-	if md.TlfID().IsPublic() != (currKeyGen == PublicKeyGen) {
+	if (md.TlfID().Type() == tlf.Public) != (currKeyGen == PublicKeyGen) {
```
jzila: The "clever" factor here is high enough that it warrants a comment, something like "inconsistent indicators that this TLF is or isn't public".

libkbfs/tlf_handle.go
```diff
@@ -44,32 +44,39 @@ type TlfHandle struct {
...
+	return h.tlfType
 }
 
 // IsWriter returns whether or not the given user is a writer for the
 // top-level folder represented by this TlfHandle.
 func (h TlfHandle) IsWriter(user keybase1.UID) bool {
-	_, ok := h.resolvedWriters[user]
+	// TODO(KBFS-2185) relax this?
+	if h.tlfType == tlf.SingleTeam {
+		panic("Can't check whether a user is a writer on a team TLF")
```
jzila: I don't understand why this statement is true. On the metadata structures we allow this check, why not on the handle?

tlf/id.go
```diff
@@ -22,8 +23,38 @@ const (
...
+)
+
+func (t Type) String() string {
+	switch t {
+	case Private:
+		return "private"
+	case Public:
+		return "public"
+	case SingleTeam:
+		return "singleTeam"
```
jzila: I'd rather we use snake case for these strings: `single_team`, but I won't argue past just making that assertion :)

PR #996, opened 2017-05-24
libkbfs: add methods for getting team TLF keys and setting keygen

strib:
This PR adds interface methods, and local implementations, for methods to get the set of crypt keys and latest key gen for team TLFs from the Keybase service.

It vendors in keybase/client#7132 (currently still in review), and depends on #995.

Issue: KBFS-2185


vendor/github.com/keybase/client/go/protocol/keybase1/teams.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+type TeamGetArg struct {
+	SessionID int    `codec:"sessionID" json:"sessionID"`
+	Name      string `codec:"name" json:"name"`
+}
+
+type TeamsInterface interface {
+	TeamCreate(context.Context, TeamCreateArg) error
+	TeamGet(context.Context, TeamGetArg) error
```
jzila: Shouldn't these be giving back a team?

strib: CORE hasn't yet built the interface we need to really get team info, that's in CORE-5372. This is just some random API they implemented to make the CLI work.

PR #997, opened 2017-05-24
libkbfs: save the current Keybase Merkle root in mdv3

strib:
Though in a live setting, this will actually only stored an omitted 0 for now, since the service doesn't expose this yet.

Issue: KBFS-2185

### DIFF
--- a/libkbfs/bare_root_metadata_v3.go
+++ b/libkbfs/bare_root_metadata_v3.go
@@ -88,6 +88,16 @@ type BareRootMetadataV3 struct {
 	// of writing to the given folder.
 	FinalizedInfo *tlf.HandleExtension `codec:"fi,omitempty"`
 
+	// The sequence number of the global Keybase Merkle tree at the
+	// time this update was created (from the writer's perspective).
+	// This field was added to V3 after it was live for a while, and
+	// older clients that don't know about this field yet might copy
+	// it into new updates via the unknown fields copier. Which means
+	// new MD updates might end up referring to older Merkle roots.
+	// That's ok since this is just a hint anyway, and shouldn't be
+	// fully trusted when checking MD updates against the Merkle tree.
+	KBMerkleSeqNo MerkleSeqNo `codec:"msn,omitempty"`
+
 	codec.UnknownFieldSetHandler
 }
 
@@ -968,6 +978,12 @@ func (md *BareRootMetadataV3) RevisionNumber() kbfsmd.Revision {
 	return md.Revision
 }
 
+// MerkleSeqNo implements the BareRootMetadata interface for
+// BareRootMetadataV3.
+func (md *BareRootMetadataV3) MerkleSeqNo() MerkleSeqNo {
+	return md.KBMerkleSeqNo
+}
+
 // BID implements the BareRootMetadata interface for BareRootMetadataV3.
 func (md *BareRootMetadataV3) BID() BranchID {
 	return md.WriterMetadata.BID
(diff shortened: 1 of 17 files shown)


libkbfs/bare_root_metadata_v3.go
```diff
@@ -88,6 +88,16 @@ type BareRootMetadataV3 struct {
 	// of writing to the given folder.
 	FinalizedInfo *tlf.HandleExtension `codec:"fi,omitempty"`
 
+	// The sequence number of the global Keybase Merkle tree at the
+	// time this update was created (from the writer's perspective).
+	// This field was added to V3 after it was live for a while, and
+	// older clients that don't know about this field yet might copy
+	// it into new updates via the unknown fields copier. Which means
+	// new MD updates might end up referring to older Merkle roots.
+	// That's ok since this is just a hint anyway, and shouldn't be
```
songgao: What are the implication of this once we start checking merkle tree in KBFS clients? Does this make verification more expensive?

libkbfs/root_metadata.go
```diff
@@ -286,6 +285,13 @@ func (md *RootMetadata) MakeSuccessor(
 		return nil, errors.New("MD with invalid revision")
 	}
 	newMd.SetRevision(md.Revision() + 1)
+
+	merkleSeqNo, err := merkleGetter.GetCurrentMerkleSeqNo(ctx)
+	if err != nil {
+		return nil, err
+	}
+	newMd.SetMerkleSeqNo(merkleSeqNo)
```
songgao: Should this be `+1`'ed?

strib:
@songgao: storing the global Merkle sequence number is an optimization suggested by @maxtaco here: https://keybase.atlassian.net/wiki/display/SPEC/KBFS+Teams+Design+Discussion?focusedCommentId=48496646#comment-48496646

It's not solely related to teams, it will be useful for the other types of TLFs as well, once we start doing Merkle-checking on the client side for KBFS.  But I think it is more _immediately_ helpful for team TLFs, because to verify updates written by users that are no longer team members, we'll need some notion of a timestamp for when the user was removed from the team, and the service might not bother implementing that (in favor of doing the more correct Merkle checking from the beginning).

libkbfs/root_metadata.go
strib: No, the sequence number is increased according to the server-side global Merkle tree updating.  The client can't update it -- from the KBFS client's perspective, it's just a monotonically-increasing magic number given to it from the service.

libkbfs/bare_root_metadata_v3.go
strib: It will likely make verification of updates written by _revoked_ devices, or by users that are no longer writers of the team, more expensive, yes.  The normal, happy path verification should be unchanged.

songgao:
Thanks for the link and explanations! That's good background.

libkbfs/bare_root_metadata_v3.go
songgao: Can non-legitimate writer with old client have old seq number included in the metadata, making their non-legitimate write seem legitimate? Or more importantly, can an attacker use a modified client to write to a TLF using an identity that is not a writer of the TLF any more? I just realized these two are the same case, so it doesn't really matter that much that we have old clients that simply copy the field without ever updating it.

Is the solution to these concerns (in the future when we implement it) what makes the verification more expensive?

strib: Yes to all questions.  We can't just rely on this hint since the clients could be lying about it.  We can only use it as a starting point for finding the right place in the merkle trees for comparing the update to the team membership around the time of the update.  Fetching those Merkle records and verifying them is where the expense will be, I think.

songgao: Makes sense; Thanks!

Issue #998, opened 2017-05-25
Type mismatches in disk_limits_unix.go on freebsd/amd64


cobaugh:
Found this when trying to build the latest master (8f19a38ad8bb185b215f7a87702af1ac71a0d1af) on FreeBSD 11. Looks like Statfs_T defines Bavail as int64 and Bsize as uint64 on freebsd/amd64.

<pre>$ go version
go version go1.8.1 freebsd/amd64
$ go install
# github.com/keybase/kbfs/libkbfs
../libkbfs/disk_limits_unix.go:27: invalid operation: stat.Bavail * uint64(stat.Bsize) (mismatched types int64 and uint64)
../libkbfs/disk_limits_unix.go:32: cannot use stat.Ffree (type int64) as type uint64 in assignment</pre>

strib:
Thanks @cobaugh, I had no idea!  I wonder how you'd ever have negative available bytes.  Anyway, I put up #999 to fix it.

strib:
Should be fixed now.  @cobaugh can you confirm master compiles for you now, and close this out if it looks ok?  Thanks!

cobaugh:
Yep, it's fixed. Still doesn't build, it's getting stuck on something else, but it's at least progress. Thanks!

PR #999, opened 2017-05-25
disk_limits_unix: on FreeBSD, apparently Bavail is an int64

strib:
Issue: #998


strib:
@jzila: see #998 -- master doesn't compile on freebsd/amd64.

songgao:
Our other stuff wouldn't work on FreeBSD either right? `libfuse` has a bunch of platform specific stuff and `_freebsd` is never included. It's probably not gonna be too much effort to make it work tho

jzila:
@songgao: most of our platform specific *nix stuff is `+build !windows`, so it should capture Unix and Linux.

PR #1005, opened 2017-05-26
key_manager: get team TLF keys

strib:
Issue: KBFS-2192

### DIFF
--- a/libkbfs/key_manager.go
+++ b/libkbfs/key_manager.go
@@ -119,6 +119,31 @@ func (km *KeyManagerStandard) getTLFCryptKey(ctx context.Context,
 		return kbfscrypto.TLFCryptKey{}, err
 	}
 
+	// Team TLF keys come from the service.
+	if tlfID.Type() == tlf.SingleTeam {
+		tid, err := kmd.GetTlfHandle().FirstResolvedWriter().AsTeam()
+		if err != nil {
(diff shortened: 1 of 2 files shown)


libkbfs/key_manager.go
```diff
@@ -119,6 +119,31 @@ func (km *KeyManagerStandard) getTLFCryptKey(ctx context.Context,
 		return kbfscrypto.TLFCryptKey{}, err
 	}
 
+	// Team TLF keys come from the service.
+	if tlfID.Type() == tlf.SingleTeam {
+		tid, err := kmd.GetTlfHandle().FirstResolvedWriter().AsTeam()
+		if err != nil {
+			return kbfscrypto.TLFCryptKey{}, err
+		}
+		keys, _, err := km.config.KBPKI().GetTeamTLFCryptKeys(ctx, tid)
```
jzila: We should be caching this in some way, IMO.

strib: Yep, we will be, in `KeybaseServiceBase`, once there's a service RPC we can call.

Or are you saying you want to put each of the individual keys into the key cache right away?  I'm not sure that's useful since we may not be using them, and they could evict other keys.

jzila: The first one.

PR #1006, opened 2017-05-26
libkbfs: check team members and resolve team IDs

strib:
To check writer/reader privileges (and MD signature validity) at the
level of the local servers, we need to pass contexts and a
TeamMembershipChecker to a lot more places.

Note that this isn't quite right, because it is using _current_ team
membership to check the validity of a signature on possibly-historic
MD updates.  Once we figure out the exact story with checking historic
team membership, this will need a bit of updating.

This makes a full libkbfs test of a team TLF possible!

This revendors client/go/libkb to get keybase/client#7181, which forced us to update a ton of other code as well.

Issue: KBFS-2192


strib:
Suggested review order:

1. `bare_root_metadata_v3.go`
2. `kbpki.go` and `keybase_daemon_local.go`
3. `mdserver_local_shared.go` and `mdserver_memory.go`
4. `data_types.go`
5. `kbfs_ops_test.go`
6. Everything else is pretty much mechanical plumbing.

libkbfs/data_types.go
```diff
@@ -180,9 +180,7 @@ const (
 	// storage of key bundles.
 	SegregatedKeyBundlesVer MetadataVer = 3
 
-	// MDv3 TODO: Change the below to SegregatedKeyBundlesVer when
-	// the client is ready for MDv3.
-	defaultClientMetadataVer MetadataVer = InitialExtraMetadataVer
+	defaultClientMetadataVer MetadataVer = SegregatedKeyBundlesVer
```
strib: Oops!  Can't believe we hadn't done this yet.  (Only affects tests, but still.)

libkbfs/kbfs_ops_test.go
```diff
@@ -3430,3 +3430,89 @@ func TestDirtyPathsAfterRemoveDir(t *testing.T) {
...
+	config3 := ConfigAsUser(config1, u3)
+	defer CheckConfigAndShutdown(ctx, t, config3)
+	session3, err := config3.KBPKI().GetCurrentSession(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	uid3 := session3.UID
+
+	// These are deterministic, and should add the same TeamInfos for
+	// both user configs.
```
jzila: Could you `t.Log` these comments instead? That way they'll appear in the test run log, which makes it easier to locate the failure.

libkbfs/mdserver_local_config.go
```diff
@@ -28,3 +29,7 @@ type mdServerLocalConfigAdapter struct {
 func (ca mdServerLocalConfigAdapter) currentSessionGetter() CurrentSessionGetter {
 	return ca.Config.KBPKI()
 }
+
+func (ca mdServerLocalConfigAdapter) teamMemChecker() TeamMembershipChecker {
```
jzila: Let's go with the full word here: `teamMembershipChecker`

libkbfs/kbfs_ops_test.go
strib: Happy to `t.Log` them, but don't the line numbers of the `require` error messages already say exactly where the failure happened?  I don't think I've ever found myself wishing for more logging within a test function.

PR #1009, opened 2017-05-29
libfuse: support team favorites and expose team list in dev mode

strib:
This vendors in keybase/client#7184, and supports a "team" top-level folderlist showing team favorites.  (Only in "devel" runmode for now.)

It also auto-favorites team folders in the local daemon (this is eventually what the API server should do as well).

With these changes, it's possible to access team folders in local mode, and when #1007 they work properly across restarts (while changing users, etc).

Issue: KBFS-2215


libfuse/dir.go
```diff
@@ -349,6 +349,17 @@ func (f *Folder) isWriter(ctx context.Context) (bool, error) {
 	}
 	f.handleMu.RLock()
 	defer f.handleMu.RUnlock()
+	if f.h.Type() == tlf.SingleTeam {
+		tid, err := f.h.FirstResolvedWriter().AsTeam()
+		if err != nil {
+			return false, err
+		}
+		isWriter, err := f.fs.config.KBPKI().IsTeamWriter(ctx, tid, session.UID)
```
songgao: It seems this will be called pretty often. Do we need a short TTL cache in `libkbfs`?

strib: I'm planning to cache team membership stuff within `keybase_service_base`, so these calls won't be real RPCs most of the time.  Does that address your concern?

songgao: Yes, perfectly!

libkbfs/keybase_daemon.go
```diff
@@ -59,13 +60,44 @@ func (k keybaseDaemon) NewKeybaseService(config Config, params InitParams, ctx C
 	localUID := localUsers[userIndex].UID
 	codec := config.Codec()
 
+	teams := MakeLocalTeams([]libkb.NormalizedUsername{"kbfs", "core", "dokan"})
+	for i := range teams {
+		teams[i].Writers = make(map[keybase1.UID]bool)
+		teams[i].Readers = make(map[keybase1.UID]bool)
+		switch teams[i].Name {
+		case "kbfs":
+			teams[i].Writers[localUsers[0].UID] = true // strib
```
jzila: Where are these defined?

strib: Just above here in the same function.

jzila: Hah, I guess all I had to do was click.

PR #1010, opened 2017-05-29
libdokan: support top-level team folderlist in devel mode

strib:
Issue: KBFS-2215

### DIFF
--- a/libdokan/fs.go
+++ b/libdokan/fs.go
@@ -16,6 +16,7 @@ import (
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/kbfs/dokan"
 	"github.com/keybase/kbfs/dokan/winacl"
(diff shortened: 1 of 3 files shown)


libdokan/fs.go
```diff
@@ -370,6 +377,20 @@ func (f *FS) open(ctx context.Context, oc *openContext, ps []string) (dokan.File
 			return nil, false, dokan.ErrAccessDenied
 		}
 		return f.root.private.open(ctx, oc, ps[1:])
+	case `TEAM` == ps[0]:
```
songgao: Not necessarily related to this PR, but I was hoping `TEAM`, `PRIVATE`, and `PUBLIC` would be defined in libdokan/common.go like their lower-cased friends :)

PR #1013, opened 2017-05-31
increase duration in EnableDelayedCancellationWithGracePeriod

songgao:
The test passed `repeat 100 ./libkbfs.test -test.run=TestDelayedCancellationEnabled` on my machine, so I assume this is just due to CI slowness (or high parallelism). This PR increases the delayed duration so the flake is less likely to happen. It increases (the major wait time of) the test from 15ms to 50ms, which is not too bad for a single test.

But in the future if this is still flaky, and we don't want to bump it to e.g. 200ms, we should probably just skip this test.

### DIFF
--- a/libkbfs/delayed_cancellation_test.go
+++ b/libkbfs/delayed_cancellation_test.go
@@ -141,7 +141,7 @@ func TestDelayedCancellationEnabled(t *testing.T) {
 	select {
 	case <-ctx.Done():
 		t.Fatalf("Cancellation is not delayed")
-	case <-time.After(20 * time.Millisecond):
+	case <-time.After(10 * time.Millisecond):
 	}
 
 	<-ctx.Done()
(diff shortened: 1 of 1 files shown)


libkbfs/delayed_cancellation_test.go
```diff
@@ -134,14 +134,14 @@ func TestDelayedCancellationEnabled(t *testing.T) {
...
-	EnableDelayedCancellationWithGracePeriod(ctx, 15*time.Millisecond)
+	EnableDelayedCancellationWithGracePeriod(ctx, 50*time.Millisecond)
 
 	cancel()
 
 	select {
 	case <-ctx.Done():
 		t.Fatalf("Cancellation is not delayed")
-	case <-time.After(10 * time.Millisecond):
+	case <-time.After(20 * time.Millisecond):
```
strib: Can we keep this at 10 to reduce the probability of a flake?  Or are you afraid this would be too likely to give a false positive if delayed cancellation is every actually broken?

songgao: Yeah I was concerned about false positive. But I guess if it happens even 1 out of 5, it's still effective. So I'll change this back to 10.

songgao:
@jzila Yeah it's all over the place :( ; I was hoping we could retire the delayed cancellation stuff sometime soon

PR #1015, opened 2017-06-01
improve testing for single-team TLFs

strib:
This adds handle and DSL tests for single team TLFs, including some new error checking code for handle creation.

Issue: KBFS-2238

### DIFF
--- a/tlf/handle_test.go
+++ b/tlf/handle_test.go
@@ -102,6 +102,50 @@ func TestMakeHandleFailures(t *testing.T) {
 
 	_, err = MakeHandle(w, r[:1], nil, ur, nil)
 	assert.Equal(t, errInvalidReader, err)
+
+	// Any handle with a team must be a single-team handle.
+	w = []keybase1.UserOrTeamID{
+		keybase1.MakeTestTeamID(3).AsUserOrTeam(),
+		keybase1.MakeTestTeamID(4).AsUserOrTeam(),
+	}
+	r = []keybase1.UserOrTeamID{}
+	_, err = MakeHandle(w, r, nil, nil, nil)
+	assert.Equal(t, errInvalidWriter, err)
+
+	w = []keybase1.UserOrTeamID{
+		keybase1.MakeTestUID(4).AsUserOrTeam(),
+		keybase1.MakeTestTeamID(3).AsUserOrTeam(),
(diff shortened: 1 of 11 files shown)


tlf/handle_test.go
```diff
@@ -241,6 +285,30 @@ func TestHandleAccessorsPublic(t *testing.T) {
...
+		keybase1.MakeTestTeamID(4).AsUserOrTeam(),
+	}
+
+	h, err := MakeHandle(
+		w, nil, nil, nil, nil)
+	require.NoError(t, err)
+
+	require.Equal(t, SingleTeam, h.Type())
+
+	// All types of IsWriter/IsReader calls should panic for team TLFs.
```
songgao: Not really related to this PR; just thinking out loud:

Is this the plan (that tlf/handle.go panics on team) for the future as well? Do we use separate code for handling team membership through service?

If so, should they still live in the `tlf` package? Maybe it can accept an interface which does the actual checks on the membership. WDYT?

strib: Yeah, my tentative plan is that `tlf/handle.go` would continue to panic, to avoid having to expand that package to import contexts and deal with potentially blocking RPC calls.  Maybe that's a bad call, I don't really know, but I think it was initially designed to be as small as possible.

songgao: I see. That makes sense.

Although it seems better to have one place that does the check (including deciding whether we need to query service or API server, or just based on the handle itself), instead of having to do the judgement everywhere `IsWriter`/`IsReader` is called. If not `tlf`, perhaps a `tlfutils` package?

strib: Yeah, I thought about that.  Not really sure how to consolidate them.  There are three `IsTeamWriter()` calls in libkbfs right now that all do slightly different things: `bare_root_metadata_v3.go` (checks that the actual _device_ is keyed, not just that the user is a writer), `root_metadata.go` (pretty standard check using a `libkbfs.TlfHandle`) and `mdserver_local_shared.go` (checks for a writer OR a rekeyer, using a `tlf.Handle`).

Once the server is ready to do these checks, we can think about factoring one of these out into a nice function, but I'm not sure which one is best yet. But if you have a good idea how to improve it now, I'm open to it.

PR #1016, opened 2017-06-05
libdokan: Provide fake root directory for other SIDs

taruti:
Provides a fake root directory for different SIDs. Should help with:
https://github.com/keybase/client/issues/5490 and more recent issues.

The basic issue is with corporate AV products doing:
1) try to access K:
2) access denied because AV running with different user/permissions
3) goto 1 in a hot loop

To test this patch using  `-mount-flags 0` when mounting kbfsdokan + switch user is the simplest way.

@zanderz is the error folder name ok for you?

### DIFF
--- a/libdokan/fs.go
+++ b/libdokan/fs.go
@@ -352,7 +352,7 @@ func (f *FS) open(ctx context.Context, oc *openContext, ps []string) (dokan.File
 	case ".kbfs_unmount" == ps[0]:
 		os.Exit(0)
 	case ".kbfs_number_of_handles" == ps[0]:
-		x := f.stringReadFile(strconv.Itoa(int(oc.fi.NumberOfFileHandles())))
+		x := stringReadFile(strconv.Itoa(int(oc.fi.NumberOfFileHandles())))
 		return oc.returnFileNoCleanup(x)
 	// TODO
 	// Unfortunately sometimes we end up in this case while using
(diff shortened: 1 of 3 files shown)


libdokan/fs.go
```diff
@@ -538,6 +545,18 @@ func (f *FS) MoveFile(ctx context.Context, source *dokan.FileInfo, targetPath st
 	return nil
 }
 
+// isPotentialRenamePath filters out some special paths
```
strib: nit: can you please move this up to above the use?

libdokan/fs.go
```diff
@@ -272,8 +272,8 @@ func newSyntheticOpenContext() *openContext {
 func (f *FS) CreateFile(ctx context.Context, fi *dokan.FileInfo, cd *dokan.CreateData) (dokan.File, bool, error) {
 	// Only allow the current user access
 	if !fi.IsRequestorUserSidEqualTo(currentUserSID) {
-		f.log.Errorf("Refusing access: SID match error")
-		return nil, false, dokan.ErrAccessDenied
+		f.log.Errorf("FS CreateFile - Refusing real access: SID match error")
+		return openFakeRoot(ctx, f, fi)
```
strib: We're always guaranteed on Windows that the root will get a `CreateFile` call first on an empty string, before file listing is allowed or any other specific files can be looked up?

taruti: The only ops not needing an opened file are CreateFile, GetDiskFreeSpace and GetVolumeInformation. MoveFile (rename) is also interesting from a security pov, but it needs an open file. Fakeroot takes care that only \ and \errorfilename succeed at CreateFile.

taruti:
Also changed from a folder to a file as per CR and moved the one comment around.

PR #1017, opened 2017-06-05
libkbfs: don't set the PublicTopLevelFolder arg anymore

strib:
As of keybase/client#7270, the client no longer uses this field, and
we need to remove it from the protocol now that team TLFs are a thing.
This removes KBFS usage, so that we can remove it from the protocol in
a future PR.

Issue: KBFS-2116


strib:
> Will we be adding a TopLevelFolderType arg instead?

Yeah probably a good idea, although right now all the service code just infers it from the canonical pathname for the file, so it's not strictly needed.

PR #1021, opened 2017-06-06
fix paper key prompt spamming and race condition during shutdown

songgao:
The paper key prompt spamming is because when FBO concludes it would need paper key for rekey to proceed, it not only requests for the paper key, but also writes a MD to mdserver with rekey set at the same time. This causes the mdserver to call `FoldersNeedRekey` into KBFS client, which sends another rekey request into the rekey FSM, which kicks off the FSM into `rekeyStateStarted`. This PR makes sure `prmoptPaper` is not set before possibly shortening the rekey wait timer.

Also I found a bug with `atomic` usage for closing the `reqs` channel. Should've used `CompareAndSwapInt32`. This PR changes it to use a `RWMutex` which is less likely to become a foot gun.

### DIFF
--- a/libkbfs/rekey_fsm.go
+++ b/libkbfs/rekey_fsm.go
@@ -262,11 +262,11 @@ func (r *rekeyStateScheduled) reactToEvent(event RekeyEvent) rekeyState {
 	case rekeyRequestEvent:
 		if r.task.promptPaper && !event.request.promptPaper {
 			// KBFS-2251: If fbo concludes that paper key would be needed in
-			// order for rekey to proceed, it write a MD to mdserver with rekey
-			// set at the same time. To prevent the FSM from being kicked of to
-			// rekeyStateStarted right away after receivng this update (through
-			// FoldersNeedRekey) from mdserver, we just reuse the same timer if
-			// r.task.promptPaper is set.
+			// order for rekey to proceed, it writes a MD to mdserver with
+			// rekey set at the same time. To prevent the FSM from being kicked
+			// of to rekeyStateStarted right away after receiving this update
+			// (through FoldersNeedRekey) from mdserver, we just reuse the same
+			// timer if r.task.promptPaper is set.
 			//
 			// If the request has promptPaper set, then it's from the KBFS
 			// client, likely due to a read request. In this case, we should
@@ -406,8 +406,8 @@ type rekeyFSMListener struct {
 }
 
 type rekeyFSM struct {
-	muReqs sync.RWMutex
-	reqs   chan RekeyEvent
+	reqs       chan RekeyEvent
+	shutdownCh chan struct{} // closed on rekeyShutdownEvent; never written
 
 	fbo *folderBranchOps
 	log logger.Logger
@@ -418,19 +418,13 @@ type rekeyFSM struct {
 	listeners   map[rekeyEventType][]rekeyFSMListener
 }
 
-func (m *rekeyFSM) closeReqsCh() {
-	m.muReqs.Lock()
-	defer m.muReqs.Unlock()
-	close(m.reqs)
-	m.reqs = nil
-}
-
 // NewRekeyFSM creates a new rekey FSM.
 func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
 	fsm := &rekeyFSM{
-		reqs: make(chan RekeyEvent, rekeyQueueSize),
-		fbo:  fbo,
-		log:  fbo.config.MakeLogger("RekeyFSM"),
+		reqs:       make(chan RekeyEvent, rekeyQueueSize),
+		shutdownCh: make(chan struct{}),
+		fbo:        fbo,
+		log:        fbo.config.MakeLogger("RekeyFSM"),
 
 		listeners: make(map[rekeyEventType][]rekeyFSMListener),
 	}
@@ -440,38 +434,40 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
 }
 
 func (m *rekeyFSM) loop() {
-	for e := range m.reqs {
-		if e.eventType == rekeyShutdownEvent {
-			m.closeReqsCh()
+	for {
+		select {
+		case e := <-m.reqs:
+			if e.eventType == rekeyShutdownEvent {
+				close(m.shutdownCh)
+			}
+
+			next := m.current.reactToEvent(e)
+			m.log.Debug("RekeyFSM transition: %T + %s -> %T",
+				m.current, e, next)
+			m.current = next
+
+			m.triggerCallbacksForTest(e)
+			continue
+		default:
 		}
-
-		next := m.current.reactToEvent(e)
-		m.log.Debug("RekeyFSM transition: %T + %s -> %T",
-			m.current, e, next)
-		m.current = next
-
-		m.triggerCallbacksForTest(e)
+		select {
+		case <-m.shutdownCh:
+			return
+		default:
+		}
 	}
 }
 
 // Event implements RekeyFSM interface for rekeyFSM.
 func (m *rekeyFSM) Event(event RekeyEvent) {
-	for {
-		// We use a spinning loop here to avoid doing a naked `m.reqs <-
-		// event`, which in racy conditions could cause a panic due to sending
-		// to closed channel. The spinning loop is fine here since reactToEvent
-		// implementations of states are supposed to consume events from m.reqs
-		// pretty fast.
-		m.muReqs.RLock()
-		if m.reqs == nil {
-			return
-		}
-		select {
-		case m.reqs <- event:
-			return
-		default:
-		}
-		m.muReqs.RUnlock()
+	select {
+	case <-m.shutdownCh:
+		return
+	default:
+	}
+	select {
+	case <-m.shutdownCh:
+	case m.reqs <- event:
 	}
 }
 
--- a/libkbfs/rekey_fsm.go
+++ b/libkbfs/rekey_fsm.go
@@ -406,8 +406,11 @@ type rekeyFSMListener struct {
 }
 
 type rekeyFSM struct {
-	reqs       chan RekeyEvent
-	shutdownCh chan struct{} // closed on rekeyShutdownEvent; never written
+	reqsMtx sync.Mutex
+	reqs    chan RekeyEvent
+
+	shutdownMtx sync.Mutex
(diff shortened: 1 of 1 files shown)


libkbfs/rekey_fsm.go
```diff
@@ -261,6 +260,22 @@ func (r *rekeyStateScheduled) reactToEvent(event RekeyEvent) rekeyState {
 	case rekeyTimeupEvent:
 		return newRekeyStateStarted(r.fsm, r.task)
 	case rekeyRequestEvent:
+		if r.task.promptPaper && !event.request.promptPaper {
+			// KBFS-2251: If fbo concludes that paper key would be needed in
+			// order for rekey to proceed, it write a MD to mdserver with rekey
```
jzila: s/write/writes

libkbfs/rekey_fsm.go
```diff
@@ -261,6 +260,22 @@ func (r *rekeyStateScheduled) reactToEvent(event RekeyEvent) rekeyState {
...
+			// set at the same time. To prevent the FSM from being kicked of to
+			// rekeyStateStarted right away after receivng this update (through
```
jzila: s/receivng/receiving

libkbfs/rekey_fsm.go
```diff
@@ -391,8 +406,8 @@ type rekeyFSMListener struct {
 }
 
 type rekeyFSM struct {
-	reqs       chan RekeyEvent
-	reqsClosed int32
+	muReqs sync.RWMutex
```
jzila: I believe `reqsMtx` is our standard naming scheme.

libkbfs/rekey_fsm.go
```diff
@@ -435,25 +456,23 @@ func (m *rekeyFSM) loop() {
...
+		// to closed channel. The spinning loop is fine here since reactToEvent
 		// implementations of states are supposed to consume events from m.reqs
 		// pretty fast.
-		for atomic.LoadInt32(&m.reqsClosed) == 0 {
-			select {
-			case m.reqs <- event:
-				return
-			default:
-			}
+		m.muReqs.RLock()
```
jzila:
```go
for {
    func() {
        m.reqsMtx.RLock()
        defer m.reqsMtx.RUnlock()
        ...
    }()
}
```
Otherwise it won't get unlocked when it should.

libkbfs/rekey_fsm.go
```diff
@@ -435,25 +456,23 @@ func (m *rekeyFSM) loop() {
...
-	default:
-	}
-	go func() {
-		// We use a spinning loop here to avoid doing a naked `m.reqs <- event`,
-		// which in racy conditions could cause a panic due to sending to closed
-		// channel. The spinning loop is fine here since reactToEvent
+	for {
+		// We use a spinning loop here to avoid doing a naked `m.reqs <-
+		// event`, which in racy conditions could cause a panic due to sending
+		// to closed channel. The spinning loop is fine here since reactToEvent
```
jzila: This would be cleaner if you used a `closedCh` instead of a mutex, because you could select on that channel in the read loop, and never close the write channel. That way you could avoid the whole spin loop.

libkbfs/rekey_fsm.go
songgao: oops thanks!

libkbfs/rekey_fsm.go
songgao: Hmm but if we are fine with leaving a channel open after shutdown, we could also just leave the read loop waiting on it; just don't write any new stuff into it. Maybe we should do that. What do you think?

jzila: Leaving the read loop will leak a goroutine, but leaving a channel open won't. Plus, using channels & channels for synchronization is clean because they can go in the same `select`. Mutexes can't.

In general, if you're guarding channels with mutexes, there's probably something wrong.

songgao: Good point; will do. Thanks!

libkbfs/rekey_fsm.go
```diff
@@ -420,8 +442,7 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
 func (m *rekeyFSM) loop() {
 	for e := range m.reqs {
 		if e.eventType == rekeyShutdownEvent {
-			atomic.StoreInt32(&m.reqsClosed, 1)
-			close(m.reqs)
+			m.closeReqsCh()
```
jzila:
```go
for {
    select {
    case <-m.reqs:
        if e.eventType == rekeyShutdownEvent {
            close(m.closedCh)
        }
        ...
    case <-m.closedCh:
    }
}
```

libkbfs/rekey_fsm.go
songgao: what about the sender part though? If the reader loop exits, the sender would need to block at some point

jzila: Sender part looks like this:
```go
select {
case m.reqs <- foo:
    // do stuff
case <-m.closedCh:
    // don't do stuff
}
```

songgao: Oh I see!

songgao: Hmm since `select{}` selects communication that can proceed randomly, it seems this can't guarantee requests won't go in after `m.closedCh` is closed, if `m.reqs` is a buffered channel.

songgao: Ah I guess two selects then

jzila: Why do you want a buffered channel if you need to guarantee that no writes go in after it's closed?

jzila: If it's a buffered channel, then you can accumulate many requests in the channel after the shutdown event.

songgao: yes but the reader can keep consuming it. The buffered channel is pre-existing, and we are using it as a queue.

jzila: Right but why not make it unbuffered?

songgao: Its buffer size is the queue size

songgao:
> Is it terrible to have the channel theoretically accept events after it's closed?
Do you mean literally closed or shutdown? If shutdown, I don't think there's anything wrong other than leaking channel with stuff in it. I don't know enough about whether GC collects it in this case; if not, isn't that not much better than leaking gorouthines?

songgao:
I don't think it's racy in this case. It guarantees everything in the channel is consumed. Maybe I missed something?

libkbfs/rekey_fsm.go
```diff
@@ -418,42 +434,41 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
...
-			select {
-			case m.reqs <- event:
-				return
-			default:
-			}
-		}
-	}()
+	select {
+	case <-m.shutdownCh:
+	case m.reqs <- event:
```
jzila: `m.shutdownCh` is closed between the above select and this one, so in this select there's a probability that either an event is written or the shutdown is respected.

Anyway I'm still not happy with this approach: we're working around the fact that you have a buffer that you want, which you want to control the input to, but you want to guarantee that all events have been flushed.

Seems like you need both:
```go
shutdownCh chan struct{}
doneCh chan struct{}
```

jzila: `shutdownCh` gets closed when the shutdown is noticed, and `doneCh` gets closed after the last events have been flushed.

songgao: > so in this select there's a probability that either an event is written or the shutdown is respected

Yes there is, but the event that's written will be consumed by the reader for sure (given that there are two selects in reader's loop). The `shutdownCh` case here is just to prevent it from blocking if it's already shutdown.

I don't feel like this approach either. We do need a buffered channel here as a queue. I guess an alternative is to have a mutex guarded `closed` variable; but that's no different than having the mutex guarding the channel (reference).

jzila: Right so if you have a `shutdownCh` and a `doneCh`, you get your 3-state synchronization.

songgao:
@jzila I can't think of a good way to make it clean and race free at the same time, so here's a version using `sync.Mutex`. I think the ultimate issue is there can be event coming in after shutdown and it's hard to prevent them for sure.

jzila:
I'll take a look again in the morning.

songgao:
OK thanks @jzila !

songgao:
@jzila I realized GC on channels has nothing to do about whether the channel has content in it, so there's no point preventing future sends. This version uses your approach with the `shutdownCh`. Sorry for my ignorance earlier!

jzila:
Yeah I was confused about your reasoning there which is why I punted.

libkbfs/rekey_fsm.go
```diff
@@ -418,42 +434,30 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
...
-		// implementations of states are supposed to consume events from m.reqs
-		// pretty fast.
-		for atomic.LoadInt32(&m.reqsClosed) == 0 {
-			select {
-			case m.reqs <- event:
-				return
-			default:
-			}
+		if event.eventType == rekeyShutdownEvent {
+			close(m.shutdownCh)
```
jzila: Why did you choose to close in the writer rather than the reader? To give the reader potentially longer to flush remaining events?

songgao: This was because of the principle that writer is supposed to signal the end. But I just realized this means multiple calls to `close(m.shutdownCh)` could happen multiple times, so will move it into the reader loop.

libkbfs/rekey_fsm.go
```diff
@@ -437,6 +437,15 @@ func (m *rekeyFSM) loop() {
 	for {
 		select {
 		case e := <-m.reqs:
+			if e.eventType == rekeyShutdownEvent {
+				// Set m.reqs to nil so on next iteration, we will skip any
+				// content in m.reqs. So if there are multiple
+				// rekeyShutdownEvent, we won't close m.shutdownCh multiple
+				// times.
+				m.reqs = nil
```
jzila: I'm not sure this will work the way you want it to; if another goroutine tries to send to `m.reqs` at the same time, it seems like there would be a race there. `close` is atomic, assignment is not.

songgao: Ah you are right. Thanks!

PR #1023, opened 2017-06-07
libfuse: don't send error notifications for ._ files

strib:
If something on the client tries to make a metadata file under
/keybase or /keybase/{private,public}, it will fail.  But there's no
reason to bother the user with a notification about that, so just
return the bare error code instead of the typed error that will cause
a notification.

Issue: keybase/client#7306

### DIFF
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -151,11 +151,6 @@ func (fl *FolderList) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (_ fs.N
 	fl.fs.log.CDebugf(ctx, "FL Mkdir")
 	tlfName := libkbfs.CanonicalTlfName(req.Name)
 	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
-	if strings.HasPrefix(req.Name, "._") {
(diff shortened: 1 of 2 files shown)


libfuse/folderlist.go
```diff
@@ -146,6 +151,11 @@ func (fl *FolderList) Mkdir(ctx context.Context, req *fuse.MkdirRequest) (_ fs.N
 	fl.fs.log.CDebugf(ctx, "FL Mkdir")
 	tlfName := libkbfs.CanonicalTlfName(req.Name)
 	defer func() { fl.reportErr(ctx, libkbfs.WriteMode, tlfName, err) }()
+	if strings.HasPrefix(req.Name, "._") {
```
songgao: I think the `._`s are always files and never directories, no matter the thing it corresponds to is a directory or a file.

(same for `fs.go`)

strib: Ok . . . are you saying you want to have the notification if someone does try to make a `._` directory?  I don't care much either way.

songgao: Yes, since that would be an intentional operation, rather than macOS kernel. The latter could happen when user does stuff in Finder on legitimate directories, but the former can only be explicitly done by user (or a program)

PR #1024, opened 2017-06-07
key_manager_test: fix tests now that some rekey events are ignored

strib:
Now that the rekey FSM ignores rekey requests when the promptPaper
flag is already set, it caused some tests to wait for long timeouts
for the events they sent to simulate the server returning their own
rekey-bit-set update.  So instead, don't wait for a complete rekey
event to finish in that case, just wait for the simulation event to be
processed.

### DIFF
--- a/libkbfs/key_manager_test.go
+++ b/libkbfs/key_manager_test.go
@@ -1522,6 +1522,33 @@ func testKeyManagerReaderRekeyAndRevoke(t *testing.T, ver MetadataVer) {
 	}
 }
 
+func keyManagerTestSimulateSelfRekeyBit(
+	t *testing.T, ctx context.Context, config Config, tlfID tlf.ID) {
+	// Simulate the mdserver sending back this node's own rekey
+	// request.  This shouldn't increase the MD version.  Since this
+	// doesn't kick off a rekey request, we don't need to wait for the
+	// whole rekey to finish; we just wait for the event to be
+	// processed.
+	rekeyWaiter := make(chan struct{})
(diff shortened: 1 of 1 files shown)


libkbfs/key_manager_test.go
```diff
@@ -1638,6 +1670,14 @@ func testKeyManagerRekeyBit(t *testing.T, ver MetadataVer) {
 		t.Fatalf("Device 2 couldn't read a: %+v", err)
 	}
 
+	// Cancel the original promptPaper timer for user2 now that the
+	// TLF is readable; otherwise the rekey later on will hang because
+	// it looks like prompt for paper was already set.  TODO: should
```
strib: @songgao: what do you think about this TODO?  Are we going to have problems where device1 requests a rekey, then device2 completes the rekey and goes offline, then later device1 tries to rekey again for a new device, and has to wait for the timer to expire before it happens?

songgao: Good point! Do you think it's better to just stop the timer, or unset the promptPaper flag in this case? If we just unset promptPaper, it would be a no-op when the timer is up if it's not needed, but still covers the case where a rekey is actually needed again. But I guess that's rare enough to ignore here?

strib: I guess stopping when you see a rekey from another device is fine, because if another one is needed, presumably that device will check for it again in 10s.  I'll file a nonurgent ticket for this.

songgao: OK sounds good thanks!

PR #1030, opened 2017-06-08
[KBFS-2208] block_retrieval_worker_test: Add logging to track down flake

jzila:


libkbfs/block_retrieval_worker_test.go
```diff
@@ -263,7 +263,9 @@ func TestBlockRetrievalWorkerShutdown(t *testing.T) {
 	shutdown := false
 	select {
 	case <-ch:
+		t.Fatal("Expected not to retrieve a result from the Request.")
```
songgao: This doesn't fix the flake does it? It still flakes the same way I think?

songgao:
Also I was a bit confused by https://github.com/keybase/kbfs/pull/1030/files#diff-e3ef24ce4d67246c2920d7d276db5648R273 and checking the `shutdown` variable again. The variable doesn't change after that call

libkbfs/block_retrieval_worker_test.go
jzila: Yeah but at least we know which scenario is happening.

jzila:
Yeah that check doesn't really do anything.

libkbfs/block_retrieval_worker_test.go
songgao: Ah got you. I didn't read the PR title fully

libkbfs/block_retrieval_worker_test.go
```diff
@@ -263,13 +263,26 @@ func TestBlockRetrievalWorkerShutdown(t *testing.T) {
...
 	case continueCh <- nil:
+		t.Fatal("Expected the block getter not to be receiving.")
 	default:
 		shutdown = true
 	}
 	require.True(t, shutdown)
-	w.Shutdown()
-	require.True(t, shutdown)
+
+	// Ensure the test completes in a reasonable time.
```
jzila: @songgao: I replaced the last check with the right one.

songgao: nice!

libkbfs/block_retrieval_worker_test.go
```diff
@@ -263,13 +263,26 @@ func TestBlockRetrievalWorkerShutdown(t *testing.T) {
...
+	timer := time.NewTimer(10 * time.Second)
+	doneCh := make(chan struct{})
+	go func() {
+		w.Shutdown()
```
songgao: Is `w.Shutdown()` synchronous? https://github.com/keybase/kbfs/pull/1030/files#diff-e3ef24ce4d67246c2920d7d276db5648R264 seems to suggest it's async?

jzila: Yes. This tests that.

jzila: Hence the `t.Fatal("Expected another Shutdown not to block.")`

PR #1031, opened 2017-06-09
tlf_journal: enable single-team journals with team ID

strib:
In a future PR, the `tlfJournal` will need to know the team ID
corresponding to a single-team TLF, so that it can check the team's
quota when limiting space, rather than the user's quota.
Unfortunately, getting the team ID here is not very convenient.

This PR plumbs it through when the journal is created in the first
place, if possible.  If not possible, it trusts the MD server to
provide the handle, given a TLF ID.  It saves the team ID to the
journal info file on disk, so it can be used without needing server
access after a restart.

Issue: KBFS-2217

### DIFF
--- a/libkbfs/journal_server.go
+++ b/libkbfs/journal_server.go
@@ -216,7 +216,8 @@ func (j *JournalServer) getEnableAutoLocked() (
 	return j.serverConfig.getEnableAuto(j.currentUID)
 }
 
-func (j *JournalServer) getTLFJournal(tlfID tlf.ID) (*tlfJournal, bool) {
+func (j *JournalServer) getTLFJournal(
+	tlfID tlf.ID, h *TlfHandle) (*tlfJournal, bool) {
 	getJournalFn := func() (*tlfJournal, bool, bool, bool) {
 		j.lock.RLock()
 		defer j.lock.RUnlock()
@@ -229,7 +230,7 @@ func (j *JournalServer) getTLFJournal(tlfID tlf.ID) (*tlfJournal, bool) {
 		ctx := context.TODO() // plumb through from callers
 		j.log.CDebugf(ctx, "Enabling a new journal for %s (enableAuto=%t, set by user=%t)",
 			tlfID, enableAuto, enableAutoSetByUser)
-		err := j.Enable(ctx, tlfID, TLFJournalBackgroundWorkEnabled)
+		err := j.Enable(ctx, tlfID, h, TLFJournalBackgroundWorkEnabled)
 		if err != nil {
 			j.log.CWarningf(ctx, "Couldn't enable journal for %s: %+v", tlfID, err)
 			return nil, false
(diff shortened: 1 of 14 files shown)


libkbfs/journal_server.go
```diff
@@ -482,12 +484,37 @@ func (j *JournalServer) enableLocked(
...
+	h *TlfHandle, bws TLFJournalBackgroundWorkStatus) (err error) {
 	j.lock.Lock()
 	defer j.lock.Unlock()
-	tj, err := j.enableLocked(ctx, tlfID, bws, false)
+	var tid keybase1.TeamID
+	if tlfID.Type() == tlf.SingleTeam {
+		if h == nil {
+			// We must have a handle for SingleTeam TLFs, so try to
+			// get one from the server.  This is an unstrusted mapping
+			// since it's just using whatever the server tells it.
```
songgao: s/unstrusted/untrusted

Also, would it be worthwhile to store whether the handle is trusted somewhere in the handle?

strib:
> Does it mean all newly created journal will have trusted mapping stored on device?

It means the first time a device access a single-team TLF, it will write the team ID corresponding to the TLF to disk, and it will live there for the lifetime of that journal.

> If so, since there can't be any existing journal for teams yet, is it possible if we require a none-nil handle to be passed in for team TLFs and never fetch it from mdserver?

The non-nil handle can only occur when the TLF is accessed for the first time by TLF ID, rather than through its complete name.  This is pretty rare.  The specific cases I can think of now are for rekey, where the server tells the device to rekey a TLF by ID, and that device may never have accessed it before.  That shouldn't happen for single-team TLFs since KBFS isn't involved in the rekeying.  Maybe there are no other cases like that, but I wasn't 100% sure and wanted to be safe about it.

So it's _probably_ safe to remove the part that fetches it, if you feel that would be better.  Let me know.

songgao:
> The non-nil handle can only occur when the TLF is accessed for the first time by TLF ID, rather than through its complete name.

Did you mean "nil handle can only occur ..."?

strib:
> Did you mean "nil handle can only occur ..."?

Oops, yes!  Will edit.

songgao:
What about the `enableAuto` [here](https://github.com/keybase/kbfs/pull/1031/files#diff-f816c61024359d85bb62ba228c9aa6b9R229)? Does it rely on TLF ID to create the journal? (I guess I'm just a little bit confused)

strib:
`enableAuto` is now on by default for every user, it was mostly useful when testing journaling.  In that function, `getTlfJournal()`, the handle is passed in by the caller, usually in `journal_md_ops` or `journal_block_server`.  Probably every time a user first accesses a TLF after a restart, it will go through `GetForHandle` first, which provides a non-nil handle to `getTlfJournal()`.  I don't think, but am not 100% sure, if there are other cases that first access the journal on other paths, besides the rekey one that doesn't happen for team TLFs.

songgao:
Ah I see; I saw a lot of `j.getTLFJournal(tlfID, nil)` calls but I guess they all (we think) happen after the first access through (as you said) `GetForHandle` so the journal should have already existed by then. Does it sounds correct?

We'll be testing the team feature for a while before shipping right? Perhaps we can disable the untrusted fetching for now and generate a loud error to know for sure?

strib:
@songgao: ok made it panic instead.  Please take another look!

PR #1033, opened 2017-06-09
backpressure_disk_limiter: track quotas by UserOrTeamID

strib:
So that we can compute pressure differently for private TLFs vs team TLFs.

The `tlfJournal` now stores a `ChargedTo` field, rather than a `TID`, so we can treat private TLFs and team TLFs symmetrically.

Also adds team quota support to the quota usage helper class.

Issue: KBFS-2217

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/keybase/client/go/logger"
+	"github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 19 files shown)


libkbfs/backpressure_disk_limiter.go
```diff
@@ -401,13 +404,38 @@ func newJournalTracker(
 		return journalTracker{}, err
 	}
 
-	journalQuotaTracker, err := newQuotaBackpressureTracker(
-		quotaMinThreshold, quotaMaxThreshold)
+	// Test quota parameters -- actually quota trackers will be
```
songgao: s/actually/actual ?

PR #1035, opened 2017-06-13
[KBFS-1895] Prefetching Workers

jzila:
I believe this is ready for review. I refactored the code a bit to spawn fewer goroutines in general, and I made the channel semantics more sane. The worker actually does the work now.

I'm also running the tests a bunch of times to minimize the likelihood of flakes.

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -176,6 +176,22 @@ func (brq *blockRetrievalQueue) shutdownRetrieval() {
 
 // notifyWorker notifies workers that there is a new request for processing.
 func (brq *blockRetrievalQueue) notifyWorker(priority int) {
+	// On-demand workers and prefetch workers share the priority queue. This
+	// allows maximum time for requests to jump the queue, at least until the
+	// worker actually begins working on it.
+	//
+	// Note that the worker being notified won't necessarily work on the exact
+	// request that caused the notification. It's just a counter. That means
+	// that sometimes on-demand workers will work on prefetch requests, and
+	// vice versa. But the numbers should match.
+	//
+	// However, there are some pathological scenarios where if all the workers
+	// of one type are making progress but the other type are not (which is
+	// highly improbable), requests of one type could starve the other. By
+	// design, on-demand requests _should_ starve prefetch requests, so this is
+	// a problem only if prefetch requests can starve on-demand workers. But
+	// because there are far more on-demand workers than prefetch workers, this
+	// should never acdtually happen.
 	workerCh := brq.workerCh
 	if priority < defaultOnDemandRequestPriority {
 		workerCh = brq.prefetchWorkerCh
@@ -186,15 +202,8 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
 	// Notify the next queued worker.
 	case workerCh <- struct{}{}:
 	default:
-		// Only launch a goroutine if we would otherwise block.
-		go func() {
-			select {
-			case <-brq.doneCh:
-				brq.shutdownRetrieval()
-			// Notify the next queued worker.
-			case workerCh <- struct{}{}:
-			}
-		}()
+		panic("notifyWorker() would have blocked, which means we somehow " +
+			"have around MaxInt32 requests already waiting.")
 	}
 }
 
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -191,7 +191,7 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
 	// design, on-demand requests _should_ starve prefetch requests, so this is
 	// a problem only if prefetch requests can starve on-demand workers. But
 	// because there are far more on-demand workers than prefetch workers, this
-	// should never acdtually happen.
+	// should never actually happen.
 	workerCh := brq.workerCh
 	if priority < defaultOnDemandRequestPriority {
 		workerCh = brq.prefetchWorkerCh
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -109,6 +109,10 @@ type blockRetrievalQueue struct {
 	// in the heap, allowing preemption as long as possible. This way, a
 	// request only exits the heap once a worker is ready.
 	workerQueue chan chan<- *blockRetrieval
+	// This is a channel of channels to maximize the time that each request is
+	// in the heap, allowing preemption as long as possible. This way, a
+	// request only exits the heap once a worker is ready.
+	prefetchWorkerQueue chan chan<- *blockRetrieval
 	// slices to store the workers so we can terminate them when we're done
 	workers []*blockRetrievalWorker
 	// channel to be closed when we're done accepting requests
@@ -125,7 +129,7 @@ var _ BlockRetriever = (*blockRetrievalQueue)(nil)
 // newBlockRetrievalQueue creates a new block retrieval queue. The numWorkers
 // parameter determines how many workers can concurrently call Work (more than
 // numWorkers will block).
-func newBlockRetrievalQueue(numWorkers int,
+func newBlockRetrievalQueue(numWorkers int, numPrefetchWorkers int,
 	config blockRetrievalConfig) *blockRetrievalQueue {
 	q := &blockRetrievalQueue{
 		config:      config,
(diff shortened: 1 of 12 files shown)


libkbfs/block_retrieval_worker.go
```diff
@@ -48,12 +48,12 @@ func newBlockRetrievalWorker(bg blockGetter,
 // blockGetter.getBlock, and responds to the subscribed requestors with the
 // results.
 func (brw *blockRetrievalWorker) HandleRequest() (err error) {
-	brw.queue.Work(brw.workCh)
 	var retrieval *blockRetrieval
 	select {
-	case retrieval = <-brw.workCh:
+	case <-brw.workCh:
+		retrieval = brw.queue.popIfNotEmpty()
```
strib: Since they share one queue, in theory prefetch workers could end up servicing on-demand requests, and once those are all serviced, an on-demand worker would end up servicing the prefetch requests that originally caused work to go on the prefetch work channel.  Is that right?

I guess it's not really a problem except that it if there were N prefecth requests that happened concurrently with a bunch of on-demand requests, they might end up lighting up N on-demand workers, which would tie them up for future on-demand requests and also cause more background CPU/network usage than we intended.

It might not even be that rare, since on-demand requests usually trigger the prefetch requests, and so if there's a batch of on-demand requests that's greater than the overall size of the on-demand worker pool, we could get into this situation pretty easily.

It's probably not worth worrying about, but I think it's worth a comment somewhere to remind ourselves that we're aware of it.  I guess to fix it you'd need two different queues and the ability to atomically move requests from one to the other when the priority changes.

libkbfs/block_retrieval_queue.go
```diff
@@ -16,10 +16,15 @@ import (
...
-	testBlockRetrievalWorkerQueueSize    int = 5
-	defaultOnDemandRequestPriority       int = 100
+	defaultBlockRetrievalWorkerQueueSize int  = 100
+	defaultPrefetchWorkerQueueSize       int  = 2
+	minimalBlockRetrievalWorkerQueueSize int  = 2
+	minimalPrefetchWorkerQueueSize       int  = 1
+	testBlockRetrievalWorkerQueueSize    int  = 5
+	testPrefetchWorkerQueueSize          int  = 1
+	defaultOnDemandRequestPriority       int  = 100
+	prefetchWorker                       bool = true
```
strib: I think these two bools are unused.

libkbfs/block_retrieval_queue.go
```diff
@@ -125,21 +131,27 @@ var _ BlockRetriever = (*blockRetrievalQueue)(nil)
...
-		ptrs:        make(map[blockPtrLookup]*blockRetrieval),
-		heap:        &blockRetrievalHeap{},
-		workerQueue: make(chan chan<- *blockRetrieval, numWorkers),
-		workers:     make([]*blockRetrievalWorker, 0, numWorkers),
-		doneCh:      make(chan struct{}),
+		config:           config,
+		log:              config.MakeLogger(""),
+		ptrs:             make(map[blockPtrLookup]*blockRetrieval),
+		heap:             &blockRetrievalHeap{},
+		workerCh:         make(chan struct{}, numWorkers),
```
jzila: I don't really understand the question...the "randomness" isn't real, because it's just a notification to start doing work. The request that actually gets picked up is determined in `popIfNotEmpty`, which only happens after a worker has started working.

libkbfs/block_retrieval_queue.go
```diff
@@ -154,17 +166,32 @@ func (brq *blockRetrievalQueue) popIfNotEmpty() *blockRetrieval {
...
-		retrieval := brq.popIfNotEmpty()
-		ch <- retrieval
+	// Notify the next queued worker.
+	case workerCh <- struct{}{}:
+	default:
+		// Only launch a goroutine if we would otherwise block.
+		go func() {
+			select {
+			case <-brq.doneCh:
+				retrieval := brq.popIfNotEmpty()
```
jzila: Hah I was afraid someone was going to suggest that.

It's 4 lines, but I can wrap it. I can't wrap the whole `select`, because the first one has a `default` and the second one does not.

libkbfs/block_retrieval_queue.go
```diff
@@ -340,19 +367,24 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context,
...
+		if br.index != -1 && priority > oldPriority {
 			br.priority = priority
 			heap.Fix(brq.heap, br.index)
+			if oldPriority < defaultOnDemandRequestPriority &&
+				priority >= defaultOnDemandRequestPriority {
+				// We've crossed the priority threshold for prefetch workers,
+				// so we now need an on-demand worker to pick up the request.
+				// This means that we might have up to two worker goroutines
+				// per request. However, they won't leak because if a worker
+				// sees an empty queue, it ends its goroutine.
```
jzila: Correct.

libkbfs/block_retrieval_worker.go
jzila: Yeah exactly, it's to reduce complexity. We're relying on the extra worker receiving a `nil` request from the queue.

libkbfs/block_retrieval_queue.go
strib: Good point about the request order.  But I still don't see why we'd rather have a thousand goroutines waiting to put on that channel, rather than a channel but a buffer of length 1000.  At the very least, it'll make debugging stack dumps easier.

libkbfs/block_retrieval_worker.go
strib: In the case I'm talking about, there's no extra worker that receives a `nil`.  No requests have changed priority -- it's just that prefetch workers happened to have started working on on-demand requests before an on-demand worker did, leading to a situation where the on-demand worker services the prefetch requests because that's all that's left when it checks the queue.

libkbfs/block_retrieval_queue.go
jzila: Because we have no idea what the actual limit is. And we need it to never block.

jzila: Oh I see your point. You want a multiplier on the channel buffer independent of the number of workers, so that we have to revert to goroutines less often.

jzila: Any ideas for the right size?

strib: Right.  I think buffers are fairly cheap -- 1000 seems reasonable to me (picked fairly arbitrarily).

libkbfs/block_retrieval_worker.go
jzila: I think that's okay as long as there's a quid-pro-quo. I work on your request and you'll work on mine. The same number of workers is allocated on both on-demand requests and prefetch requests, so no starvation should be introduced.

strib: There is potential starvation of the _next_ set of on-demand requests that come in, since some workers will be tied up with the previous prefetch requests already.  Also, extra background CPU and network usage.  I don't think it's a big deal, just want to make sure we acknowledge it somewhere and are aware of it.

This is still way better than the current situation.

libkbfs/block_retrieval_queue.go
jzila: K, will do.

libkbfs/block_retrieval_worker.go
jzila: Sort of, those on-demand workers "should" have been working on on-demand requests in this cycle, but a prefetch worker did it instead.

I don't follow how we introduce extra CPU/network usage (other than to flush the extra requests).

libkbfs/block_retrieval_queue.go
jzila: Ooh also now that we use `chan struct{}` instead of actual data, I think buffers are free:
https://dave.cheney.net/2014/03/25/the-empty-struct

libkbfs/block_retrieval_worker.go
strib: 1. 100 on-demand requests come in.
2. 2 prefetch requests come in.
3. The 2 prefetch workers pick up 2 of the on-demand requests.
4. 98 on-demand requests and 2 prefetch requests are picked by on-demand requests.
5. The prefetch workers finish.
6. 2 more on-demand requests come in.
7. 2 prefetch requests come in.
8. The 2 prefetch workers pick up the 2 on-demand requests.
9. 2 of the on-demand workers free up from servicing on-demand requests and pick up the 2 prefetch requests.
10. At this point, we now have 4 simultaneous prefetch requests in flight, using up CPU and network that we'd rather be using for the on-demand requests.  The point of this patch is to limit the amount of CPU and network used by prefetching, right?
11. Repeat this pattern over and over, until all on-demand workers are doing prefetching, while new on-demand requests block.

jzila: OK I agree with the logic, but I don't agree that it can realistically happen. It'd require the prefetching workers to be successfully completing every request and none of the on-demand workers to be. On average I suspect that this would be rare, with a small percentage of prefetch requests addressed by on-demand workers.

I could instrument it to see how often it happens though.

strib: Agree it would be very rare, just thought it was worth acknowledging that this scheme isn't foolproof.

No need to instrument anything, would be a pain to plumb the log through.  A comment somewhere is good enough.

jzila: Comment added [here](https://github.com/keybase/kbfs/pull/1035/commits/f8c1388e205840cc3db70c751f41911197989c66)

PR #1037, opened 2017-06-14
test: wait for journal to flush in TestTeamsTwoWritersJournal

strib:
At the end of the `as` clause, bob does a SyncAll.  But we also need
to do a flush before alice can see the new data, since bob is using a
journal.

Issue: KBFS-2260

### DIFF
--- a/test/teams_test.go
+++ b/test/teams_test.go
@@ -75,10 +75,18 @@ func TestTeamsTwoWritersJournal(t *testing.T) {
 			enableJournal(),
 			mkfile("a", "hello"),
 		),
+		as(alice,
+			// Wait for the flush, after doing a SyncAll().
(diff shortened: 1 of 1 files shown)


jzila:
Are there any other tests that could flake this way? Should we make the flush automatic at the end of the `as` clause?

strib:
> Are there any other tests that could flake this way? Should we make the flush automatic at the end of the as clause?

@jzila good question. The other journal tests all do explicit flushing because they disable automatic flushing as part of the test, and only want it flushed at a certain time.  So I'm leery of doing it automatically, for fear of making the tests less flexible.  But happy to revisit if we accumulate more tests like this.

jzila:
Understood. I'll watch out for it.

PR #1040, opened 2017-06-15
folder_branch_ops: remove unlinked files from dirty list

strib:
Once we clear the cache info for the file, they shouldn't be considered dirty anymore.

Also add test infrastructure to be able to check dirty paths from the DSL tests.

Issue: KBFS-2272

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -3753,8 +3753,12 @@ func (fbo *folderBranchOps) startSyncLocked(ctx context.Context,
 		// perfectly accurate (but at the same time, we'd then have to
 		// fix up the intentional panic in the background flusher to
 		// be more tolerant of long-lived dirty, removed files).
-		return false, true, nil, nil, nil, fileSyncState{}, nil,
-			fbo.blocks.ClearCacheInfo(lState, file)
+		err := fbo.blocks.ClearCacheInfo(lState, file)
+		if err != nil {
+			return false, false, nil, nil, nil, fileSyncState{}, nil, err
+		}
+		fbo.status.rmDirtyNode(node)
+		return false, true, nil, nil, nil, fileSyncState{}, nil, nil
 	}
 
 	if file.isValidForNotification() {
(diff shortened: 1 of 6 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -3753,8 +3753,12 @@ func (fbo *folderBranchOps) startSyncLocked(ctx context.Context,
...
 		// fix up the intentional panic in the background flusher to
 		// be more tolerant of long-lived dirty, removed files).
-		return false, true, nil, nil, nil, fileSyncState{}, nil,
-			fbo.blocks.ClearCacheInfo(lState, file)
+		err := fbo.blocks.ClearCacheInfo(lState, file)
+		if err != nil {
+			return false, false, nil, nil, nil, fileSyncState{}, nil, nil
+		}
+		fbo.status.rmDirtyNode(node)
+		return false, true, nil, nil, nil, fileSyncState{}, nil, nil
```
songgao: Should the second variable be `false` instead of `true`?

strib: If it's `false`, the node will be removed by the caller redundantly.

I was thinking about getting rid of that second return param entirely, not sure if it's still meaningful, but I didn't want to change too much in this PR.

songgao: I see. Thanks

libkbfs/folder_branch_ops.go
```diff
@@ -3753,8 +3753,12 @@ func (fbo *folderBranchOps) startSyncLocked(ctx context.Context,
 		// perfectly accurate (but at the same time, we'd then have to
 		// fix up the intentional panic in the background flusher to
 		// be more tolerant of long-lived dirty, removed files).
-		return false, true, nil, nil, nil, fileSyncState{}, nil,
-			fbo.blocks.ClearCacheInfo(lState, file)
+		err := fbo.blocks.ClearCacheInfo(lState, file)
+		if err != nil {
+			return false, false, nil, nil, nil, fileSyncState{}, nil, nil
```
songgao: Is the `err` ignored here intentionally? Could you explain why?

strib: Oops, that's a bug.  Nice catch!

songgao: ha OK!

PR #1041, opened 2017-06-16
keybase_service_base: integrate with service for team info and merkle seqno

strib:
Note that the merkle root call is on the critical foreground write path, and will result in a server RTT if the merkle root isn't fresh enough.  I filed KBFS-2271 to move it into the background write path if possible, maybe by moving it into `RootMetadataSigned` rather than having it in `BareRootMetadata`.

Issue: KBFS-2228


libkbfs/keybase_service_base.go
```diff
@@ -405,16 +415,61 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
...
 
 // GetCurrentMerkleSeqNo implements the KeybaseService interface for
 // KeybaseServiceBase.
 func (k *KeybaseServiceBase) GetCurrentMerkleSeqNo(ctx context.Context) (
 	MerkleSeqNo, error) {
-	// TODO: call into the service once they implement a way to get
-	// the current seqno.
-	return UnknownMerkleSeqNo, nil
+	const merkleFreshness = int(time.Second * 60 / time.Millisecond)
+	res, err := k.merkleClient.GetCurrentMerkleRoot(ctx, merkleFreshness)
```
songgao: Does this guarantee that anything returned with no error is from the past minute?

kbfsfuse/kbfsfuse.sh
```diff
@@ -12,7 +12,7 @@ if [ -z "$KEYBASE_TEST_ROOT_CERT_PEM" ]; then
     export KEYBASE_TEST_ROOT_CERT_PEM="$(echo $KEYBASE_TEST_ROOT_CERT_PEM_B64 | base64 -d)";
 fi
 if [ -z "$KBFS_METADATA_VERSION" ]; then
-    export KBFS_METADATA_VERSION=2
+    export KBFS_METADATA_VERSION=3
```
strib: FYI @jzila: I'm going to turn on mdv3 for all kbfsdocker tests (assuming CI passes). It's required for team tests, and I don't see an easy way to turn it on on a per-test basis.

libkbfs/keybase_service_base.go
strib: > Does this guarantee that anything returned with no error is from the past minute?
￼ 
Yes.

> So merkleFreshness is in ms?

Yes.  I'll rename the var to be clear.

kbfsfuse/kbfsfuse.sh
jzila: We run docker tests with both MDv2 and MDv3

strib: Oh, I was just trying it with `make test`, and it didn't work.  I guess I'll have to skip my team test when the version is 2.  But that's for another PR.  If you're not ok with changing it here by default, so it makes `make test` work, let me know.

jzila: Go ahead and change it here. That's the env var that the Jenkinsfile overrides to get both to run.

strib: Right.  I think 3 should probably be the default everywhere now.  I'll make my serverside PR to do that as well.

PR #1042, opened 2017-06-16
libkbfs: identify/resolve team names with a "team:" prefix

strib:
@maxtaco says it's better if we always resolve/identify team names with a "team:" prefix, to make things easier for the service, and to allow us to split the namespaces someday in the future if we want.

Also adds a hack in vendor that we need until CORE-5376 is complete (at which point I'll revert the hack's commit and revendor libkb).

Issue: KBFS-2275

### DIFF
--- a/libkbfs/keybase_daemon_local.go
+++ b/libkbfs/keybase_daemon_local.go
@@ -178,7 +178,7 @@ func (k *KeybaseDaemonLocal) assertionToIDLocked(ctx context.Context,
 		} else {
 			key, val := url.ToKeyValuePair()
 			a := fmt.Sprintf("%s@%s", val, key)
-			if url.IsKeybase() {
+			if url.IsKeybase() && key != "team" {
 				a = val
 			}
 			var ok bool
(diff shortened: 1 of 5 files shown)


libkbfs/keybase_daemon_local.go
```diff
@@ -535,7 +535,7 @@ func (k *KeybaseDaemonLocal) addTeamKeyForTest(
 func (k *KeybaseDaemonLocal) addTeamsForTestLocked(teams []TeamInfo) {
 	for _, t := range teams {
 		k.localTeams[t.TID] = t
-		k.asserts[string(t.Name)] = t.TID.AsUserOrTeam()
+		k.asserts[string(t.Name)+"@team"] = t.TID.AsUserOrTeam()
```
jzila: I'm not really following where we can have such an assertion in a TLF.

strib: The issue is [here](https://github.com/keybase/kbfs/blob/a768f81ba72da33f0d82d5c20d407337e540063e/libkbfs/keybase_daemon_local.go#L180).  When checking any kind of key/value assertion, this code changes it into the '@' format.  This local daemon doesn't really support "key:value" style assertions, but it seems more natural to do that for teams.

songgao: Does this go to the same place as `t_alice@twitter` or is it a completely different thing?

strib: The local daemon only has one map tracking asserts, and both types go into that map.

songgao: Got you thanks!

PR #1045, opened 2017-06-17
[KBFS-1786] Prefetch all child blocks for an indirect block

jzila:


jzila:
@songgao: it should end up being the same, since those blocks are likely to be requested soon afterwards anyway.

songgao:
OK thanks

PR #1047, opened 2017-06-19
unset paperkey prompt if another deivce has serverd the rekey request

songgao:
Approach as discussed on [slack](https://keybase.slack.com/archives/C055SBA3H/p1497911382463613).

Also, remove the explicit rekey cancellation in key manager tests.

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4705,6 +4705,13 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 			}
 		}
 		if rmd.IsRekeySet() {
+			// One might have concern that a MD update written by the device
+			// itself can slip in here, for example during the rekey after
+			// setting paper prompt, and the event may cause the paper prompt
+			// to be unset. This is not a problem because 1) the revision check
+			// above shouldn't allow MD update written by this device to reach
+			// here; 2) the rekey FSM doesn't touch anything if it has the
+			// paper prompt set and is in scheduled state.
 			fbo.rekeyFSM.Event(NewRekeyRequestEvent())
 		} else {
 			fbo.rekeyFSM.Event(NewRekeyNotNeededEvent())
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4704,6 +4704,11 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 				return err
 			}
 		}
+		if rmd.IsRekeySet() {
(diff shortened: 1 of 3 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -4704,6 +4704,11 @@ func (fbo *folderBranchOps) applyMDUpdatesLocked(ctx context.Context,
 				return err
 			}
 		}
+		if rmd.IsRekeySet() {
```
strib: In slack I mentioned we should only do this if the update is from some other device, to make sure our own updates don't cause issues.  Maybe the revision check above is good enough to prevent this, but I just want to make sure you thought about it.

songgao: But this PR doesn't trigger a rekey immediately on MD updates anymore; instead, it only unsets the paper prompts. Could you be more specific how the device's own update can interfere it?

strib: If the device sends an MD update and sets rekey=true and sets a paper prompt, I worry that its own update could unset the paper prompt that it just itself set.

libkbfs/rekey_fsm.go
```diff
@@ -290,6 +305,17 @@ func (r *rekeyStateScheduled) reactToEvent(event RekeyEvent) rekeyState {
 		}
 		r.timer.Stop()
 		return newRekeyStateScheduled(r.fsm, event.request.delay, task)
+	case rekeyNotNeededEvent:
+		// KBFS-2254: if another device finished rekey, we should unset the
+		// paperkey prompt so that if this other deivce goes offline before a
```
strib: "deivce" -> "device"

libkbfs/folder_branch_ops.go
strib: I'm pretty certain the check for the revision above takes care of the issue I was worried about.  But it's probably worth a comment here explaining that we should never be processing our own revisions here because of the check above, and because `mdWriterLock` makes it impossible for them to arrive via this path before we've applied them locally already.

songgao: Thanks for checking strib! Although I'm still not sure I understand the concern. If a MD update with `rekey=true` comes in, it generates a rekey request, and the rekey FSM doesn't do anything other than printing a log if the paper prompt is set: https://github.com/keybase/kbfs/blob/1faa48ab5973de7974186b1b84e71c0026921a56/libkbfs/rekey_fsm.go#L278-L292

strib: Ok great, I forget we already handled that case.  See the need for a comment?

songgao: Yup; will do and include both arguments!

PR #1048, opened 2017-06-21
Refactor mounting code on all platforms and make normal mount failure…

taruti:
…s non-fatal

Also fix race conditions in the original code and create a new abstraction in libfs.

### DIFF
--- a/libfs/mount_interrupter.go
+++ b/libfs/mount_interrupter.go
@@ -6,7 +6,17 @@ package libfs
 
 import (
 	"sync"
+
+	"github.com/keybase/client/go/logger"
 )
+
+// Mounter is the interface for filesystems to be mounted by MountInterrupter.
+type Mounter interface {
+	// Mount a  filesystem.
+	Mount() error
+	// Unmount a mounted filesystem.
+	Unmount() error
+}
 
 // MountInterrupter is for managing mounts with cancelation.
 type MountInterrupter struct {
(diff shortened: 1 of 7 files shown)


libfuse/start.go
```diff
@@ -42,52 +46,54 @@ func Start(mounter Mounter, options StartOptions, kbCtx libkbfs.Context) *libfs.
...
+				mi.Done()
+				return libfs.MountError(err.Error())
+			}
 		}
 	}
+	mi.Wait()
+	return nil
+}
+
+func startMounting(config libkbfs.Config, options StartOptions,
```
songgao: This is only called by `func Start`; would you mind moving this above it?

libfuse/start.go
```diff
@@ -42,52 +46,54 @@ func Start(mounter Mounter, options StartOptions, kbCtx libkbfs.Context) *libfs.
...
+	log logger.Logger, mi *libfs.MountInterrupter) error {
+	log.Debug("Mounting: %q", options.MountPoint)
+	c, err := fuseMount(options)
+	if err != nil {
+		return err
+	}
+	mi.SetOnceFun(func() { fuseUnmount(options) })
```
songgao: There could be a race condition here where `mi.Done` was called after the `fuseMount` above, but before the `SetOnceFun` here manages to grab to lock, and the mount point will never be unmounted.

Perhaps we can have a channel `doneCalled` that `mi.Done` close in a separate`Once` wrapper, before the `mi.once.Done` is called, and `SetOnceFun` would call the Fun (through `mi.once.Done`) if the channel is already closed?

libfs/mount_interrupter.go
```diff
@@ -0,0 +1,49 @@
...
+// license that can be found in the LICENSE file.
+
+package libfs
+
+import (
+	"sync"
+)
+
+// MountInterrupter is for managing mounts with cancelation.
+type MountInterrupter struct {
```
songgao: This type isn't really mounter specific, so I think a more generic name like `LazyOnceWaitable` would work as well. But I'm OK with the `MountInterrupter` too; just thought I'd point it out.

taruti:
I ended up rewriting some parts to avoid the "interrupt just after mount" and "interrupt before mount" issues.

PR #1049, opened 2017-06-22
kbfs_ops: Status() should always return the status channel

strib:
Even on a status error (which is ignored otherwise).  Otherwise the
listeners won't get notifications when the status changes, and they
won't force-fast-forward cleared folders when a login happens.

Issue: keybase/client#7475


libkbfs/kbfs_ops.go
```diff
@@ -678,7 +678,7 @@ func (fs *KBFSOpsStandard) Status(ctx context.Context) (
 		err := fillInJournalStatusUnflushedPaths(
 			ctx, fs.config, jServerStatus, tlfIDs)
 		if err != nil {
-			return KBFSStatus{}, nil, err
+			return KBFSStatus{}, ch, err
```
songgao: Maybe add a comment explaining that even `err` is `nil` the caller in `remote_status.go` still relies on `ch`?

PR #1050, opened 2017-06-23
folder_block_ops: update attr using the BlockPointer, not name

strib:
In a batch update, a file can later be renamed, so use the
setAttrOp.Name field to search for the Node to update doesn't make
sense.  Instead, just use setAttrOp.File, which is the BlockPointer
for the file whose attributes are being changed.

Issue: KBFS-2286

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -3263,21 +3263,14 @@ func (fbo *folderBlockOps) setCachedAttr(
 func (fbo *folderBlockOps) UpdateCachedEntryAttributes(
 	ctx context.Context, lState *lockState, kmd KeyMetadata,
 	dir path, op *setAttrOp) (Node, error) {
-	childPath := dir.ChildPathNoPtr(op.Name)
-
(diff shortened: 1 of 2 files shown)


libkbfs/folder_block_ops.go
```diff
@@ -3263,22 +3263,15 @@ func (fbo *folderBlockOps) setCachedAttr(
 func (fbo *folderBlockOps) UpdateCachedEntryAttributes(
 	ctx context.Context, lState *lockState, kmd KeyMetadata,
 	dir path, op *setAttrOp) (Node, error) {
-	childPath := dir.ChildPathNoPtr(op.Name)
-
-	// find the node for the actual change; requires looking up
-	// the child entry to get the BlockPointer, unfortunately.
```
jzila: Why is this no longer true?

strib: I think originally we didn't have `setAttrOp.File`, it was added later (commit db47e897).  Probably just forgot to streamline this code at the time.

PR #1051, opened 2017-06-23
config: turn off the metrics registry in minimal mode

strib:
So it doesn't waste CPU on mobile.

Suggested by patrickxb.

Issue: KBFS-2288

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -310,8 +310,9 @@ func NewConfigLocal(mode InitMode, loggerFn func(module string) logger.Logger,
 	config.qrUnrefAge = qrUnrefAgeDefault
 	config.qrMinHeadAge = qrMinHeadAgeDefault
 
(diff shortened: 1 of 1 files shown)


libkbfs/config_local.go
```diff
@@ -310,8 +310,9 @@ func NewConfigLocal(mode InitMode, loggerFn func(module string) logger.Logger,
 	config.qrUnrefAge = qrUnrefAgeDefault
 	config.qrMinHeadAge = qrMinHeadAgeDefault
 
-	// Don't bother creating the registry if UseNilMetrics is set.
-	if !metrics.UseNilMetrics {
+	// Don't bother creating the registry if UseNilMetrics is set, or
+	// if we're in minimal mode.
+	if !metrics.UseNilMetrics && mode != InitMinimal {
```
jzila: The following two lines can be consolidated into one, but it's fine otherwise.

PR #1054, opened 2017-06-27
ioutil: add WriteSerializeFile helper function

strib:
ioutil.WriteFile can lead to empty files if there's a crash or a disk
error (like out of space) at the wrong time.  Instead, overwrite and
truncate the existing file.  This has the same risk of partial OS
writes leading to corrupted files, without the risk of truncating a
non-empty file.

Issue: KBFS-2289

### DIFF
--- a/ioutil/write_serialized_file.go
+++ b/ioutil/write_serialized_file.go
@@ -1,0 +1,49 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package ioutil
+
+import (
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+// WriteSerializedFile writes (or overwrites) `data` into `filename`.
+// If `filename` doesn't exist, it's created with `perm` permissions.
+// If `filename` does exist, the data is first overwritten to the
+// file, and then the file is truncated to the length of the data.  If
+// the data represents a serialized data structure where the length is
+// explicitly stated in, or implicitly calculated from, the data
+// itself on a read, then this is approximately an atomic write
+// (without the performance overhead of writing to a temp file and
+// renaming it).  NOTE: it's technically possible a partial OS write
+// could lead to a corrupted file, though in practice this seems much
+// more rare than ioutil.WriteFile() leaving behind an empty file.
+func WriteSerializedFile(
+	filename string, data []byte, perm os.FileMode) (err error) {
+	// Don't use ioutil.WriteFile because it truncates the file first,
+	// and if there's a crash it will leave the file in an unknown
+	// state.
+	f, err := OpenFile(filename, os.O_WRONLY|os.O_CREATE, 0600)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


ioutil/write_serialized_file.go
```diff
@@ -0,0 +1,49 @@
...
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+// WriteSerializedFile writes (or overwrites) `data` into `filename`.
+// If `filename` doesn't exist, it's created with `perm` permissions.
+// If `filename` does exist, the data is first overwritten to the
+// file, and then the file is truncated to the length of the data.  If
```
jzila: What if the data is longer than the existing file? `io.ErrShortWrite`?

strib: The `File.Write()` call should just extend the file in that case, no problem.  (Just like if the file was brand new and empty.)

strib:
> Code looks good me, but I don't fully understand the issue here. What do you mean io.WriteFile can leave an empty file behind?

https://golang.org/src/io/ioutil/ioutil.go?s=2518:2586#L66

If the `os.OpenFile(O_TRUNC)` succeeds, but the `f.Write()` fails (say because the disk is full), then you have an empty file.  If the file previously had data in it, this is a problem.

songgao:
I see; thanks. So this is like reserving the space while writing so other processes can't take over the space that this one claimed. What if we need to write more than what's already in there? If this is a concern, should we "reserve" a sufficiently large file for it and never truncate it?

strib:
> I see; thanks. So this is like reserving the space while writing so other processes can't take over the space that this one claimed. What if we need to write more than what's already in there? If this is a concern, should we "reserve" a sufficiently large file for it and never truncate it?

It's not really about reserving space.  It's more about avoiding corruption, and in this case if a non-empty file turns into an empty file, that counts as corruption and the journal can't function.  I don't think "reserving" a larger file really helps with that, though perhaps it cuts down the probability of a partial write?

I consider partial writes a separate issue, and they are an issue with or without `ioutil.WriteFile`.  Here I'm just banking on the fact that they're a lot more rare than a crash or a write error.  My first attempt was to write to a temp file and rename it to get real atomic writes, but unfortunately that seems to come with a huge performance penalty.

songgao:
> It's more about avoiding corruption, and in this case if a non-empty file turns into an empty file, that counts as corruption and the journal can't function.

But even with the changes in this PR, we could still get failed writes which has already resulted file changes. We wouldn't truncate the file but it's still a corrupted file rather than the original, right?

strib:
> But even with the changes in this PR, we could still get failed writes which has already resulted file changes. We wouldn't truncate the file but it's still a corrupted file rather than the original, right?

Yes.  That can happen both before and after this PR, and I'm not trying to address that issue (which I've never seen in practice, as far as I know, because os writes of small sizes are usually all-or-nothing in practice).  I'm only trying to address the "empty file" type of corruption, which seems pretty common.

songgao:
Got you. Thanks for the context!

PR #1055, opened 2017-06-29
libfuse/libdokan: enable teams for admins on staging

strib:
Also add "core.keybase.io" to the list of prod domains.

### DIFF
--- a/libkbfs/keybase_service_util.go
+++ b/libkbfs/keybase_service_util.go
@@ -5,7 +5,9 @@
 package libkbfs
 
 import (
+	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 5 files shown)


strib:
FYI just updated this to not add KBFS favorites for team TLFs, since the CORE side isn't ready yet.

libkbfs/keybase_service_util.go
```diff
@@ -32,6 +34,21 @@ var adminFeatureList = map[keybase1.UID]bool{
 	"eb08cb06e608ea41bd893946445d7919": true, // Miles Steele, "mlsteele"
```
songgao: Should we add Taru, Michal, and Andrey?

strib: Good point, will do.

PR #1058, opened 2017-07-05
libfuse/libdokan: enable teams for admins on prod

strib:
And enable them for all users on staging.

(To be merged after we deploy prod.)

### DIFF
--- a/libdokan/fs.go
+++ b/libdokan/fs.go
@@ -377,11 +377,11 @@ func (f *FS) open(ctx context.Context, oc *openContext, ps []string) (dokan.File
 			return nil, false, dokan.ErrAccessDenied
 		}
 		return f.root.private.open(ctx, oc, ps[1:])
-	case `TEAM` == ps[0]:
+	case strings.ToUpper(TeamName) == ps[0]:
 		oc.isUppercasePath = true
 		fallthrough
 	case TeamName == ps[0]:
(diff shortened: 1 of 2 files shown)


libdokan/fs.go
```diff
@@ -377,11 +377,11 @@ func (f *FS) open(ctx context.Context, oc *openContext, ps []string) (dokan.File
 			return nil, false, dokan.ErrAccessDenied
 		}
 		return f.root.private.open(ctx, oc, ps[1:])
-	case `TEAM` == ps[0]:
+	case strings.ToUpper(TeamName) == ps[0]:
```
jzila: Do you think the Go compiler will optimize out this `ToUpper` call on a constant?

strib:
> Not mentioned, but this also enables teams by default for all staging/devel users.

Not mentioned in the subject, but it's mentioned on the first line of the PR description.  (It was already enabled for all devel users before this PR.)

libdokan/fs.go
strib: Dunno!  I was just making it consistent with the other places above.  I think this one happened concurrently with switchover to `ToUpper`.  I'm not worried about performance here, but if you are we can revert to using constants in a separate PR.

jzila: Naw it's fine

PR #1059, opened 2017-07-06
kbfscrypto: core.keybase.io uses the API root CA cert

strib:


songgao:
Thanks for working it out!

strib:
I should mention I tested it using `openssl s_client -CAfile XXX -connect XXX` from ctl.kbfs, and verified that the old cert failed while the new one worked.

strib:
> Could you add a comment explaining the cert situation?

What should the comment say that is not already clear from the code?

songgao:
(talked offline; no need for comment)

PR #1060, opened 2017-07-06
libkbfs: avoid deadlock between `CreateFile` and `backgroundFlusher`

strib:
KBFS-2295 details a race that happens when `CreateFile` calls `folderBlockOps.Write()` while holding `mdWriterLock` and blocks due to a full dirty bcache, and the background flusher can't sync anything because it needs to grab `mdWriterLock`.  This PR fixes it two ways:

* Blocked writes now time out based on contexts timing out (not sure how we missed this before).
* 0-byte writes, like the ones made from `CreateFile`, don't need to block at all.

I toyed with the idea of moving the `Write` call outside of the `CreateFile`, but a bunch of stuff depended on transactional semantics there and it seemed too complicated.

Issue: KBFS-2295

### DIFF
--- a/libkbfs/dirty_bcache.go
+++ b/libkbfs/dirty_bcache.go
@@ -490,6 +490,13 @@ func (d *DirtyBlockCacheStandard) RequestPermissionToDirty(
 		panic("Must request permission for a non-negative number of bytes.")
 	}
 	c := make(chan struct{})
+
+	// No need to wait to write 0 bytes.
+	if estimatedDirtyBytes == 0 {
+		close(c)
+		return c, nil
+	}
+
 	now := d.clock.Now()
 	deadline, ok := ctx.Deadline()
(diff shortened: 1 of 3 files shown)


jzila:
> 0-byte writes, like the ones made from CreateFile, don't need to block at all.
Doesn't this create a potential conflict scenario if two writers try to create the same file at the same time?

strib:
> Doesn't this create a potential conflict scenario if two writers try to create the same file at the same time?

Sorry for not being clear.  The "blocking" I'm referring to here is locally, within a single client, the backpressure-blocking done by `dirtyBlockCache`.  This is trying to constrain the amount of memory that we use to buffer writes in-memory while waiting for them to flush to disk or the remote server.  It has no effect on conflicts, unless I'm missing something.

jzila:
Sounds good. 🚢

PR #1061, opened 2017-07-07
libkbfs: Avoid flake in unstageAfterFailedResolution

taruti:


libkbfs/folder_branch_ops.go
```diff
@@ -5950,6 +5950,7 @@ func (fbo *folderBranchOps) unstageAfterFailedResolution(ctx context.Context,
 	default:
 	}
 
+	ctx = newLinkedContext(ctx)
```
strib: Can you please add a comment here explaining why this is needed, with a mention of KBFS-2150?  Why is this the right fix, rather than archiving any unmerged pointers as I suggested in the ticket?

taruti: I'll add an explanation to the code, but:
1) If the error is caused by a cancelled context then the recovery (archiving) will need to use a separate context anyways.
2) In such cases we have to be very careful where the error occurs and what to archive, making that solution more complicated.
3) The other "common" error case is losing server connection and after detecting that we won't have much luck archiving things.

Thus this seems like the simplest way to avoid missing the blocks.

PR #1062, opened 2017-07-07
mount_test: retry TestRemoveTLF a few times on OSX

strib:
Since the OS can reinstate a favorite with an ill-timed lookup and
there's not much we can do about it except retry.

I ran the test 10K times on OSX 10.10, and without this PR it reproduces the flake, and with it I couldn't get a reproduction.

Issue: KBFS-1370

### DIFF
--- a/libfuse/mount_test.go
+++ b/libfuse/mount_test.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path"
 	"reflect"
+	"runtime"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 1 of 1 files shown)


libfuse/mount_test.go
```diff
@@ -105,7 +106,8 @@ func mustBeDir(fi os.FileInfo) error {
 	return nil
 }
 
-func checkDir(t testing.TB, dir string, want map[string]fileInfoCheck) {
+func checkDirNoTestError(
+	t testing.TB, dir string, want map[string]fileInfoCheck) error {
```
songgao: Looks like we can remove the `t` parameter here?

strib: There's still a `t.Fatal` call below.

songgao: Oh right; it's collapsed. Thanks

PR #1063, opened 2017-07-07
libkbfs: caching team info and reacting to team renames

strib:
This should bring team TLFs up to performance parity with other TLF types.  And it also handle subteam renames, whenever that is implemented on the core side.

Issue: KBFS-2214


libkbfs/folder_branch_ops.go
```diff
@@ -6073,6 +6073,73 @@ func (fbo *folderBranchOps) onMDFlush(bid BranchID, rev kbfsmd.Revision) {
 	}()
 }
 
+// TeamNameChanged implements the KBFSOps interface for folderBranchOps
+func (fbo *folderBranchOps) TeamNameChanged(
+	ctx context.Context, tid keybase1.TeamID) {
+	fbo.log.CDebugf(ctx, "Team name changed for team %s", tid)
+
+	go func() {
```
jzila: Given that the goroutine is the whole function, I'd suggest making a method that you can call like so: `go fbo.teamNameChanged(ctx, tid)`. Also, why is it a goroutine?

libkbfs/kbfs_ops.go
```diff
@@ -754,6 +755,35 @@ func (fs *KBFSOpsStandard) GetNodeMetadata(ctx context.Context, node Node) (
...
 }
 
+// TeamNameChanged implements the KBFSOps interface for KBFSOpsStandard
+func (fs *KBFSOpsStandard) TeamNameChanged(
+	ctx context.Context, tid keybase1.TeamID) {
+	fs.log.CDebugf(ctx, "Got TeamNameChanged for %s", tid)
+	fs.opsLock.Lock()
+	defer fs.opsLock.Unlock()
+	// We have to search for the tid since we don't know the old name
+	// of the team here.  TODO: add an index for this?
```
jzila: Ticket for the TODO please.

libkbfs/keybase_service_base.go
```diff
@@ -414,7 +441,10 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
 // KeybaseServiceBase.
 func (k *KeybaseServiceBase) LoadTeamPlusKeys(
 	ctx context.Context, tid keybase1.TeamID) (TeamInfo, error) {
-	// No caching until the invalidations are ready.
+	cachedTeamInfo := k.getCachedTeamInfo(tid)
+	if cachedTeamInfo.Name != libkb.NormalizedUsername("") {
```
jzila: I guess we can't cache bad team IDs since there's no way to tell the API server to notify us. But have we thought about doing that? Would it even be worth it?

libkbfs/folder_branch_ops.go
strib: I'll break it out into a separate function.

It's a goroutine because we shouldn't block the service notification RPC while waiting on the lock accesses and outgoing service calls (which could lead to deadlock potential).

libkbfs/kbfs_ops.go
strib: This is a question, not a real TODO yet.  I'm not comfortable littering JIRA with something I'm not sure is a good idea.  I'll remove the "TODO" label and just make it a question if you prefer.

libkbfs/keybase_service_base.go
strib: Sorry, I don't follow the question.  You mean if `tid` isn't a valid team ID, we should remember that and save ourselves the cost of checking it later?  I'm confused, because this line of code has nothing to do with team IDs.  It's only checking whether the cache is populated for this team ID or not.

(This works the same way as the user cache.  Though it's probably better to actually delete stuff from the cache, rather than setting them to a zero-value, not sure why we don't do that.)

jzila: Nvmind. Your interpretation of my question was accurate, but I don't think it can really happen.

libkbfs/kbfs_ops.go
```diff
@@ -754,6 +755,36 @@ func (fs *KBFSOpsStandard) GetNodeMetadata(ctx context.Context, node Node) (
...
+func (fs *KBFSOpsStandard) TeamNameChanged(
+	ctx context.Context, tid keybase1.TeamID) {
+	fs.log.CDebugf(ctx, "Got TeamNameChanged for %s", tid)
+	fs.opsLock.Lock()
+	defer fs.opsLock.Unlock()
+	// We have to search for the tid since we don't know the old name
+	// of the team here.  Should we add an index for this?
+	for fb, fbo := range fs.ops {
+		if fb.Tlf.Type() != tlf.SingleTeam {
+			continue
```
songgao: Perhaps log a warning here?

libkbfs/keybase_service_base.go
```diff
@@ -234,12 +239,34 @@ func (k *KeybaseServiceBase) clearCachedUnverifiedKeys(uid keybase1.UID) {
...
-	k.userCacheLock.Lock()
-	defer k.userCacheLock.Unlock()
-	k.userCache = make(map[keybase1.UID]UserInfo)
-	k.userCacheUnverifiedKeys = make(map[keybase1.UID][]keybase1.PublicKey)
+	func() {
+		k.userCacheLock.Lock()
+		defer k.userCacheLock.Unlock()
+		k.userCache = make(map[keybase1.UID]UserInfo)
+		k.userCacheUnverifiedKeys = make(map[keybase1.UID][]keybase1.PublicKey)
+	}()
```
songgao: You might have considered this but just making sure, could there be any race conditions after `userCache` is nuked but before `teamCache` is?

libkbfs/keybase_service_base.go
```diff
@@ -741,6 +772,22 @@ func (k *KeybaseServiceBase) FSSyncStatusRequest(ctx context.Context,
...
+// KeybaseServiceBase.
+func (k *KeybaseServiceBase) TeamChanged(
+	ctx context.Context, arg keybase1.TeamChangedArg) error {
+	k.log.CDebugf(ctx, "Flushing cache for team %s/%s "+
+		"(membershipChange=%t, keyRotated=%t, renamed=%t)",
+		arg.TeamName, arg.TeamID, arg.Changes.MembershipChanged,
+		arg.Changes.KeyRotated, arg.Changes.Renamed)
+	k.setCachedTeamInfo(arg.TeamID, TeamInfo{})
+
+	if arg.Changes.Renamed {
```
songgao: Unrelated to this PR, but is this also how we get notified when user is changed from writer to reader, and thus know to stop pushing any outstanding MD writes to `mdserver`?

libkbfs/kbfs_ops.go
strib: This will iterate over all the public and private TLFs too, so I wouldn't want to warn on such a common occurrence.

libkbfs/keybase_service_base.go
strib: As far as I know, the two caches are totally independent and used in different calls, so there shouldn't be any races possible.  If a user call comes races with this function (and hence with the logout), it's undefined whether they will be able to read from the cache or not, and I think that's fine/expected.

libkbfs/kbfs_ops.go
songgao: Oh that makes sense. Somehow I was (mistakenly) thinking this was iterating through a set of notifications for different teams.

libkbfs/keybase_service_base.go
songgao: I see thanks!

libkbfs/keybase_service_base.go
strib: `TeamChanged` is how we learn of role upgrades, yes.  But so far I haven't done anything to plumb that down to the journal level.  The mdserver will start rejecting the writes of a non-writer being put by the journal, and the journal will see that error and start backing off its attempts.  Eventually if the reader is upgraded, the journal writes will succeed again.  I'm not sure we want it to even permanently stop.

If someone else writes to the folder while we're in this state, we should be able to run CR locally and "rebase" the outstanding changes onto the TLF, so the user will see both the local and remote updates, but we'll be unable to push the changes.  I haven't tested this though.

songgao: All these sound good to me. Thanks! Will move to #kbfs for followup questions.

PR #1064, opened 2017-07-10
[KBFS-2291] Disk Limiter Refactor

jzila:
This cleans up the disk limiter to be less tightly coupled to its usage.
* Methods now take as a parameter the tracker type they affect, and have been renamed to describe what they do to the tracker rather than where they are called.
* Some methods have been consolidated so that we don't have duplicated functionality across tracker types.
* There is now an overall tracker that tracks overall Keybase usage. This means trackers no longer depend on each other, instead treating the `overallTracker` as a single source of truth for the aggregate Keybase data usage.

Sorry this took so long; I had lots of corner case accounting bugs that I needed to fix because I put too many changes in at once before testing (consolidating and introducing the overall tracker accounting).

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -107,7 +107,7 @@ func (bt backpressureTracker) currLimit() float64 {
 	freeFloat := float64(bt.free)
 	limit := bt.limitFrac * (usedFloat + freeFloat)
 	minLimit := math.Min(limit, float64(bt.limit))
-	// Magic number of 512 gets us past overflow issues at the limit due to
+	// The magic number of 512 gets us past overflow issues at the limit due to
 	// floating point precision.
 	maxFloatForInt64 := float64(math.MaxInt64 - 512)
 	if minLimit > maxFloatForInt64 {
@@ -615,7 +615,7 @@ type backpressureDiskLimiter struct {
 	// values of the semaphores, but not the actual semaphores
 	// themselves.
 	lock sync.RWMutex
-	// overallTracker tracks the overall number of bytes used by Keybase.
+	// overallByteTracker tracks the overall number of bytes used by Keybase.
 	overallByteTracker *backpressureTracker
 	// journalTracker tracks the journal bytes and files used.
 	journalTracker journalTracker
@@ -753,7 +753,7 @@ func newBackpressureDiskLimiter(
 	// that the disk cache should consume. Add 0.5 for rounding.
 	diskCacheByteLimit := int64(
 		(float64(params.byteLimit) * params.diskCacheFrac) + 0.5)
-	overallTracker, err := newBackpressureTracker(
+	overallByteTracker, err := newBackpressureTracker(
 		1.0, 1.0, 1.0, params.byteLimit, freeBytes)
 	diskCacheByteTracker, err := newBackpressureTracker(
 		1.0, 1.0, params.diskCacheFrac, diskCacheByteLimit, freeBytes)
@@ -768,7 +768,7 @@ func newBackpressureDiskLimiter(
 		freeBytesAndFilesFn:  params.freeBytesAndFilesFn,
 		quotaFn:              params.quotaFn,
 		lock:                 sync.RWMutex{},
-		overallByteTracker:   overallTracker,
+		overallByteTracker:   overallByteTracker,
 		journalTracker:       journalTracker,
 		diskCacheByteTracker: diskCacheByteTracker,
 		syncCacheByteTracker: syncCacheByteTracker,
@@ -841,7 +841,7 @@ func (bdl *backpressureDiskLimiter) onJournalDisable(
 	bdl.overallByteTracker.onDisable(journalStoredBytes)
 }
 
-func (bdl *backpressureDiskLimiter) onDiskBlockCacheEnable(ctx context.Context,
+func (bdl *backpressureDiskLimiter) onByteTrackerEnable(ctx context.Context,
 	diskCacheBytes int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
@@ -849,7 +849,7 @@ func (bdl *backpressureDiskLimiter) onDiskBlockCacheEnable(ctx context.Context,
 	bdl.diskCacheByteTracker.onEnable(diskCacheBytes)
 }
 
-func (bdl *backpressureDiskLimiter) onDiskBlockCacheDisable(ctx context.Context,
+func (bdl *backpressureDiskLimiter) onByteTrackerDisable(ctx context.Context,
 	diskCacheBytes int64) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
@@ -964,9 +964,6 @@ func (bdl *backpressureDiskLimiter) onBlocksDelete(
 
 func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
 	ctx context.Context, blockBytes int64) {
-	if blockBytes == 0 {
-		return
-	}
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
 	bdl.diskCacheByteTracker.releaseAndCommit(blockBytes)
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -514,7 +514,7 @@ func (jt journalTracker) getSemaphoreCounts() (byteCount, fileCount int64) {
 	return jt.byte.semaphore.Count(), jt.file.semaphore.Count()
 }
 
-func (jt journalTracker) beforeBlockPut(
+func (jt journalTracker) reserve(
 	ctx context.Context, blockBytes, blockFiles int64) (
 	availableBytes, availableFiles int64, err error) {
 	availableBytes, err = jt.byte.reserve(ctx, blockBytes)
@@ -536,7 +536,7 @@ func (jt journalTracker) beforeBlockPut(
 	return availableBytes, availableFiles, nil
 }
 
-func (jt journalTracker) afterBlockPut(
+func (jt journalTracker) commitOrRollback(
 	blockBytes, blockFiles int64, putData bool,
 	chargedTo keybase1.UserOrTeamID) {
 	jt.byte.commitOrRollback(blockBytes, putData)
@@ -921,6 +921,7 @@ func (bdl *backpressureDiskLimiter) beforeBlockPut(
 			return 0, err
 		}
 
+		bdl.overallByteTracker.updateFree(freeBytes)
 		bdl.journalTracker.updateFree(freeBytes, bdl.overallByteTracker.used,
 			freeFiles)
 
@@ -947,7 +948,8 @@ func (bdl *backpressureDiskLimiter) beforeBlockPut(
 		return bdl.onBeforeBlockPutError(err)
 	}
 
-	return bdl.journalTracker.beforeBlockPut(ctx, blockBytes, blockFiles)
+	bdl.overallByteTracker.reserve(ctx, blockBytes)
+	return bdl.journalTracker.reserve(ctx, blockBytes, blockFiles)
 }
 
 func (bdl *backpressureDiskLimiter) afterBlockPut(
@@ -955,7 +957,7 @@ func (bdl *backpressureDiskLimiter) afterBlockPut(
 	chargedTo keybase1.UserOrTeamID) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
-	bdl.journalTracker.afterBlockPut(blockBytes, blockFiles, putData, chargedTo)
+	bdl.journalTracker.commitOrRollback(blockBytes, blockFiles, putData, chargedTo)
 	bdl.overallByteTracker.commitOrRollback(blockBytes, putData)
 }
 
@@ -1022,6 +1024,7 @@ func (bdl *backpressureDiskLimiter) afterDiskBlockCachePut(
 	ctx context.Context, blockBytes int64, putData bool) {
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
+	bdl.overallByteTracker.commitOrRollback(blockBytes, putData)
 	bdl.diskCacheByteTracker.commitOrRollback(blockBytes, putData)
 }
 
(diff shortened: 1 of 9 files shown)


libkbfs/backpressure_disk_limiter.go
```diff
@@ -812,20 +854,31 @@ func (bdl *backpressureDiskLimiter) onJournalDisable(
 	defer bdl.lock.Unlock()
 	bdl.journalTracker.onDisable(
 		journalStoredBytes, journalUnflushedBytes, journalFiles, chargedTo)
+	bdl.overallByteTracker.onDisable(journalStoredBytes)
 }
 
-func (bdl *backpressureDiskLimiter) onDiskBlockCacheEnable(ctx context.Context,
-	diskCacheBytes int64) {
+func (bdl *backpressureDiskLimiter) onByteTrackerEnable(ctx context.Context,
```
strib: Again, the naming here and below seems to imply that the journal would a valid type.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -846,28 +899,33 @@ func (bdl *backpressureDiskLimiter) getDelayLocked(
 	return time.Duration(delayScale * float64(maxDelay))
 }
 
-func (bdl *backpressureDiskLimiter) onBeforeBlockPutError(err error) (
+func (bdl *backpressureDiskLimiter) reserveError(err error) (
```
strib: I know this is a pre-existing condition, but do you think we need to take `bdl.lock` in this function, in order to get a consistent view of the bytes and files together?

libkbfs/disk_limiter.go
```diff
@@ -48,24 +65,32 @@ type DiskLimiter interface {
...
+	reserve(ctx context.Context, typ diskLimitTrackerType, blockBytes int64) (
+		availableBytes int64, err error)
+
+	// commitOrRollback is called after using disk storage of the given byte
+	// and file count, which must match the corresponding call to
+	// beforeBlockPut. putData reflects whether or not the data was actually
+	// put; if it's false, it's either because of an error or because the block
+	// already existed.
+	commitOrRollback(ctx context.Context, typ diskLimitTrackerType, blockBytes,
+		blockFiles int64, shouldCommit bool, chargedTo keybase1.UserOrTeamID)
```
strib: Mention that `blockFiles` is ignored for tracker types that don't track files?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -924,30 +997,31 @@ func (bdl *backpressureDiskLimiter) onBlocksFlush(
...
-}
-
-func (bdl *backpressureDiskLimiter) onDiskBlockCacheDelete(
-	ctx context.Context, blockBytes int64) {
-	if blockBytes == 0 {
-		return
+	switch typ {
+	case journalLimitTracker:
+		bdl.journalTracker.releaseAndCommit(blockBytes, blockFiles)
+	case diskCacheLimitTracker:
```
strib: Why not use a `default` case that uses `trackerFromType()`, as above?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -924,30 +997,31 @@ func (bdl *backpressureDiskLimiter) onBlocksFlush(
...
 	availableBytes int64, err error) {
 	if blockBytes == 0 {
 		// Better to return an error than to panic in ForceAcquire.
-		return 0, errors.New("backpressureDiskLimiter.beforeDiskBlockCachePut" +
-			" called with 0 blockBytes")
+		return 0, errors.New("reserve called with 0 blockBytes")
+	}
+	tracker, err := bdl.trackerFromType(typ)
+	if err != nil {
+		return 0, errors.New("reserve called with invalid tracker type")
```
strib: Why construct a new error instead of using `err`?  Theoretically the error returned by `trackerFromType` could be anything, not just about an invalid tracker type.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -959,23 +1033,23 @@ func (bdl *backpressureDiskLimiter) beforeDiskBlockCachePut(
...
+	}
 	// We calculate the total free bytes by adding up the reported
 	// free bytes, the disk cache used bytes, *and* any other used
 	// bytes (e.g., the journal cache). For now, we hack this by
 	// lumping the other used bytes with the reported free bytes.
-	//
-	// TODO: Keep track of other used bytes separately.
-	bdl.diskCacheByteTracker.updateFree(
-		freeBytes + bdl.journalTracker.getUsedBytes())
+	tracker.updateFree(freeBytes + bdl.overallByteTracker.used -
```
strib: Is the comment above still accurate? This seems like less of a hack now, and doesn't directly take journal bytes into account.

libkbfs/disk_limiter.go
```diff
@@ -1,37 +1,54 @@
...
+	commitOrRollback(resources int64, shouldCommit bool)
+	releaseAndCommit(resources int64)
 }
 
 // DiskLimiter is an interface for limiting disk usage.
 type DiskLimiter interface {
-	diskBlockCacheLimiter
+	// onByteTrackerEnable is called when a byte tracker is enabled to begin
+	// accounting.
+	onByteTrackerEnable(ctx context.Context, typ diskLimitTrackerType,
```
strib: These two methods shouldn't be called with the journal tracker type, right?  Might be worth changing the name and documenting that to make it more clear.

libkbfs/disk_limiter.go
```diff
@@ -1,37 +1,54 @@
...
+	unknownLimitTracker diskLimitTrackerType = iota
+	journalLimitTracker
+	diskCacheLimitTracker
+	syncCacheLimitTracker
+)
 
-	// onDiskBlockCacheDisable is called when the disk block cache is disabled to
-	// stop accounting for its blocks.
-	onDiskBlockCacheDisable(ctx context.Context, cacheBytes int64)
+type diskLimitByteTracker interface {
```
jzila: I was trying to differentiate byte-only trackers from the journal tracker, which has its own custom functionality. The key part of the name here is "byteTracker", so if you have a suggestion for how to better represent that I'm open to it.

libkbfs/disk_limiter.go
```diff
@@ -48,24 +65,32 @@ type DiskLimiter interface {
...
 
-	// afterBlockPut is called after putting a block of the given
-	// byte and file count, which must match the corresponding call to
-	// beforeBlockPut. putData reflects whether or not the data
-	// was actually put; if it's false, it's either because of an
-	// error or because the block already existed.
-	afterBlockPut(ctx context.Context,
-		blockBytes, blockFiles int64, putData bool,
-		chargedTo keybase1.UserOrTeamID)
+	// reserve is called by the disk block cache before using disk storage with
```
jzila: `typ` will differentiate the LRU disk cache from the synced folder byte cache.

libkbfs/disk_limiter.go
```diff
@@ -48,24 +65,32 @@ type DiskLimiter interface {
...
+	// and file count, which must match the corresponding call to
+	// beforeBlockPut. putData reflects whether or not the data was actually
+	// put; if it's false, it's either because of an error or because the block
+	// already existed.
+	commitOrRollback(ctx context.Context, typ diskLimitTrackerType, blockBytes,
+		blockFiles int64, shouldCommit bool, chargedTo keybase1.UserOrTeamID)
+
+	// releaseAndCommit is called after deleting blocks for a given tracker of
+	// the given byte and file count, both of which must be >= 0.
+	releaseAndCommit(ctx context.Context, typ diskLimitTrackerType, blockBytes,
```
jzila: This was an attempt to counter the asymmetry from the `reserve` then `commitOrRollback` flow, where `releaseAndCommit` only has one call, but I can change it to just `release`.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -106,13 +106,24 @@ func (bt backpressureTracker) currLimit() float64 {
 	usedFloat := float64(bt.used)
 	freeFloat := float64(bt.free)
 	limit := bt.limitFrac * (usedFloat + freeFloat)
-	return math.Min(limit, float64(bt.limit))
+	minLimit := math.Min(limit, float64(bt.limit))
+	// The magic number of 512 gets us past overflow issues at the limit due to
```
jzila: I tested it. How would you like me to cite said test?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -785,6 +814,18 @@ func defaultGetFreeBytesAndFiles(path string) (int64, int64, error) {
 	return int64(freeBytes), int64(freeFiles), nil
 }
 
+func (bdl *backpressureDiskLimiter) trackerFromType(typ diskLimitTrackerType) (
```
jzila: Shouldn't that be implied from the return value type? (`diskLimitByteTracker`)

libkbfs/backpressure_disk_limiter.go
jzila: The key here is _byte_ Tracker (journal is not a byte tracker)

libkbfs/backpressure_disk_limiter.go
jzila: Honestly because it was less code to do it this way, but I can do that too.

libkbfs/disk_limiter.go
strib: I find that a bit confusing since the journal does actually track bytes -- it just additionally tracks files.  So this name seems like it should apply to the journal tracker as well, until you understand all the details.

Maybe just "simpleByteTracker", with no mention of disks needed, might be better?

Another alternative is changing the interface to be able to track a custom set of values, so it can also be used by the journal tracker.  But that's obviously a lot more work and probably just confuses things more.

jzila: Yeah I had that kind of refactor halfway done and it only made things worse rather than better, so I rolled back to what we see here. I'll rename to `simpleByteTracker`.

libkbfs/disk_limiter.go
strib: Ah right, I realized that later and forgot to edit this comment.  Thanks.

libkbfs/disk_limiter.go
strib: Understood.  But to me, the name "reserve" implies that there needs to be a followup action, while "release" doesn't.  I don't feel strongly about it though.

libkbfs/backpressure_disk_limiter.go
strib: Ok, I thought you might have found an algorithmic explanation somewhere.  Maybe just put a "(according to local tests)" somewhere?

libkbfs/backpressure_disk_limiter.go
strib: But `journalLimitTracker` is a valid value of the `diskLimitByteTracker` enum, isn't it?

libkbfs/backpressure_disk_limiter.go
strib: Either way is fine with me, I just think the two functions should be consistent.

libkbfs/backpressure_disk_limiter.go
jzila: Renamed to `simpleResourceTrackerFromType`.

libkbfs/backpressure_disk_limiter.go
jzila: Renamed to `onSimpleByteTrackerEnable`.

jzila:
@strib: ready for another look.

libkbfs/disk_limiter.go
```diff
@@ -1,37 +1,58 @@
...
-	// onDiskBlockCacheDisable is called when the disk block cache is disabled to
-	// stop accounting for its blocks.
-	onDiskBlockCacheDisable(ctx context.Context, cacheBytes int64)
+// simpleResourceTracker is an interface for limiting a single resource type.
+// It is mostly used to limit bytes.
+type simpleResourceTracker interface {
+	onEnable(usedResources int64) int64
+	onDisable(usedResources int64)
+	updateFree(freeResources int64)
+	usedBytes() int64
```
strib: `usedResources`?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -846,28 +895,35 @@ func (bdl *backpressureDiskLimiter) getDelayLocked(
 	return time.Duration(delayScale * float64(maxDelay))
 }
 
-func (bdl *backpressureDiskLimiter) onBeforeBlockPutError(err error) (
+func (bdl *backpressureDiskLimiter) reserveError(err error) (
 	availableBytes, availableFiles int64, _ error) {
+	bdl.lock.RLock()
 	availableBytes, availableFiles =
 		bdl.journalTracker.getSemaphoreCounts()
+	bdl.lock.RUnlock()
```
strib: Why no `defer`?

libkbfs/backpressure_disk_limiter.go
```diff
@@ -785,6 +810,18 @@ func defaultGetFreeBytesAndFiles(path string) (int64, int64, error) {
 	return int64(freeBytes), int64(freeFiles), nil
 }
 
+func (bdl *backpressureDiskLimiter) simpleByteTrackerFromType(typ diskLimitTrackerType) (
```
strib: `simpleResourceTrackerFromType`?  As I noted with `usedBytes`, there's some confusion here.  Can we make them consistent?  I think we should go all in with either "resources" or "bytes".

libkbfs/backpressure_disk_limiter.go
jzila: I was thinking to shorten the lock by one instruction, but yeah. Defer it is.

libkbfs/backpressure_disk_limiter.go
jzila: The tracker talks about resources (since it can track anything), and the limiter talks about bytes. I can change it, but there is a clear separation there in my mind.

jzila: @strib this is the last blocker on this PR. Want me to rename it or can I merge?

strib: Merging is fine with me.

PR #1065, opened 2017-07-11
tlf: make a panic-free way to access the TLF type

strib:
Issue: KBFS-2297

### DIFF
--- a/tlf/id.go
+++ b/tlf/id.go
@@ -147,8 +147,8 @@ func (id *ID) UnmarshalText(buf []byte) error {
 // SafeType returns the type of TLF represented by this ID.  If the ID
 // isn't valid, it returns tlf.Unknown along with an error.
 func (id ID) SafeType() (Type, error) {
-	if len(id.id) == 0 {
-		return Unknown, errors.New("Invalid empty TLF ID")
+	if len(id.id) != idByteLen {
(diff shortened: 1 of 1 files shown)


tlf/id.go
```diff
@@ -144,21 +144,34 @@ func (id *ID) UnmarshalText(buf []byte) error {
...
 
-// Type returns the type of TLF represented by this ID.
-//
-// Note that this function panics if the ID suffix is unknown, rather than
-// returning tlf.Unkonwn.
-func (id ID) Type() Type {
+// SafeType returns the type of TLF represented by this ID.  If the ID
+// isn't valid, it returns tlf.Unknown along with an error.
+func (id ID) SafeType() (Type, error) {
+	if len(id.id) == 0 {
```
songgao: Make this `if len(id.id) != idByteLen {`? It would be rare, but a malicious would still be able to crash our servers by passing in a non-empty short ID.

strib: I'll do that, but how would it be possible to crash this function that way?  It's only checking the last byte.

songgao: Yes, because it's checking the last byte by using `idByteLen-1` below at line 153

strib: D'oh, you're right, I misread that line.  Thanks.

PR #1066, opened 2017-07-12
[KBFS-2309] Count Meter

jzila:
This correctly tracks disk block cache stats over the past 1, 5, 15 minutes, and overall.

### DIFF
--- a/libkbfs/count_meter.go
+++ b/libkbfs/count_meter.go
@@ -7,39 +7,6 @@ import (
 	metrics "github.com/rcrowley/go-metrics"
 )
 
-// MeterSnapshot is a read-only copy of another Meter.
-type MeterSnapshot struct {
-	count                          int64
-	rate1, rate5, rate15, rateMean float64
-}
-
-// Count returns the count of events at the time the snapshot was taken.
-func (m *MeterSnapshot) Count() int64 { return m.count }
-
-// Mark panics.
-func (*MeterSnapshot) Mark(n int64) {
-	panic("Mark called on a MeterSnapshot")
-}
-
-// Rate1 returns the one-minute moving average rate of events per second at the
-// time the snapshot was taken.
-func (m *MeterSnapshot) Rate1() float64 { return m.rate1 }
-
-// Rate5 returns the five-minute moving average rate of events per second at
-// the time the snapshot was taken.
-func (m *MeterSnapshot) Rate5() float64 { return m.rate5 }
-
-// Rate15 returns the fifteen-minute moving average rate of events per second
-// at the time the snapshot was taken.
-func (m *MeterSnapshot) Rate15() float64 { return m.rate15 }
-
-// RateMean returns the meter's mean rate of events per second at the time the
-// snapshot was taken.
-func (m *MeterSnapshot) RateMean() float64 { return m.rateMean }
-
-// Snapshot returns the snapshot.
-func (m *MeterSnapshot) Snapshot() metrics.Meter { return m }
-
 // CountMeter counts ticks with a sliding window into the past.
 type CountMeter struct {
 	mtx      sync.RWMutex
@@ -48,6 +15,8 @@ type CountMeter struct {
 
 	shutdownCh chan struct{}
 }
+
+var _ metrics.Meter = (*CountMeter)(nil)
 
 func (m *CountMeter) tick() {
 	m.mtx.Lock()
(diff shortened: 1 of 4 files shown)


libkbfs/count_meter.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"sync"
+	"time"
+
+	metrics "github.com/rcrowley/go-metrics"
+)
+
+// MeterSnapshot is a read-only copy of another Meter.
+type MeterSnapshot struct {
```
strib: Please add this line somewhere:

```go
var _ metrics.Meter = (*MeterSnapshot)(nil)
```

Also, if this is copied from go-metrics, please say so, and say why it's copied (presumably because you can't instantiate one of those from another package?).

libkbfs/count_meter.go
```diff
@@ -0,0 +1,172 @@
...
+
+// RateMean returns the meter's mean rate of events per second at the time the
+// snapshot was taken.
+func (m *MeterSnapshot) RateMean() float64 { return m.rateMean }
+
+// Snapshot returns the snapshot.
+func (m *MeterSnapshot) Snapshot() metrics.Meter { return m }
+
+// CountMeter counts ticks with a sliding window into the past.
+type CountMeter struct {
```
strib: Please add this line somewhere:

```go
var _ metrics.Meter = (*CountMeter)(nil)
```

libkbfs/count_meter.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// Rate15 returns the number of ticks in the last 15 minutes.
+func (m *CountMeter) Rate15() float64 {
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+	return m.rate15()
+}
+
+// RateMean returns the overall mean rate of ticks.
```
strib: How does `m.overall` represent a mean?  Seems to be a total.

I guess the whole concept of a rate doesn't make much sense for a counter in the first place.

libkbfs/disk_block_cache.go
```diff
@@ -96,11 +96,12 @@ type MeterStatus struct {
...
+	s := m.Snapshot()
 	return MeterStatus{
-		m.Rate1(),
-		m.Rate5(),
-		m.Rate15(),
-		m.RateMean(),
+		s.Rate1(),
+		s.Rate5(),
+		s.Rate15(),
+		s.RateMean(),
```
strib: Maybe we should change `MeterStatus.Mean` to `MeterStatus.Count`, and use `s.Count()` here?

libkbfs/count_meter.go
```diff
@@ -0,0 +1,172 @@
...
+// Rate15 returns the fifteen-minute moving average rate of events per second
+// at the time the snapshot was taken.
+func (m *MeterSnapshot) Rate15() float64 { return m.rate15 }
+
+// RateMean returns the meter's mean rate of events per second at the time the
+// snapshot was taken.
+func (m *MeterSnapshot) RateMean() float64 { return m.rateMean }
+
+// Snapshot returns the snapshot.
+func (m *MeterSnapshot) Snapshot() metrics.Meter { return m }
```
strib: Actually, do you need this function (and do you need to import `metrics` at all)?  Unless you're passing this in somewhere that expects a `metrics.Meter`, this function is useless and you can disregard my comment above.

libkbfs/count_meter.go
```diff
@@ -0,0 +1,172 @@
...
+
+// RateMean returns the overall mean rate of ticks.
+func (m *CountMeter) RateMean() float64 {
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+	return m.rateMean()
+}
+
+// Snapshot returns the snapshot in time of this CountMeter.
+func (m *CountMeter) Snapshot() metrics.Meter {
```
strib: Same here, this could just return `MeterSnapshot` unless you really needed the interface-ness for some reason.

libkbfs/count_meter.go
jzila: I want to keep feature parity with go-metrics. I might make a PR to backport it to them.

libkbfs/count_meter.go
jzila: As I mentioned above, I want to maintain the `metrics.Meter` interface so we can backport it.

PR #1067, opened 2017-07-13
keybase_service_base: use refreshers to bust service and local team caches

strib:
Both KBFS and the service have team caches.  Cached info is generally invalidated by lazy messages sent via the server, but these may not arrive in time or at all in some scenarios.  In those cases, the client can specify exactly what information it is trying to validate, and if that info doesn't match what's in the cache, it can force the service to check with the server to get the most up-to-date information.

In particular:
* When trying to decrypt a team TLF MD encryted with a particular keygen, we need to check whether that's a known keygen yet.
* When trying to verify that a user + verifying key is a valid writer for a team TLF, we need to verify both that the verifying key is currently valid for the user, _and_ that the version of the user that contains that verifying key is a member of the team (using the "eldest seqno" currently associated with the user).  This means we have to plumb the verifying key into the `IsTeamWriter` call, which is kind of annoying, but it'll make future Merkle checks a bit easier I think.
* When trying to verify that the current signed-in user is a valid reader for a team TLF, we need to verify that the current version of the user is currently a reader for the team.  In this case, we don't need to know the eldest seqno, we can just set it to 0 and let the service use the current cached eldest seqno for that user.
* We also need to cache-bust based on the expected role of the user, in case the user role gets changed.

Issue: KBFS-2308

### DIFF
--- a/libkbfs/bare_root_metadata_v3.go
+++ b/libkbfs/bare_root_metadata_v3.go
@@ -397,7 +397,7 @@ func (md *BareRootMetadataV3) isNonTeamWriter(
 		}
 		return wkb.IsWriter(user, cryptKey), nil
 	default:
-		panic(fmt.Sprintf("Unknown TLF type: %s", md.TlfID().Type()))
+		return false, errors.Errorf("Unknown TLF type: %s", md.TlfID().Type())
 	}
 }
 
--- a/libkbfs/bare_root_metadata_v3.go
+++ b/libkbfs/bare_root_metadata_v3.go
@@ -374,11 +374,9 @@ func (md *BareRootMetadataV3) getTLFKeyBundles(extra ExtraMetadata) (
 	return &extraV3.wkb, &extraV3.rkb, nil
 }
 
-// IsWriter implements the BareRootMetadata interface for BareRootMetadataV3.
-func (md *BareRootMetadataV3) IsWriter(
+func (md *BareRootMetadataV3) isNonTeamWriter(
 	ctx context.Context, user keybase1.UID,
-	deviceKey kbfscrypto.CryptPublicKey, teamMemChecker TeamMembershipChecker,
-	extra ExtraMetadata) (bool, error) {
+	cryptKey kbfscrypto.CryptPublicKey, extra ExtraMetadata) (bool, error) {
 	switch md.TlfID().Type() {
 	case tlf.Public:
 		err := md.checkNonPrivateExtra(extra)
@@ -397,7 +395,18 @@ func (md *BareRootMetadataV3) IsWriter(
 		if err != nil {
 			return false, err
 		}
-		return wkb.IsWriter(user, deviceKey), nil
+		return wkb.IsWriter(user, cryptKey), nil
+	default:
+		panic(fmt.Sprintf("Unknown TLF type: %s", md.TlfID().Type()))
+	}
+}
+
+// IsWriter implements the BareRootMetadata interface for BareRootMetadataV3.
+func (md *BareRootMetadataV3) IsWriter(
+	ctx context.Context, user keybase1.UID,
+	cryptKey kbfscrypto.CryptPublicKey, verifyingKey kbfscrypto.VerifyingKey,
+	teamMemChecker TeamMembershipChecker, extra ExtraMetadata) (bool, error) {
+	switch md.TlfID().Type() {
 	case tlf.SingleTeam:
 		err := md.checkNonPrivateExtra(extra)
 		if err != nil {
@@ -411,30 +420,22 @@ func (md *BareRootMetadataV3) IsWriter(
 
 		// TODO: Eventually this will have to use a Merkle sequence
 		// number to check historic versions.
-		isWriter, err := teamMemChecker.IsTeamWriter(ctx, tid, user)
+		isWriter, err := teamMemChecker.IsTeamWriter(
+			ctx, tid, user, verifyingKey)
 		if err != nil {
 			return false, err
 		}
 		return isWriter, nil
 	default:
-		panic(fmt.Sprintf("Unknown TLF type: %s", md.TlfID().Type()))
+		return md.isNonTeamWriter(ctx, user, cryptKey, extra)
 	}
 }
 
 // IsReader implements the BareRootMetadata interface for BareRootMetadataV3.
 func (md *BareRootMetadataV3) IsReader(
 	ctx context.Context, user keybase1.UID,
-	deviceKey kbfscrypto.CryptPublicKey, teamMemChecker TeamMembershipChecker,
+	cryptKey kbfscrypto.CryptPublicKey, teamMemChecker TeamMembershipChecker,
 	extra ExtraMetadata) (bool, error) {
-	// Writers are also readers.
-	isWriter, err := md.IsWriter(ctx, user, deviceKey, teamMemChecker, extra)
-	if err != nil {
-		return false, err
-	}
-	if isWriter {
-		return true, nil
-	}
-
 	switch md.TlfID().Type() {
 	case tlf.Public:
 		err := md.checkNonPrivateExtra(extra)
@@ -443,11 +444,20 @@ func (md *BareRootMetadataV3) IsReader(
 		}
 		return true, nil
 	case tlf.Private:
+		// Writers are also readers.
+		isWriter, err := md.isNonTeamWriter(ctx, user, cryptKey, extra)
+		if err != nil {
+			return false, err
+		}
+		if isWriter {
+			return true, nil
+		}
+
 		_, rkb, err := md.getTLFKeyBundles(extra)
 		if err != nil {
 			return false, err
 		}
-		return rkb.IsReader(user, deviceKey), nil
+		return rkb.IsReader(user, cryptKey), nil
 	case tlf.SingleTeam:
 		err := md.checkNonPrivateExtra(extra)
 		if err != nil {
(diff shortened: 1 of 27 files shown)


strib:
Suggested reviewing order:

* `interfaces.go`
* `keybase_service_base.go`
* `kbpki_client.go`
* `*metadata*.go`
* Everything else.

libkbfs/keybase_service_base.go
```diff
@@ -425,7 +425,15 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
 	uid keybase1.UID, pollForKID keybase1.KID) (UserInfo, error) {
 	cachedUserInfo := k.getCachedUserInfo(uid)
 	if cachedUserInfo.Name != libkb.NormalizedUsername("") {
-		return cachedUserInfo, nil
+		if pollForKID == keybase1.KID("") {
+			return cachedUserInfo, nil
+		}
+		// Skip the cache if pollForKID isn't present.
```
strib: This was sort of a pre-existing bug -- we should be busting our local cache if the key isn't present.  However, `KBPKIClient.HasVerifyingKey` already took care of it on that path.  This way is better though.

libkbfs/bare_root_metadata_v3.go
```diff
@@ -397,7 +395,18 @@ func (md *BareRootMetadataV3) IsWriter(
 		if err != nil {
 			return false, err
 		}
-		return wkb.IsWriter(user, deviceKey), nil
+		return wkb.IsWriter(user, cryptKey), nil
+	default:
+		panic(fmt.Sprintf("Unknown TLF type: %s", md.TlfID().Type()))
```
jzila: I'm worried that a panic here will cause breakages if we introduce something other than `SingleTeam`, since this is called in a `default` case below. Wouldn't it be better to return an error?

libkbfs/interfaces.go
```diff
@@ -1893,7 +1905,8 @@ type BareRootMetadata interface {
 	IsFinal() bool
 	// IsWriter returns whether or not the user+device is an authorized writer.
 	IsWriter(ctx context.Context, user keybase1.UID,
-		deviceKey kbfscrypto.CryptPublicKey,
+		cryptKey kbfscrypto.CryptPublicKey,
+		verifyingKey kbfscrypto.VerifyingKey,
 		teamMemChecker TeamMembershipChecker, extra ExtraMetadata) (bool, error)
 	// IsReader returns whether or not the user+device is an authorized reader.
 	IsReader(ctx context.Context, user keybase1.UID,
```
jzila: Can you rename `deviceKey` to `cryptKey` here as well?

libkbfs/kbpki_client.go
```diff
@@ -216,8 +217,36 @@ func (k *KBPKIClient) GetCurrentMerkleSeqNo(ctx context.Context) (
...
+	ctx context.Context, tid keybase1.TeamID, uid keybase1.UID,
+	verifyingKey kbfscrypto.VerifyingKey) (bool, error) {
+	// Use the verifying key to find out the eldest seqno of the user.
+	userInfo, err := k.loadUserPlusKeys(ctx, uid, verifyingKey.KID())
+	if err != nil {
+		return false, err
+	}
+
+	found := false
+	for _, key := range userInfo.VerifyingKeys {
```
jzila: Could this be a map instead, allowing an O(1) lookup by `verifyingKey.KID()`?

libkbfs/keybase_service_base.go
```diff
@@ -425,7 +425,15 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
...
+		for _, key := range cachedUserInfo.VerifyingKeys {
```
jzila: As above, this could be a faster lookup if this was a map indexed by KID.

libkbfs/keybase_service_base.go
```diff
@@ -440,16 +448,58 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
...
 // KeybaseServiceBase.
 func (k *KeybaseServiceBase) LoadTeamPlusKeys(
-	ctx context.Context, tid keybase1.TeamID) (TeamInfo, error) {
+	ctx context.Context, tid keybase1.TeamID, desiredKeyGen KeyGen,
+	desiredUser keybase1.UserVersion, desiredRole keybase1.TeamRole) (
+	TeamInfo, error) {
+	if desiredRole != keybase1.TeamRole_NONE &&
+		desiredRole != keybase1.TeamRole_WRITER &&
+		desiredRole != keybase1.TeamRole_READER {
+		panic("KBFS shouldn't care about non-writer/reader roles")
```
jzila: But we're allowing `TeamRole_NONE`? Also, can we create a static `map[keybase1.TeamRole]bool` for allowed team roles that we can check rather than doing a 3-way comparison?

libkbfs/keybase_service_base.go
```diff
@@ -440,16 +448,58 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
...
+	}
+
 	cachedTeamInfo := k.getCachedTeamInfo(tid)
 	if cachedTeamInfo.Name != libkb.NormalizedUsername("") {
-		return cachedTeamInfo, nil
+		// If the cached team info doesn't satisfy our desires, don't
+		// use it.
+		satisfiesDesires := true
+		if desiredKeyGen >= FirstValidKeyGen {
```
jzila: This might be an obvious comment, but it saves having to imagine the scenario represented by the code.

`// If the desiredKeyGen is beyond the cached latest team info, then our cached key must be out of date.`

libkbfs/keybase_service_base.go
```diff
@@ -440,16 +448,58 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
...
+		if satisfiesDesires && desiredUser.Uid.Exists() {
+			// If the user is in the writer map, that satisfies none, reader
+			// or writer desires.
+			satisfiesDesires = cachedTeamInfo.Writers[desiredUser.Uid]
+			// If not, and the desire role is a reader, we need to
+			// check the reader map explicitly.
+			if !satisfiesDesires &&
+				(desiredRole == keybase1.TeamRole_NONE ||
+					desiredRole == keybase1.TeamRole_READER) {
+				satisfiesDesires = cachedTeamInfo.Readers[desiredUser.Uid]
```
jzila: You could do less checking by doing the `cachedTeamInfo.Readers` check first, and then the writers check second.

libkbfs/bare_root_metadata_v3.go
strib: I didn't introduce this, it already existed before.  I could fix it, but it seems like a panic on the client side is a good thing, to make it obvious when we miss adding cases to `select` like this.

However I do see it's used in one place on the server side, though only for public TLFs.  I guess I'll make it an error and hope that we'll be able to catch all these cases through simple errors next time.

libkbfs/kbpki_client.go
strib: It could be, but that's kind of a large-ish change within the package.  Do you think it's worth it?

jzila: I'll leave the call up to you.

strib: For reviewability I'll leave it out of this PR, and see if I can do it in a followup today.

libkbfs/interfaces.go
strib: Done.

libkbfs/keybase_service_base.go
jzila: Ah, I was initially going to ask if Writers subsumed Readers. Fair enough.

jzila: You can still remove the `||`:
```go
satisfiesDesires = cachedTeamInfo.Readers[desiredUser.Uid]
if !satisfiesDesires && desiredRole == keybase1.TeamRole_Writer) {
    satisfiesDesires = cachedTeamInfo.Writes[desiredUser.Uid]
}
```

strib: That doesn't work, because it doesn't match a writer-user if the role is READER or NONE.  (Because writers aren't in the reader map.)

jzila: I see. OK, sorry for my confusion here.

libkbfs/keybase_service_base.go
songgao: The if condition above seems to mean "just use the cached value if `pollForKID` isn't present"?

songgao: Assuming this review comment was meant to be at line 436: Do I understand correctly that we are not busting the local cache because HasVerifyingKey takes care of that already, even tho busting here would be better?

libkbfs/keybase_service_base.go
```diff
@@ -437,19 +445,68 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
 	return k.processUserPlusKeys(res)
 }
 
+var allowedLoadTeamRoles = map[keybase1.TeamRole]bool{
+	keybase1.TeamRole_NONE:   true,
```
songgao: When would `TeamRole_NONE` be used? The code below seems to suggest it's the same as `TeamRole_READER`?

libkbfs/keybase_service_base.go
strib: It was supposed to mean "isn't present _in VerifyingKeys_".  I'll update.

libkbfs/keybase_service_base.go
strib: NONE is used when the caller doesn't care what role the user has.  Right now it's treated the same as READER, but that might not be the case in the future.  It's also used as a zero value when the desiredUID is unset.

libkbfs/keybase_service_base.go
strib: Yes, that's the way it was before this PR.  Now, with this PR, we're busting the local cache here, even in `HasVerifyingKey`.

libkbfs/keybase_service_base.go
songgao: I see; thanks. I guess I should've asked what role `TeamRole_NONE` represents. But you just answered that it's used as a zero value when the UID is not set.

libkbfs/keybase_service_base.go
songgao: OIC thanks!

songgao: oooh got you thanks!

PR #1068, opened 2017-07-13
add a bytes capped cache and use it for key bundles cache

songgao:


libkbfs/key_bundle_cache.go
```diff
@@ -24,21 +25,27 @@ type keyBundleCacheKey struct {
...
+
+// NewKeyBundleCacheLRU constructs a new KeyBundleCacheStandard with LRU
+// eviction strategy. The capacity of the cache is set to capacityBytes bytes.
+func NewKeyBundleCacheLRU(capacityBytes int) *KeyBundleCacheStandard {
+	return &KeyBundleCacheStandard{cache.NewLRUEvictedCache(capacityBytes)}
+}
+
+// NewKeyBundleCacheRandom constructs a new KeyBundleCacheStandard with random
+// eviction strategy. The capacity of the cache is set to capacityBytes bytes.
+func NewKeyBundleCacheRandom(capacityBytes int) *KeyBundleCacheStandard {
```
songgao: Planning on using this for `mdmerkle`, since it always goes through all latest metadata, and if we ever need to evict cache entries, LRU might cause us to always have a cache miss assuming scanning order is deterministic.

vendor/vendor.json
```diff
@@ -94,6 +94,12 @@
 			"tree": true
 		},
 		{
+			"checksumSHA1": "nbCxiVT48CWEg6l8L4SEJ5VfN9c=",
+			"path": "github.com/golang/groupcache/lru",
```
songgao: Vendoring this in since the existing vendored `lru` package doesn't support uncapped LRUs, which is needed for our own bytes-counting based cache.

libkbfs/key_bundle_v3.go
```diff
@@ -214,6 +253,32 @@ func DeserializeTLFWriterKeyBundleV3(codec kbfscodec.Codec, path string) (
...
+	bytes += int(unsafe.Sizeof(wkb.TLFPublicKey))
+
+	// TLFEphemeralPublicKeys
+	bytes += tlfEphemeralPublicKeysSize(wkb.TLFEphemeralPublicKeys)
+
+	// EncryptedHistoricTLFCryptKeys
+	bytes += wkb.EncryptedHistoricTLFCryptKeys.encryptedData.Size()
+
+	// For codec.UnknownFieldSetHandler. It has a private map field which we
+	// can inspect unless extending the codec package. Just assume it's empty
```
strib: "can" -> "can't"

libkbfs/key_bundle_v3.go
```diff
@@ -326,6 +391,21 @@ func DeserializeTLFReaderKeyBundleV3(codec kbfscodec.Codec, path string) (
...
 
+// Size implements the cache.Measurable interface.
+func (rkb TLFReaderKeyBundleV3) Size() (bytes int) {
+	bytes += cache.PtrSize + rkb.Keys.Size() // Keys
+
+	// TLFEphemeralPublicKeys
+	bytes += tlfEphemeralPublicKeysSize(rkb.TLFEphemeralPublicKeys)
+
+	// For codec.UnknownFieldSetHandler. It has a private map field which we
+	// can inspect unless extending the codec package. Just assume it's empty
```
strib: "can" -> "can't"

libkbfs/cache/size_utils.go
```diff
@@ -0,0 +1,108 @@
...
+
+	hmapStructSize = IntSize + // count int
+		1 + 1 + 2 + 4 + // flags, B, noverflow, hash0
+		PtrSize*3 + // buckets, oldbuckets, nevacuate
+		PtrSize + 2*PtrSize // overflow (estimate; not counting the slice)
+
+	bucketSizeWithoutIndirectPointerOverhead = 8 // tophash only
+)
+
+func sizeWithIndirectPointerOverhead(zeroValue interface{}) int {
```
strib: This is actually computing `mapKeySize`, not just `size`, so the name should probably reflect that.

libkbfs/cache/size_utils.go
```diff
@@ -0,0 +1,108 @@
...
+	rawSize := int(reflect.TypeOf(zeroValue).Size())
```
strib: Is using `reflect` in this way expensive?  I don't know the answer.  Since this is on the critical path, if it's going to be expensive it might be better to pass the in from the caller, where it can easily be saved and reused for the same type.

libkbfs/cache/cache.go
```diff
@@ -0,0 +1,157 @@
...
+	mu          sync.RWMutex
+	cachedBytes int
+	data        map[string]memorizedMeasurable
+}
+
+// NewRandomEvictedCache returns a Cache that uses random eviction strategy.
+// The cache will have a capacity of maxBytes bytes. A zero-byte capacity cache
+// is valid.
+//
+// Internally we store a memorizing wrapper for the raw Mesurable to avoid
```
strib: Both this and the LRU cache have an implicit assumption that the size of the object isn't going to change after it's cached.  That's not mentioned anywhere, here or in the `Measurable` interface definition, which does mention that the item can have "dynamic" parts, which sounds to me like it can be changed after its cached.  I know that's not what you meant, but it's not obvious and should probably be documented somewhere.

One option is to get rid of `Measurable` and just have the caller to `Add` pass in a size, which should make it clear that the size shouldn't change after the `Add`.  But if you want to keep the current way, I think clearer documentation is needed.

songgao: I do prefer the interface way here. Will add the documentation for it!

libkbfs/cache/cache.go
```diff
@@ -0,0 +1,157 @@
...
+	if len(key)+memorized.Size() > c.maxBytes {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if _, ok := c.data[key]; !ok {
+		c.cachedBytes += len(key) + memorized.Size()
+	}
+	if c.cachedBytes > c.maxBytes {
+		for k, v := range c.data {
```
songgao: hmm somehow I thought that was in the spec. Thanks for bringing it up!

We could probably duplicate Go's `map` implementation and guarantee the random iteration ourselves, but that seems to be a pretty bad idea. I'll add comment for now.

libkbfs/cache/cache.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+	c.cachedBytes += len(key) + memorized.Size()
+	for c.cachedBytes > c.maxBytes {
+		c.data.RemoveOldest()
+	}
+	c.data.Add(lru.Key(key), memorized)
+}
+
+// notReallyACache is not really a cache but satisfies the cache interface.
```
songgao: Yup will do!

libkbfs/key_bundle_cache.go
```diff
@@ -24,21 +25,27 @@ type keyBundleCacheKey struct {
...
+	return &KeyBundleCacheStandard{cache.NewRandomEvictedCache(capacityBytes)}
 }
 
 // GetTLFReaderKeyBundle implements the KeyBundleCache interface for KeyBundleCacheStandard.
 func (k *KeyBundleCacheStandard) GetTLFReaderKeyBundle(
 	tlf tlf.ID, bundleID TLFReaderKeyBundleID) (*TLFReaderKeyBundleV3, error) {
-	cacheKey := keyBundleCacheKey{tlf, bundleID.String(), false}
-	if entry, ok := k.lru.Get(cacheKey); ok {
+	cacheKey := keyBundleCacheKey{tlf, bundleID.String(), false}.str()
```
songgao: It's to satisfy the interface on `Cache`'s method which accepts `string` keys. I forgot to annotate; if we are confident `bundleID` uniquely identities a bundle, perhaps we can just use the bundle ID?

libkbfs/cache/size_utils.go
songgao: We are using `memorizedMeasurable`, which should cache the size. In other words, the size itself is only re-calculated in the same situation when we need to fetch it through RPC. I think it's way less expensive than the RPC call itself?

libkbfs/key_bundle_v3.go
```diff
@@ -214,6 +253,32 @@ func DeserializeTLFWriterKeyBundleV3(codec kbfscodec.Codec, path string) (
 	return wkb, nil
 }
 
+func tlfEphemeralPublicKeysSize(keys kbfscrypto.TLFEphemeralPublicKeys) int {
+	pkSize := int(unsafe.Sizeof(kbfscrypto.TLFEphemeralPublicKey{}))
```
songgao: I guess we could ... although I don't know enough to really understand the risk of using `unsafe.Size` here. Could you elaborate?

strib: From https://golang.org/pkg/unsafe/:

> Package unsafe contains operations that step around the type safety of Go programs.
> 
> Packages that import unsafe may be non-portable and are not protected by the Go 1 compatibility guidelines.

It doesn't seem like a good thing to have in production code unless absolutely necessary.

songgao: Oh I see. Thanks!

libkbfs/cache/size_utils.go
strib: I'm talking about across different key bundles.  `memorizedMeasurable` will only cache it for a given key bundle, but the static sizes are always the same for the lifetime of the program.

I guess in this case you're right about the cache add being tied to an RPC, but nothing about this cache interface requires it to be used that way.

songgao: ooooh I see. I chose `zeroValue` over a int size for the interface to make it easier to use (so that caller need not to figure out how to calculate key/value size themselves. I guess we can split this into two separate functions one of which accept two `int`s for size.

libkbfs/key_bundle_cache.go
strib: Ah, I guess I was thinking the key was a `Measurable` also, but it's not.  Why not?

songgao:
Ready for review again. Thanks!

cache/cache.go
```diff
@@ -0,0 +1,157 @@
...
+//
+// Internally we store a memorizing wrapper for the raw Mesurable to avoid
+// unnecessarily frequent size calculations.
+//
+// Note:
+//
+// 1) Memorizing size means once the entry is in the cache, we never bother
+//    recalculating their size. It's fine if the size changes, but the cache
+//    eviction will continue using the old size.
+// 2) We are relying the fact that Go's map iteration is random to make
```
strib: "relying" -> "relying on"

cache/cache.go
```diff
@@ -0,0 +1,157 @@
...
+// eviction will continue using the old size.
+func NewLRUEvictedCache(maxBytes int) Cache {
+	c := &lruEvictedCache{
+		maxBytes: maxBytes,
+	}
+	c.data = &lru.Cache{
+		OnEvicted: func(key lru.Key, value interface{}) {
+			if memorized, ok := value.(memorizedMeasurable); ok {
+				if k, ok := key.(string); ok {
+					// No locking here as we do them in public methods Get/Add.
```
strib: See my comment from before about adding something here that clarifies that the locks should already be taken when this is called.  (It's not necessarily obvious that evictions only happen in those two methods; for all the reader knows, `lru.Cache` could have some background eviction goroutine.)

libkbfs/key_bundle_v3.go
```diff
@@ -25,6 +26,28 @@ import (
 // key information.
 type DeviceKeyInfoMapV3 map[kbfscrypto.CryptPublicKey]TLFCryptKeyInfo
 
+// Size implements the cache.Measurable interface.
+func (dkimV3 DeviceKeyInfoMapV3) Size() int {
+	// statically-sized part
+	mapSize := cache.StaticSizeOfMap(
```
strib: Can we make this `StaticSizeOfMapWithSize` instead and do the reflect only once and save it to a const?  Same below.

cache/cache.go
songgao: Oh sorry I missed that; not intentional!

songgao: Would you prefer a background eviction goroutine as well?

libkbfs/key_bundle_v3.go
songgao: good idea; sure!

cache/cache.go
strib: Nope, wasn't trying to say that.  Was just pointing out that the existing comment doesn't really give enough info to be confident about what's going on here.

songgao: ah OK; will update comment. Thanks!

libkbfs/key_bundle_cache.go
songgao: Oh sorry just realized I missed this comment. I guess I felt `string` was slightly lighter than a `Measurable` interface. Also an interface value means it can potentially be not comparable, making then not suitable for key type for `map`, and would cause a panic if such keys are passed in. We could check if they are comparable but that would require reflect.

cache/size_utils.go
```diff
@@ -0,0 +1,116 @@
...
+
+// StaticSizeOfMapWithSize is a slightly more efficient version of
+// StaticSizeOfMap for when the caller knows the static size of key and value
+// without having to use `reflect`.
+func StaticSizeOfMapWithSize(
+	keyStaticSize, valueStaticSize int, count int) (bytes int) {
+	keySize := mapKeyOrValueSizeWithIndirectPointerOverhead(keyStaticSize)
+	valueSize := mapKeyOrValueSizeWithIndirectPointerOverhead(valueStaticSize)
+
+	B := math.Ceil(math.Log2(float64(count) / 6.5 /* load factor */))
```
jzila: Can you explain this math?

cache/cache.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/golang/groupcache/lru"
+)
+
+// Cache defines an interface for a cache that stores Measurable content.
+// Eviction only happens when Add() is called, and there's no background
+// goroutine for eviction.
+type Cache interface {
+	// Get tries to find and return data assiciated with key.
+	Get(key string) (data Measurable, ok bool)
```
songgao: As described [above](https://github.com/keybase/kbfs/pull/1068#discussion_r127488940), it's to avoid panics without having to use `reflect` here. Our key is always a string anyway. But if you feel strongly that this should be an `interface{}`, I can change it into that. Let me know!

cache/cache.go
```diff
@@ -0,0 +1,161 @@
...
+	mu          sync.RWMutex
+	cachedBytes int
+	data        map[string]memorizedMeasurable
+}
+
+// NewRandomEvictedCache returns a Cache that uses random eviction strategy.
+// The cache will have a capacity of maxBytes bytes. A zero-byte capacity cache
+// is valid.
+//
+// Internally we store a memorizing wrapper for the raw Mesurable to avoid
```
songgao: ooooh I've always thought the word was "memorize". Will change all of them to memoizing. Thanks

cache/cache.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(key)+memorized.Size() > c.maxBytes {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if _, ok := c.data[key]; !ok {
+		c.cachedBytes += len(key) + memorized.Size()
+	}
+	if c.cachedBytes > c.maxBytes {
+		for k, v := range c.data {
```
songgao: Shuffling seems to be kind of expensive. The `map` iteration is random for now. Can we change to shuffling in the future if it's not random anymore?

cache/size_utils.go
```diff
@@ -0,0 +1,116 @@
...
+
+package cache
+
+import (
+	"math"
+	"reflect"
+)
+
+/*
+From src/runtime/hashmap.go:
```
songgao: We are not using these structs; we just use them to estimate the map size. Having them here in the same file helps with understanding the code.

cache/cache.go
jzila: I don't follow how it's random. It's not deterministic, but you can't rely on its randomness. Additionally, is it ok to delete elements from a map while iterating through it?

cache/size_utils.go
jzila: Can you put that in a comment?

cache/size_utils.go
songgao: It's the reverse of how `map` allocates: https://github.com/keybase/kbfs/pull/1068/files#diff-ceba16b135943d7b0876b767cef0f52cR31

cache/size_utils.go
songgao: Sure; will do

jzila: Also there's a potential `loadFactor` constant, but then you use the magic number below rather than a `loadFactor` constant. I'd rather you used the constant.

cache/cache.go
songgao: > but you can't rely on its randomness

Are you worried this would impact performance (because it's not randomly distributed) or if the behavior will change in future Go updates?

> is it ok to delete elements from a map while iterating through it

Yeah I think so. It's just not predictable if the deleted item will show up in iteration if we are deleting the one that's not iterated yet. But in this case we are deleting iterated so there shouldn't be any problem.

cache/size_utils.go
songgao: Sure; will do

cache/cache.go
jzila: > Are you worried this would impact performance (because it's not randomly distributed) 

Yes

songgao: OK looks like this is indeed a bad idea: https://github.com/golang/go/issues/5362#issuecomment-66078617

I'll make the change. Thanks!

songgao:
Ready for review again. Thanks!

cache/cache.go
```diff
@@ -24,63 +27,70 @@ type randomEvictedCache struct {
...
 // unnecessarily frequent size calculations.
 //
 // Note:
 //
-// 1) Memorizing size means once the entry is in the cache, we never bother
+// 1) Memoizing size means once the entry is in the cache, we never bother
 //    recalculating their size. It's fine if the size changes, but the cache
 //    eviction will continue using the old size.
-// 2) We are relying the fact that Go's map iteration is random to make
+// 2) We are relying on the fact that Go's map iteration is random to make
```
strib: Is this comment true anymore?

cache/cache.go
```diff
@@ -110,10 +120,12 @@ func NewLRUEvictedCache(maxBytes int) Cache {
 	}
 	c.data = &lru.Cache{
 		OnEvicted: func(key lru.Key, value interface{}) {
-			if memorized, ok := value.(memorizedMeasurable); ok {
+			// No locking is needed in this function because we do them in
+			// public methods Get/Add, and that RemoveOldest() is only called
```
strib: "and that" -> "and"

cache/cache.go
songgao: Nope will remove; thanks!

cache/cache.go
```diff
@@ -47,42 +49,48 @@ type randomEvictedCache struct {
...
-	if c.cachedBytes > c.maxBytes {
-		for k, v := range c.data {
-			delete(c.data, key)
-			c.cachedBytes -= len(k) + v.Size()
-			if c.cachedBytes <= c.maxBytes {
-				break
-			}
-		}
+	for c.cachedBytes > c.maxBytes {
+		c.evictOneLocked()
```
jzila: FWIW this isn't a uniform distribution: the items that are left after each removal get disproportionate consideration for removal. It should be fine in the long run though.

cache/size_utils.go
```diff
@@ -107,7 +111,8 @@ func StaticSizeOfMapWithSize(
 	keySize := mapKeyOrValueSizeWithIndirectPointerOverhead(keyStaticSize)
 	valueSize := mapKeyOrValueSizeWithIndirectPointerOverhead(valueStaticSize)
 
-	B := math.Ceil(math.Log2(float64(count) / 6.5 /* load factor */))
+	// See the comment of `B` field of `hmap` struct above.
+	B := math.Ceil(math.Log2(float64(count) / mapLoadFactor))
```
jzila: much better, thanks!

cache/cache.go
songgao: Yup; I went for a shuffle approach (partially shuffle on-demand) but realized just randomly draw from them was gonna be a lot easier and faster.

songgao:
Thanks!

PR #1070, opened 2017-07-14
libkbfs: fix two quota issues on startup and first block put

strib:
This fixes two quota issues noticed by @jzila:
1. The very first call to `EventuallyConsistentQuotaUsage.Get()` wasn't blocking when `blockTolerance` is very high.
2. If `.kbfs_status` was checked before any writes, the quota status wouldn't be filled in.

Issue: KBFS-2314

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -1075,19 +1075,27 @@ type backpressureDiskLimiterStatus struct {
 }
 
 func (bdl *backpressureDiskLimiter) getStatus(
-	chargedTo keybase1.UserOrTeamID) interface{} {
+	ctx context.Context, chargedTo keybase1.UserOrTeamID) interface{} {
 	bdl.lock.RLock()
 	defer bdl.lock.RUnlock()
 
 	currentDelay := bdl.getDelayLocked(
 		context.Background(), time.Now(), chargedTo)
 
+	jStatus := bdl.journalTracker.getStatus(chargedTo)
+	// If we haven't updated the quota limit yet, update it now.
+	if jStatus.QuotaStatus.QuotaBytes == math.MaxInt64 {
+		remoteUsedBytes, quotaBytes := bdl.quotaFn(ctx, chargedTo)
+		bdl.journalTracker.updateRemote(remoteUsedBytes, quotaBytes, chargedTo)
+		jStatus = bdl.journalTracker.getStatus(chargedTo)
+	}
+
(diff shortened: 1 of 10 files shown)


libkbfs/backpressure_disk_limiter.go
```diff
@@ -1075,19 +1075,27 @@ type backpressureDiskLimiterStatus struct {
...
+	ctx context.Context, chargedTo keybase1.UserOrTeamID) interface{} {
 	bdl.lock.RLock()
 	defer bdl.lock.RUnlock()
 
 	currentDelay := bdl.getDelayLocked(
 		context.Background(), time.Now(), chargedTo)
 
+	jStatus := bdl.journalTracker.getStatus(chargedTo)
+	// If we haven't updated the quota limit yet, update it now.
+	if jStatus.QuotaStatus.QuotaBytes == math.MaxInt64 {
```
jzila: I'm wondering if we should use an error status instead. But I don't feel strongly about it.

strib: You mean instead of fetching it synchronously, showing an error (something like "unknown") in the status instead?  Or do you mean something else?

jzila: I mean `getStatus` should maybe return an error, instead of using `math.MaxUint64` to branch.

strib: I see.  It's not really error though -- it's just that no one has happened to put a block into a journal yet.  I think it should be valid to call `Status()` before that happens, right?

jzila: Yeah. As I said, I don't feel strongly. Carry on!

libkbfs/quota_usage.go
```diff
@@ -118,7 +118,7 @@ func (q *EventuallyConsistentQuotaUsage) Get(
 	}()
 	past := q.config.Clock().Now().Sub(c.timestamp)
 	switch {
-	case past > blockTolerance:
+	case past > blockTolerance || c.timestamp.IsZero():
```
songgao: When `c.timestamp` is zero value, shouldn't `past > blockTolerance` be `true` almost for sure?

strib: No, note the `math.MaxInt64` [here](https://github.com/keybase/kbfs/blob/7331354d03eb7b5d59cdc88610190974908958bc/libkbfs/backpressure_disk_limiter.go#L719).

songgao: Oh I see; thanks!

PR #1071, opened 2017-07-14
data_types: add `EntryInfo.TeamWriter` field

strib:
To help track which user wrote which file in a team TLF.

Noticed by @jzila.

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -628,6 +628,9 @@ type EntryInfo struct {
 	Mtime int64
 	// Ctime is in unix nanoseconds
 	Ctime int64
+	// If this is a team TLF, we want to track the last writer of an
+	// entry, since in the block, only the team ID will be tracked.
+	TeamWriter keybase1.UID `codec:"tw,omitempty"`
 }
 
 // ReportedError represents an error reported by KBFS.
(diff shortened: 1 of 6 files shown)


libkbfs/data_types.go
```diff
@@ -628,6 +628,9 @@ type EntryInfo struct {
 	Mtime int64
 	// Ctime is in unix nanoseconds
 	Ctime int64
+	// If this is a team TLF, we want to track the last writer of an
+	// entry, since in the block, only the team ID will be tracked.
+	TeamWriter keybase1.UID `codec:"tw,omitempty"`
```
songgao: Do we maybe want this to be `Writer keybase.UID`? In other words, use this for both team folders and user folders so that 1) we can use the same code path when we get latest writer; 2) when we eventually upgrade to the same design for non-team TLFs this will work for that too.

strib: @songgao: Anywhere we have an `EntryInfo`, we will have a `BlockInfo` -- specifically, in `DirEntry`.  There's already a `BlockInfo.Writer keybase1.UserOrTeamID` field there, so we definitely can't use the `Writer` field name here.  For non-team TLFs, it doesn't seem useful to list the same UID twice, so in my opinion it's better to keep this field empty for non-TLF cases, to save bytes.

As for (2), this field won't be needed for non-team TLFs in the new design, just like it's not needed for non-team TLFs now.  The new design only changes where keys come from -- it shouldn't change anything about these data structures, or who writes what block.

songgao: Makes sense. Thanks for the explanation!

PR #1073, opened 2017-07-17
keybase_service_base: call `serviceLoggedIn` on new session

strib:
In at least two user reports, the client has gotten into a state where
journaling is not enabled, because the journal thinks it doesn't have
a current UID even though the user is logged in.  This happens when
the KBFS init code can't connect to the service to check the current
session (e.g., because the service is still restarting or down/slow
for some reason).  Later when KBFS does connect to the service, there
is no "logged in" notification (because the user was already logged
in), and so journal-enabling is never reenabled.

This PR executes the logged-in code whenever a new session is noticed,
not just when a new "logged in" notification comes in.

Issue: KBFS-2315

### DIFF
--- a/libkbfs/kbpki_client.go
+++ b/libkbfs/kbpki_client.go
@@ -86,31 +86,29 @@ func (k *KBPKIClient) hasVerifyingKey(ctx context.Context, uid keybase1.UID,
 		}
 	}
 
-	for key, t := range userInfo.RevokedVerifyingKeys {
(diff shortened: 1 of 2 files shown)


strib:
Arg, still struggling to get this to pass kbfs-server tests.  Removing reviewers now, will re-add you when it's in better shape.  Sorry for wasting anyone's time.

strib:
Arg, I accidentally pushed `master` onto this branch, which auto-closes the PR.  Derp. Opened #1074 instead.

PR #1075, opened 2017-07-19
journal_md_ops: be careful with caching in `GetForHandle`

strib:
We hit a test flake where the journal contained an unmerged MD update that conflicted with the server's MD update (but the conflict wasn't yet known).  Then the test infrastructure called `kbfsOps.getMaybeCreateRootNode`, which called `JournalMDOps.GetForHandle()`, which fetched the server's version for that revision and overwrote the value we had already cached.  Then CR ran and based the new branch on the server's MD by reading it out of the cache, even though it should have been based on the local version, and there was a failure.

This PR deals the problem in two different ways.

1. `JournalMDOps.GetForHandle` shouldn't cause the server's MD to be cached when it's just trying to look up the TLF ID.
2. We shouldn't overwrite existing items in the cache that were created locally; upper layers should explicitly delete/replace things when managing MDs.  Instead, local MDs should be put into the cache using `Replace`, while fetched MDs should use `Put`, which does no overwriting.

Issue: KBFS-2310

### DIFF
--- a/libkbfs/mdcache.go
+++ b/libkbfs/mdcache.go
@@ -15,8 +15,8 @@ import (
 // MDCacheStandard implements a simple LRU cache for per-folder
 // metadata objects.
 type MDCacheStandard struct {
-	// lock protects `lru` from atomic operations that need to
-	// atomicity across multiple `lru` calls.
+	// lock protects `lru` from atomic operations that need atomicity
+	// across multiple `lru` calls.
 	lock sync.RWMutex
 	lru  *lru.Cache
 }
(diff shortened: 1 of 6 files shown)


libkbfs/mdcache.go
```diff
@@ -13,7 +15,10 @@ import (
 // MDCacheStandard implements a simple LRU cache for per-folder
 // metadata objects.
 type MDCacheStandard struct {
-	lru *lru.Cache
+	// lock protects `lru` from atomic operations that need to
```
songgao: s/need to/need/ ?

libkbfs/mdcache.go
```diff
@@ -13,7 +15,10 @@ import (
...
+	// atomicity across multiple `lru` calls.
+	lock sync.RWMutex
+	lru  *lru.Cache
```
songgao: Not for this PR, but should we swap this out with the bytes capped LRU?

strib: Yes, we should at some point, definitely as part of another PR/issue.

PR #1077, opened 2017-07-19
[KBFS-2317] Only start the disk cache if it doesn't exist

jzila:


libkbfs/interfaces.go
```diff
@@ -65,7 +65,8 @@ type diskBlockCacheGetter interface {
 }
 
 type diskBlockCacheSetter interface {
-	SetDiskBlockCache(DiskBlockCache)
+	ResetDiskBlockCache(ctx context.Context) error
```
strib: I guess this function is not actually used anywhere?  Should we get rid of it?

jzila: I left it there to give us the option of restarting it, but I can remove it if you'd prefer.

strib: Slight preference for keeping the interface to a minimum, since we can easily re-add this function whenever we do need it.  But I don't care that much, so if you have plans for this in the near future, I'm fine with keeping it.

(Either way I'm happy to keep the logic split off into a reusable `*Locked` function.)

jzila: All right, I'll remove it.

PR #1079, opened 2017-07-19
libkbfs: store the full Merkle root in the MD update

strib:
Instead of just the seqno, which can be predicted/guessed.

This PR assumes we want to include the merkle root in the MD ID forever.  If this goes in, a future PR (for KBFS-2271) will add an "eventually consistent" caching layer for the merkle root to avoid a server RTT when setting this fields during foreground writes, in most cases.

Issue: KBFS-2296

### DIFF
--- a/libkbfs/bare_root_metadata_v3.go
+++ b/libkbfs/bare_root_metadata_v3.go
@@ -88,15 +88,18 @@ type BareRootMetadataV3 struct {
 	// of writing to the given folder.
 	FinalizedInfo *tlf.HandleExtension `codec:"fi,omitempty"`
 
-	// The sequence number of the global Keybase Merkle tree at the
-	// time this update was created (from the writer's perspective).
-	// This field was added to V3 after it was live for a while, and
-	// older clients that don't know about this field yet might copy
-	// it into new updates via the unknown fields copier. Which means
-	// new MD updates might end up referring to older Merkle roots.
-	// That's ok since this is just a hint anyway, and shouldn't be
-	// fully trusted when checking MD updates against the Merkle tree.
+	// DEPRECATED -- see KBMerkleRoot.Seqno instead.
 	KBMerkleSeqNo MerkleSeqNo `codec:"msn,omitempty"`
+
+	// The root of the global Keybase Merkle tree at the time this
+	// update was created (from the writer's perspective).  This field
+	// was added to V3 after it was live for a while, and older
+	// clients that don't know about this field yet might copy it into
+	// new updates via the unknown fields copier. Which means new MD
+	// updates might end up referring to older Merkle roots.  That's
+	// ok since this is just a hint anyway, and shouldn't be fully
+	// trusted when checking MD updates against the Merkle tree.
+	KBMerkleRoot keybase1.MerkleRootV2 `codec:"mr,omitempty"`
 
 	codec.UnknownFieldSetHandler
 }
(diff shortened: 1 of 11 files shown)


libkbfs/bare_root_metadata_v3.go
```diff
@@ -88,16 +88,19 @@ type BareRootMetadataV3 struct {
...
 
-	// The sequence number of the global Keybase Merkle tree at the
-	// time this update was created (from the writer's perspective).
-	// This field was added to V3 after it was live for a while, and
-	// older clients that don't know about this field yet might copy
-	// it into new updates via the unknown fields copier. Which means
-	// new MD updates might end up referring to older Merkle roots.
-	// That's ok since this is just a hint anyway, and shouldn't be
-	// fully trusted when checking MD updates against the Merkle tree.
+	// DEPRECATED -- see KBMerkleRoot.Seqno instead.
```
songgao: I guess alternatively we could also just add `keybase1.HashMeta` for on-the-wire message, because `keybase1.MerkleRootV2` is essentially `keybase1.Seqno` plus `keybase1.HashMeta`. The only difference is `keybase1.Seqno` is `int64`, while `libkbfs.Seqno` is `uint64`. Not sure if it makes any difference for msgpack tho.

libkbfs/keybase_service_base.go
```diff
@@ -552,24 +552,17 @@ func (k *KeybaseServiceBase) LoadTeamPlusKeys(
...
 
-	if res.Root.Seqno < 0 {
-		return 0, fmt.Errorf(
-			"Illegal negative merkle seqno: %d", res.Root.Seqno)
-	}
-
-	// NOTE: `res.Seqno` is an int64, while `MerkleSeqNo` is a uint64,
-	// so casting in this direction should be safe.
-	return MerkleSeqNo(res.Root.Seqno), nil
+	return res.Root, nil
```
songgao: hmm two question here:

1) Why was the `res.Root.Seqno < 0` check needed before this PR?

2) Why not check here anymore? For example, `res.Root.HashMeta` can still be `nil`.

libkbfs/bare_root_metadata_v3.go
strib: Yeah fair point.  I was originally thinking that, if we were going to ignore the merkle stuff during ID computation, it would be best to keep it all in a single field.  Though one other advantage to this way is that if other fields are added to `MerkleRootV2` in the future, we'll get them for free.

But actually, I don't think I need to deprecate the field at all; I can just nuke it.  The unknown fields should take care of it.  Does that sound ok?

libkbfs/keybase_service_base.go
strib: > Why was the res.Root.Seqno < 0 check needed before this PR?

In the current code, I'm converting an int to a uint, and now I'm not doing that anymore, I'm just copying the root exactly as the service tells us.  So I don't think any checks are needed.  I guess we could be defensive and add checks, but I don't think it's really necessary.

libkbfs/bare_root_metadata_v3.go
songgao: Oh good point. The unknown handler would populate the zero value, which is also defined as unknown, right? If so, sounds good to me!

> one other advantage to this way is that if other fields are added to MerkleRootV2 in the future, we'll get them for free.

Good point. Any updates to the `MerkleRootV2` would be backward compatible, so it's safe to have that also here!

libkbfs/keybase_service_base.go
songgao: I see; thanks!

libkbfs/bare_root_metadata_v3.go
strib: > The unknown handler would populate the zero value, which is also defined as unknown, right?

If a new client sees an old MD update, the `KBMerkleSeqNo` field would go into the unknown fields set -- there would be no `KBMerkleSeqNo` to set.  If an old client sees a new MD update, there would be no such field, so it would stay as the zero value, i.e. the defined unknown value.

strib:
I had to change the merkle root field to be a pointer, to get the right `omitempty` behavior.

PR #1080, opened 2017-07-20
libkbfs: introduce eventually-consistent merkle root cache

strib:
This PR factors out the existing logic in `EventuallyConsistentQuotaUsage` into a common helper struct, and uses it to construct a cache for the service Merkle root.  The goal is to keep the server RTT for fetching the merkle root from blocking foreground writes.  Currently, `SyncAll` absorbs this RTT when the service needs to fetch from the server, and `SyncAll` blocks foreground writes.

This also improves the fetching logic by making sure there's only ever one outstanding fetch; subsequent blocking fetches wait for the outstanding fetch to complete.

`folderBranchOps` signals that it needs the merkle root as soon as a write comes in.  Hopefully, 1 second later when `SyncAll` is called, the merkle root will be in the cache and ready to be used.

Issue: KBFS-2271

### DIFF
--- a/libkbfs/fetch_decider.go
+++ b/libkbfs/fetch_decider.go
@@ -1,0 +1,135 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import (
+	"sync"
+	"time"
+
+	"github.com/keybase/client/go/logger"
+	"golang.org/x/net/context"
+)
+
+// fetchDecider is a struct that helps avoid having too frequent calls
+// into a remote server.
+type fetchDecider struct {
+	log     logger.Logger
+	fetcher func(ctx context.Context) error
+	tagKey  interface{}
+	tagName string
+	clock   clockGetter
+
+	blockingForTest chan<- struct{}
+
+	lock    sync.Mutex
+	readyCh chan struct{}
+	errPtr  *error
+}
+
+func newFetchDecider(
+	log logger.Logger, fetcher func(ctx context.Context) error,
+	tagKey interface{}, tagName string,
+	clock clockGetter) *fetchDecider {
+	return &fetchDecider{
+		log:     log,
+		fetcher: fetcher,
+		tagKey:  tagKey,
+		tagName: tagName,
+		clock:   clock,
+	}
+}
+
+func (fd *fetchDecider) launchBackgroundFetch(ctx context.Context) (
+	readyCh <-chan struct{}, errPtr *error) {
+	fd.lock.Lock()
+	defer fd.lock.Unlock()
+
+	if fd.readyCh != nil {
+		fd.log.CDebugf(ctx, "Waiting on existing fetch")
+		// There's already a fetch in progress.
+		return fd.readyCh, fd.errPtr
+	}
+
+	fd.readyCh = make(chan struct{})
+	fd.errPtr = new(error)
+
+	id, err := MakeRandomRequestID()
+	if err != nil {
+		fd.log.Warning("Couldn't generate a random request ID: %v", err)
+	}
+	fd.log.CDebugf(
+		ctx, "Spawning fetch in background with tag:%s=%v", fd.tagName, id)
+	go func() {
+		// Make a new context so that it doesn't get canceled
+		// when returned.
+		logTags := make(logger.CtxLogTags)
+		logTags[fd.tagKey] = fd.tagName
+		bgCtx := logger.NewContextWithLogTags(
+			context.Background(), logTags)
+		bgCtx = context.WithValue(bgCtx, fd.tagKey, id)
+		// Make sure a timeout is on the context, in case the
+		// RPC blocks forever somehow, where we'd end up with
+		// never resetting backgroundInProcess flag again.
+		bgCtx, cancel := context.WithTimeout(bgCtx, 10*time.Second)
+		defer cancel()
+		err := fd.fetcher(bgCtx)
+
+		// Notify everyone we're done fetching.
+		fd.lock.Lock()
+		defer fd.lock.Unlock()
+		fd.log.CDebugf(bgCtx, "Finished fetch")
+		*fd.errPtr = err
+		readyCh := fd.readyCh
+		fd.readyCh = nil
+		fd.errPtr = nil
+		close(readyCh)
+	}()
+	return fd.readyCh, fd.errPtr
+}
+
+// Do decides whether to block on a fetch, launch a background fetch
+// and use existing cached value, or simply use the existing cached
+// value with no more fetching. The caller can provide a positive
+// tolerance, to accept stale LimitBytes and UsageBytes data. If
+// tolerance is 0 or negative, this always makes a blocking RPC to
+// bserver and return latest quota usage.
+//
+// 1) If the age of cached data is more than blockTolerance, it blocks
+// until a new value is fetched and ready in the caller's cache.
+// 2) Otherwise, if the age of cached data is more than bgTolerance,
+// a background RPC is spawned to refresh cached data using `fd.fetcher`,
+// but returns immediately to let the caller use stale data.
+// 3) Otherwise, it returns immediately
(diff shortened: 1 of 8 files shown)


libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+	}()
+	return fd.readyCh, fd.errPtr
+}
+
+// Do decides whether to block on a fetch, launch a background fetch
+// and use existing cached value, or simply use the existing cached
+// value with no more fetching. The caller can provide a positive
+// tolerance, to accept stale LimitBytes and UsageBytes data. If
+// tolerance is 0 or negative, this always makes a blocking RPC to
+// bserver and return latest quota usage.
```
songgao: This line needs to be generalized now that this is used for non-quota stuff.

libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+// 3) Otherwise, it returns immediately
+func (fd *fetchDecider) Do(
+	ctx context.Context, bgTolerance, blockTolerance time.Duration,
+	cachedTimestamp time.Time) (err error) {
+	past := fd.clock.Clock().Now().Sub(cachedTimestamp)
+	switch {
+	case past > blockTolerance || cachedTimestamp.IsZero():
+		fd.log.CDebugf(
+			ctx, "Blocking on fetch; cached data is %s old", past)
+		readyCh, errPtr := fd.launchBackgroundFetch(ctx)
```
songgao: This means even blocking fetches will have a (new) background tag. We are logging the background tag with the original tag, so it's still possible to trace. I don't think it'd be a problem but just thought I'd point it out.

strib: Yeah true.  I thought about that a little bit.  We could use a `linkedContext` to keep the debug IDs the same, but it doesn't really seem worth doing since we probably don't want to do that for the `bgTolerance` case.  And as you mentioned, we have the debug message linking the two already.  So I'll probably keep it as is, unless someone feels strongly about it.

songgao: Sounds good thanks!

libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+		// when returned.
+		logTags := make(logger.CtxLogTags)
+		logTags[fd.tagKey] = fd.tagName
+		bgCtx := logger.NewContextWithLogTags(
+			context.Background(), logTags)
+		bgCtx = context.WithValue(bgCtx, fd.tagKey, id)
+		// Make sure a timeout is on the context, in case the
+		// RPC blocks forever somehow, where we'd end up with
+		// never resetting backgroundInProcess flag again.
+		bgCtx, cancel := context.WithTimeout(bgCtx, 10*time.Second)
```
jzila: Should this duration be a constant, or a parameter to the `fetchDecider`?

libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+
+		// Notify everyone we're done fetching.
+		fd.lock.Lock()
+		defer fd.lock.Unlock()
+		fd.log.CDebugf(bgCtx, "Finished fetch: %+v", err)
+		*fd.errPtr = err
+		readyCh := fd.readyCh
+		fd.readyCh = nil
+		fd.errPtr = nil
+		close(readyCh)
```
jzila: We're under lock, so why can't these lines be:
```go
close(fd.readyCh)
fd.readyCh = nil
fd.errPtr = nil
```

libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+// fetchDecider is a struct that helps avoid having too frequent calls
+// into a remote server.
+type fetchDecider struct {
+	log     logger.Logger
+	fetcher func(ctx context.Context) error
+	tagKey  interface{}
+	tagName string
+	clock   clockGetter
```
jzila: Is there a reason this can't just be an embedded struct, allowing us to do `fd.Clock()` rather than `fd.clock.Clock()`?

libkbfs/fetch_decider.go
jzila: I see some risk below with `*errPtr` being evaluated immediately after the channel is closed, but I think that's okay because you've already set `*fd.errPtr` before closing the channel even with my above suggestion.

libkbfs/fetch_decider.go
```diff
@@ -0,0 +1,135 @@
...
+// value with no more fetching. The caller can provide a positive
+// tolerance, to accept stale LimitBytes and UsageBytes data. If
+// tolerance is 0 or negative, this always makes a blocking call using
+// `fd.fetcher`.
+//
+// 1) If the age of cached data is more than blockTolerance, it blocks
+// until a new value is fetched and ready in the caller's cache.
+// 2) Otherwise, if the age of cached data is more than bgTolerance,
+// a background RPC is spawned to refresh cached data using `fd.fetcher`,
+// but returns immediately to let the caller use stale data.
```
jzila: I'm confused, why are we launching a background fetch if we're just going to use the cached value anyway? Why not lazy evaluate that fetch once the age is above `bgTolerance`?

libkbfs/merkle_root.go
```diff
@@ -0,0 +1,106 @@
...
+		ecmr.log, ecmr.getAndCache, ECMRCtxTagKey{}, ECMRID, ecmr.config)
+	return ecmr
+}
+
+func (ecmr *EventuallyConsistentMerkleRoot) getAndCache(
+	ctx context.Context) (err error) {
+	defer func() {
+		ecmr.log.CDebugf(ctx, "getAndCache: error=%v", err)
+	}()
+	// Go through the
```
jzila: Go through the...?

libkbfs/fetch_decider.go
strib: I don't really follow.  `clockGetter` is an interface, not a struct, so it can't really be embedded. Do you want something like this?

```go
type fetchDecider struct {
Clock

...
}
```

That would mean we couldn't swap out the clock at runtime, which defeats the purpose of passing in the getter.

But I might be misunderstanding what you're asking.

libkbfs/fetch_decider.go
strib: Sure, made it a constant for now.

libkbfs/fetch_decider.go
strib: Sure, I'll do that.

libkbfs/fetch_decider.go
strib: Yeah, the point is to try to refresh the value before `blockTolerance` is hit, so that future callers don't ever block while waiting.

(This is the behavior that `EventuallyConsistentQuotaUsage` has always had, I just factored it out here to be reusable.)

libkbfs/merkle_root.go
strib: I must have passed out.

songgao: I blame Desmond

libkbfs/fetch_decider.go
jzila: You can still swap out the clock at runtime by assigning to `fd.Clock`, but yes, either embedding `Clock` or `clockGetter` should work.

libkbfs/fetch_decider.go
jzila: What's to stop multiple background goroutines at once in this condition?

strib: `launchBackgroundFetch` guarantees there's only one fetch outstanding at once, even if multiple bg  goroutines try to initiate a bg fetch at once.

libkbfs/fetch_decider.go
strib: We could technically swap it out if it were embedded, but then `ConfigLocal.SetClock` would have to somehow know where all the users of `fetchDecider` were, in order to swap it out.  And it only has access to other interfaces, so it would have to do casting and a bunch of other stuff to get it done.

Ok, I embedded `clockGetter`.

PR #1083, opened 2017-07-21
revendor go-framed-msgpack-rpc

oconnor663:
One more callsite for res_init.

r? @strib


oconnor663:
In parallel with: https://github.com/keybase/client/pull/7861

PR #1084, opened 2017-07-21
properly update size in randomEvicted cache when replacing entries

songgao:


cache/cache.go
```diff
@@ -78,9 +78,11 @@ func (c *randomEvictedCache) Add(key string, data Measurable) {
 	}
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if _, ok := c.data[key]; !ok {
-		c.cachedBytes += increase
+	if v, ok := c.data[key]; ok {
+		decrease := 2*len(key) + v.Size()
```
jzila: It seems like the pattern `2*len(key) + val.Size()` is used a bunch: should it be a function?

songgao: sure; will do

cache/cache.go
```diff
@@ -48,12 +48,22 @@ func NewRandomEvictedCache(maxBytes int) Cache {
 	}
 }
 
+func (c *randomEvictedCache) entrySize(key *string, value Measurable) int {
```
jzila: Why is the string a pointer? Strings are copy-on-write, aren't they?

songgao: Really? I had no idea! I thought it was always copied

songgao: Updated; thanks!

jzila: Looks like calling it copy-on-write is not strictly accurate, even if the semantics are the same:
https://stackoverflow.com/a/8532749

songgao: Yeah it's immutable, so there's no "write". But for the same reason it's not copied, so still good point!

PR #1085, opened 2017-07-22
folder_branch_ops: fix the timestamps in update history

strib:
The previous one was just supposed to be a placeholder; it only tracks
the actual update time if the root directory was updated (and the
user's clock is synced).

Issue: KBFS-2321

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -6195,7 +6195,7 @@ func (fbo *folderBranchOps) GetUpdateHistory(ctx context.Context,
 		}
 		updateSummary := UpdateSummary{
 			Revision:  rmd.Revision(),
-			Date:      time.Unix(0, rmd.data.Dir.Mtime),
+			Date:      rmd.localTimestamp,
 			Writer:    writer,
 			LiveBytes: rmd.DiskUsage(),
 			Ops:       make([]OpSummary, 0, len(rmd.data.Changes.Ops)),


libkbfs/folder_branch_ops.go
```diff
@@ -6195,7 +6195,7 @@ func (fbo *folderBranchOps) GetUpdateHistory(ctx context.Context,
 		}
 		updateSummary := UpdateSummary{
 			Revision:  rmd.Revision(),
-			Date:      time.Unix(0, rmd.data.Dir.Mtime),
```
songgao: What's the `rmd.data.dir.Mtime`? Is it the TLF root's mtime?

libkbfs/folder_branch_ops.go
```diff
@@ -6195,7 +6195,7 @@ func (fbo *folderBranchOps) GetUpdateHistory(ctx context.Context,
...
+			Date:      rmd.localTimestamp,
```
songgao: We are using `localTimestamp` here because the server timestamp is not trusted? Although this history can include updates from other devices, so there can be clock issues, and the time when the updates are merged to master also may matter. Do you think we should include server timestamp in addition to the local one?

strib:
> How do we handle timezone changes?

`time.Time` handles all that for us.  When printed in `.kbfs_update_history`, it prints out the timezone.  (On the wire and in the server DB, it's just an epoch time.)

libkbfs/folder_branch_ops.go
strib: Yes.

libkbfs/folder_branch_ops.go
strib: This is the server's timestamp, adjusted for the local clock according to ping latency measurements: https://github.com/keybase/kbfs/blob/master/libkbfs/root_metadata.go#L948-L957

songgao: ooooh I see; thanks!

PR #1089, opened 2017-07-24
jenkinsfile: Generate mocks and fail if it causes any diffs.

jzila:


Jenkinsfile
```diff
@@ -69,11 +69,10 @@ helpers.rootLinuxNode(env, {
...
                         env.COMMIT_HASH = readFile('kbfsfuse/revision')
-                        sh 'echo -n $(git --no-pager show -s --format="%an" HEAD) > .author_name'
-                        sh 'echo -n $(git --no-pager show -s --format="%ae" HEAD) > .author_email'
-                        env.AUTHOR_NAME = readFile('.author_name')
-                        env.AUTHOR_EMAIL = readFile('.author_email')
-                        sh 'rm .author_name .author_email'
+                        env.AUTHOR_NAME = sh(returnStdout: true, script: 'git --no-pager show -s --format="%an" HEAD').trim()
+                        env.AUTHOR_EMAIL = sh(returnStdout: true, script: 'git --no-pager show -s --format="%ae" HEAD').trim()
+                        sh 'git add kbfsfuse/revision'
+                        sh "git -c user.name='Jenkins' -c user.email='ci@keyba.se' commit -m 'revision'"
```
strib: I don't follow why `kbfsfuse/revision` has to be committed locally.  Can you just `rm` it after you get `env.COMMIT_HASH`?  Or use the same `env.COMMIT_HASH = sh(...)` trick you use for the other vars?

jzila: I need it to go into the docker.

strib: Ah ok.  Add a comment explaining what's going on?

jzila: Will do.

PR #1091, opened 2017-07-27
[KBFS-2219] Sync Cache

jzila:
This is an initial PR for the sync cache. Still need some integration tests.
- [x] `Config` interface and implementation for determining whether a TLF is to be synced.
- [x] Separate working set cache from sync cache.
- [x] Wrapped cache so the main components don't have to care if a TLF is being synced.
- [x] Provide a way for the prefetcher and cache to know if the immediate child level's prefetches are done.
- [x] Refactor current priorities such that we have a much greater range of priorities that can be prefetched but are lower than on-demand. On-demand requests run on on-demand retrieval workers, and prefetch requests run on prefetch workers. Priorities:
  * [INT_MIN, 0]: Will run on prefetch worker and won't trigger prefetch.
  * [1, 2^30-1]: Will run on prefetch worker and will trigger prefetch.
  * [2^30, INT_MAX]: Will run on on-demand workers and will trigger prefetch.
- [x] Handle and test a full sync cache (don't evict).
- [x] Traslate blocks from the working set cache to the sync cache on a `Put`.
- [x] Allow the user to enable the sync cache for a TLF via the `.kbfs_sync_enable` and `.kbfs_sync_disable` files in a TLF.
- [x] Tracking metrics: display the status for both caches.
- [ ] KBFS Docker tests to make sure that syncing works: one for enabling/disabling, and one for remote update for an existing shared TLF across devices.
- [ ] TODO Ticket: prefetcher should have a priority reservation system for prefetching, such that each triggered prefetch is lower priority than all prior triggered prefetches. This ensures that the tree is synced in somewhat depth-first traversal.
- [ ] TODO Ticket: more intelligent disk limiting that allows the sync cache to be more flexible.


libfs/sync_control_file.go
```diff
@@ -0,0 +1,59 @@
...
+		c.SetTlfSyncState(fb.Tlf, true)
+
+	case SyncDisable:
+		c.SetTlfSyncState(fb.Tlf, false)
+
+	default:
+		return fmt.Errorf("Unknown action %s", a)
+	}
+	// Re-trigger prefetches.
+	c.KBFSOps().GetRootNode(ctx, h, fb.Branch)
```
strib: Also, just so I understand, this doesn't actually cause a full prefetch yet, as of this PR, right? You just have this in here in preparation for a future PR that will do full prefetching when syncing is turned on for the TLF?

Once that is ready, in this case of going from enable -> disable, this probably won't do much I'm guessing, other than move the root block and maybe its immediate descendants over to the working set cache, right? There will still be lots of other blocks stuck in the sync cache?

libkbfs/block_retrieval_queue.go
```diff
@@ -221,47 +225,62 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
...
 	default:
 		// We should return the error here because otherwise we could thrash
 		// the prefetcher.
 		return err
 	}
 	// This must be called in a goroutine to prevent deadlock in case this
 	// CacheAndPrefetch call was triggered by the prefetcher itself.
-	go brq.Prefetcher().PrefetchAfterBlockRetrieved(block, ptr, kmd)
+	go func() {
+		<-brq.Prefetcher().PrefetchAfterBlockRetrieved(block,
```
strib: This seems to mean that it will be marked as `donePrefetched` once the immediate descendants are fetched.  Depending on how that value is used in the future, does it mean that if a TLF is only partially prefetched, the unprefetched blocks will never be attempted again (unless they are requested explicitly)?

libfs/sync_control_file.go
jzila: Yeah I wonder whether I should trigger a traversal that blasts away the sync cache blocks.

libkbfs/block_retrieval_queue.go
```diff
@@ -221,47 +225,62 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
...
 	err = brq.config.BlockCache().PutWithPrefetch(ptr, kmd.TlfID(), block,
 		lifetime, hasPrefetched)
 	switch err.(type) {
 	case nil:
 	case cachePutCacheFullError:
-		brq.log.CDebugf(ctx, "Skipping prefetch because the cache "+
-			"is full")
-		return err
+		brq.log.CWarningf(ctx, "Cache is full, but prefetching anyway to "+
+			"populate the disk cache.")
```
jzila: That fix was from before the disk cache. We'll get errors when trying to place the blocks into the in-memory cache, but at least we can put them in the disk cache so that we don't have to re-retrieve them later.

libkbfs/block_retrieval_queue.go
jzila: Yeah. As you've identified, this isn't currently used, and I wrote the functionality speculatively. It currently gets set even when there are errors retrieving the underlying blocks.

The idea here is to provide a fast way to determine whether child blocks have been prefetched, so that we can provide an API for it. As you've identified, the implementation currently fails to do that. I'll update it to behave correctly.

libkbfs/disk_block_cache.go
```diff
@@ -466,31 +463,41 @@ func (cache *DiskBlockCacheStandard) encodeBlockCacheEntry(buf []byte,
...
+	}
+	return nil
+}
+
+// Get implements the DiskBlockCache interface for DiskBlockCacheStandard.
+func (cache *DiskBlockCacheStandard) Get(ctx context.Context, tlfID tlf.ID,
+	blockID kbfsblock.ID) (buf []byte,
+	serverHalf kbfscrypto.BlockCryptKeyServerHalf, hasPrefetched bool,
+	err error) {
+
```
jzila: Deliberate for readability.

libkbfs/disk_block_cache.go
```diff
@@ -619,60 +646,56 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
...
-	default:
-		// If the cache hasn't started yet, return 0.
-		return 0
-	}
-	cache.lock.RLock()
-	defer cache.lock.RUnlock()
-	// shutdownCh has to be checked under lock, otherwise we can race.
-	select {
-	case <-cache.shutdownCh:
+
```
jzila: I'm using the blank line to separate out the locking and checking from the rest of the functionality for each of the interface methods.

libkbfs/disk_block_cache_test.go
```diff
@@ -519,3 +531,48 @@ func TestDiskBlockCacheWithRetrievalQueue(t *testing.T) {
...
+			clock.Add(time.Second)
+		}
+	}
+
+	t.Log("Set the cache maximum bytes to the current total.")
+	currBytes := int64(cache.currBytes)
+	limiter := config.DiskLimiter().(*backpressureDiskLimiter)
+	limiter.syncCacheByteTracker.limit = currBytes
+
+	t.Log("Add a block to the cache. Verify that blocks were evicted.")
```
jzila: Yep, copy-paste error.

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return cache.workingSetCache.GetMetadata(ctx, blockID)
+}
+
+// Put implements the DiskBlockCache interface for diskBlockCacheWrapped.
+func (cache *diskBlockCacheWrapped) Put(ctx context.Context, tlfID tlf.ID,
+	blockID kbfsblock.ID, buf []byte,
+	serverHalf kbfscrypto.BlockCryptKeyServerHalf) error {
+	if cache.config.IsSyncedTlf(tlfID) {
+		cache.workingSetCache.Delete(ctx, []kbfsblock.ID{blockID})
```
jzila: I was torn about this: on the one hand I wanted it to be atomic to minimize disk duplication, but I think you're right, especially since LevelDB doesn't guarantee immediate disk cleanup. I'll put the `Delete`s in a goroutine.

libkbfs/interfaces.go
```diff
@@ -925,13 +930,17 @@ type DiskBlockCache interface {
 	// Delete deletes some blocks from the disk cache.
 	Delete(ctx context.Context, blockIDs []kbfsblock.ID) (numRemoved int,
 		sizeRemoved int64, err error)
-	// UpdateMetadata updates the LRU time to Now() for a given block.
+	// UpdateMetadata updates metadata for a given block in the disk cache.
 	UpdateMetadata(ctx context.Context, blockID kbfsblock.ID,
-		hasPrefetched bool) error
+		hasPrefetched, donePrefetch bool) error
```
jzila: Yeah. I originally envisioned this as a single-level cache of the sync state of a given folder. But without the full depth it doesn't seem that helpful. WDYT? Should I spend some effort thinking about how to do `finishedPrefetch` in depth?

libkbfs/prefetcher.go
```diff
@@ -186,27 +217,44 @@ func (p *blockPrefetcher) PrefetchBlock(
...
 		} else {
-			p.prefetchDirectDirBlock(ptr, b, kmd)
+			wg = p.prefetchDirectDirBlock(ptr, b, kmd)
 		}
 	default:
 		// Skipping prefetch for block of unknown type (likely CommonBlock)
+		close(doneCh)
+		return doneCh
 	}
+	go func() {
```
jzila: Hah I had it that way and changed to the current semantics. I'll change it to match your expectations.

libkbfs/disk_block_cache.go
strib: None of the rest of the code does it though. For me, inconsistency is way worse for readability.  I find it confusing because I think there's supposed to be real code above the blank line, and I get distracted looking for it.

libkbfs/disk_block_cache.go
strib: This blank line is above the locking though, after the signature.

libkbfs/interfaces.go
strib: Yeah I think that would be a useful thing to know, while I can't really think of a use for the current semantics.  If it's going to take a while though, it might not be worth blocking this PR on.  But I think we should have it in place before turning it on for non-admins.

jzila:
@strib: ready for a second look.

strib:
I added the test failure to KBFS-2183, looks unrelated to the PR to me at quick glance.

libkbfs/block_retrieval_queue.go
```diff
@@ -213,6 +215,8 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
 func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 	ptr BlockPointer, block Block, kmd KeyMetadata, priority int,
 	lifetime BlockCacheLifetime, hasPrefetched bool) (err error) {
+	didUpdateCh := make(chan struct{})
+	donePrefetch := hasPrefetched
```
strib: Please update these `donePrefetched` variables as well, for consistency.  Although maybe for now we should always be setting it to `false` anyway?

Can we also change `hasPrefetched` to `triggeredPrefetch` like in the interface?

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	syncNumRemoved, syncSizeRemoved, err :=
+		cache.syncCache.Delete(ctx, blockIDs)
+	return numRemoved + syncNumRemoved, sizeRemoved + syncSizeRemoved, err
+}
+
+// UpdateMetadata implements the DiskBlockCache interface for
+// diskBlockCacheWrapped.
+func (cache *diskBlockCacheWrapped) UpdateMetadata(ctx context.Context,
+	blockID kbfsblock.ID, hasPrefetched, donePrefetch bool) error {
```
strib: Please update `donePrefetch` here too.

libkbfs/disk_block_cache_helper.go
```diff
@@ -26,6 +26,8 @@ type diskBlockCacheMetadata struct {
 	BlockSize uint32
 	// whether the block has triggered prefetches
 	HasPrefetched bool
+	// whether the block's triggered prefetches are complete
+	DonePrefetch bool
```
strib: Can we update this name?  I know we can't update `HasPrefetched` since it's serialized, but it would be nice if this matched the name in the interface.

libkbfs/disk_block_cache.go
```diff
@@ -619,60 +640,54 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
...
-	case <-cache.shutdownCh:
-		return DiskCacheClosedError{"UpdateMetadata"}
-	default:
+	return cache.getMetadataLocked(blockID)
+}
+
+// UpdateMetadata implements the DiskBlockCache interface for
+// DiskBlockCacheStandard.
+func (cache *DiskBlockCacheStandard) UpdateMetadata(ctx context.Context,
+	blockID kbfsblock.ID, hasPrefetched, donePrefetch bool) (err error) {
```
strib: Please update these var names to match the interface.

libkbfs/disk_block_cache.go
```diff
@@ -401,12 +393,14 @@ func (*DiskBlockCacheStandard) tlfKey(tlfID tlf.ID, blockKey []byte) []byte {
 // updateMetadataLocked updates the LRU time of a block in the LRU cache to
 // the current time.
 func (cache *DiskBlockCacheStandard) updateMetadataLocked(ctx context.Context,
-	tlfID tlf.ID, blockKey []byte, encodeLen int, hasPrefetched bool) error {
-	metadata := diskBlockCacheMetadata{
+	tlfID tlf.ID, blockKey []byte, encodeLen int, hasPrefetched,
```
strib: Please update the var names here to match the interface.

libkbfs/block_retrieval_queue.go
jzila: I can do s/hasPrefetched/triggeredPrefetch everywhere except in the main struct, since users' caches will already have `HasPrefetched` encoded into their disk cache metadata. WDYT?

strib: Yep, figured, sounds good.

I guess technically you could use the encoding options to specify that the encoded name is `HasPrefetched`, while the struct name could be `TriggeredPrefetch`.  But that's probably overkill.  Up to you.

jzila:
@strib: I did the renaming changes you requested. I've also added an auto-enabler when someone explicitly enables a TLF for syncing, to allow testing. WDYT?
https://github.com/keybase/kbfs/pull/1091/files/4089f433596786ce0753f4947e71d5891391c7e6..47f12c179750c0ed47abe5328e030f672db7cbc1

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -46,40 +47,48 @@ func newDiskBlockCacheWrapped(config diskBlockCacheConfig, storageRoot string) (
...
+	// _ = cache.enableSyncCache()
+	return cache, nil
+}
+
+func (cache *diskBlockCacheWrapped) enableSyncCache() (err error) {
+	if cache.syncCache != nil {
+		return nil
+	}
+	syncCacheRoot := filepath.Join(cache.storageRoot, syncCacheFolderName)
+	cache.syncCache, err = newDiskBlockCacheStandard(cache.config,
```
strib: Seems like this would need locking, to prevent two LRU caches from being created for the same folder.

PR #1098, opened 2017-08-02
revendor resinit

oconnor663:
r? @strib 

This should fix https://keybase.atlassian.net/browse/CORE-5761. The CircleCI build might break because it tries to merge vendored files from client, and client is trying to land its own versions of these changes (https://github.com/keybase/client/pull/7970), but I need to confirm that.


oconnor663:
I'm almost certain that the build failure here is spurious:

```
/tmp/go-android/src/github.com/keybase/vendor/github.com/keybase/go-framed-msgpack-rpc/rpc/resinit/resinit_android.go:13: resInit redeclared in this block
	previous declaration at /tmp/go-android/src/github.com/keybase/vendor/github.com/keybase/go-framed-msgpack-rpc/rpc/resinit/resinit_other.go:8
```

The vendored file `resinit_android.go` doesn't exist in my branch, and this error only happens because we're merging in the master (not yet revendored) branch from client. If I land this and https://github.com/keybase/client/pull/7970 at the same time, we should be fine.

oconnor663:
Local build passes when I run `LOCAL_KBFS=1 yarn run rn-gobuild-android`.

PR #1099, opened 2017-08-04
libkbfs: support subteam TLFs

strib:
* The `chargedTo` of a subteam TLF is the subteam's root team ID
* When using libkb to parse subteam names, they must be prefixed with "team:", otherwise the parser gets confused since usernames can't have dots in them.

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -248,7 +248,7 @@ func MakeLocalUsers(users []libkb.NormalizedUsername) []LocalUser {
 }
 
 // MakeLocalTeams is a helper function to generate a list of local
-// teams suitable to use with KBPKILocal.  Any subteams must came
+// teams suitable to use with KBPKILocal.  Any subteams must come
 // after their root team names in the `teams` slice.
 func MakeLocalTeams(teams []libkb.NormalizedUsername) []TeamInfo {
 	localTeams := make([]TeamInfo, len(teams))
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"os"
+	"strings"
 	"sync"
 	"time"
 
@@ -246,8 +247,9 @@ func MakeLocalUsers(users []libkb.NormalizedUsername) []LocalUser {
 	return localUsers
 }
 
-// MakeLocalTeams is a helper function to generate a list of
-// local teams suitable to use with KBPKILocal.
+// MakeLocalTeams is a helper function to generate a list of local
+// teams suitable to use with KBPKILocal.  Any subteams must came
+// after their root team names in the `teams` slice.
 func MakeLocalTeams(teams []libkb.NormalizedUsername) []TeamInfo {
 	localTeams := make([]TeamInfo, len(teams))
 	for i := 0; i < len(teams); i++ {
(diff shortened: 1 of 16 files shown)


libkbfs/config_local.go
```diff
@@ -261,6 +263,16 @@ func MakeLocalTeams(teams []libkb.NormalizedUsername) []TeamInfo {
 			},
 			LatestKeyGen: FirstValidKeyGen,
 		}
+		// If this is a subteam, set the root ID.
```
strib: Correct.  Only the top-level root team has a quota, no matter how deep the subteam tree is.

libkbfs/data_types.go
```diff
@@ -58,6 +58,7 @@ type TeamInfo struct {
 	TID          keybase1.TeamID
 	CryptKeys    map[KeyGen]kbfscrypto.TLFCryptKey
 	LatestKeyGen KeyGen
+	RootID       keybase1.TeamID // for subteams only
```
songgao: Are we using `RootID` being non-empty as an indication for a team is a sub-team?

libkbfs/kbpki_client.go
```diff
@@ -271,6 +271,18 @@ func (k *KBPKIClient) IsTeamReader(
...
+// GetTeamRootID implements the KBPKI interface for KBPKIClient.
+func (k *KBPKIClient) GetTeamRootID(ctx context.Context, tid keybase1.TeamID) (
+	keybase1.TeamID, error) {
+	teamInfo, err := k.serviceOwner.KeybaseService().LoadTeamPlusKeys(
+		ctx, tid, UnspecifiedKeyGen, keybase1.UserVersion{},
+		keybase1.TeamRole_NONE)
+	if err != nil {
+		return keybase1.TeamID(""), err
+	}
+	return teamInfo.RootID, nil
```
songgao: `LoadTeamPlusKeys` doesn't populate `RootID` for sub-teams, so this would be empty for root teams. It seems contradictory with the interface comment (https://github.com/keybase/kbfs/pull/1099/files#diff-9bdcfee86ede222f52d83f6b7f9f7f28R547) which suggests the method can be used for both root teams and sub-teams?

libkbfs/config_local.go
```diff
@@ -245,8 +246,9 @@ func MakeLocalUsers(users []libkb.NormalizedUsername) []LocalUser {
 	return localUsers
 }
 
-// MakeLocalTeams is a helper function to generate a list of
-// local teams suitable to use with KBPKILocal.
+// MakeLocalTeams is a helper function to generate a list of local
+// teams suitable to use with KBPKILocal.  Any subteams must came
+// after their root team names in the `teams` slice.
 func MakeLocalTeams(teams []libkb.NormalizedUsername) []TeamInfo {
```
strib: It's probably best in `test_common.go`, but I have it here because `MakeLocalUsers` is right above it.  We can reorganize things in a later PR if we want.

libkbfs/data_types.go
strib: I don't think I use it that way anywhere, but it should reliably work that way if we want.  Though I think it's more clear to do `TID.IsSubTeam()`.

libkbfs/kbpki_client.go
strib: Why do you say it doesn't set it?  It's set right here: https://github.com/keybase/kbfs/pull/1099/files#diff-f385c6b3885c40be40101a1ac02efdcaR577

strib: Oh, do you mean it doesn't populate it for _root_ teams.  True, I can fix that.

songgao: Oh right; meant to say root teams. Sorry for the typo!

libkbfs/data_types.go
songgao: Now that `RootID` is populated for root teams, I guess this comment should be changed accordingly?

PR #1101, opened 2017-08-04
debugging: add entry size logs for directory listing

strib:
Issue: KBFS-2359


strib:
> Do we need it in libdokan too or is this for a fuse specific issue?

We're only seen it on macOS so far, so I think `libfuse` is good enough for now.

PR #1103, opened 2017-08-07
libkbfs: use field in `ImmutableRootMetadata` to track whether an MD has been put

strib:
Previously, we were relying on the last writer's device key, and whether or not journaling was enabled, to decide whether an MD had been fully merged to the server.  But once we have multiple processes writing to the same TLF, using the verifying key will no longer be an effective way to decide whether the MD has been merged to the server or not.

Instead, keep track of this status in a local field in each `ImmutableRootMetadata`, which can be set to `false` for MDs written to or read from the journal, and switched to `true` once the MD is successfully put to the server.

Issue: KBFS-2341

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -982,6 +982,10 @@ func (fbm *folderBlockManager) isQRNecessary(
 		fbm.log.CWarningf(ctx, "Couldn't get the current session: %+v", err)
 		return false
 	}
+	// It's ok to treat both MDs written by this process on this
+	// device, and MDs written by other processes (e.g., kbgit) in the
+	// same way.  Other processes are likely to be short-lived, and
+	// probably won't do their own QR, so a conflict is unlikely here.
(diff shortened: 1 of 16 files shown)


libkbfs/folder_block_manager.go
```diff
@@ -982,6 +982,10 @@ func (fbm *folderBlockManager) isQRNecessary(
 		fbm.log.CWarningf(ctx, "Couldn't get the current session: %+v", err)
 		return false
 	}
+	// It's ok to treat both MDs written by this process on this
+	// device, and MDs written by other processes (e.g., kbgit) in the
+	// same way.  Other processes are likely to be short-lived, and
+	// probably won't do their own QR, so a conflict is unlikely here.
```
songgao: If the unlikely happen, does it just mean a more expensive operation, but not affect correctness?

strib: That's right.  It will cause conflict resolution to kick in, when normally QR would try to avoid that within the same device.  But because we will have multiple processes acting on behalf of the same device, it's much harder to avoid that now.  But yes, correctness won't be affected.

PR #1104, opened 2017-08-07
libkbfs: add a new "single op" init mode

strib:
This mode is to be used by standalone KBFS programs, like the upcoming git remote helper, that don't need to keep a TLF update to date or rekey anything.

"single op" mode differs from default mode by:
* Not subscribing to folder updates.
* Not checking for rekeys.
* Not suppressing prefixes that are normally disallowed, on the assumption that the names will be controlled by the program implementing the single operation.

Issue: KBFS-2342

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -806,6 +806,11 @@ const (
 	// InitMinimal is for when KBFS will only be used as a MD lookup
 	// layer (e.g., for chat on mobile).
 	InitMinimal
+	// InitSingleOp is a mode for when KBFS is only needed for a
+	// single logical operation; no rekeys or update subscriptions is
+	// needed, and some naming restrictions are lifted (e.g., `.kbfs_`
+	// filenames are allowed).
+	InitSingleOp
 )
 
 func (im InitMode) String() string {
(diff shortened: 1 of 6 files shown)


strib:
> Do you want to also make changes to the goroutine counts for the block retriever?

Since the current plan is to proxy at the disk cache layer, I figured we would want to keep the same number of fetchers by default at the block ops layer.  But I could be convinced otherwise if you have ideas on what number would be better.  Of course, we can always change it later if it doesn't seem right.

jzila:
@strib but we're _only_ proxying the disk cache, right?

strib:
> @strib but we're only proxying the disk cache, right?

Right, exactly.  So the BlockOps layer will still be sending requests to an in-git-process BlockServerRemote when data is not in the disk block cache.  So that means it will be useful to have lots of goroutines ready to perform those fetches if they're ever needed.

On the other hand, hopefully prefetching will mean that they're very rarely needed.  However, the goroutines also handle the disk block cache lookups, so I think it will still be useful to have lots of them ready to do.

jzila:
Sounds good.

PR #1106, opened 2017-08-09
kbfsgit: barebones git remote helper process

strib:
Builds an executable called `git-remote-keybase` that git will use for "keybase://" repos if it's in the user's $PATH.  Currently doesn't do much besides respond to the "capabilities" command with an empty set of capabilities.

Each invocation has its own unique journal directory, since they shouldn't interact with each other at all locally.

This also threads a context through the initialization code, because all instances of this executable will share a single log, and it will be nice to have separate debug tags for each instance during the initialization log messages.

Issue: KBFS-2337


strib:
Circle won't pass until keybase/client#8058 is in, needs to be merged at the same time as this PR.

kbfsgit/git-remote-keybase/main.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func start() *libfs.Error {
+	kbCtx := env.NewContext()
+
+	// TODO: Also remove all kbfsgit directories older than 24h.
+	storageRoot, err := ioutil.TempDir(kbCtx.GetDataDir(), "kbfsgit")
+	if err != nil {
+		return libfs.InitError(err.Error())
+	}
```
jzila: `defer os.RemoveAll(storageRoot)`?

libkbfs/init.go
```diff
@@ -425,8 +444,10 @@ func InitLog(params InitParams, ctx Context) (logger.Logger, error) {
 // The keybaseServiceCn argument is to specify a custom service and
 // crypto (for non-RPC environments) like mobile. If this is nil, we'll
 // use the default RPC implementation.
-func Init(ctx Context, params InitParams, keybaseServiceCn KeybaseServiceCn,
-	onInterruptFn func(), log logger.Logger) (cfg Config, err error) {
+func InitWithPrefix(
```
jzila: `InitWithLogPrefix` would make more sense to me.

kbfsgit/git-remote-keybase/main.go
strib: I currently have that `defer` in `start.go`, but yeah it probably makes more sense here.  I'll move it.

Jenkinsfile
```diff
@@ -327,6 +327,13 @@ def runNixTest(prefix) {
             sh './simplefs.test -test.timeout 2m'
         }
     }
+    tests[prefix+'kbfsgit'] = {
+        dir('kbfsgit') {
+            sh 'go test -i'
+            sh 'go test -race -c'
+            sh './kbfsgit.test -test.timeout 10m'
```
strib: This command doesn't build it, but all binaries are built with a `go install` command above.

songgao: Oh I see; thanks!

PR #1108, opened 2017-08-10
libfs: implement fs.OpenFile() and File for billy interface

strib:
This is part 1 of the billy filesystem interface for KBFS.

I'll implement the rest of the methods in subsequent PRs.

Let me know if you think it would be better in a separate subdirectory, since it doesn't have much to do with git.

Issue: KBFS-2338

### DIFF
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -78,7 +78,7 @@ func (f *File) ReadAt(p []byte, off int64) (n int, err error) {
 	if err != nil {
 		return 0, err
 	}
-	if int(readBytes) < n {
+	if int(readBytes) < len(p) {
 		// ReadAt is more strict than Read.
 		return 0, errors.Errorf("Could only read %d bytes", readBytes)
 	}
@@ -110,7 +110,7 @@ func (f *File) Seek(offset int64, whence int) (n int64, err error) {
 		return 0, errors.Errorf("Cannot seek to offset %d", newOffset)
 	}
 
-	_ = atomic.SwapInt64(&f.offset, offset)
+	_ = atomic.SwapInt64(&f.offset, newOffset)
 	return newOffset, nil
 }
 
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD
 // license that can be found in the LICENSE file.
 
-package kbfsgit
+package libfs
 
 import (
 	"io"
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD
 // license that can be found in the LICENSE file.
 
-package kbfsgit
+package libfs
 
 import (
 	"context"
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -136,6 +136,8 @@ func (fs *FS) OpenFile(filename string, flag int, perm os.FileMode) (
 
 	parts := strings.Split(filename, "/")
 	n := fs.root
+	// Iterate through each of the parent directories of the file, but
+	// not the file itself.
 	for i := 0; i < len(parts)-1; i++ {
 		p := parts[i]
 		var ei libkbfs.EntryInfo
@@ -164,7 +166,9 @@ func (fs *FS) OpenFile(filename string, flag int, perm os.FileMode) (
 
 	offset := int64(0)
 	if flag&os.O_APPEND != 0 {
-		// TODO: worry about overflow.
+		if ei.Size >= uint64(1<<63) {
+			return nil, errors.New("offset too large")
+		}
 		offset = int64(ei.Size)
 	}
 
@@ -204,7 +208,7 @@ func (fs *FS) Remove(filename string) error {
 
 // Join implements the billy.Filesystem interface for FS.
 func (fs *FS) Join(elem ...string) string {
-	return path.Join(elem...)
+	return path.Clean(path.Join(elem...))
 }
 
 // TempFile implements the billy.Filesystem interface for FS.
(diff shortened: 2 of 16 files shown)


strib:
> One Q: would it be worth forking the library to allow context.Context to be passed into each method rather than the current way of passing it into the struct?

In my opinion, no.  I think it would be nice to have a standard billy FS interface that other projects can use if they want to.  Also, the thought of maintaining go-git if we've changed all the function signatures to pass around ctx gives me nightmares.

strib:
> Would it be possible to move File, FileInfo, FileSystem to libkbfs? I guess we could list out individual interface in form of var _ io.Reader = (*File)(nil) if we don't want to import billy into libkbfs.

I'd rather not put them in `libkbfs/`, because it doesn't have anything to do with KBFS internals.  But I'd be happy to move it into it's own package, say `github.com/keybase/kbfs/billy`, if you'd like that better.  I was thinking about doing that anyway.

songgao:
Ah I just edited a minute ago to change `libkbfs` to `libfs`. Would `libfs` work? If not, separate package sounds good too!

strib:
`libfs` is fine with me too.  I'll move it over.

libfs/fs.go
```diff
@@ -0,0 +1,273 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if ei.Type != libkbfs.Dir {
+			return nil, errors.Errorf("%s is not a directory",
+				path.Join(parts[:i]...))
+		}
+	}
+
+	fName := parts[len(parts)-1]
```
jzila: Are we guaranteed that this is nonempty?

libfs/fs.go
```diff
@@ -0,0 +1,273 @@
...
+	if flag&os.O_TRUNC != 0 {
+		err := fs.config.KBFSOps().Truncate(fs.ctx, n, 0)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	offset := int64(0)
+	if flag&os.O_APPEND != 0 {
+		// TODO: worry about overflow.
```
jzila: TODO: wow! That's a big offset!

Kidding aside, this seems worth it to do right away rather than a TODO:
```go
if ei.Size >= uint64(1<<63) {
  return nil, errors.New("offset too large")
}
```

strib:
Thanks for the review.  I didn't end up having time to respond to them yesterday, sorry.  One thing about `ToSlash` though -- @songgao is right that the paths passed in should always be the result of `FS.Join()`, presumably constructed as a result of taking an existing path and trying to add a child to it, so we don't have to worry about slash difference on different platforms.  The Windows comment applies only to "UNC" paths like "\\printer", which we'll never be given.  So the whole interface is assuming that the paths to it will already be in the expected format, so I'm not going to use `ToSlash` anywhere.

I should probably use `path.Clean` in the `Join` method though, just in case.

libfs/file.go
```diff
@@ -0,0 +1,121 @@
...
+	_ = atomic.SwapInt64(&f.offset, origOffset+advanceBytes)
+}
+
+// Write implements the billy.File interface for File.
+func (f *File) Write(p []byte) (n int, err error) {
+	if f.readOnly {
+		return 0, errors.New("Trying to write a read-only file")
+	}
+
+	origOffset := atomic.LoadInt64(&f.offset)
```
strib: I didn't want to hold the mutex over the blocking Read/Write operations, and I was finding it annoying to read the value under the lock early in the function, because I couldn't use `defer` to release the lock.  This way seemed like less code, and was still clear.  But if you don't like it, I could change it to mutex.

libfs/file.go
```diff
@@ -0,0 +1,121 @@
...
+// ReadAt implements the billy.File interface for File.
+func (f *File) ReadAt(p []byte, off int64) (n int, err error) {
+	f.fs.log.CDebugf(f.fs.ctx, "Read %d bytes at offset %d", len(p), off)
+
+	// ReadAt doesn't affect the underlying offset.
+	readBytes, err := f.fs.config.KBFSOps().Read(f.fs.ctx, f.node, p, off)
+	if err != nil {
+		return 0, err
+	}
+	if int(readBytes) < n {
```
strib: Oops, yes, that's a bug.  Fixed.

libfs/file.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return 0, err
+		}
+		newOffset = int64(ei.Size) + offset
+	}
+	if newOffset < 0 {
+		return 0, errors.Errorf("Cannot seek to offset %d", newOffset)
+	}
+
+	_ = atomic.SwapInt64(&f.offset, offset)
```
strib: Oops, yes, that's a bug.

libfs/fs.go
```diff
@@ -0,0 +1,273 @@
...
+		if flag&os.O_CREATE == 0 {
+			return nil, libkbfs.EntryInfo{}, err
+		}
+		fs.log.CDebugf(
+			fs.ctx, "Creating %s since it doesn't exist yet", filename)
+		excl := libkbfs.NoExcl
+		if flag&os.O_EXCL != 0 {
+			excl = libkbfs.WithExcl
+		}
+		isExec := (perm & 0100) != 0
```
strib: KBFS only supports executable permissions for the owner.  This is the same way libfuse behaves.

libfs/fs.go
strib: We are if the caller is using `FS.Join()` for all pathnames, like it is supposed to.

libfs/file.go
jzila: It's fine, I was mostly confused why we wouldn't want to hold the mute over the whole thing.

PR #1109, opened 2017-08-10
libfs: support FileInfo methods in the billy FS layer

strib:
Depends on #1108.

Issue: KBFS-2338

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -247,13 +247,13 @@ func (fs *FS) TempFile(dir, prefix string) (billy.File, error) {
 }
 
 // ReadDir implements the billy.Filesystem interface for FS.
-func (fs *FS) ReadDir(path string) (fis []os.FileInfo, err error) {
-	fs.log.CDebugf(fs.ctx, "ReadDir %s", path)
+func (fs *FS) ReadDir(p string) (fis []os.FileInfo, err error) {
+	fs.log.CDebugf(fs.ctx, "ReadDir %s", p)
 	defer func() {
 		fs.deferLog.CDebugf(fs.ctx, "ReadDir done: %+v", err)
 	}()
 
-	n, base, err := fs.lookupParent(path)
+	n, base, err := fs.lookupParent(p)
 	if err != nil {
 		return nil, err
 	}
@@ -273,7 +273,7 @@ func (fs *FS) ReadDir(path string) (fis []os.FileInfo, err error) {
 		fis = append(fis, &FileInfo{
 			fs:       fs,
 			ei:       ei,
-			fullpath: path + "/" + name,
+			fullpath: path.Join(p, name),
 		})
 	}
 	return fis, nil
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -220,9 +220,9 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 	}
 
 	return &FileInfo{
-		fs:       fs,
-		ei:       ei,
-		fullpath: filename,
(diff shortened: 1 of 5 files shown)


kbfsgit/fs.go
```diff
@@ -213,8 +243,36 @@ func (fs *FS) TempFile(dir, prefix string) (billy.File, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	fis = make([]os.FileInfo, 0, len(children))
+	for name, ei := range children {
+		fis = append(fis, &FileInfo{
+			fs:       fs,
+			ei:       ei,
+			fullpath: path + "/" + name,
```
songgao: Maybe use `path.Join` here?

kbfsgit/file_info.go
```diff
@@ -7,41 +7,63 @@ package kbfsgit
...
+	ei       libkbfs.EntryInfo
+	fullpath string
 }
 
 var _ os.FileInfo = (*FileInfo)(nil)
 
 // Name implements the os.FileInfo interface for FileInfo.
 func (fi *FileInfo) Name() string {
-	return ""
+	return fi.fullpath
```
songgao: I think this is supposed to be base name so `path.Base(fi.fullpath)`?

kbfsgit/fs.go
```diff
@@ -188,8 +199,27 @@ func (fs *FS) Open(filename string) (billy.File, error) {
...
+
+	_, ei, err := fs.config.KBFSOps().Lookup(fs.ctx, n, base)
+	if err != nil {
+		return nil, err
+	}
+
+	return &FileInfo{
+		fs:       fs,
+		ei:       ei,
+		fullpath: filename,
```
songgao: Did you intend to have `fullpath` always include full path or does it not matter? It seems it's only used by logging though.

kbfsgit/file_info.go
strib: The billy comments are pretty clear that it should be the path that's passed to `Open`, which in billy-land is the full path: https://github.com/src-d/go-billy/blob/d0dc16de10d86398cc9aba4513a2453b401eceb3/fs.go#L131

kbfsgit/fs.go
strib: As mentioned in the other comment, this needs to be the full path passed to `Open`.  It's used in the `FileInfo.Name()` implementation.

kbfsgit/file_info.go
songgao: Oh I see; thanks!

songgao: Actually, that's the comment for `File.Name`, not `FileInfo.Name`. I think `os.FileInfo` interface still require this to be the base name?

strib: Oh, d'oh, you're right.  I'll change it.

libfs/fs.go
```diff
@@ -125,31 +127,40 @@ func (fs *FS) lookupOrCreateFile(
...
-	f billy.File, err error) {
-	fs.log.CDebugf(
-		fs.ctx, "OpenFile %s, flag=%d, perm=%o", filename, flag, perm)
-	defer func() {
-		fs.deferLog.CDebugf(fs.ctx, "OpenFile done: %+v", err)
-	}()
-
+func (fs *FS) lookupParent(filename string) (
+	parent libkbfs.Node, base string, err error) {
 	parts := strings.Split(filename, "/")
```
jzila: Not sure if I've already suggested this for this instance, but `filepath.ToSlash(filename)` before splitting.

PR #1110, opened 2017-08-10
libfs: add dir-related operations for the billy FS layer

strib:
Issue: KBFS-2338

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -227,13 +227,47 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 }
 
 // Rename implements the billy.Filesystem interface for FS.
-func (fs *FS) Rename(oldpath, newpath string) error {
-	return nil
+func (fs *FS) Rename(oldpath, newpath string) (err error) {
+	fs.log.CDebugf(fs.ctx, "Rename %s -> %s", oldpath, newpath)
+	defer func() {
+		fs.deferLog.CDebugf(fs.ctx, "Rename done: %+v", err)
+	}()
+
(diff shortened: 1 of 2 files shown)


libfs/fs.go
```diff
@@ -276,7 +311,42 @@ func (fs *FS) ReadDir(p string) (fis []os.FileInfo, err error) {
...
 
 // MkdirAll implements the billy.Filesystem interface for FS.
-func (fs *FS) MkdirAll(filename string, perm os.FileMode) error {
+func (fs *FS) MkdirAll(filename string, perm os.FileMode) (err error) {
+	fs.log.CDebugf(fs.ctx, "MkdirAll %s", filename)
+	defer func() {
+		fs.deferLog.CDebugf(fs.ctx, "MkdirAll done: %+v", err)
+	}()
+
+	parts := strings.Split(filename, "/")
```
jzila: `filepath.ToSlash(filename)` first

songgao: hmm I was under the impression that it's gonna be `"/"` no matter the platform, like https://golang.org/pkg/net/http/#FileSystem . But I can't seem to find relevant doc for `billy`

jzila: ¯\_(ツ)_/¯ I just wanted to make sure it was brought up.

songgao: Yup it's a good point!

songgao: Oh there's a `Filesystem.Join` [method](https://godoc.org/github.com/src-d/go-billy#Basic) so I guess that's up to us. But the comment on `Join` seems to suggest we should treat Windows and Unix differently.

PR #1111, opened 2017-08-11
[KBFS-2364] Track whether prefetches are done for a block

jzila:
This PR passes through whether prefetches for a block are done.

It's not pretty, but to make it nice it'll need a much larger refactor where we combine the block retriever and prefetcher. The components are too tightly coupled right now and there's too much back-and-forth.

Missing:
* Persisting the sync state for a TLF (incoming PR soon)
* Correctly updating the `finishedPrefetch` state of a block when it wouldn't trigger prefetches itself (as in unsynced TLFs). This is not as high priority because blocks in the unsynced disk cache can be deleted arbitrarily.

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -216,7 +216,9 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 	ptr BlockPointer, block Block, kmd KeyMetadata, priority int,
 	lifetime BlockCacheLifetime, triggeredPrefetch bool) (err error) {
 	didUpdateCh := make(chan struct{})
-	finishedPrefetch := triggeredPrefetch
+	// TODO: Currently this is wrong; we need to pipe the real value through
+	// the caches.
+	finishedPrefetch := false
 	defer func() {
 		if err != nil {
 			brq.log.CWarningf(ctx, "Error Putting into the block cache: %+v",
@@ -238,6 +240,9 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 			close(didUpdateCh)
 		}
 	}()
+	// TODO: If the working set cache has triggered a prefetch in the past, and
+	// we've switched to a sync cache, we still need to re-trigger prefetches
+	// so that it goes down the tree.
 	if triggeredPrefetch {
 		return brq.config.BlockCache().PutWithPrefetch(ptr, kmd.TlfID(), block,
 			lifetime, triggeredPrefetch)
@@ -268,20 +273,39 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 	// This must be called in a goroutine to prevent deadlock in case this
 	// CacheAndPrefetch call was triggered by the prefetcher itself.
 	go func() {
-		<-brq.Prefetcher().PrefetchAfterBlockRetrieved(block,
-			ptr, kmd)
+		prefetcher := brq.Prefetcher()
+		doneCh, errCh, numBlocks :=
+			prefetcher.PrefetchAfterBlockRetrieved(block, ptr, kmd)
+
+		// If we have child blocks to prefetch, wait for them.
+		if numBlocks > 0 {
+			for i := 0; i < numBlocks; i++ {
+				select {
+				case <-doneCh:
+				case <-errCh:
+					// One error means this block didn't finish prefetching, so
+					// we don't want to overwrite whatever the current
+					// finishedPrefetch state is.
+					return
+				case <-prefetcher.ShutdownCh():
+					return
+				}
+			}
+		}
+
 		// Make sure we've already updated the metadata once to avoid a race.
 		<-didUpdateCh
 		// Prefetches are done. Update the disk cache metadata.
-		// TODO: implement deep prefetch done tracking.
-		//dbc := brq.config.DiskBlockCache()
-		//if dbc != nil {
-		//	err := dbc.UpdateMetadata(ctx, ptr.ID, true, true)
-		//	if err != nil {
-		//		brq.log.CWarningf(ctx, "Error updating metadata after "+
-		//			"prefetch: %+v", err)
-		//	}
-		//}
+		dbc := brq.config.DiskBlockCache()
+		if dbc != nil {
+			// TODO: set finishedPrefetch to true here once it correctly
+			// represents the doneness of the prefetching.
+			err := dbc.UpdateMetadata(ctx, ptr.ID, true, false)
+			if err != nil {
+				brq.log.CWarningf(ctx, "Error updating metadata after "+
+					"prefetch: %+v", err)
+			}
+		}
 	}()
 	return nil
 }
@@ -291,10 +315,10 @@ func (brq *blockRetrievalQueue) checkCaches(ctx context.Context,
 	lifetime BlockCacheLifetime) error {
 	// Attempt to retrieve the block from the cache. This might be a specific
 	// type where the request blocks are CommonBlocks, but that direction can
-	// Set correctly. The cache will never have CommonBlocks.  TODO: verify
-	// that the returned lifetime here matches `lifetime` (which should always
-	// be TransientEntry, since a PermanentEntry would have been served
-	// directly from the cache elsewhere)?
+	// Set correctly. The cache will never have CommonBlocks.
+	// TODO: verify that the returned lifetime here matches `lifetime` (which
+	// should always be TransientEntry, since a PermanentEntry would have been
+	// served directly from the cache elsewhere)?
 	cachedBlock, triggeredPrefetch, _, err :=
 		brq.config.BlockCache().GetWithPrefetch(ptr)
 	if err == nil && cachedBlock != nil {
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -52,8 +52,10 @@ func (c *realBlockRetrievalConfig) blockGetter() blockGetter {
 
 // blockRetrievalRequest represents one consumer's request for a block.
 type blockRetrievalRequest struct {
-	block  Block
-	doneCh chan error
+	block          Block
+	doneCh         chan error
+	prefetchDoneCh chan<- struct{}
+	prefetchErrCh  chan<- struct{}
 }
 
 // blockRetrieval contains the metadata for a given block retrieval. May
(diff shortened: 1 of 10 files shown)


jzila:
@songgao: this is to support:
* Eventually allowing a client to query whether a folder subtree is done syncing
* Correctly tracking whether an on-demand request needs to re-trigger a subtree prefetch.

libkbfs/block_retrieval_queue.go
```diff
@@ -263,44 +279,69 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
...
+
+		// If we have child blocks to prefetch, wait for them.
+		if numBlocks > 0 {
+			for i := 0; i < numBlocks; i++ {
+				select {
+				case <-childPrefetchDoneCh:
+					// Passthrough
+				case <-childPrefetchErrCh:
+					// One error means this block didn't finish prefetching.
+					prefetchErrCh <- struct{}{}
```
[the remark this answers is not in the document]
jzila: Definitely a valid concern. It almost seems like the prefetcher needs a cache to track every block currently being prefetched. Still digesting...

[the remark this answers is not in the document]
jzila: The idea here is that when `prefetchErrCh` returns, the receiver does nothing to the state of `DonePrefetch` on a block. We need such a mechanism because we don't want to waltz all over other prefetches that are happening. But as you identified in another comment, this can cause havoc when multiple prefetches might need to affect the same block. Essentially, these are the scenarios we need to support:
1. First prefetch for block `c` from parent `a`:
    * A counting semaphore (currently a `chan struct{}`) that reaches 0 when all child block trees are fully prefetched. When this happens, the state of the block is updated to `DonePrefetch = true`.
    * A semaphore that is notified as soon as any child block fails to prefetch. Now it doesn't matter if other blocks finish, so we can short circuit this process, leave `DonePrefetch` alone, and notify the parent's error semaphore.
2. Second prefetch for already-prefetched block `c` from `a'`:
    * Notice that this block already has `DonePrefetch` set, so we send to the parent's `done` semaphore. This is currently incorrect in the code, we are only checking for `triggeredPrefetch`.
3. Second prefetch for triggered but not done for block `c`, originally triggered by `a` and now triggered by `a'`:
    1. Prefetch from `a` is ongoing:
        * We probably need to track this, notice it, and add the `a'`'s done semaphore to a slice of parent channels to notify (this is new code). This also needs to be done atomically, such that we can't race with the completion of the existing prefetch.
    2. Prefetch from `a` failed:
        * Once we check `donePrefetch` instead of `triggeredPrefetch`, this should just work.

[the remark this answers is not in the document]
jzila: @strib: I'll work on fixing that scaffolding. Sorry this has been such a mess; it's proven to be quite a challenge to keep the existing prefetch behavior working along the same codepath as the new deep tree syncing.

libkbfs/block_retrieval_queue.go
```diff
@@ -69,13 +69,16 @@ type blockRetrieval struct {
...
 
-	// protects requests and lifetime
+	// protects requests, cacheLifetime, and the prefetch channels
 	reqMtx sync.RWMutex
 	// the individual requests for this block pointer: they must be notified
 	// once the block is returned
 	requests []*blockRetrievalRequest
 	// the cache lifetime for the retrieval
 	cacheLifetime BlockCacheLifetime
+	// channels to notify when this retrieval and its prefetches are done
```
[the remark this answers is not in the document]
jzila: I was trying to keep the channel as a semaphore (`chan struct{}` instead of `chan foo`), to avoid the memory overhead of the buffer for each level of the block tree. I commented below regarding how I think I can fix this behavior.

libkbfs/block_retrieval_queue.go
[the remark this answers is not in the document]
strib: This sounds like a good plan to me.  Thanks for writing it out!  It would probably be helpful as a comment in the code somewhere too.

jzila:
@strib: ready for another round with all the fixes we talked about. Here's the link to the commits: https://github.com/keybase/kbfs/pull/1111/files/c6dda32115e650ffc3bc2fcb74f35d8c7f18d544..e395353e1d2cb0efea689d483747d4ca740cf284

libkbfs/block_retrieval_queue.go
```diff
@@ -249,27 +305,42 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
...
+		return brq.config.BlockCache().PutWithPrefetch(ptr, kmd.TlfID(),
+			block, lifetime, FinishedPrefetch)
+	}
+	if brq.config.IsSyncedTlf(kmd.TlfID()) && dbc != nil {
+		switch prefetchStatus {
+		case NoPrefetch:
+		case TriggeredPrefetch:
+			// We still need to register this requestor for receiving
+			// notifications about this prefetch.
+			// TODO: handle
```
jzila: Actually I can remove this TODO. Since I gave you the commit range link, I'll leave it alone for now.

libkbfs/block_retrieval_queue.go
```diff
@@ -455,7 +559,25 @@ func (brq *blockRetrievalQueue) FinalizeRequest(
 		// Since we created this channel with a buffer size of 1, this won't
 		// block.
 		req.doneCh <- err
+		doneChans = append(doneChans, r.prefetchDoneCh)
+		errChans = append(errChans, r.prefetchErrCh)
 	}
+
+	go func() {
+		// Fan-out the prefetch result to the channel of each individual
+		// request. Since the slice was created before the request mutex was
```
strib: Above there's a comment that says the request mutex is symbolic, but that seems like it's not the case anymore.

libkbfs/block_retrieval_queue.go
```diff
@@ -455,7 +559,25 @@ func (brq *blockRetrievalQueue) FinalizeRequest(
...
+		// released, this won't race.
+		var prefetchChans []chan<- struct{}
+		select {
+		case <-prefetchDoneCh:
+			prefetchChans = doneChans
+		case <-prefetchErrCh:
+			prefetchChans = errChans
+		}
+		for _, ch := range prefetchChans {
+			ch <- struct{}{}
```
strib: `RequestWithPrefetch`, which appends requests to the retrieval object, seems to allow `nil` channels.  Do you need to check for `nil` here?

libkbfs/interfaces.go
```diff
@@ -1205,20 +1218,23 @@ type KeyOps interface {
...
-		priority int) error
+		priority int) (doneCh, errCh <-chan struct{}, err error)
 	// PrefetchAfterBlockRetrieved allows the prefetcher to trigger prefetches
 	// after a block has been retrieved. Whichever component is responsible for
 	// retrieving blocks will call this method once it's done retrieving a
 	// block. Returns a channel that is closed once the all the underlying
 	// prefetches are complete.
 	PrefetchAfterBlockRetrieved(b Block, blockPtr BlockPointer,
-		kmd KeyMetadata) <-chan struct{}
+		kmd KeyMetadata) (doneCh, errCh <-chan struct{}, numBlocks int)
```
strib: The comment here and for `PrefetchBlock` should be updated to describe the new return channels.

libkbfs/block_retrieval_queue.go
```diff
@@ -209,23 +215,84 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
...
+	lifetime BlockCacheLifetime, prefetchDoneCh, prefetchErrCh chan<- struct{},
+	didUpdateCh <-chan struct{}) {
+	prefetcher := brq.Prefetcher()
+	childPrefetchDoneCh, childPrefetchErrCh, numBlocks :=
+		prefetcher.PrefetchAfterBlockRetrieved(block, ptr, kmd)
+
+	// If we have child blocks to prefetch, wait for them.
+	if numBlocks > 0 {
+		for i := 0; i < numBlocks; i++ {
+			select {
```
strib: Need a `<-ctx.Done()` case here?

libkbfs/prefetcher_test.go
```diff
@@ -292,22 +310,27 @@ func TestPrefetcherAlreadyCached(t *testing.T) {
...
+		prefetchDoneCh, prefetchErrCh)
 	err = <-ch
 	require.NoError(t, err)
 	require.Equal(t, dirA, block)
 
 	t.Log("Release the prefetch for fileB.")
 	continueChFileB <- nil
-	t.Log("Shutdown the prefetcher and wait until it's done prefetching.")
+	t.Log("Wait for the prefetch to finish. This time, the prefetchErrCh " +
+		" returns immediately, so we need to shutdown the prefetcher to find " +
```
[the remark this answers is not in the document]
strib: Ping.

libkbfs/block_retrieval_queue.go
jzila: Hmm, yeah. I used to have `FinishedPrefetch` in here, but yes the switch is redundant now. Do you think I should leave the `if` here with a comment, or invert the logic and `AND` it with the `else` block?

strib: Whatever you think the clearest comment structure is would be fine with me, though w/ a very slight preference to a single `if` statement if the comment is clear enough.  Now I'm confused about the `else` block though.  Why is it ok to signal an error for prefetches that are still in progress?  If the caller supplied a channel, wouldn't they prefer to learn when the prefetch completes?

jzila: All callers supply channels via `CacheAndPrefetch`. The `else` block is for non-synced TLFs.

strib: Why would someone requesting a non-synced prefetch not want to know when the existing prefetch has finished?  Not sure what the TLF type has to do with it.

jzila: Because the `prefetchDoneCh` is only to signal that the full tree has finished prefetching.

strib: Ok, guess that makes sense for now, though it might need to change for git when full-tree prefetches in an unsynced TLF will be possible.

libkbfs/block_retrieval_queue.go
jzila: `brq.mtx` protects the overall queue. As soon as we've deleted this pointer from `brq.ptrs` (protected by `brq.mtx`), nobody else can interfere with this request. However, `retrieval.requests` is technically unprotected by a mutex after the `brq.mtx` unlock, which I found flakily tripping the race detector (an unprotected read where there were protected writes). IMO this is actually a bug in the race detector, but I addressed it by protecting this particular resource with its own mutex.

strib: I'm just arguing that we should remove the word "symbolic" above if the lock is important in some way (even for the race detector), because otherwise some future person might just decide to remove it.

PR #1113, opened 2017-08-17
[KBFS-2370] Persist Synced TLFs

jzila:


libkbfs/config_local.go
```diff
@@ -1213,6 +1227,48 @@ func (c *ConfigLocal) MakeDiskBlockCacheIfNotExists() error {
...
+	}
+	if c.storageRoot == "" {
+		return errors.New("empty storageRoot specified for non-test run")
+	}
+	ldb, err := c.openConfigLevelDB()
+	if err != nil {
+		return err
+	}
+	defer ldb.Close()
+	tlfPrefix := []byte(syncedTlfsConfigKeyPrefix)
```
strib: Why not just have a dedicated sync config db, rather than this key prefix thing?  Is there an advantage to having fewer levelDBs on disk if we decide to add other options later?

libkbfs/config_local.go
```diff
@@ -1227,9 +1283,32 @@ func (c *ConfigLocal) SetTlfSyncState(tlfID tlf.ID, isSynced bool) error {
...
 		diskCacheWrapped, ok := c.diskBlockCache.(*diskBlockCacheWrapped)
 		if !ok {
-			return errors.Errorf("Invalid disk cache type to set TLF sync state.")
+			return errors.New("invalid disk cache type to set TLF sync state")
+		}
+		if !diskCacheWrapped.IsSyncCacheEnabled() {
+			return errors.New("sync block cache is not enabled")
+		}
+	}
+	if !c.IsTestMode() {
```
strib: Might be nice to avoid indents below with:

```go
if c.IsTestMode() {
  return
}
```
instead.

libkbfs/data_types.go
```diff
@@ -814,12 +819,24 @@ const (
...
+	return im&flags > 0
+}
+
 func (im InitMode) String() string {
 	switch im {
 	case InitDefault:
 		return InitDefaultString
 	case InitMinimal:
 		return InitMinimalString
+	case InitSingleOp:
```
strib: Oops, thanks!

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -42,38 +42,59 @@ type diskBlockCacheWrapped struct {
...
+		// We already have a cache of the desired type. Thus, this method is
+		// idempotent.
+		return nil
 	}
-	cache = &diskBlockCacheWrapped{
-		config:          config,
-		storageRoot:     storageRoot,
-		workingSetCache: workingSetCache,
-		syncCache:       nil,
+	if cache.storageRoot == "" {
```
strib: This is kind of a weird check -- could you use the test mode flag you introduced in this PR?  Or be even more explicit by having a `newDiskBlockCacheWrappedForTest` that passes some special flag into this function.

libkbfs/leveldb.go
```diff
@@ -14,3 +21,34 @@ var leveldbOptions = &opt.Options{
...
 }
+
+type levelDb struct {
+	*leveldb.DB
+	closer io.Closer
+}
+
+func (ldb *levelDb) Close() (err error) {
+	err = ldb.DB.Close()
+	// Hide the closer error
```
strib: Should end in period.

libkbfs/data_types.go
```diff
@@ -814,12 +819,24 @@ const (
...
+	return im & InitModeMask
+}
+
+// HasFlags returns whether all the specified flags are set.
+func (im InitMode) HasFlags(flags InitMode) bool {
+	return im&flags > 0
+}
+
 func (im InitMode) String() string {
 	switch im {
```
songgao: Should this be `im.Mode()` now that it can contain flags?

jzila: Nice catch! Yes it should.

libkbfs/config_local.go
jzila: We still want to set the synced state in memory for test mode.

PR #1115, opened 2017-08-18
libfs: support changing attrs in the billy FS layer

strib:
Issue: KBFS-2338

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -29,6 +29,7 @@ type FS struct {
 	config   libkbfs.Config
 	root     libkbfs.Node
 	h        *libkbfs.TlfHandle
+	subdir   string
 	log      logger.Logger
 	deferLog logger.Logger
 }
@@ -77,6 +78,7 @@ func NewFS(ctx context.Context, config libkbfs.Config,
 		config:   config,
 		root:     n,
 		h:        tlfHandle,
+		subdir:   subdir,
 		log:      log,
 		deferLog: log.CloneWithAddedDepth(1),
 	}, nil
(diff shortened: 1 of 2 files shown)


libfs/fs.go
```diff
@@ -484,33 +486,81 @@ func (fs *FS) Readlink(link string) (target string, err error) {
...
 // Chmod implements the billy.Filesystem interface for FS.
-func (fs *FS) Chmod(name string, mode os.FileMode) error {
-	return nil
+func (fs *FS) Chmod(name string, mode os.FileMode) (err error) {
+	fs.log.CDebugf(fs.ctx, "Chmod %s %s", name, mode)
+	defer func() {
+		fs.deferLog.CDebugf(fs.ctx, "Chmod done: %+v", err)
+	}()
+
+	n, _, err := fs.lookupOrCreateEntry(name, os.O_RDONLY, 0)
```
songgao: It seems `os.O_RDONLY` doesn't make a difference; perhaps just pass in `0` to avoid confusion? Same for below.

strib: I personally thought O_RDONLY was less confusing, since it makes it clear that the file wouldn't be created if it doesn't exist yet.  Keeping it that way for now, but if it becomes confusing we can change it later.

songgao: Sounds good; although isn’t it the lack of O_CREATE that indicates the file won’t be created?

strib: Yep, exactly!  With just a bare `0`, there's no indication that `O_CREATE` is even an option.  With an explicit `O_RDONLY` (which is defined as 0, by the way), it seems more clear to me that creation is being explicitly ruled out.  Just a personal preference I guess.

songgao: Oh I see! I didn’t realize O_RDONLY was defined as 0. Makes perfect sense now.

PR #1116, opened 2017-08-18
kbfsgit: a working push/pull test with a KBFS repo

strib:
Also it seems go-git does depend on temp files, so I had to add support for that.  It relies on the user to pass in a unique ID to guarantee exclusivity.  I'm thinking GRH will use the device ID + the local temp directory name.

Issue: KBFS-2338


songgao:
Exciting PR title!

strib:
I'm still trying to get this to work on windows, so removing reviewers for now.  Will put you back on once it's ready.

strib:
Open for business again!

libfs/fs.go
```diff
@@ -366,15 +387,22 @@ func (fs *FS) Join(elem ...string) string {
 
 // TempFile implements the billy.Filesystem interface for FS.
 func (fs *FS) TempFile(dir, prefix string) (billy.File, error) {
-	// We'd have to turn off journaling to support TempFile fully.
-	return nil, errors.New("TempFile is not currently supported")
+	// We'd have to turn off journaling to support TempFile perfectly,
+	// but the given uniq ID and a random number should be good
+	// enough.  Especially since most users will end up renaming the
+	// temp file before journal flushing even happens.
+	r := rand.Int63()
```
songgao: hmm why not have more than 8 bytes, from `crypto/rand`?

kbfsgit/fs_test.go
```diff
@@ -0,0 +1,162 @@
...
+	h, err := libkbfs.ParseTlfHandle(ctx, config.KBPKI(), "user1", tlf.Private)
+	require.NoError(t, err)
+	fs, err := libfs.NewFS(ctx, config, h, subdir, "")
+	require.NoError(t, err)
+	return ctx, h, fs
+}
+
+// configInMemoryStorer keeps the git config in memory, to work around
+// a gcfg bug (used by go-git when reading configs from disk) that
+// causes a freakout when it sees backslashes in git file URLs.
```
songgao: What's the freakout here? It still needs to read it in the first time right? Also, does the `*gogitcfg.Config` hold content of the entire config?

libfs/fs.go
```diff
@@ -30,6 +31,7 @@ type FS struct {
 	root     libkbfs.Node
 	h        *libkbfs.TlfHandle
 	subdir   string
+	uniqID   string
```
strib: No, this is a unique ID for only this instance of the remote-helper.  There could be multiple GRH processes on the same device, and they all need to have different uniqIDs.  It's used for making temporary files in KBFS that need to have globally unique names.

LockIDs need to be agreed on by all devices and all processes, otherwise they wouldn't be very effective locks!

libfs/fs.go
strib: 8 bytes seems like plenty for the purposes of making unique filenames within a KBFS folder, but I'm open to suggestions.  I guess longer file names will take slightly longer to process internally, but it probably wouldn't affect performance much.  What do you suggest for an appropriate number of bytes?

libfs/fs.go
songgao: Oh that's true! I was confused by myself.

kbfsgit/fs_test.go
strib: When reading the config from disk, gcfg fails if it sees any backslashes in any of the remote names.  When there's no config on disk, it makes a new one, so it can handle when the config is missing.

But, in a future PR that I'll put up soon, I realized we do need to write the config to disk, if only to get the "bare = true" line.  So that PR changes this class into something that just filters out the remotes before writing it to disk.  We want to do that anyway, since we don't want to start any remotes permanently in the bare repo.

libfs/fs.go
songgao: I didn't know the answer, so I looked up how Go's `ioutil.TempFile` does it, and it turns out it's just an integer too: https://golang.org/src/io/ioutil/tempfile.go?s=#L26

So I guess it's fine. Although do we call `rand.Seed()` somewhere? Otherwise it's always the same sequence and we'd end up with the same file name. Also I'm curious why `rand` instead of `crypto/rand`? Is it performance concern too?

kbfsgit/fs_test.go
songgao: I see; Thanks for the explanation! I still don't quiet understand the need for a remote in the bare repo though (also asked [here](https://github.com/keybase/kbfs/pull/1120#discussion_r134866980))

libfs/fs.go
strib: Oops, good point about seeding.  I can change to crypto/rand -- I just find it annoying to have to initialize a byte buffer before calling it.  But I guess that's easy enough.

PR #1119, opened 2017-08-22
limit number of ongoing rekeys

songgao:
cc @maxtaco

### DIFF
--- a/libkbfs/ongoing_work_limiter.go
+++ b/libkbfs/ongoing_work_limiter.go
@@ -1,0 +1,40 @@
+package libkbfs
+
+import "sync"
+
+// OngoingWorkLimiter limits maximum number of routines that can work on a same
+// type of thing at the same time. For example, it can be used for limiting
+// number of ongoing rekeys.
+type OngoingWorkLimiter struct {
+	cond     sync.Cond
+	capacity int64
+}
+
+// NewOngoingWorkLimiter creates a new *OngoingWorkLimiter with capacity of
+// maxNumOngoingWorks.
+func NewOngoingWorkLimiter(maxNumOngoingWorks int) *OngoingWorkLimiter {
+	return &OngoingWorkLimiter{
+		capacity: int64(maxNumOngoingWorks),
+		cond:     sync.Cond{L: new(sync.Mutex)},
+	}
+}
+
+// WaitToStart blocks until the limiter would allow one more routine to start
+// working on the thing.
+func (owl *OngoingWorkLimiter) WaitToStart() {
+	owl.cond.L.Lock()
+	for owl.capacity <= 0 {
+		owl.cond.Wait()
+	}
+	owl.capacity--
+	owl.cond.L.Unlock()
+}
+
+// Done tells the limiter that the caller is done working on the thing, and
+// somebody else is free to start work.
+func (owl *OngoingWorkLimiter) Done() {
+	owl.cond.L.Lock()
+	defer owl.cond.L.Unlock()
+	owl.capacity++
+	owl.cond.Signal()
+}
--- a/libkbfs/ongoing_work_limiter.go
+++ b/libkbfs/ongoing_work_limiter.go
@@ -1,40 +1,35 @@
 package libkbfs
 
-import "sync"
+import "context"
 
 // OngoingWorkLimiter limits maximum number of routines that can work on a same
 // type of thing at the same time. For example, it can be used for limiting
 // number of ongoing rekeys.
 type OngoingWorkLimiter struct {
-	cond     sync.Cond
-	capacity int64
+	permits chan struct{}
 }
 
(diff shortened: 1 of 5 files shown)


libkbfs/ongoing_work_limiter.go
```diff
@@ -0,0 +1,40 @@
...
+type OngoingWorkLimiter struct {
+	cond     sync.Cond
+	capacity int64
+}
+
+// NewOngoingWorkLimiter creates a new *OngoingWorkLimiter with capacity of
+// maxNumOngoingWorks.
+func NewOngoingWorkLimiter(maxNumOngoingWorks int) *OngoingWorkLimiter {
+	return &OngoingWorkLimiter{
+		capacity: int64(maxNumOngoingWorks),
```
strib: Why not just require an `int64` to be passed in?  (And why an `int64` instead of an `int` anyway?  Do you think a 32-bit machine will fire up more than 2^32 workers?)

libkbfs/ongoing_work_limiter.go
```diff
@@ -0,0 +1,40 @@
...
+		cond:     sync.Cond{L: new(sync.Mutex)},
+	}
+}
+
+// WaitToStart blocks until the limiter would allow one more routine to start
+// working on the thing.
+func (owl *OngoingWorkLimiter) WaitToStart() {
+	owl.cond.L.Lock()
+	for owl.capacity <= 0 {
+		owl.cond.Wait()
```
strib: It would be nice if there was an option to wait for a `ctx.Done()` here as well.  Rekey task contexts can be canceled, right?  (It would have to fork a goroutine for the cond wait though, of course.)

libkbfs/ongoing_work_limiter.go
```diff
@@ -0,0 +1,40 @@
+package libkbfs
+
+import "sync"
+
+// OngoingWorkLimiter limits maximum number of routines that can work on a same
+// type of thing at the same time. For example, it can be used for limiting
+// number of ongoing rekeys.
+type OngoingWorkLimiter struct {
+	cond     sync.Cond
```
strib: I think it might be more Go-like (and would integrate better with contexts) to use a fixed-capacity channel instead.  Then `Wait` could try to send on it, and `Done()` would drain from it.  Then you don't have to track capacity yourself, and get context cancellation without having to fork a goroutine (see comment below).

libkbfs/rekey_fsm.go
```diff
@@ -454,6 +460,19 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
...
+		// In normal desktop app, we limit to 16 routines.
+		fsm.limiter = NewOngoingWorkLimiter(16)
+	case InitMinimal:
+		// This is likely mobile. Limit it to 4.
+		fsm.limiter = NewOngoingWorkLimiter(4)
+	case InitSingleOp:
+		// Just block all rekeys and don't bother cleaning up requests since the process is short lived anyway.
+		fsm.limiter = NewOngoingWorkLimiter(0)
+	default:
+		panic("😱")
```
strib: Cute, but please print out the unknown mode here at least.

libkbfs/rekey_fsm.go
```diff
@@ -454,6 +460,19 @@ func NewRekeyFSM(fbo *folderBranchOps) RekeyFSM {
 
 		listeners: make(map[rekeyEventType][]rekeyFSMListener),
 	}
+	switch fbo.config.Mode() {
+	case InitDefault:
+		// In normal desktop app, we limit to 16 routines.
+		fsm.limiter = NewOngoingWorkLimiter(16)
```
strib: Isn't there a `rekeyFSM` per-`folderBranchOps`?  We need to limit rekeys across all the different folder branch ops, right?  I think a global limiter would need to be passed in to make that happen.

songgao: Oops; Yes and yes. Good catch thanks!

libkbfs/ongoing_work_limiter.go
songgao: hmmm OK!

libkbfs/ongoing_work_limiter.go
songgao: hmmm I don't think rekey can be canceled; we use context for timeouts.

strib: Ah ok, same thing for `ctx.Done()` purposes.

songgao:
Open for review again; thanks!

patrickxb:
Did tests of this on real devices bring CPU usage down?

songgao:
Just tried provisioning a paper key on my laptop, on staging with 500+ folders; it's mostly around 20% now, with occasional bump to 30%-60%. But I didn't see it jumping above 100% anymore (as seen previously). So I think it's effective!

patrickxb:
Ok, cool!

PR #1120, opened 2017-08-22
kbfsgit: initialize bare repo on a git `list` command

strib:
This causes a bare repo to be initialized under `.kbfs_git` whenever the `list` command is passed to the git remote helper.  That means you can run something like this:

```sh
git clone keybase://private/strib/test
```

And there will be a new bare repo in `/keybase/private/strib/.kbfs_git/test`.  This is a bare repo that is compatible with the git binary from the command line.

Things it doesn't do yet:
* Take any locks while initializing the repo, so it wouldn't be great if there were multiple concurrent creates (there would probably be at least a HEAD.conflict file).
* Work well with existing kbfsfuse/kbfsdokan processes.  It's best to kill the existing process, run the git command, and then re-mount to check your work.
* Guarantee that everything is committed within a single MD update, so an error at the wrong time might leave things in a weird state.
* Any kind of server-side prioritization.
* Put git blocks under their own quota.

Issue: KBFS-2346
Issue: KBFS-2350


kbfsgit/runner.go
```diff
@@ -0,0 +1,287 @@
...
+
+	fs, err := libfs.NewFS(
+		ctx, r.config, r.h, path.Join(kbfsRepoDir, r.repo), r.uniqID)
+	if err != nil {
+		return nil, err
+	}
+
+	// We store the config in memory for two reasons. 1) gogit/gcfg
+	// has a bug where it can't handle backslashes in remote URLs, and
+	// 2) we don't want to flush the remotes since they'll contain
```
songgao: > we don't want to flush the remotes

Is this the remote configured in the bare repo? In a typical use, bare repos in KBFS shouldn't have remotes configured right? I thought it worked as remotes for other repos which to or fetch from it.

kbfsgit/storer.go
```diff
@@ -0,0 +1,49 @@
...
+// configInMemoryStorer keeps the git config in memory, to work around
+// a gcfg bug (used by go-git when reading configs from disk) that
+// causes a freakout when it sees backslashes in git file URLs.
+type configInMemoryStorer struct {
+	*filesystem.Storage
+	cfg *gogitcfg.Config
+}
+
+func newConfigInMemoryStorer(fs *libfs.FS) (*configInMemoryStorer, error) {
+	fsStorer, err := filesystem.NewStorage(fs)
```
strib: I don't think so -- whatever we make needs to implement the full `Storer` interface, because that's what gogit expects in the `Init` call.  Unless you see a way around that?

kbfsgit/runner.go
strib: In go-git, as far as I know, there is no way to hook two `go-git.Repository` instances together to do fetches/pulls between them.  One of the two repos has to be defined as a "remote" to the other one in order to use the nice `Fetch` and `Pull` commands.  (There might be other more involved ways to transfer objects manually one-by-one, but that seems like it would be pretty sad.)

Since there is no standard remote protocol for keybase yet (that's what we're building!), it's not supported by go-git itself.  That means our only option is to treat the local on-disk repo as a "remote" with respect to the bare KBFS repo, and do everything in reverse: for example, when a user does a push, we actually fetch from the local repo and write the objects into the bare repo.

It's confusing, but I haven't seen a way around this yet.  Let me know if you think of one.

FYI, the next PR in this series will be changing the storer to filter out the remotes, rather than store them in memory, since we still need some of the config file to be written to KBFS.  But these temporary remotes should never leave the local process, they're only for local convenience to work around go-git limitations.

kbfsgit/storer.go
songgao: `Storage.Init` simply calls `Initialize()` on `dir`. I guess we can just directly call that instead of going through the `Storage` type?

strib: We don't pass it to `Storage.Init`, we pass it to `go-git.Init()`, which requires a full `Storer` interface.  It gets saved in the returned `Repository` instance and used all over the place.

kbfsgit/runner.go
songgao: Understood. That's actually pretty creative! Nice.

Would we ever end up in a recursive situation where a GRH calls GRH itself to use fetch for push or the other way around? I can't think of a way in which that could happen, but thought I'd bring it up.

kbfsgit/storer.go
songgao: Oh I see this is the storer that's used for everything. I thought it was just config. NVM then!

kbfsgit/runner.go
strib: Good question. I think it's impossible, because the GRH will always be using a local directory for the "remote", and so it will always operate directly on the file system without invoking other remote helpers.  (Also, I don't think go-git even has a notion of invoking remote-helpers, it implements https/ssh/etc in the library directly, as far as I know.)

songgao: Sounds good; thanks!

> https

I wonder how they implement that? It can't use the reverse approach and needs to actually implement pushing/fetching objects in that case right? Can we reuse some of the code there?

strib: It simply implements the git pack HTTP protocol, which is well-defined and pretty simple: https://github.com/src-d/go-git/tree/master/plumbing/transport/http.

That's basically the same thing the "file" transport layer does, which is what we use.

songgao: Oh I see; so if there were an API for pushing to a `gogit.Repository` instead of a remote, it would have been the same mechanism as the reversed approach. So despite using a different API, we are essentially doing the same thing in exactly the same way. Does it sound correct?

strib: Yeah, that sounds right.

PR #1122, opened 2017-08-23
kbfsgit: support list and simple fetches

strib:
This may not cover all corner cases, but it works for simple cases. It sets the local repo as a "remote" in the bare repo, and pushes from the bare repo into this "remote".  go-git requires pushing to a ref name, but `git` gets upset if you push into the refname it's expecting, so we have to use a temporary name and then delete it afterwards.

Issue: KBFS-2353

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -19,6 +19,7 @@ import (
 	"github.com/keybase/kbfs/tlf"
 	"github.com/pkg/errors"
 	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
 const (
@@ -119,6 +120,12 @@ func newRunner(ctx context.Context, config libkbfs.Config, repo string,
 		output: output}, nil
 }
 
+// handleCapabilities: from https://git-scm.com/docs/git-remote-helpers
+//
+// Lists the capabilities of the helper, one per line, ending with a
+// blank line. Each capability may be preceded with *, which marks
+// them mandatory for git versions using the remote helper to
+// understand. Any unknown mandatory capability is a fatal error.
 func (r *runner) handleCapabilities() error {
 	caps := []string{
 		gitCmdFetch,
@@ -177,7 +184,7 @@ func (r *runner) initRepoIfNeeded(ctx context.Context) (
 	// has a bug where it can't handle backslashes in remote URLs, and
 	// 2) we don't want to flush the remotes since they'll contain
 	// local paths.
-	storer, err := newConfigInMemoryStorer(fs)
+	storer, err := newConfigWithoutRemotesStorer(fs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


kbfsgit/runner.go
```diff
@@ -37,6 +40,26 @@ const (
...
+	// one-by-one, but that seems like it would be pretty sad.)
+	//
+	// Since there is no standard remote protocol for keybase yet
+	// (that's what we're building!), it's not supported by go-git
+	// itself. That means our only option is to treat the local
+	// on-disk repo as a "remote" with respect to the bare KBFS repo,
+	// and do everything in reverse: for example, when a user does a
+	// push, we actually fetch from the local repo and write the
+	// objects into the bare repo.
+	localRepoRemoteName = "local"
```
songgao: I just realized a thing: in a normal git setup, we'd have have `origin/master` ref stored in local repo (which allows you do things like `git reset --hard origin/master`). In this reverse setup, would go-git store refs like `local/master` in the bare repo, which points to the local repo on disk in the bare repo? If so, do we need to clean them up after each push/fetch?

kbfsgit/runner.go
```diff
@@ -232,56 +267,184 @@ func (r *runner) waitForJournal(ctx context.Context) error {
...
 		return errors.New("Lists for non-fetches unsupported for now")
 	}
 
-	_, err := r.initRepoIfNeeded(ctx)
+	repo, err := r.initRepoIfNeeded(ctx)
 	if err != nil {
 		return err
 	}
 
+	refs, err := repo.References()
```
songgao: Does this include remote references or only local ones?

kbfsgit/runner.go
```diff
@@ -232,56 +267,184 @@ func (r *runner) waitForJournal(ctx context.Context) error {
...
+	remote, err := repo.CreateRemote(&gogitcfg.RemoteConfig{
+		Name: localRepoRemoteName,
+		URL:  r.gitDir,
+	})
+
+	for _, fetch := range args {
+		if len(fetch) != 2 {
+			return errors.Errorf("Bad fetch request: %v", fetch)
+		}
+		refInBareRepo := fetch[1]
```
songgao: This is the `<sha1>` in the doc above right?

kbfsgit/runner.go
```diff
@@ -232,56 +267,184 @@ func (r *runner) waitForJournal(ctx context.Context) error {
...
+			return err
+		}
+
+		// Delete the temporary refspec now that the objects are
+		// safely stored in the local repo.
+		refSpec = fmt.Sprintf(":refs/remotes/%s/%s", r.remote, localTempRef)
+		err = remote.Push(&gogit.PushOptions{
+			RemoteName: localRepoRemoteName,
+			RefSpecs:   []gogitcfg.RefSpec{gogitcfg.RefSpec(refSpec)},
+		})
```
songgao: Assuming `fetch[2]` is the local ref name for the fetch (or "remote" ref name in reverse speaking), I was under the impression that we'd need to  create a ref under that name in the on-disk repo. Then I saw the comment above:

> fetching directly via the runner doesn't leave any refs, those would normally be created by the `git` process that invokes the runner

Is this what the comment is talking about? Does it mean `fetch[2]` is useless then?

kbfsgit/start.go
```diff
@@ -60,7 +61,9 @@ func Start(ctx context.Context, options StartOptions,
 	}
 	defer config.Shutdown(ctx)
 
-	r, err := newRunner(ctx, config, options.Repo, input, output)
+	r, err := newRunner(
+		ctx, config, options.Remote, options.Repo, options.GitDir,
```
songgao: Do you think it'd be clearer if we name stuff as "OnDisk" and "BareRepo" rather than "Remote"?

kbfsgit/runner.go
strib: No, from my tests it doesn't seem like go-git does that when you simply call `Fetch`.  I think you have to explicitly add a reference to get it to save it to the repo.  (Or maybe that's a bare repo thing?)

In any case, in this PR we're not even calling `Fetch` on the remote, we're calling `Push`, so it doesn't modify the bare repo at all (except for adding the remote).

kbfsgit/runner.go
strib: The bare repo doesn't have any remote references.

kbfsgit/runner.go
strib: No, it's the `<name>`.

kbfsgit/runner.go
strib: As above, `fetch[1]` is the name of the reference the user wants to fetch from the bare repo.  `fetch[0]` is the sha1 hash, which I'm not using here.

The comment is trying to say that calling `Push` on the remote (the local disk in this case) doesn't actually create any ref in the remote, it only populates the objects.  The `git` command line process later creates the ref to be whatever the user specified it should be (which is not passed to the GRH).  In the test, where that comment was taken from, we're not using the command line, so no ref is created, which means we have to check out using the sha1 directly.

kbfsgit/start.go
strib: I'm not sure, I was using the names from the doc to keep things consistent: https://git-scm.com/docs/git-remote-helpers.  For now I'll leave it as is, and add comments explaining what these names mean to us.

PR #1123, opened 2017-08-24
root_metadata: fix v2->v3 upconversion for conflict TLFs

strib:
Oops.

The test fails without this fix on the final line.

Issue: KBFS-2381
Issue: keybase/client#8188


libkbfs/bare_root_metadata_v2.go
```diff
@@ -443,7 +447,7 @@ func (md *BareRootMetadataV2) makeSuccessorCopyV3(
 
 	if md.ConflictInfo != nil {
 		ci := *md.ConflictInfo
-		md.ConflictInfo = &ci
+		mdV3.ConflictInfo = &ci
```
strib: This is the bug.

PR #1124, opened 2017-08-24
govendor update github.com/keybase/client/go/protocol/keybase1/...

mlsteele:


vendor/github.com/keybase/client/go/protocol/keybase1/extras.go
```diff
@@ -1664,6 +1664,7 @@ func (u UserPlusKeysV2) ToUserVersion() UserVersion {
 	}
 }
 
+// Can return nil.
```
songgao: Perhaps name it `GetLatestPerUserKeyIfPossible()` to suggest this in name directly?

mlsteele: That's a good suggestion. I hope you don't mind if I ignore it just to shuttle this PR along. I don't think these end up on the kbfs side of the wire

songgao: OK no problem!

PR #1125, opened 2017-08-24
govendor update github.com/keybase/client/go/protocol/keybase1/...

mlsteele:
Vendor from https://github.com/keybase/client/pull/8211


mlsteele:
client side merged

PR #1126, opened 2017-08-24
kbfsgit: support pushes and branch deletes

strib:
Issue: KBFS-2354


kbfsgit/runner.go
```diff
@@ -276,8 +276,10 @@ func (r *runner) waitForJournal(ctx context.Context) error {
 // unrecognized attributes are ignored. The list ends with a blank
 // line.
 func (r *runner) handleList(ctx context.Context, args []string) (err error) {
-	if len(args) > 0 {
-		return errors.New("Lists for non-fetches unsupported for now")
+	if len(args) == 1 && args[0] == "for-push" {
+		r.log.CDebugf(ctx, "Treating for-push the same as a regular list")
```
songgao: Git doc seems to suggest `for-push` is a optimization. Do we have a plan/ticket for treating them differently?

strib: No; I don't actually know what such an optimization would look like for us.  I guess maybe it could let you prepare a packfile slightly ahead of time, before the batch completes?  Maybe there's something else I'm not getting.  If we dream up a practical optimization, I'll add a ticket.

PR #1130, opened 2017-08-28
Jenkins: prebuild gogit dependencies

strib:
There seems to be a go build parallelism issue when both the `git-remote-helper` and `kbfsgit` directories are built in parallel.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -262,7 +262,7 @@ def runNixTest(prefix) {
     }
 
     // Build out the vendored gogit dependency first, otherwise the
-    // fit-remote-helper binary and the kbfsgit tests might have
+    // git-remote-helper binary and the kbfsgit tests might have
     // concurrent build issues when running in parallel.
     dir('kbfsgit') {
         sh 'go test -i'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -260,6 +260,13 @@ def runNixTest(prefix) {
     dir('test') {
         sh 'go test -i -tags fuse'
     }
+
+    // Build out the vendored gogit dependency first, otherwise the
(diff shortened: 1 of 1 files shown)


strib:
cc: @mmaxim

Jenkinsfile
```diff
@@ -260,6 +260,13 @@ def runNixTest(prefix) {
...
+
+    // Build out the vendored gogit dependency first, otherwise the
+    fit-remote-helper binary and the kbfsgit tests might have
+    concurrent build issues when running in parallel.
+    dir('kbfsgit') {
+        sh 'go test -i'
+    }
     tests = [:]
     tests[prefix+'install'] = {
         sh 'go install github.com/keybase/kbfs/...'
```
songgao: Shall we move this out of parallel `tests` too in case it happens for other stuff? Also it seems this would cover the `kbfsgit` too for free.

strib: Will defer to @jzila on that one.

Jenkinsfile
```diff
@@ -260,6 +260,13 @@ def runNixTest(prefix) {
     dir('test') {
         sh 'go test -i -tags fuse'
     }
+
+    // Build out the vendored gogit dependency first, otherwise the
+    // fit-remote-helper binary and the kbfsgit tests might have
```
jzila: s/fit/git

Jenkinsfile
jzila: AFAIK Go is supposed to handle parallel builds properly. But we can make a ticket to move the dependency building (`go test -i`/`go test -c -o test.test`) out of the parallel nodes. That's how client has it: https://github.com/keybase/client/blob/21ec2ed7dc4b449a94b431ed33b9a0f5a33fab7a/Jenkinsfile#L388-L403

PR #1132, opened 2017-08-28
kbfsgit: simple status messages during operation

strib:
Git remote helpers can print status messages on stderr that will be
displayed to the user.  This should have been easy, but
`client/go/logger` duplicates and closes the stderr file descriptor
that the program starts with.  So, before that happens,
`git-remote-keybase` needs to duplicate the original stderr in a
platform-specific way, before it gets redirected into a log file.

A future PR will print more useful status messages (e.g., "M/N objects
uploaded", etc).  But these simple messages are still useful.  Example
output on Linux:

```
$ KEYBASE_RUN_MODE=staging git clone keybase://private/strib/test12
Cloning into 'test12'...
Initializing KBFS... done.
Syncing to Keybase... done.
Fetching data from Keybase... done.
Checking connectivity... done.
```

Issue: KBFS-2385

### DIFF
--- a/kbfsgit/git-remote-keybase/dup_stderr_dummy.go
+++ b/kbfsgit/git-remote-keybase/dup_stderr_dummy.go
@@ -1,0 +1,15 @@
+// Copyright 2017 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+// +build android
+
+package main
+
+import (
+	"errors"
+	"os"
+)
+
+func dupStderr() (*os.File, error) {
+	return nil, errors.New("Duplicating stderr is not supported on this operating system")
+}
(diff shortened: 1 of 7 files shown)


songgao:
Also this looks great: https://github.com/gosuri/uiprogress

strib:
> LGTM! Is there a reason we are doing calling dupStderr in start() rather than in kbfsgit.Start() right before init'ing the logger?

I wanted to keep the knowledge of how kbfsgit is invoked out of the kbfsgit library.  There's no need for kbfsgit to know that it should be writing status messages specifically to os.Stderr, in my opinion.  Maybe someday we'd want to start this from a GUI, for example, and we'd want to pass in a custom io.Writer to Start.

songgao:
Makes sense. Thanks for explaining!

PR #1133, opened 2017-08-29
Fix Dokan mount failure propagation

taruti:


taruti:
Yes, it will keep running.

PR #1136, opened 2017-08-30
fix panic during conflict resolution

songgao:
Steve's KBFS was in a crash loop. He said he revoked a device hours ago. I think this is the correct fix for the crash as shown in the log, but I'm not sure if it'd crash somewhere afterwards.

Log send: https://keybase.io/_/admin/logdump_viewer?39510239#acad1c98d3388ec687e99b1c

### DIFF
--- a/tlf/id.go
+++ b/tlf/id.go
@@ -147,9 +147,6 @@ func (id *ID) UnmarshalText(buf []byte) error {
 // SafeType returns the type of TLF represented by this ID.  If the ID
 // isn't valid, it returns tlf.Unknown along with an error.
 func (id ID) SafeType() (Type, error) {
-	if len(id.id) != idByteLen {
-		return Unknown, fmt.Errorf("Invalid ID length: %d", len(id.id))
-	}
 	switch id.id[idByteLen-1] {
 	case idSuffix:
 		return Private, nil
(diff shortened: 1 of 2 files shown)


tlf/id.go
```diff
@@ -147,9 +147,6 @@ func (id *ID) UnmarshalText(buf []byte) error {
 // SafeType returns the type of TLF represented by this ID.  If the ID
 // isn't valid, it returns tlf.Unknown along with an error.
 func (id ID) SafeType() (Type, error) {
-	if len(id.id) != idByteLen {
```
songgao: Removed this since it's not necessary as `id.id` is an array rather than slice.

strib:
Thanks for looking into this!  I prefer not to use `SafeType` on the client side, because it would hide exactly these kind of problems.


The problem seems to be that's there's a `createOp` in the set of merged ops coming from the server, and the `op.finalPath` never gets set in that case.

In this case, I think a better fix is to do this:

```go
if p.FolderBranch.Tlf == tlf.NullID {
    return false
}
```

in this function: https://github.com/keybase/kbfs/blob/76a8bafb605d7227df72f2d55d7afdec7b4c0b3b/libkbfs/path.go#L98-L107

The best fix is probably to make sure that `op.finalPath` is always set, even for merged ops.  Or at least that `op.finalPath.FolderBranch` is set.  But that would be a bit more involved and probably not worth it.

songgao:
Sounds good; thanks!

PR #1140, opened 2017-08-31
add SuggestedRetryIn to throttle error

songgao:


songgao:
I'm thinking about adding a `ServerErrorSmartThrottle` since `ServerErrorThrottle` is retried in RPC layer. An alternative is to keep using `ServerErrorThrottle` and branch out in `ShouldRetry` but that looks ugly. If we introduce new error type would old client freak out?

strib:
> I'm thinking about adding a ServerErrorSmartThrottle since ServerErrorThrottle is retried in RPC layer. An alternative is to keep using ServerErrorThrottle and branch out in ShouldRetry but that looks ugly. If we introduce new error type would old client freak out?

As I mentioned in chat, I think the right move is to make `ShouldRetry` return an optional, suggested retry duration that the RPC layer can use.  Then you can use the same error or a different one, it wouldn't matter.

However, if this is a pain, I personally don't think it's a big priority to have the server direct the retry time.  It's probably ok if the client retries on the normal schedule, even though it's a bit wasteful.

songgao:
Just RPC retried throttle error I think. I haven't extended the rpc library to take the suggested wait time though.

jzila:
OK

PR #1141, opened 2017-08-31
kbfsblock: fix `UsageStat.AccumOne` for git blocks

strib:
Issue: KBFS-2366

### DIFF
--- a/kbfsblock/quota.go
+++ b/kbfsblock/quota.go
@@ -59,7 +59,7 @@ func (u *UsageStat) AccumOne(change int, usage UsageType) {
 	if change == 0 {
 		return
 	}
-	if usage < UsageWrite || usage > UsageRead {
+	if usage == UsageMDWrite || usage >= NumUsage {
 		return
 	}
 	u.Bytes[usage] += int64(change)


jzila:
@songgao: we haven't, and we will continue not to.

strib:
> Hmmm does it mean we haven't been counting MD quota usage?

Hmm, it does kinda seem that way, doesn't it?  But we do have a non-zero stat for it, although the stat hasn't changed in months!  https://www.stathat.com/v/stats/VOmk/tf/1y17h1w

@jzila: we should be tracking the MD usage in the DB though.  Hrm.

songgao:
@jzila WDYM? I thought we wanted to count them, since there's MD quota usage in DB. These are for MD blocks from MDv3 I think.

jzila:
Oh I thought we agreed that for the time being we won't count MD quota.

strib:
It's true that there's unlimited quota, but we are still supposed to track it for stat purposes, so we can tell if someone starts abusing it.

jzila:
Ah right, maybe we should fix it then.

songgao:
But I don't understand how that bug was introduced over 7 months yet we still have stats change until May.

strib:
Yeah that's what I'm trying to figure out too.  Will look more tonight or tomorrow.

jzila:
@songgao: old clients?

songgao:
This package is used by bserver not clients I think

strib:
I finally figured out why this happened.  kbfs-server commit 39c9012f2 directly edited the vendored copy of `quota.go` to allow usage stats for stats bigger than `UsageRead`, but it was never committed to `kbfs/kbfsblock`.  Then, on the deploy that happened on May 25th, we re-vendored `kbfs/kbfsblock` into bserver, erasing this change.  Sigh.

The tests didn't catch it, because the tests also use this function to determine what the expected stats should be.  Double sigh.

I'll update this PR to fix it both for MD and git blocks.  But it means we've lost track of all MD blocks written since May 25th.  We'd have to scan S3 to recover that info.

strib:
> I'll update this PR to fix it both for MD and git blocks

Oh nevermind, duh, this PR is already correct.  Merging.

songgao:
Ah that's a sneaky bug. Good job finding it!

> But it means we've lost track of all MD blocks written since May 25th.

Just to be sure, this'd be just that MD quota values are off, and we can let it go if we don't care right?

strib:
> Just to be sure, this'd be just that MD quota values are off, and we can let it go if we don't care right?

Correct, we can let it go.  We just have to remember that if someday, we ever want to track down what/who is taking up all our S3 storage, the numbers won't add up until we re-scan S3.

songgao:
I see. I'll try to keep that in mind as redundancy to yours.

jzila:
I'll be the parity bit.

PR #1142, opened 2017-09-01
libkbfs: allow a default block type and use everywhere

strib:
The journal also needs to calculate its limits based on the default block type.

kbfsgit uses the GIT block type as its default.  Everything else uses DATA.

Note that this shouldn't get merged until keybase/kbfs-server#383 is merged and deployed.  But it can be reviewed now.

Issue: KBFS-2366

### DIFF
--- a/kbfsgit/start.go
+++ b/kbfsgit/start.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"io"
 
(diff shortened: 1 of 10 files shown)


songgao:
Ugh silly GitHub AJAX made me think my previous approve didn't go through.

PR #1144, opened 2017-09-03
plumb lockID/lockContext/priority

songgao:


libkbfs/journal_md_ops.go
```diff
@@ -222,8 +222,8 @@ func (j journalMDOps) getRangeFromJournal(
 	return irmds, nil
 }
 
-func (j journalMDOps) GetForHandle(
-	ctx context.Context, handle *TlfHandle, mStatus MergeStatus) (
+func (j journalMDOps) GetForHandle(ctx context.Context, handle *TlfHandle,
+	mStatus MergeStatus, lockBeforeLock *keybase1.LockID) (
```
strib: `lockBeforeLock` -> `lockBeforeGet` (here and everywhere below).

libkbfs/md_ops_test.go
```diff
@@ -210,7 +211,8 @@ func putMDForPrivate(config *ConfigMock, rmd *RootMetadata) {
 		EncryptedPrivateMetadata{}, nil)
 	config.mockBsplit.EXPECT().ShouldEmbedBlockChanges(gomock.Any()).
 		Return(true)
-	config.mockMdserv.EXPECT().Put(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+	config.mockMdserv.EXPECT().Put(gomock.Any(), gomock.Any(), gomock.Any(),
+		gomock.Any(), gomock.Any()).Return(nil)
```
strib: Since you're passing `nil` everywhere, shouldn't the mock server be expecting a `nil`?  That said, it would be nice to make sure that the right non-`nil` thing gets plumbed through to the server. but I'm ok with doing that as part of a higher-level test if you want.

libkbfs/mdserver_remote.go
```diff
@@ -530,6 +530,10 @@ func (md *MDServerRemote) GetForHandle(ctx context.Context,
 		BranchID:     NullBranchID.String(),
 		Unmerged:     mStatus == Unmerged,
 	}
+	if lockBeforeGet != nil {
+		copied := *lockBeforeGet
```
strib: Why is the copy needed?  Isn't this just a `long`?  Once you make the RPC, the long gets serialized, and there's nothing hanging around that can get modified asynchronously.

libkbfs/interfaces.go
```diff
@@ -1319,29 +1335,56 @@ type MDServer interface {
...
+	// releases the lock on the lock ID if the put is successful. Releasing the
+	// lock in mdserver is idempotent.
+	Put(ctx context.Context, rmds *RootMetadataSigned, extra ExtraMetadata,
+		lockContext *keybase1.LockContext, priority keybase1.MDPriority) error
+
+	// LockOps performs a locking operation with a mdserver. If isTake is true, it
+	// ensures lockID for tlfID is taken by this device, i.e., idempotently
+	// take the lock. If isTake is false, it ensures lockID for tlfID is not taken
+	// by this device, i.e., idmpotently release the lock.
+	LockOp(ctx context.Context,
```
strib: I have a slight preference for explicit `Lock` and `Unlock` operations, so we're not relying on a bool.  But I don't feel that strongly about it.

libkbfs/tlf_journal.go
```diff
@@ -1349,11 +1350,11 @@ func (j *tlfJournal) flushOneMDOp(
 
 	j.log.CDebugf(ctx, "Flushing MD for TLF=%s with id=%s, rev=%s, bid=%s",
 		rmds.MD.TlfID(), mdID, rmds.MD.RevisionNumber(), rmds.MD.BID())
-	pushErr := mdServer.Put(ctx, rmds, extra)
+	pushErr := mdServer.Put(ctx, rmds, extra, nil, keybase1.MDPriorityNormal)
```
strib: We might need to do something more here, since Git writes data through the journal and will need to mark it with the git priority and, in some cases, with lock info.  The priority could be a config option that's always applied, but the lock info might need to be serialized to the journal somehow...  Anyway, we can deal with that in a separate PR, but maybe add a TODO here?

libkbfs/interfaces.go
```diff
@@ -1319,29 +1335,56 @@ type MDServer interface {
...
+	// the time of the put, and optionally (if specified in lockContext)
+	// releases the lock on the lock ID if the put is successful. Releasing the
+	// lock in mdserver is idempotent.
+	Put(ctx context.Context, rmds *RootMetadataSigned, extra ExtraMetadata,
+		lockContext *keybase1.LockContext, priority keybase1.MDPriority) error
+
+	// LockOps performs a locking operation with a mdserver. If isTake is true, it
+	// ensures lockID for tlfID is taken by this device, i.e., idempotently
+	// take the lock. If isTake is false, it ensures lockID for tlfID is not taken
+	// by this device, i.e., idmpotently release the lock.
```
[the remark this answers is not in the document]
songgao: It [returns a throttled error](https://github.com/keybase/kbfs-server/pull/382/files#diff-ed92fae2090a5de96c87ab32d1f7b6dcR2439) if the lock is already taken, causing client to retry. The caller should observe a blocking call due to RPC layer retry as I understand. I'll add some comment here to explain that.

libkbfs/interfaces.go
songgao: OK; will do. I was just being lazy, especially thinking that we might not even use it given the locking parameters in Get/Put.

libkbfs/journal_md_ops.go
songgao: OOPS thanks!

libkbfs/md_ops_test.go
songgao: I was worried if we were gonna change some stuff in tests that break these mock tests. But I guess we can set them all to be `nil` and fix the broken ones when we start using locks.

libkbfs/mdserver_remote.go
songgao: Oh good point. I'll remove the copy. Thanks!

libkbfs/tlf_journal.go
songgao: Sounds good; will add a TODO. I thought about how these would work a little bit, but decided that would be best to stay in a separate PR, and put in `nil` everywhere.

PR #1145, opened 2017-09-05
libgit: generate repo IDs and check repo names

strib:
This PR refactors the kbfsgit code a bit to move repo directory creation into its own library, because that part will likely be used by the kbfsfuse/kbfsdokan binaries upon a request from the GUI to create a repo.

The service also requires a stable random ID per repo, and KBFS will be the source of truth for that ID.  So we write that ID to a new config file in the repo, so we can always retrieve it.

This also adds some restrictions on repo names, and normalizes names by lower-casing them.  It preserves the original casing of the name in the kbfs repo config file, in case we want to display it that way in the GUI.

Issue: KBFS-2389

### DIFF
--- a/libgit/config.go
+++ b/libgit/config.go
@@ -8,7 +8,8 @@ import "encoding/json"
 
 // Config is a KBFS git repo config file.
 type Config struct {
-	ID ID
+	ID   ID
+	Name string // the original user-supplied format of the name
 }
 
 func configFromBytes(buf []byte) (*Config, error) {
--- a/libgit/config.go
+++ b/libgit/config.go
@@ -1,0 +1,25 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import "encoding/json"
+
+// Config is a KBFS git repo config file.
+type Config struct {
+	ID ID
+}
+
+func configFromBytes(buf []byte) (*Config, error) {
+	var c Config
+	err := json.Unmarshal(buf, &c)
+	if err != nil {
+		return nil, err
+	}
+	return &c, nil
+}
+
+func (c *Config) toBytes() ([]byte, error) {
+	return json.Marshal(c)
+}
--- a/libgit/id.go
+++ b/libgit/id.go
@@ -1,0 +1,113 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import (
+	"encoding"
+	"encoding/hex"
+	"fmt"
+
+	"github.com/keybase/kbfs/kbfscrypto"
+	"github.com/pkg/errors"
+)
+
+const (
+	// idByteLen is the number of bytes in a git repo ID.
+	idByteLen = 16
+	// idStringLen is the number of characters in the string
+	// representation of a git repo ID.
+	idStringLen = 2 * idByteLen
+
+	idSuffix = 0x2c
+)
+
+// InvalidIDError indicates that a repo ID string is not parseable or
+// invalid.
+type InvalidIDError struct {
+	id string
+}
+
+func (e InvalidIDError) Error() string {
+	return fmt.Sprintf("Invalid repo ID %q", e.id)
+}
+
+// ID encapsulates a repo ID.
+type ID struct {
+	id [idByteLen]byte
+}
+
+var _ encoding.BinaryMarshaler = ID{}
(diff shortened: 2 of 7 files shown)


libgit/id.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+const (
+	// idByteLen is the number of bytes in a git repo ID.
+	idByteLen = 16
+	// idStringLen is the number of characters in the string
+	// representation of a git repo ID.
+	idStringLen = 2 * idByteLen
+
+	idSuffix = 0x2c
```
songgao: Just to be sure, did you remember to update the list in `keybase/keybase`?

libgit/config.go
```diff
@@ -0,0 +1,26 @@
...
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import "encoding/json"
+
+// Config is a KBFS git repo config file.
+type Config struct {
+	ID   ID
+	Name string // the original user-supplied format of the name
```
songgao: This means we'd be able to implement rename without breaking anything right?

libgit/repo.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// This character set is what Github supports in repo names.  It's
+// probably to avoid any problems when cloning onto filesystems that
+// have different Unicode decompression schemes
+// (https://en.wikipedia.org/wiki/Unicode_equivalence).  There's no
+// internal reason to be so restrictive, but it probably makes sense
+// to start off more restrictive and then relax things later as we
+// test.
+var repoNameRE = regexp.MustCompile(`^([a-zA-Z0-9][a-zA-Z0-9_.-]+)$`)
```
songgao: Did you intend to have `+` instead of `*`? It seems this would disallow single character names. Also if you decide to change this, would you mind changing `.` into `\.` as well while you are there? I was confused a bit and thought that was a bug :)

libgit/repo.go
```diff
@@ -0,0 +1,153 @@
...
+
+func checkValidRepoName(repoName string, config libkbfs.Config) bool {
+	return len(repoName) >= 1 &&
+		uint32(len(repoName)) <= config.MaxNameBytes() &&
+		(os.Getenv("KBFS_GIT_REPONAME_SKIP_CHECK") != "" ||
```
songgao: Thanks for adding that!

libgit/repo.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	defer f.Close()
+
+	buf, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, NullID, err
+	}
+	c, err := configFromBytes(buf)
+	if err != nil {
+		return nil, NullID, err
```
songgao: If we have a bad write for the config file, this means we won't be able to fix it without asking user to cd into the `.git` dir. But it seems simply overwriting it is a bit risky too. What do you think if we rename the old file and create a new one in this case?

libgit/id.go
strib: Max already did it, yeah.

libgit/config.go
strib: This is actually so that we can have someone make a repo called `NaMe`, and call it "name" under .kbfs_git (and prevent any other casings of "name" on future repos), but still be able to display the string "NaMe" in the GUI.

We would replace this field if a rename ever happened, to preserve the format of the new name.

libgit/repo.go
strib: If we had an error while writing, then we'll never flush the journal (since we're in single-op mode).  So I don't see how a bad config file could happen.  I guess a bug in some older version of KBFS could have led to it.  But it's not as simple as recreating a new file, because we have to have the same stable ID and Name that we had before, and we have no other way of getting that info back automatically.  In the unlikely event of a problem, it seems to me like a manual fix would be best.

libgit/config.go
songgao: I see. So repo name as we see from KBFS is always lower-cased, correct?

> We would replace this field if a rename ever happened, to preserve the format of the new name.

By replace this field, did you mean a new field in this struct, or changing what's stored in `Name`?

strib: > So repo name as we see from KBFS is always lower-cased, correct?

Yes.

> By replace this field, did you mean a new field in this struct, or changing what's stored in Name?

Change what's stored in `Name`.

libgit/repo.go
songgao: Sounds good; thanks for the explanation! Good point on journal flush only happening with good writes. I was thinking about corrupted writes when config is created, but I guess there's no good way to know if it's first write either.

libgit/config.go
songgao: Understood; thanks!

PR #1146, opened 2017-09-06
Fix a few prefetcher bugs

jzila:


libkbfs/disk_block_cache.go
```diff
@@ -633,12 +633,15 @@ func (cache *DiskBlockCacheStandard) Put(ctx context.Context, tlfID tlf.ID,
...
 	}
-	// Initially set TriggeredPrefetch and FinishedPrefetch to false; rely on
-	// the later-called UpdateMetadata to fix it.
-	return cache.updateMetadataLocked(ctx, blockKey, DiskBlockCacheMetadata{
-		TlfID:     tlfID,
-		BlockSize: uint32(encodedLen),
-	})
+	md, err := cache.getMetadataLocked(blockID)
+	if err != nil {
+		// Only set the relevant fields if we had trouble getting the metadata.
```
strib: Could you give me an explanation of the scenario this is fixing?  Is it a block being put twice was erasing the metadata set by the first put?  When would this happen?  (No need to expand the comment, just a reply here or in the PR description would probably work.)

jzila: This is a speculative fix that I noticed when I was combing through things that might reset the `TriggeredPrefetch`/`FinishedPrefetch` statuses.

PR #1147, opened 2017-09-06
[KBFS-2402] Fix working set cache status

jzila:
This is needed to make the sync cache work properly for people with content in their working set disk cache.


jzila:
@strib: IMO it's actually the correct behavior. If someone disables sync on a folder, and blocks are transferred over to the working set cache, they become eligible for eviction. Thus, the `FinishedPrefetch` flag for a block coming out of that cache should always be false.

PR #1149, opened 2017-09-06
[KBFS-2399] Report Sync Status

jzila:


libkbfs/disk_block_cache.go
```diff
@@ -490,8 +490,9 @@ func (cache *DiskBlockCacheStandard) Get(ctx context.Context, tlfID tlf.ID,
 	}
 
 	defer func() {
-		cache.log.CDebugf(ctx, "Cache Get id=%s tlf=%s bSize=%d err=%+v",
-			blockID, tlfID, len(buf), err)
+		cache.log.CDebugf(ctx, "Cache Get id=%s tlf=%s bSize=%d, "+
+			"prefetchStatus=%s err=%+v", blockID, tlfID, len(buf),
+			prefetchStatus.String(), err)
```
strib: Shouldn't need the `.String()`, but it doesn't hurt of course.

libkbfs/folder_branch_ops.go
```diff
@@ -1962,6 +1961,18 @@ func (fbo *folderBranchOps) GetNodeMetadata(ctx context.Context, node Node) (
...
 	}
+	_, prefetchStatus, _, err :=
+		fbo.config.BlockCache().GetWithPrefetch(res.BlockInfo.BlockPointer)
+	if err != nil {
+		dbc := fbo.config.DiskBlockCache()
+		if dbc != nil {
+			_, _, prefetchStatus, _ = fbo.config.DiskBlockCache().Get(ctx,
+				fbo.id(), res.BlockInfo.ID)
+		}
+		// Swallow cache errors since we don't want to communicate
```
strib: Maybe it would be good to explicitly set `prefetchStatus = NoPrefetch` here if `err != nil`, so that we don't have to rely on the return parameter in an error case.

(Same thing below.)

jzila: Done. I also factored it out because I found myself hitting compile errors due to copypasta.

PR #1150, opened 2017-09-06
kbfsgit: add memory profiles for each stage when verbosity >= 3

strib:
Also re-vendor in go-git to include some minor master performance improvements, and a new status stage.

Issue: KBFS-2404

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -453,19 +453,23 @@ func (r *runner) processGogitStatus(
 	lastByteCount := 0
 	for update := range statusChan {
 		if update.Stage != currStage {
-			if r.verbosity >= 3 {
-				f, err := os.Create(filepath.Join(
-					os.TempDir(), fmt.Sprintf("memprof.%d.prof", update.Stage)))
-				if err != nil {
-					r.log.CDebugf(ctx, err.Error())
-				} else {
-					pprof.WriteHeapProfile(f)
-					f.Close()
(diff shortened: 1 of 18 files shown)


kbfsgit/runner.go
```diff
@@ -450,6 +453,17 @@ func (r *runner) processGogitStatus(
 	lastByteCount := 0
 	for update := range statusChan {
 		if update.Stage != currStage {
+			if r.verbosity >= 3 {
+				f, err := os.Create(filepath.Join(
+					os.TempDir(), fmt.Sprintf("memprof.%d.prof", update.Stage)))
```
jzila: Can you output the location of this file as part of the status dump?

strib: By "status dump" you mean the errput data that's being printed as part of the `git` execution to the terminal?  Sure, will do.

jzila: Yep, thanks.

kbfsgit/runner.go
```diff
@@ -450,6 +453,17 @@ func (r *runner) processGogitStatus(
...
+				if err != nil {
+					r.log.CDebugf(ctx, err.Error())
+				} else {
```
patrickxb: might want to try `runtime.GC()` here, otherwise profiles could be skewed by garbage.

strib: Thanks, done.

PR #1152, opened 2017-09-07
block_retrieval_queue: Monitor prefetches in a separate context

jzila:
Do this so the monitoring doesn't get canceled.

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -377,9 +377,7 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 // TODO: Clearly the caller should be responsible for CacheAndPrefetch in both
 // scenarios.
 func (brq *blockRetrievalQueue) checkCaches(ctx context.Context,
-	priority int, kmd KeyMetadata, ptr BlockPointer, block Block,
-	lifetime BlockCacheLifetime,
-	deepPrefetchDoneCh, deepPrefetchCancelCh chan<- struct{}) error {
+	kmd KeyMetadata, ptr BlockPointer, block Block) (PrefetchStatus, error) {
 	// Attempt to retrieve the block from the cache. This might be a specific
 	// type where the request blocks are CommonBlocks, but that direction can
 	// Set correctly. The cache will never have CommonBlocks.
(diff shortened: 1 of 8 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -216,10 +216,12 @@ func (brq *blockRetrievalQueue) notifyWorker(priority int) {
...
 
-func (brq *blockRetrievalQueue) triggerAndMonitorPrefetch(
-	ctx context.Context, ptr BlockPointer, block Block, kmd KeyMetadata,
-	lifetime BlockCacheLifetime, deepPrefetchDoneCh,
-	deepPrefetchCancelCh chan<- struct{}, didUpdateCh <-chan struct{}) {
+func (brq *blockRetrievalQueue) triggerAndMonitorPrefetch(ptr BlockPointer,
+	block Block, kmd KeyMetadata, lifetime BlockCacheLifetime,
+	deepPrefetchDoneCh, deepPrefetchCancelCh chan<- struct{},
+	didUpdateCh <-chan struct{}) {
+	ctx, cancel := context.WithTimeout(context.Background(), prefetchTimeout)
```
strib: Bonus points for using `CtxWithRandomIDReplayable` to attach a debug tag and random ID to this, so we can track it to the server.

libkbfs/prefetcher.go
```diff
@@ -122,12 +122,6 @@ func (p *blockPrefetcher) run() {
 func (p *blockPrefetcher) request(priority int, kmd KeyMetadata,
 	ptr BlockPointer, block Block, entryName string,
 	doneCh, errCh chan<- struct{}) error {
-	if _, err := p.config.BlockCache().Get(ptr); err == nil {
-		return nil
-	}
```
jzila: This was a major PitA to find: it was causing some huge goroutine leaks and unfinished prefetches.

jzila:
@strib: added a few more fixes in [these](https://github.com/keybase/kbfs/pull/1152/files/16f008762766019916198a3d48ee944a0e3c9fb2..61bca98692619b75e57e4c61c8d2da9f9dda0170) commits.

libkbfs/block_retrieval_queue.go
```diff
@@ -371,10 +371,13 @@ func (brq *blockRetrievalQueue) CacheAndPrefetch(ctx context.Context,
 	return nil
 }
 
+// checkCaches returns a block if it's in one of our caches. If it's
+// successful, it triggers CacheAndPrefetch. Otherwise, it doesn't trigger it
```
strib: Is this comment out of date?  I don't see where `checkCaches` calls `CacheAndPrefetch` at all now.

libkbfs/block_retrieval_queue.go
```diff
@@ -440,9 +437,14 @@ func (brq *blockRetrievalQueue) RequestWithPrefetch(ctx context.Context,
...
 	// Check caches before locking the mutex.
-	err := brq.checkCaches(ctx, priority, kmd, ptr, block, lifetime,
-		deepPrefetchDoneCh, deepPrefetchCancelCh)
+	prefetchStatus, err := brq.checkCaches(ctx, kmd, ptr, block)
 	if err == nil {
+		err = brq.CacheAndPrefetch(ctx, ptr, block, kmd, priority, lifetime,
+			prefetchStatus, deepPrefetchDoneCh, deepPrefetchCancelCh)
+		if err != nil {
+			brq.log.CWarningf(ctx, "An error occurred when caching and/or "+
+				"prefetching disk cached block %s", ptr.ID)
```
strib: Print the error too?

libkbfs/prefetcher.go
strib: Do we have anything calling `checkDataVersion` in the prefetcher path anymore?

libkbfs/block_retrieval_queue.go
jzila: Yep, I'll update it.

libkbfs/prefetcher.go
jzila: We do not. What do you suggest would be the right approach here? I tried adding it in `BlockRetrievalQueue` after `checkCaches`, but it broke a ton of tests.

strib: Hmm, I'm not actually sure why we were checking the version that came out of the cache.  We should probably maintain an invariant that we only put useful blocks in the cache, ones that we can read.

Ideally it would be checked any time a block is fetched from a server, before it's put into the cache.

jzila: @strib: we weren't checking the version that came out of the cache, we were only checking if it _didn't_ come from the cache.

I tried to put that into the block retriever instead, but then `block_ops` tests broke because they were expecting different errors. It would probably be a pretty mechanical fix to update the `DataVersion` fields on those tests so the correct error is returned.

That said, `checkDataVersion` only checks pointers + paths, not blocks themselves. So we'd be checking _before_ retrieving from the server, not after.

strib: Oh right, good point, supposed to check them before fetching.  I think it's a good guard to have, but up to you if you want it in this PR.

jzila: I'll add it and fix the tests.

jzila:
OK I think I got almost everything: https://github.com/keybase/kbfs/pull/1152/files/3076f410a3fdd702eb43df4e83380d42d8127b1f..91f78d4a498d2f0f55df793ca1a8940f0750e2ce

PR #1154, opened 2017-09-08
kbfsgit: optimize `clone` by copying the raw object files out of KBFS

strib:
This doesn't solve the memory/performance problems of big `pull`s, but optimizing for `clone` is an easy win.

Issue: KBFS-2416

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -510,6 +510,39 @@ func (r *runner) processGogitStatus(
 	r.errput.Write([]byte("\n"))
 }
 
+func (r *runner) recursiveByteCount(
+	ctx context.Context, fs billy.Filesystem, totalSoFar int64, toErase int) (
+	bytes int64, toEraseRet int, err error) {
+	fileInfos, err := fs.ReadDir("/")
+	if err != nil {
+		return 0, 0, err
+	}
+
+	for _, fi := range fileInfos {
+		if fi.IsDir() {
+			chrootFS, err := fs.Chroot(fi.Name())
+			if err != nil {
+				return 0, 0, err
+			}
+			totalSoFar, toErase, err = r.recursiveByteCount(
+				ctx, chrootFS, totalSoFar, toErase)
+			if err != nil {
+				return 0, 0, err
+			}
+		} else {
+			totalSoFar += fi.Size()
+			if r.progress {
+				// This function only runs if r.verbosity >= 1.
+				eraseStr := strings.Repeat("\b", toErase)
+				newStr := fmt.Sprintf("%d bytes... ", totalSoFar)
+				toErase = len(newStr)
+				r.errput.Write([]byte(eraseStr + newStr))
+			}
+		}
+	}
+	return totalSoFar, toErase, nil
+}
+
 type statusWriter struct {
 	r           *runner
 	output      io.Writer
@@ -526,7 +559,7 @@ func (sw *statusWriter) Write(p []byte) (n int, err error) {
 
 	sw.soFar += int64(len(p))
 	eraseStr := strings.Repeat("\b", sw.nextToErase)
-	newStr := fmt.Sprintf("%d/%d bytes...", sw.soFar, sw.totalBytes)
+	newStr := fmt.Sprintf("%d/%d bytes... ", sw.soFar, sw.totalBytes)
 	sw.r.errput.Write([]byte(eraseStr + newStr))
 	sw.nextToErase = len(newStr)
 	return n, nil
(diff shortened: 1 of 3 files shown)


kbfsgit/runner.go
```diff
@@ -503,6 +514,196 @@ func (r *runner) processGogitStatus(
...
+	sw.soFar += int64(len(p))
+	eraseStr := strings.Repeat("\b", sw.nextToErase)
+	newStr := fmt.Sprintf("%d/%d bytes... ", sw.soFar, sw.totalBytes)
+	sw.r.errput.Write([]byte(eraseStr + newStr))
+	sw.nextToErase = len(newStr)
+	return n, nil
+}
+
+func (r *runner) copyFile(
+	ctx context.Context, fs billy.Filesystem, localFS billy.Filesystem,
```
songgao: Perhaps name them `from` and `to` rather than `fs` and `localFS` in case we need the other direction?

kbfsgit/runner.go
```diff
@@ -503,6 +514,196 @@ func (r *runner) processGogitStatus(
...
+		}
+	}
+	return nil
+}
+
+// handleClone copies all the object files of a KBFS repo directly
+// into the local git dir, instead of using go-git to calculate the
+// full set of objects that are to be transferred (which is slow and
+// memory inefficient).  If the user requested only a single branch of
+// cloning, this will copy more objects that necessary, but still only
```
songgao: In the future we can follow this with a GC to remove the unnecessary ones, correct?

kbfsgit/runner.go
```diff
@@ -747,6 +948,14 @@ func (r *runner) handleOption(ctx context.Context, args []string) (err error) {
 		r.progress = b
 		r.log.CDebugf(ctx, "Setting progress to %t", b)
 		result = "ok"
+	case gitOptionCloning:
```
songgao: I don't see a cloning related option defined here: https://git-scm.com/docs/git-remote-helpers#_options Is it something`git` passes in to the helper automatically during a clone, or something user needs to provide?

If former, how come the cloning already worked before this is added now? Is it a phase that `git` simply skips if it gets "unsupported" line from helper?

kbfsgit/runner.go
```diff
@@ -774,10 +983,18 @@ func (r *runner) processCommands(ctx context.Context) (err error) {
...
-					return err
+				if r.cloning {
+					r.log.CDebugf(ctx, "Processing clone")
+					err = r.handleClone(ctx)
+					if err != nil {
+						return err
+					}
+				} else {
+					r.log.CDebugf(ctx, "Processing fetch batch")
+					err = r.handleFetchBatch(ctx, fetchBatch)
```
songgao: Is the plan to do the copy thing with fetch as well in the future?

kbfsgit/runner.go
strib: Yes, will add a note.

kbfsgit/runner.go
strib: You don't see that option because you're reading the docs for git 1.8.1.4.  Here are the latest docs: https://git-scm.com/docs/gitremote-helpers#gitremote-helpers-emoptioncloningememtrueememfalseem

It's just a hint `git` gives to the remote helper.  `git` itself always just sends `fetch` commands, and it's up to the helper to decide whether it wants to optimize for cloning or not.

kbfsgit/runner.go
strib: I don't think it makes sense for regular fetches, because in the common case you'll end up processing way more objects that you need to.  I've done other optimizations for `fetch` in #1159.

strib:
> I found it somewhat counterintuitive for the addition to be happening down a level if it's a dir, and at this level if it's not. I'd expect the recursive function to return the total of its subtree, and then we add it at this level whether it's a file or directory.

Agreed it's counterintuitive, but it's the only way to print the total byte count so far interactively, as soon as the bytes are counted.  If I only add the bytes at the end, then I'll only update the count once and there'll be a UI hang until the whole count is done.

I guess I could do a `chan`-based thing where the byte counts are passed around, but that didn't seem worth it and could still cause UI lag.  I'll add a comment explaining that and `toErase` in a new PR.

kbfsgit/runner.go
```diff
@@ -503,6 +514,196 @@ func (r *runner) processGogitStatus(
...
+	}
+
+	for _, fi := range fileInfos {
+		if fi.IsDir() {
+			chrootFS, err := fs.Chroot(fi.Name())
+			if err != nil {
+				return 0, 0, err
+			}
+			totalSoFar, toErase, err = r.recursiveByteCount(
+				ctx, chrootFS, totalSoFar, toErase)
```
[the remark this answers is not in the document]
strib: Actually nevermind, I figured out a way that's better.  Will make a PR soon.

kbfsgit/runner.go
songgao: Oh I see; thanks for the link!

PR #1157, opened 2017-09-09
revendor keybase1 and stub out new notify function

mmaxim:


patrickxb:
Mobile build fails without this, merging.

PR #1159, opened 2017-09-09
kbfsgit: for fetches, use on-demand storer and skip compression

strib:
By default, go-git reads all data objects into memory while building a packfile to push to a remote (which is the operation we use for fetches).  If the repo is too big, this results in extraordinary memory usage.

Instead, use a shim storer to prevent loading object data into memory except when it is needed (via the `Reader()` object call, or to verify the object type or size).

go-git uses delta compression to build the packfile, but it is very slow.  It also messes up the on-demand storage plan, because go-git modifies data in memory and expects it to survive all the way to flushing it to the packfile.  If we read data on demand and then forget about it, go-git will _think_ it compressed an object, but those changes will get lost, and the resulting packfile will be corrupt.  Plus, we don't even need delta compression, because we're not transferring the objects over a network, only to a local git repo. So we turn delta compression off entirely.

Issue: KBFS-2406


strib:
> Is this whole pack thing part of GC? Does it mean the memory issue will need to be dealt with in another way eventually?

Not sure what you mean by "part of GC", but whenever the user runs `git` locally, there is some chance it will run GC/repack and fix the uncompressed packfiles without any extra work on our end.  We could speed up that process by running it ourselves as I suggest in the code comment, but I think `git` will eventually do it for us.  So no, I don't think the memory issue needs to be addressed in a completely new way in the future we just might want to call the extra command to save space until the next `git gc` runs.

> Also, so we want to pack for user git push, where we'd use go-git's fetch?

In that case, the pack file is generated by the local `git` executable, which doesn't have memory issues during the pack construction.  `git-remote-keybase` gets a stream of the already-constructed packfile, and doesn't hold it all in memory at once as far as I can tell, so there's nothing to fix for user pushes.

songgao:
Thanks for the detailed explanation! Makes perfect sense now.

PR #1164, opened 2017-09-11
kbfsgit: interrupt on `git` control-C when stdin is closed

strib:
The only reliable way to tell if the calling `git` process has been ctrl-c'd is if `stdin` is killed.  If that happens before the remote helper gets a blank line command, then the remote helper should exit immediately.

Accomplish this by processing individual commands in a separate goroutine (via a channel), and interrupt that goroutine when EOF is read from stdin.

The added test hangs without the rest of the commit.

Issue: KBFS-2408

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -1014,78 +1014,121 @@ func (r *runner) handleOption(ctx context.Context, args []string) (err error) {
 	return err
 }
 
+func (r *runner) processCommand(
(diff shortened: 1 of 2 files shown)


maxtaco:
cool thanks for fixing this!

jzila:
BTW @songgao if you still want to review this, `w=1` is your friend.

songgao:
ha thanks John

PR #1165, opened 2017-09-12
libgit: implement CreateRepo RPC

strib:
This implements the new `CreateRepo` RPC expected by the service, which creates a repo and errors if the repo already exists.  This is to be served by the existing kbfsfuse/kbfsdokan daemon.

To avoid blocking on any big transfer currently in the journal, this spins up a completely new `Config`, with a journal in a temporary directory, just like the `git-remote-keybase` command does.  This PR refactors some of that code to share between the two paths.

To avoid import cycles, I needed to use the same trick as `SimpleFS` to pass an RPC interface creation hook to the keybase daemon, rather than importing `libgit` directly.

I also added some extra data to the config file that seems like it would be useful to record in the long run, and we're now marshaling it with indentation so it's easier to read by humans.

### DIFF
--- a/libgit/rpc.go
+++ b/libgit/rpc.go
@@ -1,0 +1,151 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import (
+	"os"
+
+	"github.com/keybase/client/go/logger"
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/kbfs/env"
+	"github.com/keybase/kbfs/kbfsmd"
+	"github.com/keybase/kbfs/libkbfs"
+	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+type RPCHandler struct {
+	config libkbfs.Config
+	log    logger.Logger
+}
+
+// NewRPCHandler returns a new instance of a Git RPC handler.
+func NewRPCHandler(config libkbfs.Config) keybase1.KBFSGitInterface {
+	return &RPCHandler{
+		config: config,
+		log:    config.MakeLogger(""),
+	}
+}
+
+var _ keybase1.KBFSGitInterface = (*RPCHandler)(nil)
+
+func (rh *RPCHandler) waitForJournal(
+	ctx context.Context, config libkbfs.Config, h *libkbfs.TlfHandle) error {
+	rootNode, _, err := config.KBFSOps().GetOrCreateRootNode(
+		ctx, h, libkbfs.MasterBranch)
+	if err != nil {
+		return err
+	}
+
+	err = config.KBFSOps().SyncAll(ctx, rootNode.GetFolderBranch())
+	if err != nil {
+		return err
+	}
+
+	jServer, err := libkbfs.GetJournalServer(config)
+	if err != nil {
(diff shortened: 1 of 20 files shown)


strib:
Note that circleCI won't pass because keybase/client#8411 needs to be merged at the same time.

libgit/rpc.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/keybase/kbfs/env"
+	"github.com/keybase/kbfs/kbfsmd"
+	"github.com/keybase/kbfs/libkbfs"
+	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
+	"golang.org/x/net/context"
+)
+
+type RPCHandler struct {
+	config libkbfs.Config
```
songgao: This is the `config` shared with `kbfsfuse` or `kbfsdokan` processes right? Shall we maybe rename it to something like `sharedConfig` to differentiate from the new shortlived `config` generated in the `CreateRepo` handler?

strib: Yes and good idea, though I think I'm going to rename the non-shared ones to `gitConfig` to be consistent with having `config` as a struct field everywhere else.

libkbfs/keybase_daemon.go
```diff
@@ -25,7 +25,10 @@ func (k keybaseDaemon) NewKeybaseService(config Config, params InitParams, ctx C
 		if err != nil {
 			return nil, err
 		}
-		return NewKeybaseDaemonRPC(config, ctx, log, params.Debug, params.CreateSimpleFSInstance), nil
+		return NewKeybaseDaemonRPC(
+			config, ctx, log, params.Debug, params.CreateSimpleFSInstance,
+			params.CreateGitHandlerInstance,
+		), nil
```
jzila: Does this talk over the same socket as `kbfsfuse`? How is that multiplexed?

strib: It is receiving RPCs on the same socket, yes.  Multiplexing happens in the RPC server at the protocol level, because this is registered as a new protocol here: https://github.com/keybase/kbfs/pull/1165/files#diff-12bb5206b823967cfe8c244ffef2c727R277

jzila: All right, so `kbfsfuse` won't create `gitHandler`, but `kbgit` will, so there's a unique protocol handler for each one. What would happen if they both implemented a given handler? Are there any other handlers that `kbgit` implements that duplicate `kbfsfuse`'s, where multiplexing ambiguity over the same socket could be a problem?

strib: Actually no, this `CreateRepo` RPC is served by the running `kbfsfuse` daemon.  We don't spin up a new instance of `git-remote-keybase` just to create the repo.  There's no duplication here; `git-remote-keybase` doesn't serve any RPCs.

strib: (Specifically, `InitParams.CreateGitHandlerInstance` is never set by `git-remote-keybase`, but it is set by `kbfsfuse` and `kbfsdokan`.)

jzila: Cool, perfect, thanks.

strib: Oh, another point just for clarification: `git-remote-keybase` (or anything running in `SingleOp` mode) doesn't introduce itself to the keybase service as being a KBFS instance.  So the service knows not to route any KBFS-related RPCs over that connection.

libgit/rpc.go
songgao: sounds good; thanks!

PR #1168, opened 2017-09-12
git-remote-keybase: default to production run mode

strib:
And warn if running in a non-prod mode.

### DIFF
--- a/kbfsgit/git-remote-keybase/main.go
+++ b/kbfsgit/git-remote-keybase/main.go
@@ -62,6 +62,15 @@ func getLocalGitDir() (gitDir string, err error) {
 func start() (startErr *libfs.Error) {
 	kbCtx := env.NewContext()
 
+	switch kbCtx.GetRunMode() {
+	case libkb.StagingRunMode:
+		fmt.Fprintf(os.Stderr, "Running in staging mode\n")
+	case libkb.DevelRunMode:
+		fallthrough
+	case libkb.NoRunMode:
+		fmt.Fprintf(os.Stderr, "Running in devel mode\n")
+	}
+
 	defaultParams, storageRoot, err := libgit.Params(kbCtx, kbCtx.GetDataDir())
 	if err != nil {
 		return libfs.InitError(err.Error())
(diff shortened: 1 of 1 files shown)


kbfsgit/git-remote-keybase/main.go
```diff
@@ -62,6 +62,15 @@ func getLocalGitDir() (gitDir string, err error) {
 func start() (startErr *libfs.Error) {
 	kbCtx := env.NewContext()
 
+	switch kbCtx.GetRunMode() {
+	case libkb.StagingRunMode:
+		fmt.Fprintf(os.Stderr, "Running in staging mode\n")
+	case libkb.DevelRunMode:
+		fallthrough
+	case libkb.NoRunMode:
+		fmt.Fprintf(os.Stderr, "Running in devel mode\n")
```
songgao: With the changes below that sets default run mode, isn't it impossible to have `linkb.NoRunMode` now?

strib: You can still get it, because `KEYBASE_RUN_MODE` can be set to some unknown, non-empty string before the program is called.

songgao: Ah right. Although in that case, it's something we don't understand; should we run in `devel` mode or just panic?

strib: `GetRunMode` returns production run mode in that case, for production builds.

strib: Actually, I guess I don't need to explicitly set the run mode because of that.  Oh well, it's nice to be explicit.

songgao: > GetRunMode returns production run mode in that case, for production builds.

This is when something other than `dev`, `staging`, or `prod` is passed in as `KEYBASE_RUN_MODE`? I'm a bit confused; doesn't that mean this `libkb.NoRunMode` will never happen?

strib: Yeah . . . oh I guess you're right, because we never let the string be empty, `GetRunMode()` can't actually return `NoRunMode`.  I can panic in that case instead if you prefer.

songgao: I have slight preference for panic but not really a big deal. These questions were mostly to verify my understanding anyway. Thanks!

strib:
Added the panic.  Will merge once #1166 is reviewed and merged, then we'll be ready for testing!

PR #1170, opened 2017-09-13
re-vendor client

jzila:


jzila:
I'm setting the baseline for changes that I actually need.

PR #1172, opened 2017-09-13
Restore pre-BlockOps block cache check, and fix data ver test

strib:


libkbfs/kbfs_ops_test.go
```diff
@@ -1031,7 +1018,7 @@ func TestKBFSOpsLookupNoSuchNameFail(t *testing.T) {
 	}
 }
 
-func TestKBFSOpsLookupNewDataVersionFail(t *testing.T) {
+func TestKBFSOpsReadNewDataVersionFail(t *testing.T) {
```
jzila: Why this change?

libkbfs/folder_block_ops.go
```diff
@@ -308,11 +308,11 @@ func (fbo *folderBlockOps) getCleanEncodedBlockSizeLocked(ctx context.Context,
 		return 0, InvalidBlockRefError{ptr.Ref()}
 	}
 
-	// Try to get the encoded size from the cache before escalating to the
-	// block retriever (even though it's supposed to do a similar thing with
-	// checking caches before checking the data version).
-	// TODO: Figure out how to remove this without breaking journal tests in
-	// kbfs/test.
+	// Try to get the encoded size from the cache before escalating to
```
jzila: Are you using 8-space tabs? This shows up as <80chars with 4-space tabs in my editor.

libkbfs/kbfs_ops_test.go
strib: You removed the `checkDataVersion` call that used to be in `Lookup`, so this test doesn't work anymore when the parent block is served from the cache.

libkbfs/folder_block_ops.go
strib: Nope, 4-space.  Everything looks right to me, in my editor and in github.

jzila: What's your column limit?

libkbfs/kbfs_ops_test.go
jzila: Makes sense, thanks.

strib:
> Can we remove the redundant integrity check? Seems like the storer can remove it, unless it's dependent on it in some way. assembleBlock will do it anyway along all paths, meaning journal, disk cache, and server.

If we remove the block_journal.go one, we'd be missing the check on journal flushes, when reading the block back in from the disk before pushing it.  That's arguably not very important since the server should catch it, but it's nice to catch it sooner maybe.  In any case, it's not relevant to this PR, I'd do it in a new PR if we want that.

> Why was this slowing anything down? The blocks are in the in-memory cache, so why are they also in the journal? I thought the DirtyBlockCache is for journaled blocks, and we've already checked that.

Blocks get saved in the in-memory cache at the same time they are written to the journal, to avoid having to re-read blocks from the disk.  These are finalized, clean blocks.  The `DirtyBlockCache` is only for blocks with outstanding changes that haven't gone through `SyncAll` yet.

libkbfs/folder_block_ops.go
strib: 80.
80chars

strib: Oh, you're complaining because it's less than 80 chars?  I just used alt-q in emacs, like I always do.  It doesn't push things right to the edge.

jzila: I'm just trying to get us to a place where when I reflow something in my client, I know that your client won't do it, because often it's distracting in PRs to have reflowed text mixed with actual code changes.

strib: Ok sorry.  I just use: http://ergoemacs.org/emacs/modernization_fill-paragraph.html, which says "~70 characters".  That's what I've done since the beginning of KBFS.  Not sure how to translate that to other editors.

jzila:
> That's arguably not very important since the server should catch it, but it's nice to catch it sooner maybe.

IMO it isn't worth the performance hit; I'd rather have the server reject it in the rare cases when it happens.

> In any case, it's not relevant to this PR, I'd do it in a new PR if we want that.

Agreed.

> Blocks get saved in the in-memory cache at the same time they are written to the journal, to avoid having to re-read blocks from the disk. These are finalized, clean blocks. The DirtyBlockCache is only for blocks with outstanding changes that haven't gone through SyncAll yet.

Neat, thanks! I assume they get saved as `PermanentEntry` until they are synced? In any case, I have 2 alternative proposals:
* Modify `BlockOps.Get` such that it checks the in-memory cache first, then `checkDataVersion`, then check the journal, then pass through to the `BlockRetriever`. The `BlockRetriever` would not do `checkDataVersion` in this case.
* Modify the `BlockRetriever` such that it checks the journal after the cache and `checkDataVersion`.

With either of these proposals, we should be able to skip all the cache and `checkDataVersion` checks in `folderBlockOps`.

jzila:
On a second perusal of the code, I actually think the first proposal is better, since the journal server has its own data size lookups so it would be overkill to do a full retrieval when we're just trying to get the size.

strib:
We can move the cache check into `BlockOps.Get` I guess, but it means the read notifications (currently ignored by the GUI) will be wrong, since the data being read doesn't need to be decrypted.  We can move that too I guess, but then things are getting pretty far astray.  It also would unlock the blockLock for no reason.

I just want to restore performance to where it was ASAP and then we can worry about refactoring.  Can we push this first and I'll follow it up with the riskier one that moves stuff around?

PR #1174, opened 2017-09-14
libfs: OpenFile should create any missing parent directories

strib:
Apparently go-git assumes this, but it's not documented anywhere. Without this change, we're unable to push branches with "/" in their names.

Issue: KBFS-2430

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -267,6 +267,28 @@ func translateErr(err error) error {
 	}
 }
 
+func (fs *FS) mkdirAll(filename string, perm os.FileMode) (err error) {
+	defer func() {
+		err = translateErr(err)
+	}()
+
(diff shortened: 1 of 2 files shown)


libfs/fs.go
```diff
@@ -277,6 +299,11 @@ func (fs *FS) OpenFile(filename string, flag int, perm os.FileMode) (
 		err = translateErr(err)
 	}()
 
+	err = fs.mkdirAll(path.Dir(filename), 0755)
```
songgao: Have you considered the case where `path.Dir` would return `/` if there's only one element in a path with leading slash (e.g. `/blah`)? Looks like it's gonna cause problems in the `parts := strings.Split(leftover, "/")` later on?

strib: Good call thanks: #1175.

PR #1177, opened 2017-09-14
kbfsgit: allow debugging to be turned off with an env var

strib:
With this commit, you can turn off KBFS debugging by setting
KBFSGIT_DEBUG=0 in your environment.

Note I had to move the `NewWithCallDepth` line below the
`SetLogFileConfig` line, because the latter call erases all the
default INFO levels set by the keybase logger whenever a new log is
made.

Turning off debugging completely reduced the time of the "Preparing"
phase by 52% (31% compared to #1176), but I don't think we want to do
this by default until we're more comfortable with git.

Issue: KBFS-2390

### DIFF
--- a/libgit/init.go
+++ b/libgit/init.go
@@ -7,6 +7,7 @@ package libgit
 import (
 	"context"
 	"io/ioutil"
+	"os"
 
 	"github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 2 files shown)


songgao:
(LGTM otherwise)

PR #1179, opened 2017-09-15
libgit: implement the delete repo RPC call

strib:
And vendor in the protocol changes for it.

Depends on keybase/client#8478, will revendor once that's merged.

Issue: KBFS-2435

### DIFF
--- a/libgit/repo.go
+++ b/libgit/repo.go
@@ -20,9 +20,10 @@ import (
 )
 
 const (
-	kbfsRepoDir       = ".kbfs_git"
-	kbfsConfigName    = "kbfs_config"
-	gitSuffixToIgnore = ".git"
+	kbfsRepoDir         = ".kbfs_git"
+	kbfsConfigName      = "kbfs_config"
+	gitSuffixToIgnore   = ".git"
+	kbfsDeletedReposDir = ".kbfs_deleted_repos"
 )
 
 // This character set is what Github supports in repo names.  It's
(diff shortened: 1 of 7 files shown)


libgit/repo.go
```diff
@@ -262,3 +266,52 @@ func CreateRepoAndID(
 	}
 	return id, err
 }
+
+// DeleteRepo "deletes" the given repo in the given TLF.  RIght now it
```
jzila: s/RIght/Right

libgit/repo.go
```diff
@@ -262,3 +266,52 @@ func CreateRepoAndID(
...
+
+	deletedReposNode, err := lookupOrCreateDir(
+		ctx, config, repoNode, kbfsDeletedReposDir)
+	if err != nil {
+		return err
+	}
+
+	// For now, just rename the repo out of the way, using the uniq
+	// ID.  TODO(KBFS-2442): periodically delete old-enough repos from
+	// `kbfsDeletedReposDir`.
```
jzila: So `kbfsDeletedReposDir` is a staging area for deleted repos? Meaning anything there is good to be deleted?

libgit/repo.go
```diff
@@ -262,3 +266,52 @@ func CreateRepoAndID(
...
+func DeleteRepo(
+	ctx context.Context, config libkbfs.Config, tlfHandle *libkbfs.TlfHandle,
+	repoName string) error {
+	// Create a unique ID using the verifying key and the `config`
+	// object, which should be unique to each call in practice.
+	session, err := config.KBPKI().GetCurrentSession(ctx)
+	if err != nil {
+		return err
+	}
+	uniqID := fmt.Sprintf("%s-%p", session.VerifyingKey.String(), config)
```
jzila: I don't love using the pointer value of `config` as the unique ID. Is there a reason not to use `kbfscrypto.RandRead`?

libgit/repo.go
strib: Yes, though it's probably best to wait until the renamed repo has been deleted at least a few minutes, to make sure any concurrent writers have finished pushing to it.

libgit/repo.go
strib: It was easier, and it's guaranteed to be unique for this call.  Technically a random number isn't guaranteed to be unique.  Why don't you like it?

jzila: I don't understand how a pointer is guaranteed to be unique.

strib: It only needs to be unique "during the lifetime of this call", as the function comment explains.  A new config is passed into each of these calls, and therefore any calls happening at the same time will have unique pointers.

jzila: But the new config's pointer can be the same as the old one...

jzila: Oh nvmind I follow, different configs _at the same time_

strib: Yeah that's what I was thinking, but you're right.  By including it in the name, it needs to be unique for all time, or at least until the next delete run happens.  I'll fix it soonish.

strib:
@jzila: fixed the uniqueness issue in 6229826.  If you agree, then this can be merged once keybase/client#8478 is approved and merged.

PR #1187, opened 2017-09-22
small fix to set success flag in kbgit test

songgao:


songgao:
I was confused too ... perhaps journal maintained a file descriptor that it kept writing into and kernel kept it alive until closed?

strib:
No, it makes directory entries.  But it might just call something like "mkdir -p" whenever it makes a new entry.

songgao:
Oh OK that makes sense. Does it mean we don't actually need to create the tempdir manually in tests?

strib:
Maybe but let's not mess with it now.  Did you actually see a test fail without this fix, or just something you noticed?

songgao:
Sounds good. (No tests failing because of this; I was setting up tests for locking and noticed this.)

PR #1188, opened 2017-09-22
implement Lock/Unlock for libfs.File (and some plumbing)

songgao:
This PR makes some changes in `journal_server.go` and `tlf_journal.go` to support locking, and implements Lock/Unlock for `libfs/file.go` based on [KBFS-2347](https://keybase.atlassian.net/browse/KBFS-2347).

Note that writing MD while requiring a lock is only supported with `FinishSingleOp`, but not regular `Put`. This is because in non-singleOp mode, there can be multiple MDs, which can require different locks. This means we'd need to plumb through journal (and prepper?) to make sure MDs requiring different locks are not squashed (or aggregated/appended during squash if we decide to support multiple `LockContext` in `PutMetadata`). If we need locking support in non-singleOp mode in the future, we can add it.

Also, with this PR in, all GetRange calls that require locking will simply make a GetRange to the real mdserver directly. In the future we can probably just make a `Lock` call if those requested MDs are available locally.


strib:
Haven't reviewed yet, but FYI when we vendor in the actual go-git calls to `Lock()`, we'll need 49ddce0 (added as part of PR #1191).

songgao:
Great! Thanks for the heads up.

So #1191 and this together complete the locking support for Git in KBFS, if I understand correctly?

strib:
> So #1191 and this together complete the locking support for Git in KBFS, if I understand correctly?

Well not on purpose, I just had to vendor go-git in #1191 to get a different fix.  I did no testing of the lock stuff or fast-forward-only pushes.

libfs/file.go
```diff
@@ -114,3 +118,65 @@ func (f *File) Close() error {
 	f.node = nil
 	return nil
 }
+
+func (f *File) getLockID() keybase1.LockID {
+	// We must not change or remove this since it'd break locking guarantees.
+	return keybase1.LockIDFromBytes(
+		bytes.Join([][]byte{
+			f.fs.GetLockNamespace().Bytes(),
+			[]byte(f.Name()),
```
strib: Isn't this supposed to be a hash, so we don't expose the filename to the server?

libfs/fs_test.go
```diff
@@ -29,6 +31,44 @@ func makeFS(t *testing.T, subdir string) (
...
+		err := ioutil.RemoveAll(tempdir)
+		assert.NoError(t, err)
+	}
+
+	h, err := libkbfs.ParseTlfHandle(ctx, config.KBPKI(), "user1", tlf.Private)
+	require.NoError(t, err)
+	fs, err := NewFS(ctx, config, h, subdir, "")
+	require.NoError(t, err)
+
+	// Flush the journal first otherwise tests fail on errNoFlushedRevisions.
```
strib: Ah, this code is probably why you didn't have the problem I had in #1191.  This shouldn't need to be here, but we can remove it in #1191 if you want.

libfs/file.go
strib: Also, using `f.Name()` will work fine in practice, but technically `f.fs` could be chroot'd to some subdirectory within the repo, and `f.Name()` might not be truly unique within the repo.  You could use `f.fs.Root()` to fix that, but it would have some extra components like tlf and repo name that you wouldn't need.

libkbfs/journal_md_ops.go
```diff
@@ -403,6 +403,11 @@ func (j journalMDOps) GetRange(ctx context.Context, id tlf.ID, start,
 		j.jServer.deferLog.LazyTrace(ctx, "jMDOps: GetRange %s %d-%d done (err=%v)", id, start, stop, err)
 	}()
 
+	if lockBeforeGet != nil {
+		// We need to grab locks. So bypass journal and hit server directly.
+		return j.MDOps.GetRange(ctx, id, start, stop, lockBeforeGet)
```
strib: I don't think this is correct.  It will not return any MD entries that happen to be in the journal.  In your use case that's fine, because we only ever set the lock when fetching entries that come from the server, but in general it's dangerous if there are ever other types of callers.

I'd be more comfortable if we called the delegate MDOps directly from `folderBranchOps.getAndApplyMergedUpdates`, but that would require a bunch of annoying changes.  Maybe for now just return an if there's any revision in the journal within the requested range, when the lock is set?

libkbfs/mdserver_memory.go
```diff
@@ -548,19 +569,54 @@ func (md *MDServerMemory) Put(ctx context.Context, rmds *RootMetadataSigned,
...
+	// NOTE: taking lock would always succeed in MDServerMemory since it's
+	// always within the same session.
+	lockKey := mdLockMemKey{
+		tlfID:  tlfID,
+		lockID: lockID,
+	}
+	existing, ok := md.lockIDs[lockKey]
+	if !ok || md.config.Clock().Now().Sub(existing) >= mdLockTimeout {
+		// The lock doesn't exists or has expired.
+		md.lockIDs[lockKey] = md.config.Clock().Now()
```
strib: Why not track the expiration time instead of the lock time?  That would avoid a few of these subtractions.

libkbfs/tlf_journal.go
```diff
@@ -2361,5 +2368,36 @@ func (j *tlfJournal) finishSingleOp(ctx context.Context) error {
...
 			return err
 		}
+		_, noLock := j.lastFlushErr.(kbfsmd.ServerErrorRequiredLockIsNotHeld)
+		if noLock {
+			return j.lastFlushErr
+		}
+	}
+}
+
+func (j *tlfJournal) waitForBlockFlush(ctx context.Context) error {
```
strib: Would be nice to have a test in `tlf_journal_test.go` for this function.

libfs/file.go
```diff
@@ -114,3 +118,65 @@ func (f *File) Close() error {
 	f.node = nil
 	return nil
 }
+
+func (f *File) getLockID() keybase1.LockID {
+	// We must not change or remove this since it'd break locking guarantees.
```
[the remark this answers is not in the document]
songgao: Sounds good. Although since we don't have a way to *require* client to upgrade with current scheme, I'm not sure that'll work well. Perhaps we should include a way for `mdserver` to require minimum client version so that we can reject old clients once the "transition" is done?

libfs/file.go
songgao: Yup! `keybase1.LockIDFromBytes` hashes the passed in bytes and turns it into an integer (keybase1.LockID is an integer) to send to `mdserver`

songgao: >  in practice
Meaning we never chroot in `git` right?

In context of git, I'm concerned about including `f.fs.Root()` since repo name can change. I think we should only associate with the path inside the repo and the repo ID. But good point on chroot from billy perspective. But I don't have a better idea. Perhaps we can alter the LockNamespace in `fs.Chroot`, or just mention in the comment of Lock() that chroot breaks locking?

libfs/file.go
```diff
@@ -114,3 +118,65 @@ func (f *File) Close() error {
...
+	return jServer.FinishSingleOp(f.fs.ctx,
+		f.fs.root.GetFolderBranch().Tlf, &keybase1.LockContext{
+			RequireLockID:       f.getLockID(),
+			ReleaseAfterSuccess: true,
+		})
+}
+
+// Truncate does not implement the billy.File interface for File.
+func (f *File) Truncate(size int64) error {
+	return errors.New("not implemented")
```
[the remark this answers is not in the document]
songgao: Thanks for the reminder! I just didn't think about it yet; but I'll try to add it to this PR.

libfs/fs_test.go
songgao: Hmmm where do you think the fix would be? I think with out this flush, error comes from `SyncFromServerForTesting` called in `Lock`.

libkbfs/journal_md_ops.go
songgao: Oh I see; I didn't look closely enough and wrongly think this goes to server directly. Thanks for spotting that! 

>  Maybe for now just return an if there's any revision

Is it "just return an error if" what you meant?

libfs/file.go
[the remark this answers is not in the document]
strib: Yeah, we can always make the mdserver enforce a min client version.  But that's something we can add later.  If you want to be more clear about it, you could use the first byte of a lock ID as a lock ID version number.

libfs/file.go
strib: Oh oops, I forgot about that.  Sorry!

strib: > Meaning we never chroot in git right?

Not for the FS we pass into go-git, where the locking happens.

You could just strip off the first N (5?) elements from Root() if you want to get the path just within the repo.

libfs/fs_test.go
strib: The fix is 49ddce0.  You can move it to this PR if you want it.

libkbfs/journal_md_ops.go
strib: > I didn't look closely enough and wrongly think this goes to server directly

Your code here does go to the server directly.  But my point is that it's wrong to only go to the server if there's anything in the journal within the start/stop range.

And yeah, I meant "error".

songgao: Oh I see you point now. Thanks for clarification! Error sounds good; will do!

libfs/fs_test.go
songgao: Oh I see! I guess to avoid weird conflict it's better to remove this part in #1191 then. Thanks for the fix!

libfs/file.go
songgao: But then that's making the assumption in billy that we are using `git`, which is not ideal right?

songgao: NP!

libfs/file.go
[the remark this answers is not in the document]
songgao: Oh interesting! That's a great idea. Will do.

libfs/file.go
strib: Hmm, I guess? I don't care much though.  In the custom context object you could have another method that transforms the filename maybe, and we could have a custom one for git, just like the namespace bytes?

libfs/file.go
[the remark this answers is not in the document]
songgao: Made the change [here](https://github.com/keybase/client/pull/8630); will vendor it in once it's merged.

libfs/file.go
songgao: Since the locking namespace is set in `FS`, I think it's better to merge this into `FS`. I'm pushing a commit that just builds a unix canonical path and uses that as default namespace. In case of Chroot it appends the changed path to the original namespace. This should work for both default case and git. Let me know if it sounds good!

songgao:
Addressed some feedback, but I still need to implement `Truncate` as well as add that requested test. Will get them done tomorrow!

songgao:
Ugh sorry I didn't get to this PR today because of the certs stuff. Will make some progress tonight or tomorrow!

songgao:
I think test failure is legit. It goes to a conflict branch sometimes. Trying to fix ...

libfs/fs.go
```diff
@@ -79,16 +88,20 @@ func NewFS(ctx context.Context, config libkbfs.Config,
 	log.CDebugf(ctx, "Made new FS for TLF=%s, subdir=%s",
 		tlfHandle.GetCanonicalName(), subdir)
 
+	// Use the canonical unix path for default locking namespace, as this needs
+	// to be the same across all platforms.
+	unixFullPath := path.Join("/keybase", tlfHandle.Type().String(), subdir)
```
jzila: I know I've mentioned this in the past with one of Strib's PRs, but just make sure that `subdir` is slashed correctly on Windows vs Unix.

songgao:
Merging without waiting for CI since Linux fuse tests are hanging on CI right now.

libfs/fs.go
songgao: This is intentionally forward slashed since it has to be consistent across platforms. It's just used as part of lock ID hashing source, but never for locating files on the filesystem.

PR #1191, opened 2017-09-23
kbfsgit: enable non-force pushes, and onDemandStorer shouldn't hide missing obj error

strib:
go-git depends on getting on error from `EncodedObject()` when an object is missing, especially when constructing the set of "ignored" objects.  This is the set of objects that the local git repo already has. and there's no need for them to exist in the kbfs git repo. go-git allows objects to be missing when constructing that ignored list, but we were hiding that error in onDemandObject.cache(). Instead, we should surface it during `EncodedObject`, but without fully reading all object data from disk, which would negate the point of previous optimizations.

This is based on #1188, which vendors in go-billy master.  This also vendors in go-git master, which pulls in the locking code, requiring a small `folderBranchOps` fix.

Issue: KBFS-2445

### DIFF
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -7,6 +7,7 @@ package libfs
 import (
 	"bytes"
 	"io"
+	"sync"
 	"sync/atomic"
 
 	"github.com/keybase/client/go/protocol/keybase1"
@@ -26,6 +27,9 @@ type File struct {
 	node     libkbfs.Node
 	readOnly bool
 	offset   int64
+
+	lockedLock sync.Mutex
+	locked     bool
 }
 
 var _ billy.File = (*File)(nil)
@@ -116,6 +120,10 @@ func (f *File) Seek(offset int64, whence int) (n int64, err error) {
 
 // Close implements the billy.File interface for File.
 func (f *File) Close() error {
+	err := f.Unlock()
+	if err != nil {
+		return err
+	}
 	f.node = nil
 	return nil
 }
@@ -131,9 +139,21 @@ func (f *File) getLockID() keybase1.LockID {
 }
 
 // Lock implements the billy.File interface for File.
-func (f *File) Lock() error {
+func (f *File) Lock() (err error) {
+	f.fs.log.CDebugf(f.fs.ctx, "Locking!")
+	f.lockedLock.Lock()
+	defer f.lockedLock.Unlock()
+	if f.locked {
+		return nil
+	}
+	defer func() {
+		if err == nil {
+			f.locked = true
+		}
+	}()
+
 	// First, sync all and ask journal to flush blocks of all existing writes.
-	err := f.fs.SyncAll()
+	err = f.fs.SyncAll()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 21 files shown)


songgao:
Sorry for the delay on reviewing this PR. Will go through this soon!

strib:
@songgao: wait one minute, I'm adding a few new things to it now that your stuff is merged.

strib:
Ok this is ready to go now.  Had to do a few more things to make locking work:
* Call `Unlock` from `File.Close()`.  This means if the lock isn't held, `Unlock` must be a no-op and return `nil` since it will be very commonly-called.
* `MDServerMemory` has to do an unlock even when returning early.
* `TlfJournal` shouldn't unlock if there are multiple MDs in the journal (which only ever happens if the initial revision isn't yet flushed).

Ready for review!

songgao:
> Call Unlock from File.Close(). This means if the lock isn't held, Unlock must be a no-op and return nil since it will be very commonly-called.

Hmm.. I thought if the lock isn't held, the Unlock had to fail since it's where MD is written to the server.

strib:
> Hmm.. I thought if the lock isn't held, the Unlock had to fail since it's where MD is written to the server.

I meant if even the local code doesn't think it's locked -- i.e., if `Lock()` was never called on this `File` object.  If the local code thinks it should hold the lock, but the server disagrees, this PR still returns an error in that case.

songgao:
Oh I see. That makes sense.

libkbfs/tlf_journal.go
```diff
@@ -1435,6 +1436,32 @@ func (j *tlfJournal) flushOneMDOp(ctx context.Context,
...
 
+	// If this is the first revision, and there are more MD entries
+	// still to be flushed, don't pass in lc yet, wait for the next
+	// flush.  Normally the lock context is used in single-op mode,
+	// and there will always only be one revision to flush.  The
+	// exception is when the first revision of the TLF hasn't been
+	// pushed yet, since it can't be squashed.  In that case, there
+	// could be 2 revisions in the MD journal.  Don't unlock on the
+	// first flush, since we'll still need to hold the lock until the
+	// second flush.
```
songgao: Thanks for the explanation here! I didn't realize this was needed.

libfs/fs_test.go
```diff
@@ -597,34 +592,37 @@ func TestFileLocking(t *testing.T) {
...
 	require.NoError(t, err)
 
-	// Now we do have some stuff that needs to flush, and we don't have the
-	// lock, so this should fail like the first Unlock.
+	// Now we do have some stuff that needs to flush, but we don't
+	// have the lock, so this should complete without flushing
+	// anything.
 	err = f.Unlock()
-	require.Error(t, err)
+	require.NoError(t, err)
```
songgao: Now that it doesn't error as long as local state agrees the lock is not held, do you think we should add a case here to make sure an expired lock causes `Unlock` to error?

strib: Added a test for that, and one for non-force and force pushes for the runner.

libfs/file.go
```diff
@@ -26,6 +27,9 @@ type File struct {
 	node     libkbfs.Node
 	readOnly bool
 	offset   int64
+
+	lockedLock sync.Mutex
```
jzila: Confusing naming, maybe `fileLock`?

libkbfs/mdserver_memory.go
```diff
@@ -336,12 +336,19 @@ func (md *MDServerMemory) getCurrentDeviceKey(ctx context.Context) (
...
 	md.log.CDebugf(ctx, "GetRange %d %d (%s)", start, stop, mStatus)
-	bid, err := md.checkGetParamsRLocked(ctx, id, bid, mStatus)
+	bid, err = md.checkGetParamsRLocked(ctx, id, bid, mStatus)
 	if err != nil {
 		return nil, err
 	}
+
+	defer func() {
+		if err == nil && lockBeforeGet != nil {
+			md.lockLocked(ctx, id, *lockBeforeGet)
```
jzila: lockLockedLockLockLocked :P

libfs/file.go
strib: I hear you, but I don't think it's worth a PR right now.  Next time I touch this code I'll make it happen.

PR #1201, opened 2017-09-27
make git errors exportable

oconnor663:
This will let the GUI do nicer things with them once they come out on the other end with proper error codes.

r? @strib


oconnor663:
(initial build failed because some linter didn't like the receiver name I was using)

PR #1205, opened 2017-09-27
libkbfs: Handle GetForTLF returning an empty result in folder branch ops

taruti:
Spotted in wild.


strib:
Hmm, this is a bit strange, because if the user already has an unmerged MD update, by definition there needs to be a merged one that it conflicted against, to exist in the first place.

I wrote more details in the JIRA ticket.

jzila:
Closing for archival.

PR #1207, opened 2017-09-27
clean up git errors a little bit more

oconnor663:
r? @strib


oconnor663:
Going to do this in core instead.

PR #1208, opened 2017-09-28
remove WaitForBlockFlush and test since we don't use it anymore

songgao:
Flake seems to be caused by occasional state races from the `stateCh`. So just drain it without checking, since the test checks for block flushing using `getJournalEntryCounts` anyway.


strib:
@songgao: you might be hitting the same bug that I fixed in `testTLFJournalSingleOp`: https://github.com/keybase/kbfs/blob/3ad48ad146501203566475accf860851263932d6/libkbfs/tlf_journal_test.go#L1817-L1826

Can you please try something like that instead?

songgao:
I saw some context canceled errors so I thought it might have been a different order. But I'll try this and see if it works too. Thanks!

songgao:
I keep getting context canceled with this:

```
	go func() {
		// For waitForBlockFlush
		delegate.requireNextState(ctx, bwBusy)
		delegate.requireNextState(ctx, bwIdle)

		// For finishSingleOp.
		delegate.requireNextState(ctx, bwBusy)
		// Background loop awakens after the finish is signaled.  Should
		// now be on a conflict branch.  The pause signal sent by the
		// branch-converter races with the background work finishing
		// (KBFS-2440), and so the second state could be either idle or
		// paused, depending on what gets processed first.
		nextState := delegate.requireNextState(ctx, bwPaused, bwIdle)
		if nextState == bwIdle {
			delegate.requireNextState(ctx, bwPaused)
		}
	}()
```

@strib any ideas?

strib:
Ah, I think you should probably `pauseBackgroundWork()` at the beginning of the test, like in the other test.  Otherwise the background worker will be making transitions out of your control.

songgao:
OK this seems to work well but I don't really know why ...

strib:
I don't think this is right either.  I get errors pretty quickly when I run it in a loop 1000 times.

I spent a bunch of time thinking about this, and I don't think it's possible to write a flake-free test without signaling the delegate somehow in `waitForBlockFlush` that `j.signalWork()` has been called for sure.  Otherwise races are inevitable.  You could add a channel for that to `testBWDelegate`.  It would make the finishSingleOp test simpler as well.

Note that you can't call the `requireNextState()` functions from a goroutine, because if they fail they call `t.Fatal` and that has to be called from the main test goroutine.

If you think it's not worth it, you can completely nuke this test if you want.  Might be too much of a pain to do this right.

songgao:
Ah I put that in a loop and waited for 10~20 iterations with no error, which is why I thought that was not flaky. I’ll give it a try to implemented what you suggested above before giving up.

songgao:
@jzila we just use a `FinishSingleOp` now: https://github.com/keybase/kbfs/blob/29a90fe257faea92d6279e388ed2c72a12b6863f/libfs/file.go#L170 That's where WaitForBlockFlush was used earlier, but it turned out `FinishSingleOp` was needed actually.

PR #1209, opened 2017-09-28
 switch to the libkb git error types

oconnor663:
~~At the same time, implement ExportableError for withStack, so that it doesn't hide the exportability of the types it wraps.~~

Core takes advantage of all this export/import plumbing in https://github.com/keybase/client/pull/8703.

r? @strib 

cc @chrisnojima


vendor/github.com/pkg/errors/errors.go
```diff
@@ -174,6 +177,18 @@ func (w *withStack) Format(s fmt.State, verb rune) {
...
+func (w *withStack) ToStatus() (s keybase1.Status) {
+	if exportable, ok := w.error.(libkb.ExportableError); ok {
+		return exportable.ToStatus()
+	}
+	s.Code = int(keybase1.StatusCode_SCGeneric)
+	s.Name = "GENERIC"
+	s.Desc = w.Error()
+	return
+}
+
```
oconnor663: Woops, I made this change in a vendored package. I'll make it upstream and then revendor it here.

oconnor663: Err, hell, that's an upstream package. I'll figure something out...

oconnor663: (Fixed this by taking a dependency on `pkg/errors` in client.)

libgit/repo.go
```diff
@@ -185,7 +147,7 @@ func getOrCreateRepoAndID(
 	ctx context.Context, config libkbfs.Config, tlfHandle *libkbfs.TlfHandle,
 	repoName string, uniqID string, createOnly bool) (*libfs.FS, ID, error) {
 	if !checkValidRepoName(repoName, config) {
-		return nil, NullID, errors.WithStack(InvalidRepoNameError{repoName})
+		return nil, NullID, errors.WithStack(libkb.InvalidRepoNameError{Name: repoName})
```
oconnor663: Note that the client changes we're vendoring in include an `ExportToStatus` change than can unwrap `errors.WithStack`.

PR #1210, opened 2017-09-29
kbfsgit: optimizations for large `push --mirror/--all` initializations

strib:
This PR adds optimizations for the case where the user has a local repo with a lot of references, and then runs a `git push --mirror` or `git push --all` to initialize a new KBFS repo from this local repo.

`--mirror` syncs every single local ref to the remote, and so we can optimize by just copying all the objects/refs/packed-refs files themselves, rather than going through expensive go-git computation.  If there is a packed-refs file, this allows us to avoid having a ton of individual refs files, which is slow and might break KBFS's dir size limitations.

`--all` syncs every branch and tags (i.e., `refs/heads` and `refs/tags`). If these are the only types of refs in the local repo, then this is equivalent to `--mirror`.  If not, we can't use the above optimization because there may be some objects that shouldn't be transferred.  However, we still want the benefits of a packed-refs file as above.  Instead, when initializing a repo with more than 10 refs, we ask go-git to create a packed-refs file, rather than using individual refs files.

Also fixes a bug where a "." directory was being created in every git repo.  Oops!

Issue: KBFS-2464

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -75,6 +75,10 @@ const (
 	// push, we actually fetch from the local repo and write the
 	// objects into the bare repo.
 	localRepoRemoteName = "local"
+
+	// The minimum number of refs needed to initialize a repo with a
+	// packed-refs file.
+	packedRefsThreshDefault = 10
 )
 
 type ctxCommandTagKey int
@@ -101,6 +105,8 @@ type runner struct {
 
 	logSync     sync.Once
 	logSyncDone sync.Once
+
+	packedRefsThresh int
 }
 
 // newRunner creates a new runner for git commands.  It expects `repo`
@@ -147,18 +153,19 @@ func newRunner(ctx context.Context, config libkbfs.Config,
 	uniqID := fmt.Sprintf("%s-%d", session.VerifyingKey.String(), os.Getpid())
 
 	return &runner{
-		config:    config,
-		log:       config.MakeLogger(""),
-		h:         h,
-		remote:    remote,
-		repo:      parts[2],
-		gitDir:    gitDir,
-		uniqID:    uniqID,
-		input:     input,
-		output:    output,
-		errput:    errput,
-		verbosity: 1,
-		progress:  true,
+		config:           config,
+		log:              config.MakeLogger(""),
+		h:                h,
+		remote:           remote,
+		repo:             parts[2],
+		gitDir:           gitDir,
+		uniqID:           uniqID,
+		input:            input,
+		output:           output,
+		errput:           errput,
+		verbosity:        1,
+		progress:         true,
+		packedRefsThresh: packedRefsThreshDefault,
 	}, nil
 }
 
@@ -1077,10 +1084,10 @@ func (r *runner) handleFetchBatch(ctx context.Context, args [][]string) (
 // --all/--mirror).
 func (r *runner) canPushAll(
 	ctx context.Context, repo *gogit.Repository, args [][]string) (
-	bool, error) {
+	canPushAll, repoEmpty bool, err error) {
 	refs, err := repo.References()
 	if err != nil {
-		return false, err
+		return false, false, err
 	}
 	defer refs.Close()
 
@@ -1089,26 +1096,26 @@ func (r *runner) canPushAll(
 		if errors.Cause(err) == io.EOF {
 			break
 		} else if err != nil {
-			return false, err
+			return false, false, err
 		}
 
 		if ref.Type() != plumbing.SymbolicReference {
 			r.log.CDebugf(ctx, "Remote has at least one non-symbolic ref: %s",
 				ref.String())
-			return false, nil
+			return false, false, nil
 		}
 	}
 
 	sources := make(map[string]bool)
 	for _, push := range args {
 		if len(push) != 1 {
-			return false, errors.Errorf("Bad push request: %v", push)
+			return false, false, errors.Errorf("Bad push request: %v", push)
 		}
 		refspec := gogitcfg.RefSpec(push[0])
 		// If some ref is being pushed to a different name on the
 		// remote, we can't do a push-all.
 		if refspec.Src() != refspec.Dst("").String() {
-			return false, nil
+			return false, true, nil
 		}
 
 		src := refspec.Src()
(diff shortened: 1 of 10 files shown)


strib:
@malgorithms: when this optimization runs, the output on pushes is slightly different as it's directly copying files, similar to the `clone` optimization I did.  Right now the output looks like:

```
Counting objects: 193 bytes... done.
Preparing objects: (100.00%) 193/193 bytes... done.
Counting refs: 82 bytes... done.
Preparing refs: (100.00%) 82/82 bytes... done.
Counting packed refs: 322 bytes... done.done.
Preparing packed refs: (100.00%) 322/322 bytes... done.
Syncing data to Keybase: (100.00%) 24.00/24.00 KB... done.
```

I.e., a byte-counting phase and a preparing phase for each type of thing we're syncing into KBFS.  Let me know if you want to change these in some way, like we did for "Cryptographic cloning"

malgorithms:
hmm, what do you think about making the last step "Syncing encrypted data to Keybase" ? 

I guess we'd want to say "signed" instead of "encrypted" in the case it's not a team/private folder,  for if we ever expose that (or can people do it now anyway just outside the GUI?)?  If switching on that is annoying, we could make the last step "Crypto & syncing data to Keybase"

maxtaco:
ping @jzila  and/or @songgao !

kbfsgit/runner.go
```diff
@@ -925,40 +988,143 @@ func (r *runner) handleFetchBatch(ctx context.Context, args [][]string) (
...
+
+	sources := make(map[string]bool)
+	for _, push := range args {
+		if len(push) != 1 {
+			return false, false, errors.Errorf("Bad push request: %v", push)
+		}
+		refspec := gogitcfg.RefSpec(push[0])
+		// If some ref is being pushed to a different name on the
+		// remote, we can't do a push-all.
+		if refspec.Src() != refspec.Dst("").String() {
```
songgao: I wanted to know why `""` was used here, but I'm having a hard time understanding the `Dst` method [here](https://github.com/src-d/go-git/blob/f9879dd043f84936a1f8acb8a53b74332a7ae135/config/refspec.go#L114). Is this the correct test for ref name being same or not?

kbfsgit/runner.go
```diff
@@ -925,40 +988,143 @@ func (r *runner) handleFetchBatch(ctx context.Context, args [][]string) (
...
-// an LF.
-func (r *runner) handlePushBatch(ctx context.Context, args [][]string) (
-	err error) {
-	repo, fs, err := r.initRepoIfNeeded(ctx, gitCmdPush)
+// canPushAll returns true if a) the KBFS repo is currently empty, and
+// b) we've been asked to push all the local references (i.e.,
+// --all/--mirror).
+func (r *runner) canPushAll(
+	ctx context.Context, repo *gogit.Repository, args [][]string) (
+	canPushAll, repoEmpty bool, err error) {
```
songgao: s/repoEmtpy/kbfsRepoEmpty? Same for `pushSome`. I was a bit confused initially.

songgao:
("Crypto & syncing data to Keybase" sounds good to me!)

kbfsgit/runner.go
strib: The argument is only used for wildcard specs, and then it's used to figure out if the destination branch you're interested in matches the spec.  We don't have wildcare specs, so we don't need to worry about it.

songgao: I see; is the argument a branch name, or a remote name?

strib: Branch name.

strib:
Thanks for the reviews @songgao!

It was easy to use the word, either "encrypt" or "sign", based on TLF type.  So I went with "Syncing encrypted data"/"Syncing signed data".

songgao:
NP; thanks for getting them all done!

kbfsgit/runner.go
```diff
@@ -74,6 +75,10 @@ const (
 	// push, we actually fetch from the local repo and write the
 	// objects into the bare repo.
 	localRepoRemoteName = "local"
+
+	// The minimum number of refs needed to initialize a repo with a
+	// packed-refs file.
+	packedRefsThreshDefault = 10
```
jzila: Is this measured or arbitrary?

kbfsgit/runner.go
```diff
@@ -1014,40 +1085,151 @@ func (r *runner) handleFetchBatch(ctx context.Context, args [][]string) (
...
+		}
+
+		if ref.Type() != plumbing.SymbolicReference {
+			r.log.CDebugf(ctx, "Remote has at least one non-symbolic ref: %s",
+				ref.String())
+			return false, false, nil
+		}
+	}
+
+	sources := make(map[string]bool)
```
jzila: Comment here please describing the role of this loop. And in general it'd be nice to have a bit more commenting for blocks especially in longer methods.

kbfsgit/runner.go
```diff
@@ -1014,40 +1085,151 @@ func (r *runner) handleFetchBatch(ctx context.Context, args [][]string) (
...
-// <dst> <why>? lines to indicate success or failure of each pushed
-// ref. The status report output is terminated by a blank line. The
-// option field <why> may be quoted in a C style string if it contains
-// an LF.
-func (r *runner) handlePushBatch(ctx context.Context, args [][]string) (
-	err error) {
-	repo, fs, err := r.initRepoIfNeeded(ctx, gitCmdPush)
+// canPushAll returns true if a) the KBFS repo is currently empty, and
+// b) we've been asked to push all the local references (i.e.,
+// --all/--mirror).
```
jzila: So if you try to `push --mirror` for a non-empty repo, it won't work?

kbfsgit/runner.go
```diff
@@ -1124,6 +1315,66 @@ func (r *runner) handlePushBatch(ctx context.Context, args [][]string) (
...
+		return err
+	}
+	var results map[string]error
+	if canPushAll {
+		err = r.pushAll(ctx, fs)
+		// All refs in the batch get the same error.
+		results = make(map[string]error, len(args))
+		for _, push := range args {
+			if len(push) != 1 {
+				return errors.Errorf("Bad push request: %v", push)
```
jzila: This can't happen, since `canPushAll` would have already failed.

kbfsgit/runner.go
strib: Arbitrary, but #1211 got rid of it anyway, so the point is moot.

kbfsgit/runner.go
strib: It will work, but it won't be able to use the `pushAll` optimization.  It will have to compute the exact set of commits needed, so it doesn't overwrite anything important.

PR #1211, opened 2017-09-29
kbfsgit: lock during repo creation and on packed-refs creation

strib:
We need to take a lock before creating the repo, when generating the repo ID, so there are no conflicts.  Use ._kbfs_config as the lockfile.

Also, any time we're pushing to a repo we think is empty, we need to create a packed-refs file, even if it's only a single branch, so that we can take a lock on the packed-refs file and ensure the repo only gets that initial push once.

Finally, it turns out that calling WaitForBlockFlush() isn't enough when in single-op mode, because the subsequent call to SyncFromServerForTesting won't bother fetching anything from the server when the local journal still has data in it.  So we have to finish a complete single op, flushing all object files, before taking a lock.

Note that I tried for a long time to come up with a good test for concurrent repo creation, but because MDServerMemory always allows locking to go through, this wasn't possible.

Issue: KBFS-2477

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -76,7 +76,8 @@ const (
 	// objects into the bare repo.
 	localRepoRemoteName = "local"
 
-	packedRefsPath = "packed-refs"
+	packedRefsPath     = "packed-refs"
+	packedRefsTempPath = "._packed-refs"
 )
 
 type ctxCommandTagKey int
@@ -1166,7 +1167,7 @@ func (r *runner) canPushAll(
 	return true, true, nil
 }
 
-func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) error {
+func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
 	r.log.CDebugf(ctx, "Pushing the entire local repo")
 	localFS := osfs.New(r.gitDir)
 
@@ -1189,6 +1190,26 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) error {
 		ctx, localFSObjects, fsObjects,
 		"Counting objects", "countobj",
 		fmt.Sprintf("Preparing and %s objects", verb), "pushobj")
+	if err != nil {
+		return err
+	}
+
+	// Hold the packed refs lock file while transferring, so we don't
+	// clash with anyone else trying to push-init this repo.  go-git
+	// takes the same lock while writing packed-refs during a
+	// `Remote.Fetch()` operation (used in `pushSome()` below).
+	lockFile, err := fs.Create(packedRefsTempPath)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		closeErr := lockFile.Close()
+		if closeErr != nil && err == nil {
+			err = closeErr
+		}
+	}()
+
+	err = lockFile.Lock()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


strib:
By the way #1210 should be reviewed before this one; this one depends on it and fixes some things in it.

kbfsgit/runner.go
```diff
@@ -1111,13 +1127,14 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) error {
...
-	_, err = localFS.Stat("packed-refs")
+	_, err = localFS.Stat(packedRefsPath)
 	if os.IsNotExist(err) {
 		return nil
 	} else if err != nil {
 		return err
 	}
-	return r.copyFileWithCount(ctx, localFS, fs, "packed-refs",
+
+	return r.copyFileWithCount(ctx, localFS, fs, packedRefsPath,
```
songgao: I see a `recursiveCopyWithCounts` call on `fsObjects` above, so this only includes the refs, so it should be fairly small right?

songgao:
oops sorry haven't looked at #1210 yet. Will look soon!

kbfsgit/runner.go
strib: Yeah, first we do objects, then the loose refs, and then this is the packed-refs file if one exists.  It would probably less than 1 MB in practice I'd guess.

strib:
> Oops sorry! I made the MDServerMemory always let locking to go through because I though it wouldn't make sense to fail locking if everything is within the same process. Would you prefer to have a way for MDServerMemory to differentiate different "sessions" within the process, or do you think we should do it in docker tests?

Yeah, I remember agreeing it was ok to do it that way.  I would like to be able to differentiate sessions for locking, it would be easier and give more control than we'd get in the docker tests.  But it's not urgent.

songgao:
Sounds good; will do!

kbfsgit/runner.go
songgao: I see; thanks!

songgao:
> Also, any time we're pushing to a repo we think is empty, we need to create a packed-refs file, even if it's only a single branch, so that we can take a lock on the packed-refs file and ensure the repo only gets that initial push once.

Could you elaborate a bit on why we need to ensure this "initial push" happens only once? What makes this push special that conflict will be bad? (I guess I just don't know enough about pack-file and refs.) Thanks!

strib:
> Could you elaborate a bit on why we need to ensure this "initial push" happens only once? What makes this push special that conflict will be bad? 

If two users think they both successfully pushed a `packed-refs` file, one of them will lose in conflict resolution, and will get their file renamed to ".conflict..." and git will never look at it again.  So even though their push was successful, their data is lost.

If one user pushes a `packed-refs` file, and the other pushes a regular, "loose" ref, that loose ref will override whatever is in the `packed-refs` file, even if the user didn't use "--force".

libgit/repo.go
```diff
@@ -23,6 +23,7 @@ import (
 const (
 	kbfsRepoDir         = ".kbfs_git"
 	kbfsConfigName      = "kbfs_config"
+	kbfsConfigNameTemp  = "._kbfs_config"
```
jzila: I'm not sure this naming is a good idea; that's the format for macOS xattr metadata files on unsupported filesystems.

strib: Good point, but I don't think it matters.  Even if macOS has already created it, it means the kbfs_config file already exists, and so the TLF creation would have already happened.

That said, it would probably be better to name it something else, but now that there are testers out there already using this fliename, I don't think changing it is a good idea unless absolutely necessary.

PR #1212, opened 2017-09-29
kbfsgit: no lazy creates for team or single-user private TLFs

strib:
Issue: KBFS-2480

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -251,8 +251,14 @@ func (r *runner) initRepoIfNeeded(ctx context.Context, forCmd string) (
 		}()
 	}
 
-	fs, _, err = libgit.GetOrCreateRepoAndID(
-		ctx, r.config, r.h, r.repo, r.uniqID)
+	// Only allow lazy creates for public and multi-user TLFs.
+	if r.h.Type() == tlf.Public || len(r.h.ResolvedWriters()) > 1 {
+		fs, _, err = libgit.GetOrCreateRepoAndID(
+			ctx, r.config, r.h, r.repo, r.uniqID)
+	} else {
+		fs, _, err = libgit.GetRepoAndID(
+			ctx, r.config, r.h, r.repo, r.uniqID)
+	}
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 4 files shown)


strib:
> So the only way to create a new repo now is through GUI or keybase git right?

Yes, for your personal private repo and for team TLFs.

songgao:
Got you. Thanks! SGTM. (Yeah meant to ask for personal private repo and team TLFs)

kbfsgit/runner.go
```diff
@@ -251,8 +251,14 @@ func (r *runner) initRepoIfNeeded(ctx context.Context, forCmd string) (
 		}()
 	}
 
-	fs, _, err = libgit.GetOrCreateRepoAndID(
-		ctx, r.config, r.h, r.repo, r.uniqID)
+	// Only allow lazy creates for public and multi-user TLFs.
```
jzila: Is that because we don't provide a UI for these?

kbfsgit/runner.go
```diff
@@ -251,8 +251,14 @@ func (r *runner) initRepoIfNeeded(ctx context.Context, forCmd string) (
...
+	if r.h.Type() == tlf.Public || len(r.h.ResolvedWriters()) > 1 {
```
jzila: What about unresolved writers? I should be able to make a git repository with `jzila,foo@twitter`.

libgit/repo.go
```diff
@@ -238,7 +271,7 @@ func getOrCreateRepoAndID(
 		return nil, NullID, err
 	}
 
-	if createOnly {
+	if op == createOnly {
```
jzila: How are we locking this?

kbfsgit/runner.go
strib: Yeah, there would be no other way to create public or mutli-user repos otherwise.

kbfsgit/runner.go
strib: Good point, will fix.

libgit/repo.go
strib: I don't think what "this" means here.  If we get to this line of code, we know for sure the repo existed at the time the request started.  That's enough to fail the request.  If this is being executed concurrently with a delete of the same repo, oh well, the user will just have to retry.  Does that make sense?

jzila: Yep, thanks.

PR #1214, opened 2017-09-29
git-remote-keybase: fix windows dup handle for git bash shell

strib:
Bash shells on Windows seem to behave slightly different from the
regular windows shell, in terms of duplicated file handles.  Giving
the new handle the same access as the old handle seems to work well in
both environments.

Issue: KBFS-2481

### DIFF
--- a/kbfsgit/git-remote-keybase/dup_stderr_windows.go
+++ b/kbfsgit/git-remote-keybase/dup_stderr_windows.go
@@ -17,7 +17,8 @@ func dupStderr() (*os.File, error) {
 
 	var dupStderrFd windows.Handle
 	err = windows.DuplicateHandle(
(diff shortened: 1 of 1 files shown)


songgao:
I don't know enough about this; but if it works I don't see a reason to block merging for Monday buidl. For longer term though, does this look good to @zanderz ?

zanderz:
I don't really know about this but will point out that "git bash" and "windows bash" usually refer to very different things nowadays. "git bash" is a collection of workalike utilities that all execute as windows programs. "windows bash" is a whole linux subsystem that runs linux binaries. I don't think our helper works under windows bash, which we may want to think about someday. (or if it does, it likely doesn't know about the currently configured Windows Keybase user, I haven't tried it.)

songgao:
Thanks Steve. Windows Bash sounds exciting!

PR #1215, opened 2017-09-30
folder_block_manager: turn off QR in single op mode

strib:
Otherwise an ill-timed QR MD update can break the journal, which
assumes there will only be no more MD updates once the single op is
marked as finished.


strib:
Yeah, any `kbfsfuse` process of any user in the folder would eventually reclaim the blocks.  This should have no negative effects.  Even without this change, `git-remote-keybase` wouldn't usually last long enough to clean stuff up (blocks don't get reclaimed until they are a minute old).

songgao:
I see thanks!

PR #1216, opened 2017-09-30
print "Synching data to Keybase ..." in push

songgao:
This PR adds status message in the last step of `git push` where we are flushing data to the server. When pushing a commit that adds a 64MB file, it looks like this now:

```
$ git push origin master
Initializing Keybase... done.
Syncing with Keybase... done.
Counting and decrypting: 3 objects... done.
Preparing and encrypting: (100.00%) 3/3 objects... done.
Indexing hashes: (100.00%) 3/3 objects... done.
Indexing CRCs: (100.00%) 3/3 objects... done.
Indexing offsets: (100.00%) 3/3 objects... done.
Syncing data to Keybase: (100.00%) 63.25/63.25 MB... done.
To keybase://private/songgao,t_alice/git-test
   855f70d..9c3108c  master -> master
```

Note that it shows a slightly lower size than the actual data size, because journal flush happens before we start printing status. However to fix this we'd need a bigger refactor (and potentially design) which will need to happen after launch.

@strib It turns out the reason why I had multiple "Syncing data to Keybase ..." earlier was `Unlock` (or `Close`) was called multiple times on non-ref files (even when lock was not grabbed which I forgot), and I didn't do the event filtering properly. So now we only print journal status for `FSEventLock`.

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -626,6 +626,7 @@ func (r *runner) processGogitStatus(ctx context.Context,
 	var startTime time.Time
 	lastByteCount := 0
 	cpuProf := ""
+	donePrinted := true
 	for {
 		if fsEvents == nil && statusChan == nil {
 			// Both channels are closed.
@@ -647,8 +648,9 @@ func (r *runner) processGogitStatus(ctx context.Context,
 					// printing "done." for this status here, but print it
 					// along with a newline when FSEventLock event is received
 					// in the case below.
-					if currStage != plumbing.StatusIndexOffset {
+					if currStage != plumbing.StatusIndexOffset && !donePrinted {
 						r.errput.Write([]byte("done." + elapsedStr + "\n"))
+						donePrinted = true
 					}
 				}
 				if r.verbosity >= 4 {
@@ -664,6 +666,7 @@ func (r *runner) processGogitStatus(ctx context.Context,
 					}
 				}
 				r.errput.Write([]byte(gogitStagesToStatus[update.Stage]))
+				donePrinted = false
 				lastByteCount = 0
 				currStage = update.Stage
 				startTime = r.config.Clock().Now()
@@ -706,7 +709,10 @@ func (r *runner) processGogitStatus(ctx context.Context,
 					// event happens before IndexOffset is done, so we print
 					// "done." along with a newline here and start waiting for
 					// journal.
-					fmt.Fprintf(r.errput, "done.\n")
+					if !donePrinted {
+						fmt.Fprintf(r.errput, "done.\n")
+						donePrinted = true
+					}
 					r.printJournalStatusUntilFlushed(ctx, fsEvent.Done)
 				}
 			}
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -627,8 +627,16 @@ func (r *runner) processGogitStatus(ctx context.Context,
 	lastByteCount := 0
 	cpuProf := ""
 	for {
+		if fsEvents == nil && statusChan == nil {
+			// Both channels are closed.
+			break
+		}
 		select {
-		case update := <-statusChan:
+		case update, ok := <-statusChan:
+			if !ok {
+				statusChan = nil
+				continue
+			}
 			if update.Stage != currStage {
 				if currStage != plumbing.StatusUnknown {
 					memProf := fmt.Sprintf("mem.%d.prof", currStage)
@@ -686,7 +694,11 @@ func (r *runner) processGogitStatus(ctx context.Context,
 			}
 
 			currStage = update.Stage
-		case fsEvent := <-fsEvents:
+		case fsEvent, ok := <-fsEvents:
+			if !ok {
+				fsEvents = nil
+				continue
+			}
 			switch fsEvent.EventType {
 			case libfs.FSEventLock:
 				if currStage == plumbing.StatusIndexOffset {
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -344,6 +344,7 @@ func humanizeBytes(n int64, d int64) string {
 	return fmt.Sprintf("%.2f/%.2f GB", float64(n)/gbf, float64(d)/gbf)
 }
 
+// caller should make sure doneCh is closed when journal is all flushed.
 func (r *runner) printJournalStatus(
 	ctx context.Context, jServer *libkbfs.JournalServer, tlf tlf.ID,
 	doneCh <-chan struct{}) {
@@ -376,7 +377,7 @@ func (r *runner) printJournalStatus(
 		r.errput.Write([]byte(str))
 	}
 
-	ticker := time.NewTicker(1 * time.Millisecond)
+	ticker := time.NewTicker(1 * time.Second)
 	defer ticker.Stop()
 	for {
 		select {
@@ -395,14 +396,6 @@ func (r *runner) printJournalStatus(
 					humanizeBytes(flushed, firstStatus.UnflushedBytes))
 				lastByteCount = len(str)
 				r.errput.Write([]byte(eraseStr + str))
-			}
-
-			if status.UnflushedBytes == 0 {
-				elapsedStr := r.getElapsedStr(ctx, startTime, "mem.flush.prof", "")
-				if r.verbosity >= 1 {
-					r.errput.Write([]byte("done." + elapsedStr + "\n"))
-				}
-				return
 			}
 		case <-doneCh:
 			if r.verbosity >= 1 && r.progress {
@@ -620,6 +613,8 @@ func (r *runner) printJournalStatusUntilFlushed(
 		ctx, jServer, rootNode.GetFolderBranch().Tlf, doneCh)
 }
 
+const unlockPrintBytesStatusThreshold = time.Second / 2
+
 func (r *runner) processGogitStatus(ctx context.Context,
 	statusChan <-chan plumbing.StatusUpdate, fsEvents <-chan libfs.FSEvent) {
 	currStage := plumbing.StatusUnknown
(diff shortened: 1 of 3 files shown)


songgao:
Oops forgot to push to the billy repo. Will do soon

songgao:
Actually there's no billy changes. NVM!

songgao:
Pushed another commit to fix multi-branch push. So instead of:

```
$ git push origin master b
Initializing Keybase... done.
Syncing with Keybase... done.
Counting and decrypting: 6 objects... done.
Preparing and encrypting: (100.00%) 6/6 objects... done.
Indexing hashes: (100.00%) 6/6 objects... done.
Indexing CRCs: (100.00%) 6/6 objects... done.
Indexing offsets: (100.00%) 6/6 objects... done.
Syncing data to Keybase: (100.00%) 16.57/16.57 MB... done.
done.
Syncing data to Keybase: (100.00%) 3.33/3.33 KB... done.
To keybase://private/songgao,t_alice/git-test
   ab0f32d..e360bc0  master -> master
 * [new branch]      b -> b
```

Now we have:

```
$ git push origin master b
Initializing Keybase... done.
Syncing with Keybase... done.
Counting and decrypting: 6 objects... done.
Preparing and encrypting: (100.00%) 6/6 objects... done.
Indexing hashes: (100.00%) 6/6 objects... done.
Indexing CRCs: (100.00%) 6/6 objects... done.
Indexing offsets: (100.00%) 6/6 objects... done.
Syncing data to Keybase: (100.00%) 2.03/2.03 MB... done.
Syncing data to Keybase: (100.00%) 3.33/3.33 KB... done.
To keybase://private/songgao,t_alice/git-test
   e360bc0..7cb141b  master -> master
   f74c94b..06e9b9f  b -> b
```

kbfsgit/runner.go
```diff
@@ -376,30 +376,47 @@ func (r *runner) printJournalStatus(
...
+				eraseStr := strings.Repeat("\b", lastByteCount)
+				flushed := firstStatus.UnflushedBytes - status.UnflushedBytes
+				str := fmt.Sprintf(
+					bytesFmt, percent(flushed, firstStatus.UnflushedBytes),
+					humanizeBytes(flushed, firstStatus.UnflushedBytes))
+				lastByteCount = len(str)
+				r.errput.Write([]byte(eraseStr + str))
+			}
+
+			if status.UnflushedBytes == 0 {
```
strib: Do we still need this block of code?  Seems like we can assume `doneCh` will be closed.

kbfsgit/runner.go
```diff
@@ -579,62 +596,127 @@ func humanizeObjects(n int, d int) string {
...
-	for update := range statusChan {
-		if update.Stage != currStage {
-			if currStage != plumbing.StatusUnknown {
-				memProf := fmt.Sprintf("mem.%d.prof", currStage)
-				elapsedStr := r.getElapsedStr(ctx, startTime, memProf, cpuProf)
-				r.errput.Write([]byte("done." + elapsedStr + "\n"))
+	donePrinted := true
+	for {
+		if fsEvents == nil && statusChan == nil {
+			// Both channels are closed.
```
strib: `nil` is different from closed, right?

kbfsgit/runner.go
```diff
@@ -579,62 +596,127 @@ func humanizeObjects(n int, d int) string {
...
+					elapsedStr := r.getElapsedStr(ctx, startTime, memProf, cpuProf)
+					// go-git seems to grab the lock during this phase. So we'd
+					// be flushing the journal before
+					// plumbing.StatusIndexOffset is done. So just skip
+					// printing "done." for this status here, but print it
+					// along with a newline when FSEventLock event is received
+					// in the case below.
+					if currStage != plumbing.StatusIndexOffset && !donePrinted {
+						r.errput.Write([]byte("done." + elapsedStr + "\n"))
+						donePrinted = true
```
strib: This is immediately set to `false` right below, without being check in between.  Seems like it's not really doing anything...

kbfsgit/runner.go
```diff
@@ -579,62 +596,127 @@ func humanizeObjects(n int, d int) string {
...
 
-		currStage = update.Stage
+			currStage = update.Stage
+		case fsEvent, ok := <-fsEvents:
+			if !ok {
+				fsEvents = nil
+				continue
+			}
+			switch fsEvent.EventType {
+			case libfs.FSEventLock:
```
strib: Isn't there data syncing during the Unlock event? It should be syncing the ref file and the parent dir blocks.  If you're worried about lots of `Close()` events triggering this, just move the `FSEvent` notification in `File` until after the lock is checked.

kbfsgit/runner.go
```diff
@@ -579,62 +596,127 @@ func humanizeObjects(n int, d int) string {
...
-					r.log.CDebugf(
-						ctx, "Couldn't create CPU profile: %s", cpuProf)
-					cpuProf = ""
-				} else {
-					pprof.StartCPUProfile(f)
+			if update.Stage != currStage {
+				if currStage != plumbing.StatusUnknown {
+					memProf := fmt.Sprintf("mem.%d.prof", currStage)
+					elapsedStr := r.getElapsedStr(ctx, startTime, memProf, cpuProf)
+					// go-git seems to grab the lock during this phase. So we'd
```
strib: It actually grabs the lock _after_ this phase, but before it sends the Done status update.

Do you need to hard-code the phase?  It seems like just `!donePrinted` would be enough, and more future-proof.

kbfsgit/runner.go
```diff
@@ -376,30 +376,47 @@ func (r *runner) printJournalStatus(
 		r.errput.Write([]byte(str))
 	}
 
-	ticker := time.NewTicker(1 * time.Second)
+	ticker := time.NewTicker(1 * time.Millisecond)
```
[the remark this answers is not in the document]
songgao: Oops! That was for testing. Good catch!

kbfsgit/runner.go
songgao: Hmmm this is also used for cloning, which passes in a nil `doneCh`. Or are you suggesting to change that part to use channel as well?

kbfsgit/runner.go
songgao: They are set to `nil` when closed (in each case below). I guess I should add some more comments.

kbfsgit/runner.go
songgao: Hmm good point. `donePrinted` was added later and I didn't realize it could replace the hard coded status. I'll update that as well as comment. Thanks!

kbfsgit/runner.go
songgao: Yes in this case it's fine to leave this line out. I just wanted to make it clear to set the flag every time. But I'm fine with removing this as well; will do!

kbfsgit/runner.go
songgao: That means we'd print two lines per pushed branch/ref now, right? Since `Lock` is where most of flushing (data blocks) happens, I figured it would be better to only do it for `Lock`. It would miss a few bytes being flushed, but we'd still miss some bytes at beginning even we print status for `Unlock` as well. WDYT?

kbfsgit/runner.go
strib: I don't see anywhere that passes in nil to printJournalStatus. Certainly not handleClone, which is copying data out of KBFS and into the local filesystem. What are you referring to?

kbfsgit/runner.go
strib: Oh I see, I thought the comment was referring to the channels as they were passed in.

kbfsgit/runner.go
strib: I just don't want any pauses in the output.  If the `Unlock()` flush happens to hit a disconnected or really slow network, I want the user to have some idea about why things are slow. If the second line really bothers you, maybe you can do something where the Unlock event is delayed by a second, and if `doneCh` is closed by the time the event arrives than nothing needs to be printed?  That should cover 99% of cases, since `Unlock` doesn't push much data.

kbfsgit/runner.go
songgao: Oh sorry! I was confusing this with the added `fsEvents` in [printJournalStatusUntilFlushed ](https://github.com/keybase/kbfs/pull/1216/files#diff-b6999c82a099222a06e9532d0ff7e054R624). I'm implement your suggestion then. Thanks

kbfsgit/runner.go
songgao: The delay approach sounds great. Will do!

songgao: I ended up using the delay approach (with .5s instead of 1s) for both Lock and Unlock, so taht for multi-ref pushes on fast networks it only prints one line of bytes status. Let me know if you think it should only be for Lock!

kbfsgit/runner.go
```diff
@@ -629,7 +624,10 @@ func (r *runner) processGogitStatus(ctx context.Context,
 	donePrinted := true
 	for {
 		if fsEvents == nil && statusChan == nil {
-			// Both channels are closed.
+			// statusChan is passed in as nil. So if both of them are nil, then
+			// they have both been closed in the select/case below, because
+			// receive failed. So instead of let select block forever, we break
```
strib: "let" -> "letting"

kbfsgit/runner.go
```diff
@@ -703,17 +703,22 @@ func (r *runner) processGogitStatus(ctx context.Context,
...
+					fmt.Fprintf(r.errput, "done.\n")
+					donePrinted = true
+				}
+				// Since we flush all blocks in Lock, subsequent calls to
+				// Lock/Unlock normally don't take much time. So we only print
+				// journal status if it's been longer than
+				// unlockPrintBytesStatusThreshold and fsEvent.Done hasn't been
+				// closed.
+				timer := time.NewTimer(unlockPrintBytesStatusThreshold)
+				select {
```
strib: Might be nice to have a case for `<-ctx.Done` here, just in case (the done ch should be closed if that happens, but you never know...).

songgao: Sure; will push a separate PR. Thanks!

PR #1217, opened 2017-10-01
no more libkb G references, it's going away

maxtaco:
(let's see if this passes tests...)


maxtaco:
Ah cool, it wasn't needed!

maxtaco:
ping @jzila

PR #1221, opened 2017-10-02
fix printing

songgao:
It took me a while but I think this works now!

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -360,12 +360,29 @@ func humanizeBytes(n int64, d int64) string {
 	return fmt.Sprintf("%.2f/%.2f GB", float64(n)/gbf, float64(d)/gbf)
 }
 
+// printStageEndIfNeeded should only be used to end stages started with
+// printStageStart.
+func (r *runner) printStageEndIfNeeded(ctx context.Context) {
+	r.printStageLock.Lock()
+	defer r.printStageLock.Unlock()
+	// go-git grabs the lock right after plumbing.StatusIndexOffset, but before
+	// sending the Done status update. As a result, it would look like we are
+	// flushing the journal before plumbing.StatusIndexOffset is done. So
+	// instead, print "done." only if it's not printed yet.
+	if r.needPrintDone {
+		elapsedStr := r.getElapsedStr(ctx,
+			r.stageStartTime, r.stageMemProfName, r.stageCPUProfPath)
+		r.errput.Write([]byte("done." + elapsedStr + "\n"))
+		r.needPrintDone = false
+	}
+}
+
 func (r *runner) printStageStart(ctx context.Context,
 	toPrint []byte, memProfName, cpuProfName string) {
 	if len(toPrint) == 0 {
 		return
 	}
-	r.printStageEndMaybe(ctx)
+	r.printStageEndIfNeeded(ctx)
 
 	r.printStageLock.Lock()
 	defer r.printStageLock.Unlock()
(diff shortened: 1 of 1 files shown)


songgao:
Oh sorry for lacking context. See #kbfs on keybase chat where I had steps to reproduce. It's to fix missing "done." when pushing from a branch that's behind master:

```
Indexing offsets: (100.00%) 10/10 objects... Syncing encrypted data to Keybase: (100.00%) 72.33/72.33 MB... done.
```

It fixes it by having the `donePrinted` moved to `runner` struct (and reversing the logic making it `needPrintDone`) so that it works across functions.

songgao:
Yeah; agreed we'd need a refactor. I also wanted to make the journal status checking parallel. Will have to do it later!

kbfsgit/runner.go
```diff
@@ -629,6 +633,28 @@ func (r *runner) printJournalStatusUntilFlushed(
 
 const unlockPrintBytesStatusThreshold = time.Second / 2
 
+func (r *runner) printDoneIfNeeded(elapsedStr string) {
```
strib: Please move these two functions above where they are first used.

kbfsgit/runner.go
```diff
@@ -629,6 +633,28 @@ func (r *runner) printJournalStatusUntilFlushed(
...
+	// sending the Done status update. As a result, it would look like we are
+	// flushing the journal before plumbing.StatusIndexOffset is done. So
+	// instead, print "done." only if it's not printed yet.
+	if r.needPrintDone {
+		r.errput.Write([]byte("done." + elapsedStr + "\n"))
+		r.needPrintDone = false
+	}
+}
+
+func (r *runner) needToPrintDone(toPrint []byte) {
```
strib: This function name is confusing, because 1) it sort of sounds like a question, rather than a command, and 2) it prints a string in addition to setting the `needPrintDone` bool.

How about `printStageStart`?  (And maybe `printStageEnd` for the function above, for the sake of consistency?)

kbfsgit/runner.go
```diff
@@ -358,6 +361,7 @@ func humanizeBytes(n int64, d int64) string {
 func (r *runner) printJournalStatus(
 	ctx context.Context, jServer *libkbfs.JournalServer, tlfID tlf.ID,
 	doneCh <-chan struct{}) {
+	r.printDoneIfNeeded("")
```
strib: I guess this means we won't get timing info or other profiles for the final index phase, when "-v" is passed to the git command one or more times?

songgao:
@strib I think in this case it's because `processGogitStatus` just never returns because it needs to keep listening on the `fsEvents`. In context of `pushSome`, it's blocked [here](https://github.com/keybase/kbfs/blob/308d1e33402bbb959eb6773f1de5224e0462d20a/kbfsgit/runner.go#L1310). So as a result, after `pushSome` returns, `waitForJournal` is called [here](https://github.com/keybase/kbfs/blob/308d1e33402bbb959eb6773f1de5224e0462d20a/kbfsgit/runner.go#L1403), which then calls `printJournalStatus` and prints "Syncing to Keybase ..." line, but "done." was never printed. This is why the r.printDoneIfNeeded("") was needed at the beginning of `printJournalStatus`.

kbfsgit/runner.go
songgao: Hmmm good question. I'll need to test it.

kbfsgit/runner.go
songgao: Sounds good!

strib:
Oh right, `Lock` is never called because go-git can figure out it's not a ff ref update before it even takes the lock.  For what it's worth, we could probably also fix that by moving [this code](https://github.com/keybase/go-git/blob/9dbe76bd6a72dfed002d5d3d7bd98d61db4437f7/remote.go#L316-L320) up above the `r.updateLocalReferenceStorage` call, so that `StatusDone` is always returned.  But this fix is fine too, and cleans things up a little bit as well.

songgao:
> so that StatusDone is always returned

But the `fsEvents` is still not closed, so `processGogitStatus` would still block, right?

strib:
```go
			case plumbing.StatusDone:
				r.log.CDebugf(ctx, "Status processing done")
				return
```

songgao:
OH I totally missed that. Thanks!

kbfsgit/runner.go
songgao: Yeah looks like we'd need to move `startTime`, `profName`, and `cpuProfFullPath` into the struct as well. Otherwise we are not writing memprof, or stopping CPU profiler. Do you want me to do that in this PR, or shall we fix it in go-git as you suggested?

strib: I think moving them into the struct makes sense, if you don't mind. Thanks!

songgao: OK then; will do!

songgao:
@strib PTAL thanks!

kbfsgit/runner.go
```diff
@@ -357,11 +360,59 @@ func humanizeBytes(n int64, d int64) string {
...
+		r.stageCPUProfPath = cpuProfPath
+	}
+
+	r.errput.Write(toPrint)
+	r.needPrintDone = true
+}
+
+// printStageEndMaybe should only be used to end stages started with
+// printStageStart.
+func (r *runner) printStageEndMaybe(ctx context.Context) {
```
strib: Please move up above its first use.

strib: Also, how about `printStageEndIfNeeded()`?  That's a little more clear to me.

songgao: Hmm I think it's already above every use of it?

songgao: Yup this is above `printStageStart` :)

strib: Am I going crazy?
`printStageStart`, line 363: https://github.com/keybase/kbfs/pull/1221/files/c11d716a6163d1267380e6bd992876559905e6d5#diff-b6999c82a099222a06e9532d0ff7e054R363
`printStageEndMaybe`, line 396: https://github.com/keybase/kbfs/pull/1221/files/c11d716a6163d1267380e6bd992876559905e6d5#diff-b6999c82a099222a06e9532d0ff7e054R396

strib: Oh, looks like your last "better name" commit moved it.

songgao: No I moved it before the `better name`! Hmm unless I forgot to save before commit "review feedback"?

songgao: Ugh looks like so. Sorry!

PR #1222, opened 2017-10-02
vendor: update go-git, so we don't over-transfer on a push

strib:
If the user is doing a push to a branch in the KBFS repo, and the KBFS repo has extra commits not known to the local repo (i.e.., the local user has forked), then go-git was causing way too much data to be transferred because it doesn't fully negotiate the common commits between the two repos.  It simply sends the commits representing all references known by the KBFS repo; the local side wouldn't recognize the new commit from the KBFS branch, and would therefore conclude the user needs to push all commits in its history.  (Note: if the KBFS repo has multiple branches, and at least one of which has a reference with a common commit, then the entire history isn't sent, only the commits up to that common commit, which is still wasteful.)

The right way to improve this is to implement the multi-round git ack protocol (https://github.com/git/git/blob/master/Documentation/technical/pack-protocol.txt), which keeps iterating until the two sides agree on a set of common commits.  Unfortunately, go-git doesn't implement this, and it looks like it would be a ton of work to implement it right.

Instead, do something quick-and-dirty in go-git.  Instead of sending just the HEAD commit for each KBFS references, send the most recent 100 commits.  Hopefully, the local side will know at least one of these, and can prune the set of pushed objects appropriately.  As an optimization, we don't need to send the full 100 if we already know for sure that the local side has the reference.  In practice, this shouldn't add much latency or extra communication overhead, especially in the common case when most references match between the two repos.

In the case where the local branch is ahead of the KBFS branch (i.e., normally the case for a push), this also shouldn't cause much extra data to be sent, because almost certainly the local repo would have a "remotes/origin" reference for the branch it was based on, so the commit walker would be able to stop early once it runs into that common commit.

I filed KBFS-2487 to track the more correct approach of the multi-ack protocol.  Hopefully we can get to that eventually.

Issue: KBFS-2486

### DIFF
--- a/vendor/gopkg.in/src-d/go-git.v4/remote.go
+++ b/vendor/gopkg.in/src-d/go-git.v4/remote.go
@@ -292,7 +292,7 @@ func (r *Remote) fetch(ctx context.Context, o *FetchOptions) (storer.ReferenceSt
 
 	req.Wants, err = getWants(r.s, refs)
 	if len(req.Wants) > 0 {
-		req.Haves, err = getHaves(localRefs)
+		req.Haves, err = getHaves(localRefs, remoteRefs, r.s)
 		if err != nil {
 			return nil, err
 		}
@@ -511,8 +511,29 @@ func (r *Remote) references() ([]*plumbing.Reference, error) {
 	return localRefs, nil
 }
 
-func getHaves(localRefs []*plumbing.Reference) ([]plumbing.Hash, error) {
+func getHaves(
+	localRefs []*plumbing.Reference,
+	remoteRefs storer.ReferenceStorer,
+	s storage.Storer,
+) ([]plumbing.Hash, error) {
 	haves := map[plumbing.Hash]bool{}
+
+	// Build a map of all the remote references, to avoid loading too
+	// many parent commits for references we know don't need to be
+	// transferred.
+	stopEarlyRefs := map[plumbing.Hash]bool{}
+	iter, err := remoteRefs.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+	err = iter.ForEach(func(ref *plumbing.Reference) error {
+		stopEarlyRefs[ref.Hash()] = true
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
 	for _, ref := range localRefs {
 		if haves[ref.Hash()] == true {
 			continue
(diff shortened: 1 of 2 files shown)


vendor/gopkg.in/src-d/go-git.v4/remote.go
```diff
@@ -522,7 +544,36 @@ func getHaves(localRefs []*plumbing.Reference) ([]plumbing.Hash, error) {
...
+		if err != nil {
+			haves[ref.Hash()] = true
+			continue
+		}
+
+		// Until go-git supports proper commit negotiation during an
+		// upload pack request, include up to 100 commits from the
+		// history of each ref.
+		walker := object.NewCommitPreorderIter(commit, haves, nil)
+		const maxToVisit = 100
```
jzila: What if there isn't a common commit in the last 100? Does it degrade to transferring all the refs?

vendor/gopkg.in/src-d/go-git.v4/remote.go
```diff
@@ -511,8 +512,29 @@ func (r *Remote) references() ([]*plumbing.Reference, error) {
...
+	// Build a map of all the remote references, to avoid loading too
+	// many parent commits for references we know don't need to be
+	// transferred.
+	stopEarlyRefs := map[plumbing.Hash]bool{}
+	iter, err := remoteRefs.IterReferences()
+	if err != nil {
+		return nil, err
+	}
+	err = iter.ForEach(func(ref *plumbing.Reference) error {
+		stopEarlyRefs[ref.Hash()] = true
```
jzila: Can we cache this? Seems like we'll be rebuilding it on every push.

vendor/gopkg.in/src-d/go-git.v4/remote.go
strib: Yep, exactly.

vendor/gopkg.in/src-d/go-git.v4/remote.go
strib: Every push is a separate invocation of `git-remote-keybase` and `getHaves` should only be called once for each push, so I'm not sure where we'd cache it.  To disk maybe?  But note that the thing we're iterating over here is entirely in memory -- it just represents an answer from earlier when we asked the remote side to list all its references.  This is just a weird iterator type go-git puts in front of that.

jzila: All right let's leave it.

PR #1228, opened 2017-10-04
kbfsgit: allow readers to clone/fetch from repos

strib:
This required a few small changes:

* Ignore go-git's attempt to flush the Config file if it hasn't changed from what's already on disk.
* libfs.FS.MkdirAll ignores write-access errors if the dir being created already exists.

Issue: KBFS-2491

### DIFF
--- a/kbfsgit/config_without_remotes_storer.go
+++ b/kbfsgit/config_without_remotes_storer.go
@@ -34,8 +34,13 @@ func newConfigWithoutRemotesStorer(fs *libfs.FS) (
 		return nil, err
 	}
 	// To figure out if this config has been written already, check if
-	// the "IsBare" bit is already flipped.
-	return &configWithoutRemotesStorer{fsStorer, cfg, cfg.Core.IsBare}, nil
+	// it differs from the zero Core value (probably because the
+	// IsBare bit is flipped).
+	return &configWithoutRemotesStorer{
+		fsStorer,
+		cfg,
+		cfg.Core != gogitcfg.Config{}.Core,
+	}, nil
 }
 
 func (cwrs *configWithoutRemotesStorer) Init() error {
--- a/kbfsgit/config_without_remotes_storer.go
+++ b/kbfsgit/config_without_remotes_storer.go
@@ -19,7 +19,8 @@ import (
 // backslashes in git file URLs.
 type configWithoutRemotesStorer struct {
 	*filesystem.Storage
-	cfg *gogitcfg.Config
+	cfg    *gogitcfg.Config
+	stored bool
 }
 
 func newConfigWithoutRemotesStorer(fs *libfs.FS) (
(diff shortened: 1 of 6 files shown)


kbfsgit/runner.go
```diff
@@ -258,8 +258,9 @@ func (r *runner) initRepoIfNeeded(ctx context.Context, forCmd string) (
 	}
 
 	// Only allow lazy creates for public and multi-user TLFs.
-	if r.h.Type() == tlf.Public ||
-		len(r.h.ResolvedWriters())+len(r.h.UnresolvedWriters()) > 1 {
+	numUsers := len(r.h.ResolvedWriters()) + len(r.h.UnresolvedWriters()) +
+		len(r.h.ResolvedReaders()) + len(r.h.UnresolvedReaders())
```
jzila: It looks to me that an attempt by a reader to clone a nonexistent repo will get a `WriteAccessError`. Is that what we want?

strib: That was my plan, yeah.  Do you have another suggestion?

jzila: Ideally it'd be the same error you get when a repo doesn't exist.

strib: Hmm, not sure I agree.  The repo will show up in the GUI if it exists, so it will be confusing if we say it doesn't exist.

kbfsgit/config_without_remotes_storer.go
```diff
@@ -32,7 +33,9 @@ func newConfigWithoutRemotesStorer(fs *libfs.FS) (
 	if err != nil {
 		return nil, err
 	}
-	return &configWithoutRemotesStorer{fsStorer, cfg}, nil
+	// To figure out if this config has been written already, check if
+	// the "IsBare" bit is already flipped.
+	return &configWithoutRemotesStorer{fsStorer, cfg, cfg.Core.IsBare}, nil
```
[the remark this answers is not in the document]
strib: No, this way is just the first thing I thought of.  If you prefer that way, I'm happy to change it.

kbfsgit/runner.go
jzila: Right but it doesn't exist, and we failed to create it because you're a reader.

strib: It does exist from a Keybase perspective (i.e., `keybase git create` was run), it's just not initialized from a git perspective.

jzila: Oh hmm...so you can `keybase git create` successfully when you're just a reader? That doesn't seem intuitive to me.

strib: No, I don't think you can do that.   I thought you were talking about the scenario where a writer made a repo, then a reader tried the URL before the writer initialized the repo.

jzila: Ah, nope. I'm a reader, and I try to clone a repo that doesn't exist.

strib: Sorry, you said that pretty clearly and I missed it.

I'll see what I can do.

kbfsgit/config_without_remotes_storer.go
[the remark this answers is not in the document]
songgao: Using the whole struct's zero value seemed more future-proof to me, but I don't think it really matters much in this case. So it's your call!

kbfsgit/runner.go
strib: Done: eb42e6a

PR #1236, opened 2017-10-05
allow setting server addrs for the GRH and GitHandler; Dockerfile changes

songgao:


libgit/init.go
```diff
@@ -37,29 +43,44 @@ func Params(kbCtx libkbfs.Context, storageRoot string) (
...
+	// This is set to false in docker tests for now, but we need it. So
+	// override it to true here.
+	params.EnableJournal = true
 	params.EnableDiskCache = false
 	params.StorageRoot = tempDir
 	params.Mode = libkbfs.InitSingleOpString
 	params.TLFJournalBackgroundWorkStatus =
 		libkbfs.TLFJournalSingleOpBackgroundWorkEnabled
+
+	if baddr := os.Getenv(paramKeybaseGitBServerAddr); len(baddr) > 0 {
```
jzila: Why do we need this if we're passing in the params explicitly?

libgit/rpc.go
```diff
@@ -19,25 +19,27 @@ import (
...
+	config         libkbfs.Config
+	kbfsInitParams *libkbfs.InitParams
+	log            logger.Logger
 }
 
 // NewRPCHandlerWithCtx returns a new instance of a Git RPC handler.
-func NewRPCHandlerWithCtx(
-	kbCtx libkbfs.Context, config libkbfs.Config) keybase1.KBFSGitInterface {
+func NewRPCHandlerWithCtx(kbCtx libkbfs.Context, config libkbfs.Config,
+	kbfsInitParams *libkbfs.InitParams) keybase1.KBFSGitInterface {
```
jzila: I think this will break Circle until gobind is updated.

libkbfs/state_checker.go
```diff
@@ -93,13 +93,15 @@ func (sc *StateChecker) getLastGCData(ctx context.Context,
...
 	var latestTime time.Time
 	var latestRev kbfsmd.Revision
-	for _, c := range *config.allKnownConfigsForTesting {
-		ops := c.KBFSOps().(*KBFSOpsStandard).getOps(context.Background(),
-			FolderBranch{tlfID, MasterBranch}, FavoritesOpNoChange)
-		rt, rev := ops.fbm.getLastQRData()
-		if rt.After(latestTime) && rev > latestRev {
-			latestTime = rt
-			latestRev = rev
+	if config.allKnownConfigsForTesting != nil {
```
jzila: I _think_ this is the right answer (because kbfsdocker runs in devel mode), but I'm not sure. @strib?

Jenkinsfile
```diff
@@ -154,11 +154,13 @@ helpers.rootLinuxNode(env, {
                         // Install kbfsfuse first so we can start on dockerizing.
                         sh "go install github.com/keybase/kbfs/kbfsfuse"
                         sh "cp ${env.GOPATH}/bin/kbfsfuse ./kbfsfuse/kbfsfuse"
+                        sh "go install github.com/keybase/kbfs/kbfsgit/git-remote-keybase/git-remote-keybase"
```
jzila: This should be `go install github.com/keybase/kbfs/kbfsgit/git-remote-keybase`

libgit/init.go
songgao: This is for the `git-remote-keybase` binary. The explicitly passed-in params is for the RPC handler that lives inside the `kbfsfuse` binary, which serves git related RPCs with a separate config (`gitConfig`).

libkbfs/state_checker.go
songgao: This actually isn't necessary, I think. I noticed a panic here at first, so I added this check like elsewhere. But it turned out it was because we were using a wrong MDServer; now that it's fixed this should never happen. But I'm fine with either having this or not, so let me know if any of you want to leave it out so that we have an intentional panic if this-should-never-happen happens.

Jenkinsfile
songgao: oooh right!

PR #1237, opened 2017-10-05
folder_update_prepper: update new MD's root ID if needed

strib:
A user in the wild had this happen:
* Device A: Removed a bunch of files
* Device B: Removed a bunch of files
* Device A: Flush to server.
* Device B: Notice conflict, resolve to local journal.
* Device B: Connectivity failure, flush pauses.
* Device A: Perform QR.
* Device B: Re-run conflict resolution on top of QR.
* Device B: Flush new MD to server.

Unfortunately, the new MD pushed in the last step still had the root block pointer from the QR's MD.  Basically, the folder_update_prepper didn't do anything interesting in the final CR because the real conflicts had already been resolved in the previous MD, and as a result, it never updated the root block pointer to what it had previously computed.

To get around this, if we pull in an update from a previous unmerged resolutionOp, update the root block pointer on the new MD if it matches that pointer.

Also introduce a test that reproduces the issue without this fix.

Issue: keybase/client#8832

### DIFF
--- a/libkbfs/folder_update_prepper.go
+++ b/libkbfs/folder_update_prepper.go
@@ -1195,6 +1195,9 @@ func (fup *folderUpdatePrepper) prepUpdateForPaths(ctx context.Context,
 		// most recent block on a chain and haven't yet been involved
 		// in an update during this resolution.  Unreference any
 		// blocks that aren't the most recent blocks on their chains.
+		currMDPtr := md.data.Dir.BlockPointer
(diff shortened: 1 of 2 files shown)


strib:
@songgao: The issue was that the root block pointer listed in the current MD was already archived and deleted, because it was unreferenced.  So reading the folder resulted in a "block does not exist" error, which is translated to an EIO.

PR #1249, opened 2017-10-09
folder_branch_ops: check context during `syncAllUnlocked`

strib:
This fixes a race:

* CR input loop notices new work, creates a context, and starts a goroutine to do the CR.
* A write operation comes in from the user, grabs the mdWriter lock, finishes and calls cr.ForceCancel(), which cancels the CR context we just made.
* The CR goroutine now gets the mdWriter lock, and calls SyncAll, which includes the latest write.
* After the SyncAll finishes, it checks its context, sees that its canceled, and stops further processing.
* If no other writes happen, CR never starts up again.

Issue: KBFS-2505
Issue: keybase/client#8915

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4191,6 +4191,18 @@ func (fbo *folderBranchOps) syncAllUnlocked(
 	ctx context.Context, lState *lockState) error {
 	fbo.mdWriterLock.Lock(lState)
 	defer fbo.mdWriterLock.Unlock(lState)
+
+	select {
+	case <-ctx.Done():
+		// We've already been canceled, possibly because we're a CR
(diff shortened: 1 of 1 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -4191,6 +4191,18 @@ func (fbo *folderBranchOps) syncAllUnlocked(
 	ctx context.Context, lState *lockState) error {
 	fbo.mdWriterLock.Lock(lState)
 	defer fbo.mdWriterLock.Unlock(lState)
+
+	select {
+	case <-ctx.Done():
```
jzila: What if we get canceled after this line?

strib: `ForceCancel` is called under `mdWriterLock`, so we don't have to worry about that.

PR #1250, opened 2017-10-09
jenkinsfile: Re-enable fuse tests since we upgraded the kernel.

jzila:


strib:
I'm having issues with this on my local box as well.  I think this explains it:

https://github.com/golang/go/issues/21014 (via https://github.com/bazil/fuse/issues/183).

Looks like we might need some bazil work to fix this, but @tv42 never responded to that open issue...  I wonder if we can hit this in production or if it's only a test harness thing somehow.

jzila:
Nice find! We should probably escalate.

tv42:
I've been hoping they'd figure out a fix in runtime. It seems like that's not happening. @hanwen's code change is a racy kludge that injects a file into the filesystem namespace, I don't want to do that. This might take same thinking.

tv42:
The good news seem to be: 

- The hang only affects FUSE servers that are their own clients; typically, unit tests. Real apps are unlikely to suffer.
- (After pretty trivial fixes to unit tests,) bazil.org/fuse tests pass at least half a dozen times in a row, before triggering this race and hanging.

jzila:
FWIW this is the hanwen/go-fuse fix: https://github.com/hanwen/go-fuse/commit/4f10e248ebabd3cdf9c0aa3ae58fd15235f82a79

jzila:
@tv42: any luck figuring out a long term solution?

jzila:
Closing for archival.

PR #1253, opened 2017-10-10
Vendor new team suffixes

mlsteele:
I'd like to merge https://github.com/keybase/client/pull/8961 but it would break circle because of the added argument to `MakeTestTeamID` (circle revendors). So I'm hoping to just merge this at the same time and hope circle settles out.


vendor/vendor.json
```diff
@@ -201,10 +201,10 @@
 			"revisionTime": "2017-10-05T01:33:07Z"
 		},
 		{
-			"checksumSHA1": "F/PcyNFC/ILGiB4UME3omC0YrGA=",
+			"checksumSHA1": "ltLZ6MBJL//Z+aFwGo+MBeTjPYk=",
 			"path": "github.com/keybase/client/go/protocol/keybase1",
-			"revision": "d452e21cd5113a0ca80e1eb7fe5291f4a4859c43",
-			"revisionTime": "2017-10-05T01:33:07Z"
+			"revision": "7939031c7c349ca45d71f9b5be48d5427ffa2c74",
```
strib: It would be nice to maintain this to be a real client hash, but I know the circular dependency for circle sucks. Can you update it right after you push?  (No review needed.)

mlsteele: sure thing, thanks

mlsteele: all set

PR #1254, opened 2017-10-10
[KBFS-2513] Allow a bigger leveldb file size for the block db.

jzila:
After writing a 120MiB file:
```
bash-4.3$ cd ~/.local/share/keybase.devel/kbfs_block_cache/v1/diskCacheBlocks.leveldb/
bash-4.3$ ls -lh
total 130292
-rw-r--r--    1 keybase  keybase    50.2M Oct 10 23:43 000077.ldb
-rw-r--r--    1 keybase  keybase     3.5M Oct 10 23:43 000079.ldb
-rw-r--r--    1 keybase  keybase    50.4M Oct 10 23:43 000080.ldb
-rw-r--r--    1 keybase  keybase     3.8M Oct 10 23:43 000082.ldb
-rw-r--r--    1 keybase  keybase    13.2M Oct 10 23:43 000083.ldb
-rw-r--r--    1 keybase  keybase     2.5M Oct 10 23:44 000084.log
-rw-r--r--    1 keybase  keybase     3.6M Oct 10 23:43 000085.ldb
-rw-r--r--    1 keybase  keybase       16 Oct 10 23:37 CURRENT
-rw-r--r--    1 keybase  keybase        0 Oct 10 23:37 LOCK
-rw-r--r--    1 keybase  keybase    16.6K Oct 10 23:43 LOG
-rw-r--r--    1 keybase  keybase     5.6K Oct 10 23:43 MANIFEST-000000
```

### DIFF
--- a/libkbfs/disk_block_cache.go
+++ b/libkbfs/disk_block_cache.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pkg/errors"
 	metrics "github.com/rcrowley/go-metrics"
 	"github.com/syndtr/goleveldb/leveldb"
+	"github.com/syndtr/goleveldb/leveldb/opt"
 	"github.com/syndtr/goleveldb/leveldb/storage"
 	"github.com/syndtr/goleveldb/leveldb/util"
 	"golang.org/x/net/context"
@@ -32,6 +33,7 @@ import (
 const (
 	// 10 GB maximum storage by default
 	defaultDiskBlockCacheMaxBytes uint64 = 10 * (1 << 30)
+	defaultBlockCacheTableSize    int    = 50 * opt.MiB
 	evictionConsiderationFactor   int    = 3
 	defaultNumBlocksToEvict       int    = 10
 	maxEvictionsPerPut            int    = 4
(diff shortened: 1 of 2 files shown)


libkbfs/disk_block_cache.go
```diff
@@ -139,7 +141,9 @@ func newDiskBlockCacheStandardFromStorage(
 			closer()
 		}
 	}()
-	blockDb, err := openLevelDB(blockStorage)
+	blockDbOptions := *leveldbOptions
+	blockDbOptions.CompactionTableSize = defaultBlockCacheTableSize
```
jzila: This variable is eventually plumbed through here, so I _think_ it does what we need.
https://github.com/syndtr/goleveldb/blob/3c5717caf1475fd25964109a0fc640bd150fce43/leveldb/db_compaction.go#L399-L401

I haven't tested it yet though.

libkbfs/disk_block_cache.go
```diff
@@ -32,6 +33,7 @@ import (
 const (
 	// 10 GB maximum storage by default
 	defaultDiskBlockCacheMaxBytes uint64 = 10 * (1 << 30)
+	defaultBlockCacheTableSize    int    = 50 * opt.MiB
```
songgao: For LevelDB, this is an upper bound rather than fixed size right? Any reason not to make it even bigger?

jzila: In the average case for our disk cache, it works out to be a fixed size. I'm not sure yet what would be the consequences of making it bigger, other than definitely decreasing the performance efficiency of levelDB (inserts need to sort keys within that 50MiB, and potentially move out elements into the next "level")

songgao: I see; thanks!

PR #1255, opened 2017-10-10
libgit: introduce deleted-repo cleaner

strib:
Run it after every git operation, rather than periodically, because it would be difficult and expensive to build something that crawls *every* TLF for a user to look for repos to delete.

When a repo is deleted, this schedules a background task to clean up the repo after a set amount of time.  If KBFS is shut down before that scheduled task, any other git operation in that folder will trigger a cleanup.  However, if the cleanup blocks a git operation, it is time-limited, and will only delete as many files as it can within a 2 second timeout, so as not to block foreground operation for too long. Subsequent operations will clean up more data, etc, until the full repo is gone.

Issue: KBFS-2442

### DIFF
--- a/libgit/repo.go
+++ b/libgit/repo.go
@@ -11,7 +11,9 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/protocol/keybase1"
@@ -21,11 +23,12 @@ import (
 )
 
 const (
-	kbfsRepoDir         = ".kbfs_git"
-	kbfsConfigName      = "kbfs_config"
-	kbfsConfigNameTemp  = "._kbfs_config"
-	gitSuffixToIgnore   = ".git"
-	kbfsDeletedReposDir = ".kbfs_deleted_repos"
+	kbfsRepoDir              = ".kbfs_git"
+	kbfsConfigName           = "kbfs_config"
+	kbfsConfigNameTemp       = "._kbfs_config"
+	gitSuffixToIgnore        = ".git"
+	kbfsDeletedReposDir      = ".kbfs_deleted_repos"
+	minDeletedAgeForCleaning = 1 * time.Hour
 )
 
 // This character set is what Github supports in repo names.  It's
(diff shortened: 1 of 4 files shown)


libgit/repo.go
```diff
@@ -21,11 +23,12 @@ import (
...
-	kbfsRepoDir         = ".kbfs_git"
-	kbfsConfigName      = "kbfs_config"
-	kbfsConfigNameTemp  = "._kbfs_config"
-	gitSuffixToIgnore   = ".git"
-	kbfsDeletedReposDir = ".kbfs_deleted_repos"
+	kbfsRepoDir              = ".kbfs_git"
+	kbfsConfigName           = "kbfs_config"
+	kbfsConfigNameTemp       = "._kbfs_config"
+	gitSuffixToIgnore        = ".git"
+	kbfsDeletedReposDir      = ".kbfs_deleted_repos"
```
songgao: I know this was added in a previous PR, but I was just thinking what would  happen if `git push` takes longer than an hour (i.e. `minDeletedAgeForCleaning`), and repo delete happens in between. I suppose git-remote-keybase will find out about conflict and rename the repo folder, and only the original that was moved into `.kbfs_deleted_repos` would be deleted, meaning stuff from the eventually finished super slow `git push` would still stay inside `.kbfs_git` as a renamed folder with a conflict suffix so if user needs they can still find them back. Does it sound correct?

libgit/repo.go
```diff
@@ -44,6 +47,115 @@ func checkValidRepoName(repoName string, config libkbfs.Config) bool {
...
+}
+
+// CleanOldDeletedReposTimeLimited is the same as
+// `CleanOldDeletedRepos`, except it limits the time spent on
+// cleaning, deleting as much data as possible within the given time
+// limit (without returning an error).
+func CleanOldDeletedReposTimeLimited(
+	ctx context.Context, config libkbfs.Config,
+	tlfHandle *libkbfs.TlfHandle) error {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
songgao: Make it a const?

libgit/repo.go
strib: If a repo is deleted (i.e., renamed to the `.kbfs_deleted_repos` directory), CR will apply the push changes to that renamed directory.  No conflict names will be used, and the newly pushed data will get cleaned as normal with the rest of the repo.

If the repo is deleted AND then cleaned up all during the push (that's quite a push!), then CR will recreate the pushed files in their renamed location under `.kbfs_deleted_repos`, also without a conflict name (since there's nothing left to conflict with).  Though we don't have a test for this, so I'm not 100% sure -- it could get recreated under the old `.kbfs_git` repo name.  I can add a test later, but either way it seems super unlikely to be an issue.  If it becomes an issue, we can add a check after CR completes to make sure the repo's config file still exists, and if not, we fail the push so the user sees an error.

songgao: Thanks for the explanation. I was mostly worried about the situation where user sees the push succeeds, but it ends up in `.kbfs_deleted_repos` and later on gets cleaned up because another team member deleted the repo. But I guess in that case since delete happened, it should be expected that it's gone anyway.

PR #1258, opened 2017-10-11
 return libkb.RepoDoesntExistError when deleting a nonexistent repo

oconnor663:
r? @strib

### DIFF
--- a/libgit/repo.go
+++ b/libgit/repo.go
@@ -48,6 +48,18 @@ func checkValidRepoName(repoName string, config libkbfs.Config) bool {
 			repoNameRE.MatchString(repoName))
 }
 
+// For the common "repo doesn't exist" case, use the error type that the client can recognize.
+func castNoSuchNameError(err error, repoName string) error {
+	switch errors.Cause(err).(type) {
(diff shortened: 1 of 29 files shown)


libgit/repo.go
```diff
@@ -493,7 +493,16 @@ func DeleteRepo(
 
 	_, _, err = kbfsOps.Lookup(ctx, repoNode, normalizedRepoName)
 	if err != nil {
-		return err
+		// For the common "repo doesn't exist" case, use an error type that the
+		// client can recognize.
+		switch errors.Cause(err).(type) {
+		case libkbfs.NoSuchNameError:
+			return libkb.RepoDoesntExistError{
```
strib: Also, it might be worth returning this for the above lookup on `kbfsRepoDir` as well, which could happen if someone tries to delete a repo in a TLF that has never had any repos at all.

oconnor663: Sure, I'll add that in for consistency.

oconnor663:
@strib could you eyeball those last two commits and let me know if they look good?

PR #1259, opened 2017-10-11
libgit: always set namespace for returned FS, and refactor a bit

strib:
We weren't correctly setting the repo ID as the lock namespace when returning an FS for a newly-created.  That would only affect the first push of public and multi-user TLFs that are lazily-created.

Also, factor out the taking of the lock config file into a separate function for clarity, and reuse in a later PR.  Verify that the lock namespace is what we expect it to be, to make sure a bad FS isn't passed in by accident.

Issue: KBFS-2439

### DIFF
--- a/libgit/repo.go
+++ b/libgit/repo.go
@@ -5,6 +5,7 @@
 package libgit
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


libgit/repo.go
```diff
@@ -203,6 +205,52 @@ func UpdateRepoMD(ctx context.Context, config libkbfs.Config,
 	return nil
 }
 
+func normalizeRepoName(repoName string) string {
+	return strings.TrimSuffix(strings.ToLower(repoName), gitSuffixToIgnore)
+}
+
+func takeConfigLock(
+	fs *libfs.FS, tlfHandle *libkbfs.TlfHandle, repoName string) (
+	f billy.File, err error) {
```
jzila: I wonder whether you should just return an `io.Closer` here, to make it clear how this is to be used.

PR #1260, opened 2017-10-12
libgit: implement rename repos

strib:
We follow github's logic here:

* Create a symlink from the old repo name to the new name
* However, if someone tries to create a repo using the old name, we
  overwrite the symlink.

Issue: KBFS-2439

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -76,6 +76,20 @@ const (
 	maxSymlinkLevels = 40 // same as Linux
 )
 
+func followSymlink(parentPath, link string) (newPath string, err error) {
+	if path.IsAbs(link) {
+		return "", errors.Errorf("Can't follow absolute link: %s", link)
+	}
+
+	newPath = path.Clean(path.Join(parentPath, link))
+	if strings.HasPrefix(newPath, "..") {
+		return "", errors.Errorf(
+			"Cannot follow symlink out of chroot: %s", newPath)
+	}
+
+	return newPath, nil
+}
+
 // NewFS returns a new FS instance, chroot'd to the given TLF and
 // subdir within that TLF.  `subdir` must exist, and point to a
 // directory, before this function is called.  `uniqID` needs to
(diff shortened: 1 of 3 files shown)


libgit/repo.go
```diff
@@ -545,3 +571,167 @@ func DeleteRepo(
...
+	// Take locks in both repos during rename (same lock that's taken
+	// for new repo creation).
+	oldLockFile, err := takeConfigLock(
+		oldRepoFS.(*libfs.FS), tlfHandle, oldRepoName)
+	if err != nil {
+		return err
+	}
+	defer oldLockFile.Close()
+
+	if normalizedOldRepoName == normalizedNewRepoName {
```
songgao: So this case is for case changes, and `.git` suffix addition/removal?

Related question: should [`normalizeRepoName`](https://github.com/keybase/kbfs/blob/4ed4da1f519db79f8edbce19500a02a7bd3a7327/libgit/repo.go#L295) do the `strings.TrimSuffix` in a loop until the string doesn't change, in case something like `repo.git.git` was passed in?

libgit/repo.go
```diff
@@ -545,3 +571,167 @@ func DeleteRepo(
...
+	_, _, err = kbfsOps.Lookup(ctx, repoNode, normalizedNewRepoName)
+	switch errors.Cause(err).(type) {
+	case libkbfs.NoSuchNameError:
+		// The happy path.
+	case nil:
+		newRepoFS, err := fs.Chroot(normalizedNewRepoName)
+		if err != nil {
+			return err
+		}
+		return makeExistingRepoError(ctx, config, newRepoFS, newRepoName)
```
songgao: What if the new repo location is a symlink, i.e. the location's repo was renamed to something else? It seems we should override the symlink here too, since one of scenarios where user wants to rename repos is they might want to reuse the name for something else.

libgit/repo.go
```diff
@@ -545,3 +571,167 @@ func DeleteRepo(
...
+	default:
+		return err
+	}
+
+	// Make the new repo subdir just so we can take the lock inside
+	// the new repo.  (We'll delete the new dir after the rename.)
```
songgao: Hmmm did you mean "before the rename"?

libgit/repo.go
```diff
@@ -545,3 +571,167 @@ func DeleteRepo(
...
+		return err
+	}
+	defer newLockFile.Close()
+
+	// Rename this new dir out of the way before we rename.
+	fi, err := fs.Stat(normalizedNewRepoName)
+	if err != nil {
+		return err
+	}
+	err = recursiveDelete(ctx, fs, fi)
```
songgao: This would delete what's referenced by `newLockFile` too right? And since it's open already, we can still call `Close()` to unlock it?

libfs/fs.go
```diff
@@ -96,15 +110,41 @@ func NewFS(ctx context.Context, config libkbfs.Config,
...
-		n, ei, err = config.KBFSOps().Lookup(ctx, n, p)
-		if err != nil {
-			return nil, err
-		}
-		if ei.Type != libkbfs.Dir {
-			return nil, errors.Errorf("%s is not a directory",
-				path.Join(parts[:i]...))
+	// Loop while we follow symlinks.
+outer:
+	for {
```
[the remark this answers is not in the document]
strib: I just didn't want to have to modify the `parts` slice and the iteration index, during iteration.  Seems a little harder to get right, and this should be fast enough since everything will be cached after the first iteration.  I could be convinced to try it the other way if you really prefer it, though.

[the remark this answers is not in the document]
songgao: Got you. That's OK; I didn't think through the whole thing for the other approach, so if you think this way it's easier to get right then let's stick with this!

libgit/repo.go
strib: > So this case is for case changes, and .git suffix addition/removal?

Yes.

> Related question: should normalizeRepoName do the strings.TrimSuffix in a loop until the string doesn't change, in case something like repo.git.git was passed in?

Whoa, I didn't realize github did this, but you're right.  I guess it should (in a separate PR).

libgit/repo.go
strib: Yes, it would delete the locked file too, but that's fine because we still have a file handle open to it.  And yes Close should still unlock it.  I don't see another way to lock down the repo's new name before doing the rename.

libgit/repo.go
songgao: > I didn't realize github did this

Hmm I didn't know that!

libgit/repo.go
strib: Oh yeah, I meant to do that and forgot, good catch.  Will work on that now.

libgit/repo.go
songgao: OK sounds good. I was just making sure I understood correctly. Thanks!

PR #1263, opened 2017-10-13
kbfsgit: update go-git to fix ref deletion from packed-refs, and add test

strib:
This change also includes several src-d master fixes.

The go-git PR fixing the packed-refs deletion logic is here (self-reviewed): keybase/go-git#11

Issue: KBFS-2509

### DIFF
--- a/kbfsgit/runner_test.go
+++ b/kbfsgit/runner_test.go
@@ -633,3 +633,58 @@ func TestRunnerReaderClone(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "hello", string(data))
 }
+
+func TestRunnerDeletePackedRef(t *testing.T) {
+	ctx, config, tempdir := initConfigForRunner(t)
+	defer libkbfs.CheckConfigAndShutdown(ctx, t, config)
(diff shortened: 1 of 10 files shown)


kbfsgit/runner_test.go
```diff
@@ -633,3 +633,58 @@ func TestRunnerReaderClone(t *testing.T) {
...
+	dotgit1 := filepath.Join(git1, ".git")
+
+	makeLocalRepoWithOneFile(t, git1, "foo", "hello", "b")
+
+	// Add a different file to master.
+	cmd := exec.Command(
+		"git", "--git-dir", dotgit1, "--work-tree", git1,
+		"checkout", "-b", "master")
+	err = cmd.Run()
+	require.NoError(t, err)
```
jzila: Can you make command wrappers for these git commands? e.g.
```go
func gitExec(t *testing.T, gitDir string, workTree string, command ...string) {
    cmd := exec.Command("git", "--git-dir", gitDir, "--work-tree", workTree, command...)
    err := cmd.Run()
    require.NoError(t, err)
}
```

PR #1264, opened 2017-10-13
[KBFS-2499] Don't sync when there's a full cache

jzila:
Add a test that regresses KBFS-2499, and fix the failure.

As a bonus, we now have a test that tests the prefetcher and disk cache combined.

### DIFF
--- a/libkbfs/prefetcher_test.go
+++ b/libkbfs/prefetcher_test.go
@@ -983,7 +983,7 @@ func TestPrefetcherUnsyncedThenSyncedPrefetch(t *testing.T) {
 }
 
 func TestSyncBlockCacheWithPrefetcher(t *testing.T) {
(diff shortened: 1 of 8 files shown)


libkbfs/prefetcher_test.go
```diff
@@ -967,3 +981,95 @@ func TestPrefetcherUnsyncedThenSyncedPrefetch(t *testing.T) {
 	testPrefetcherCheckGet(t, config.BlockCache(),
 		aa.Children["aab"].BlockPointer, aab, FinishedPrefetch, TransientEntry)
 }
+
+func TestSyncBlockCacheWithPrefetcher(t *testing.T) {
+	t.Log("Test synced TLF prefetching in a more complex fetch order.")
```
strib: Wrong comment?

jzila: Yeah good call, I'll change it.

PR #1268, opened 2017-10-13
dump goroutines if operation takes longer than a minute

songgao:


libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+		fmt.Fprintf(gzipper,
+			"\n======== START Profile: %s ========\n\n", p.Name())
+		_ = p.WriteTo(gzipper, 2)
+		fmt.Fprintf(gzipper,
+			"\n======== END   Profile: %s ========\n\n", p.Name())
+	}
+	gzipper.Close()
+	base64er.Close()
+	d.logger.CDebugf(tracker.ctx,
+		"operation expired. dump>gzip>base64: %q", buf.String())
```
songgao: Compressed so it doesn't overload the logs. This can be easily read with a `pbpaste | base64 -d | gzip -d`

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	limiter *rate.Limiter
+
+	lock          sync.Mutex
+	chronological *ctxTimeTrackerList
+}
+
+const impatientDebugDumperCheckInterval = time.Second
+const impatientDebugDumperDumpMinInterval = time.Second // 1 dump per sec max
+
+func NewImpatientDebugDumper(config Config, dumpIn time.Duration) *ImpatientDebugDumper {
```
strib: Needs comment.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	if c.tail != nil {
+		c.tail.next = newNode
+	}
+	c.tail = newNode
+	if c.front == nil {
+		c.front = newNode
+	}
+}
+
+type ImpatientDebugDumper struct {
```
strib: Needs comment.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	config Config
+	logger logger.Logger
```
strib: Most structs call this just `log`.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	d := &ImpatientDebugDumper{
+		config: config,
+		logger: config.MakeLogger("IGD"),
+		dumpIn: dumpIn,
+		ticker: time.NewTicker(impatientDebugDumperCheckInterval),
+		limiter: rate.NewLimiter(
+			rate.Every(impatientDebugDumperDumpMinInterval), 1),
+		chronological: &ctxTimeTrackerList{},
+	}
+	go d.dumpLoop()
```
strib: Needs a `Shutdown` method to stop this goroutine.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+
+	ticker  *time.Ticker
+	limiter *rate.Limiter
+
+	lock          sync.Mutex
+	chronological *ctxTimeTrackerList
+}
+
+const impatientDebugDumperCheckInterval = time.Second
+const impatientDebugDumperDumpMinInterval = time.Second // 1 dump per sec max
```
strib: Maybe `time.Minute`?

libkbfs/impatient_debug_dumper.go
strib: I think "operation expired" is the wrong phrase, since the actual operation itself is still going.  Maybe "Operation exceeded max wait time"?

strib: Can you put the instructions for reading in a comment?  Or better yet, in the debug message itself?

strib: Also, I don't think "pbpaste" is a thing on Linux...

libkbfs/impatient_debug_dumper.go
songgao: I was worried 1m is too long for people stuck somewhere to wait for. How about 5s?

strib: I was talking about the rate-limiting, not the checking.  So there will still be one dump within a second of the expiration.  Isn't that good enough?  Why would we ever need one dump _per second_?

libkbfs/impatient_debug_dumper.go
songgao: I think I had a shell function that does that, but point taken; will add instructions!

libkbfs/impatient_debug_dumper.go
songgao: Yes rate-limiting. I was thinking about the first ever dump which could be dropped if user restarts the app. This isn't one per second though; it's at most one per second. But I suppose I can change it into a ticker so there's always a tick there for the first dump even if we use 1m interval.

songgao: Oh actually `rate.Limiter` works like that too. Nvm; will change!

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	tracker := &ctxTimeTracker{
+		ctx:       ctx,
+		expiresAt: d.config.Clock().Now().Add(d.dumpIn),
+	}
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	d.chronological.append(tracker)
+	return func() {
+		atomic.StoreInt32(&tracker.done, 1)
+	}
```
jzila: This should return a `*ctxTimeTracker`, whose `Finish` method can be called afterwards instead of a `func()`.

songgao:
> Also, did you test this? How large is the gzipped/b64'd output typically?

Yes I tested; with a clean `kbfsfuse` process (i.e. just started) it's around 20KB. It should happen very rarely, especially after we change the rate limit to 1 minute. But with more goroutines it would be much larger I suppose.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+	gzipper.Close()
+	base64er.Close()
+	d.logger.CDebugf(tracker.ctx,
+		"operation expired. dump>gzip>base64: %q", buf.String())
+}
+
+func (d *ImpatientDebugDumper) dumpTick() {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+	for {
```
jzila: Please add a comment that every iteration through this loop either pops from `chronological` or returns, so there is no possibility of getting stuck in an infinite loop.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,162 @@
...
+type ImpatientDebugDumper struct {
+	config Config
+	logger logger.Logger
+	dumpIn time.Duration
+
+	ticker  *time.Ticker
+	limiter *rate.Limiter
+
+	lock          sync.Mutex
+	chronological *ctxTimeTrackerList
```
jzila: `chronologicalTimeTrackerList` is clearer, IMO.

libkbfs/impatient_debug_dumper.go
songgao: Hmmm I found returning a function more clear in its mission. Is there a reason for returning the pointer to the whole thing?

songgao:
> I'm not sure how I feel about the b64/gzip situation. It creates a bunch of friction and makes it difficult to debug at a glance in context while reading the log.

It would be a goroutine dump so likely not very readable just from the log messages anyway. And it's only for unanticipated hangs, and we'd only look at it when we specifically need to.

libkbfs/impatient_debug_dumper.go
jzila: It hides concerns about underlying implementation. Forcing another component to be aware of `atomic.StoreInt32` is abstraction-breaking.

songgao: Why is it forcing another component to use `atomic`? The caller just gets a function to call in a `defer` statement right?

songgao: Oh you are saying the `atomic` shouldn't be used in this function? I guess I can have a `Finish` and just return `tracker.Finish` here

jzila: That would work too, although I don't really see why returning the tracker is so bad.

jzila: You could also make a `tracker.Close()`, and then return an `io.Closer`.

strib: @jzila: I don't follow what's wrong with the closure as is.  It doesn't leak anything to the calling components, it doesn't even take any arguments...

jzila: 1. `ImpatientDebugDumper` is having to be aware of the underlying implementation of `ctxTimeTracker`.
2. IMO `return t` then `defer t.Close()` or `return t.Close` then `defer closer()` is much clearer than `return func() { atomic.StoreInt32(&tracker.done, 1) }`. Likewise, in `dumpTick`, `t.IsDone()` is clearer than `atomic.LoadInt32(&t.done) == 1`.

strib: Oh, I guess I consider `ctxTimeTracker` to be a sub-type of `ImpatientDebugDumper`.  I thought you were talking about the users of `ImpatientDebugDumper`.  I have no opinion on the internal cleanliness, both are fine with me.

libkbfs/impatient_debug_dumper.go
```diff
@@ -0,0 +1,213 @@
...
+// the log too much by rate limit goroutine dumping based on
+// impatientDebugDumperDumpMinInterval (at most 1 per minute).
+type ImpatientDebugDumper struct {
+	config Config
+	log    logger.Logger
+	dumpIn time.Duration
+
+	ticker                 *time.Ticker
+	limiter                *rate.Limiter
+	idempotentShutdownFunc func()
```
jzila: You can do it this way if you want, but naming it `idempotent` doesn't really add anything. I'd prefer either `shutdownFunc`/`shutdownFn` or `shutdownCh` for a channel closed in a `sync.Once`.

jzila:
All shutdowns should be idempotent.

songgao:
Not as of now though. Currently when I see a `Shutdown` I have no way to tell if it's idempotent or not, unless I read its implementation.

PR #1269, opened 2017-10-14
keybase_daemon_rpc: put short timeouts on service pings

strib:
The whole point of the pings is to force quick reconnects of KBFS to the service, if the service restarts.  However, if a ping itself gets "stuck" in a connection to a daemon that has restarted, we need to make sure that pings gets canceled so the next one will be sent.  So use 1 second timeout for that.

Issue: KBFS-2478

### DIFF
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -359,6 +359,17 @@ func (k *KeybaseDaemonRPC) ShouldRetryOnConnect(err error) bool {
 	return !inputCanceled
 }
 
+func (k *KeybaseDaemonRPC) sendPing(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


libkbfs/keybase_daemon_rpc.go
```diff
@@ -359,6 +359,17 @@ func (k *KeybaseDaemonRPC) ShouldRetryOnConnect(err error) bool {
 	return !inputCanceled
 }
 
+func (k *KeybaseDaemonRPC) sendPing(ctx context.Context) {
+	const sessionID = 0
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
songgao: Could this be too short for CI?

PR #1272, opened 2017-10-16
kbfs: fix a few nits in git and FBO suggested on other PRs

strib:
* #1263: add `gitExec` test function
* #1260: clean subdir in `NewFS` if needed
* #954: when removing block states, allow cap of 0


libkbfs/folder_branch_ops.go
```diff
@@ -2014,7 +2014,7 @@ func (bps *blockPutState) removeOtherBps(other *blockPutState) {
 	// slice length.
 	newLen := len(bps.blockStates) - len(other.blockStates)
 	if newLen <= 0 {
-		newLen = 1
+		newLen = 0
```
jzila: the check can now be `if newLen < 0`

PR #1273, opened 2017-10-16
git-remote-keybase: check for a running service

strib:
Dialing it seems to be the best platform-agnostic way to check for a running service.  Just checking for the socket file doesn't work on Windows.

If the service isn't running, the user will see:

```
git-remote-keybase error: (1) The Keybase service is not running for this user.
```
cc: @malgorithms 

Issue: KBFS-2514

### DIFF
--- a/env/context.go
+++ b/env/context.go
@@ -25,6 +25,7 @@ type Context interface {
 	GetLogDir() string
 	GetDataDir() string
 	ConfigureSocketInfo() (err error)
+	GetGlobalContext() *libkb.GlobalContext
 	GetSocket(clearError bool) (net.Conn, rpc.Transporter, bool, error)
 	NewRPCLogFactory() *libkb.RPCLogFactory
 	GetKBFSSocket(clearError bool) (net.Conn, rpc.Transporter, bool, error)
@@ -91,6 +92,11 @@ func (c *KBFSContext) GetDataDir() string {
 // GetRunMode returns run mode
 func (c *KBFSContext) GetRunMode() libkb.RunMode {
 	return c.g.GetRunMode()
+}
+
+// GetGlobalContext returns the libkb global context.
+func (c *KBFSContext) GetGlobalContext() *libkb.GlobalContext {
+	return c.g
 }
 
 // GetSocket returns a socket
--- a/kbfsgit/git-remote-keybase/main.go
+++ b/kbfsgit/git-remote-keybase/main.go
@@ -59,6 +59,27 @@ func getLocalGitDir() (gitDir string, err error) {
 	return filepath.FromSlash(gitDir), nil
 }
 
+func checkService(kbCtx env.Context) *libfs.Error {
+	// Trying to dial the service seems like the best
+	// platform-agnostic way of seeing if the service is up.  Stat-ing
+	// the socket file, for example, doesn't work for Windows named
+	// pipes.
+	s, err := libkb.NewSocket(kbCtx.GetGlobalContext())
+	if err != nil {
+		return libfs.InitError(err.Error())
+	}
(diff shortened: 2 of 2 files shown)


malgorithms:
thanks for cc'ing - I assume to talk about the copy? I'm cc'ing @oconnor663 - who suggests that we should probably remind people of the exact command to start it, since they're only ever taught it once, during their initial install, and may have forgotten.

If it has to be the same on all platforms, can we just give all instructions, like:

```
Error: Keybase isn't running. Windows/Mac: open the Keybase app; on Linux: `run_keybase`
```

Although if can be different on Windows/Mac/Linux, that can obviously be simplified. I'm happy either way.

oconnor663:
@malgorithms it sounds like @strib is more worried about _detecting_ this problem in a platform-independent way, than about what to print after we detect it?

strib:
The detection is done, I'm happy to take suggestions on printing.  Was cc'ing @malgorithms because you asked for this feature.

That's fine but I actually don't know what the commands are on mac/windows, as it's all handled by packaging.  I could generically say `keybase ctl start`?

oconnor663:
Two options on Linux:

- `keybase ctl start` will start just the Core service.
- `run_keybase` will (re)start everything, including the GUI.

If the user's situation is that the Core service has crashed or something, then the first command should fix it. (Though we also plan to auto-restart it using systemd really soon.) But if the user's situation is that nothing at all is running, including kbfsfuse, then they'd need to run the second command to get going. Do you guys know if that second situation comes up in practice?

strib:
In this case they were running nothing at all.  It's pretty rare, but it happens.  I suggested `keybase ctl start` because technically `git-remote-keybase` only requires the service to run, and not anything else.  But of course they won't be able to make any repos without kbfsfuse running as well.

malgorithms:
> That's fine but I actually don't know what the commands are on mac/windows

I think the mostly likely scenario on mac/windows is just that the user quit the app (from the menu) and needs to start it back up by double-clicking its icon. So telling them to open the app is likely fine.

there is a deeper problem for some macos users (and cc @maxtaco who was just talking about it) where some mac users can't get the app to start at all due to installation/permission problems. I don't think customizing this message for them is so important, as we should be giving them errors on attempts to start up, and they'll be reporting the bugs to us elsewhere. Also, I think Max is going to have a failsafe added...

So I guess in summary for mac/windows we should just tell them to start the app, and it'll cover all the cases that are solvable.

> `keybase ctl start` will start just the Core service.

I'm going to make a core ticket to ask about this....the help suggests this is all platforms and all services...if it's making keybase half work, maybe it shouldn't be showing up in our help.

strib:
> I'm going to make a core ticket to ask about this....the help suggests this is all platforms and all services...if it's making keybase half work, maybe it shouldn't be showing up in our help.

I don't think this appears in any help text, I was just throwing it out there as a suggestion.  I don't think any changes to core are needed.  I can change the text to `run_keybase` on Linux and "open the app" everywhere else.

oconnor663:
For what it's worth, Linux users also have the option to "open the app" in the regular "find the Keybase icon inside your desktop's Applications menu" sense, which basically just runs `run_keybase` under the covers. However, I think most desktops do this clever thing where if they detect that the Keybase window is still running, they'll just foreground it instead of running the command again. (For most regular GUI apps that don't hard-restart themselves like we do, that prevents the confusing condition where you have two copies of the same app running at once. Probably Windows and Mac do something similar?) It's possible that if a user was in an oh-no-service-has-crashed state, the GUI could still be alive, and the "open the app" action might just foreground it instead of fixing the problem.

So yeah, slight bias towards telling them to `run_keybase` manually.

PR #1276, opened 2017-10-17
git-remote-keybase: add platform-specific "missing service" errors

strib:
Suggested by @malgorithms.

Issue: #1273

### DIFF
--- a/kbfsgit/git-remote-keybase/main.go
+++ b/kbfsgit/git-remote-keybase/main.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/kbfs/env"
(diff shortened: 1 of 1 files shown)


kbfsgit/git-remote-keybase/main.go
```diff
@@ -70,8 +71,13 @@ func checkService(kbCtx env.Context) *libfs.Error {
 	}
 	c, err := s.DialSocket()
 	if err != nil {
-		return libfs.InitError(
-			"The Keybase service is not running for this user.")
+		if runtime.GOOS == "darwin" || runtime.GOOS == "windows" {
+			return libfs.InitError(
+				"Keybase isn't running. Open the Keybase app.")
+		} else {
```
songgao: Perhaps specifically filter for `"linux"`? It seems we also have FreeBSD users in the wild.

strib: `run_keybase` isn't what freebsd users would use?  I thought that script was for all non-mac/win users.

songgao: Hmm it's under a `linux` directory: https://github.com/keybase/client/blob/master/packaging/linux/run_keybase

But there doesn't seem to be anything Linux-specific, so I suppose FreeBSD users can use it as well. I've not idea what they actually use. But I guess it's fine this way too.

PR #1277, opened 2017-10-17
kick off outstanding rekeys when PaperKeyCached is called

songgao:


libkbfs/keybase_service_base.go
```diff
@@ -365,6 +365,11 @@ func (k *KeybaseServiceBase) PaperKeyCached(ctx context.Context,
 	k.log.CDebugf(ctx, "Paper key for %s cached", arg.Uid)
 
 	if k.getCachedCurrentSession().UID == arg.Uid {
+		err := k.config.KBFSOps().KickoffAllOutstandingRekeys()
```
strib: Move this below the comment about ignoring errors, or add a new one?

songgao: Added a comment instead of moving since it's not really the same scenario as the one below which requires waiting on the returned channel.

PR #1278, opened 2017-10-18
keybase_service_base: don't hold the sessionCachelock across RPCs

strib:
The RPC connection might end up calling `KeybaseDaemonRPC.OnDisconnected()` during the RPC (especially if this RPC is initializing the connection for the first time), which calls `KeybaseServiceBase.clearCaches()`, which takes the `sessionCacheLock`.

Issue: keybase/client#9093

### DIFF
--- a/libkbfs/keybase_service_base.go
+++ b/libkbfs/keybase_service_base.go
@@ -37,6 +37,7 @@ type KeybaseServiceBase struct {
 	sessionCacheLock sync.RWMutex
 	// Set to the zero value when invalidated.
 	cachedCurrentSession SessionInfo
+	sessionInProgressCh  chan struct{}
 
 	userCacheLock sync.RWMutex
 	// Map entries are removed when invalidated.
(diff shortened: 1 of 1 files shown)


libkbfs/keybase_service_base.go
```diff
@@ -693,7 +723,11 @@ func (k *KeybaseServiceBase) getCurrentSession(
 		ctx, "new session with username %s, uid %s, crypt public key %s, and verifying key %s",
 		s.Name, s.UID, s.CryptPublicKey, s.VerifyingKey)
 
+	k.sessionCacheLock.Lock()
+	defer k.sessionCacheLock.Unlock()
 	k.cachedCurrentSession = s
+	close(k.sessionInProgressCh)
```
jzila: Actually, can this be nil here?

strib: It can't be `nil`, because for `doRPC` above to be `true`, this variable must have been set, and can only be cleared by that same goroutine.  I guess it should be in a `defer` actually, in case there's an error.  Will fix.

songgao: Why can it only be cleared by the same goroutine? What if getCurrentSession is called by another routine?

strib: The other goroutine will be stuck in the `for` loop above until they either see a real session, or until `doRPC` returns `true`.

PR #1279, opened 2017-10-18
folder_branch_ops: always send the initial MD straight to server

strib:
We don't want to return keys to chat that haven't yet been accepted by the server.

Issue: KBFS-2553

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -439,6 +439,28 @@ func (fs *KBFSOpsStandard) GetTLFCryptKeys(
 	if err != nil {
 		return nil, tlf.ID{}, err
 	}
+
+	// If journaling is enabled, wait for the initial revision to be
+	// accepted by the server, so we don't return keys that might
+	// later hit a conflict. (See KBFS-2553.)
+	jServer, err := GetJournalServer(fs.config)
+	if err == nil {
+		fb := FolderBranch{Tlf: rmd.TlfID(), Branch: MasterBranch}
+		ops := fs.getOpsNoAdd(ctx, fb)
+		lState := makeFBOLockState()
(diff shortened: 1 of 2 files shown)


libkbfs/kbfs_ops.go
```diff
@@ -439,6 +439,28 @@ func (fs *KBFSOpsStandard) GetTLFCryptKeys(
...
+		fb := FolderBranch{Tlf: rmd.TlfID(), Branch: MasterBranch}
+		ops := fs.getOpsNoAdd(ctx, fb)
+		lState := makeFBOLockState()
+		for ops.getLatestMergedRevision(lState) < kbfsmd.RevisionInitial {
+			err = ops.waitForJournal(ctx, lState, jServer)
+			if err != nil {
+				return nil, tlf.ID{}, err
+			}
+
+			rmd, err = fs.getMDByHandle(ctx, tlfHandle, FavoritesOpNoChange)
```
songgao: Is `rmd` here different from before the flush?

strib: Yes, it might now contain keys written by some other user who initialized the TLF simultaneously.

strib:
Actually, I think this needs to be done inside `folderBranchOps.initMDLocked` instead, so that it applies to KBFS-related creates as well.  There's already a TODO there.  I'll fix it.

libkbfs/kbfs_ops.go
songgao: Oh right. Thanks.

strib:
Ok ready for review again, sorry for the churn.

strib:
> doesn't the journal server also need to be told about the Put

Nope!  The journal server doesn't need to know about any merged revisions.  Works just the same as if this device started up after some other device had pushed the initial revision.

PR #1280, opened 2017-10-19
use keybase1.MDPriorityGit for git

songgao:
As discussed on keybase chat today, I ended up not using a `DefaultPriority` in config, but instead used a similar approach as `lockContext` in `tlf_journal.go`.


libkbfs/tlf_journal.go
```diff
@@ -864,7 +869,6 @@ func (j *tlfJournal) flush(ctx context.Context) (err error) {
 			if j.singleOpMode == singleOpFinished {
 				j.log.CDebugf(ctx, "Resetting single op mode")
 				j.singleOpMode = singleOpRunning
-				j.singleOpLockContext = nil
```
songgao: I don't remember why I had `singleOpLockContext` as part of the `tlfJournal` struct, so I moved it into this function here. If anybody knows my earlier thought process better than me, please let me know!

songgao: Ugh I think I remember now. NVM ...

PR #1284, opened 2017-10-20
add ServerErrorClassicTLFNotExists

songgao:


kbfsmd/server_errors.go
```diff
@@ -54,6 +54,12 @@ const (
 	// write to be contingent on is not held at the time server tries to commit
 	// the MD, and as a result the MD is not written.
 	StatusCodeServerErrorRequiredLockIsNotHeld = 2813
+	// StatusCodeServerErrorClassicTLFNotExists is the error code returned by a
+	// MD get operation to indicate that a classic TLF is not found, and client
+	// has specified not to create one. Normally upon this error, KBFS client
+	// should ask service to create an implicit team for the give handle, and
+	// use the i-team backed TLF.
+	StatusCodeServerErrorClassicTLFNotExists = 2814
```
jzila: `ClassicTLFDoesNotExist` sounds more correct, I think.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -402,3 +402,8 @@ func (k *KeybaseDaemonRPC) Shutdown() {
 		k.keepAliveCancel()
 	}
 }
+
+// TeamExit (does not) implement keybase1.NotifyTeamInterface.
+func (k *KeybaseDaemonRPC) TeamExit(context.Context, keybase1.TeamID) error {
+	return nil
```
songgao: Implementing this as a no-op here since it doesn't seem needed by KBFS for now. Let me know if I'm wrong!

cc @mlsteele

mlsteele: Yes good! My bad. It's an unreliable notification anyway, so fine to no-op. If you'd like to bust your cache or something later you could use it for that.

songgao: Cool; thanks for confirming!

mlsteele: Is this right? Jenkins didn't catch this because client doesn't run kbfs. But circle caught it because it revendors keybase1 into kbfs before compiling kbfs?

Also all I ever see when there's a circle failure is this 
image

songgao: I think circle catches it because on mobile we use gobind, and it freaks out there if interface and implementation don't match.

The "spins forever" thing seems like a CircleCI UI bug; I always just collapsed it and expanded again to see actual error.

PR #1286, opened 2017-10-24
kbfs_ops: avoid a deadlock on concurrent TeamNameChanged calls

strib:
Deadlock:
* First call to `KBFSOps.TeamNameChanged` takes `opsLock`.
* Calls `go FolderBranchOps.TeamNameChanged()`.
* Releases `opsLock`.
* `FolderBranchOps.TeamNameChanged()` takes `headLock`.
* Second call to `KBFSOps.TeamNameChanged` takes `opsLock`.
* Calls `FolderBranchOps.getRootNode`, which blocks trying to acquire `headLock`.
* `FolderBranchOps.TeamNameChanged()` calls `KBFSOps.changeHandle()`.
* `KBFSOps.changeHandle()` tries to acquire `opsLock`.
* DEADLOCK.

Instead, make sure we don't hold `opsLock` when calling `getRootNode`.

Issue: KBFS-2565

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -873,10 +873,17 @@ func (fs *KBFSOpsStandard) TeamNameChanged(
 
 	fs.log.CDebugf(ctx, "Got TeamNameChanged for %s", tid)
 	fs.opsLock.Lock()
-	defer fs.opsLock.Unlock()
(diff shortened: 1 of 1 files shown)


maxtaco:
thanks for fixing it so quickly. we'll look into the other issue of why two calls get fired. i think i have a theory

PR #1291, opened 2017-10-26
use default mount type by default and reinstall mountdir if needed

songgao:
In https://github.com/keybase/kbfs/commit/39534e4a6ffe95f84a5480d60b3606112fa93206 , we made default mounter do unmount-then-mount in case of failure too. So the only difference between the default mounter and the force mounter now is the force mounter returns an error causing `kbfsfuse` to exit in case of a mount failure. 

This PR changes the default mount type for production build from `force` to `default` so that we don't get stuck in a crash loop in case of mount failures.

In addition, if mount fails with the unmount-then-mount-again approach too, on darwin we call the installer to reinstall the mountdir and try again. This should cover the case where `/keybase` permission is messed up preventing us from mounting.

### DIFF
--- a/libfuse/mounter.go
+++ b/libfuse/mounter.go
@@ -20,13 +20,28 @@ type mounter struct {
 	c       *fuse.Conn
 }
 
+const darwinInstallerPath = "/Applications/Keybase.app/Contents/Resources/" +
+	"KeybaseInstaller.app/Contents/MacOS/Keybase"
+
+func (m *mounter) reinstallMountDirForDarwin() {
+	args := []string{
+		"--app-path=/Applications/Keybase.app",
+		"--run-mode=prod",
+		"--timeout=60",
+	}
+	exec.Command(darwinInstallerPath,
+		append(args, "--uninstall-mountdir")...).Run()
+	exec.Command(darwinInstallerPath,
+		append(args, "--install-mountdir")...).Run()
+}
+
 // fuseMount tries to mount the mountpoint.
 // On a force mount then unmount, re-mount if unsuccessful
 func (m *mounter) Mount() (err error) {
 	m.c, err = fuseMountDir(m.options.MountPoint, m.options.PlatformParams)
 	// Exit if we were succesful. Otherwise, try unmounting and mounting again.
 	if err == nil {
-		return err
+		return nil
 	}
 
 	// Mount failed, let's try to unmount and then try mounting again, even
@@ -34,6 +49,22 @@ func (m *mounter) Mount() (err error) {
 	m.Unmount()
 
 	m.c, err = fuseMountDir(m.options.MountPoint, m.options.PlatformParams)
+
+	if err == nil {
+		return nil
+	}
+
+	if runtime.GOOS == "darwin" {
+		// Mount failed again, and we are on darwin. So ask the installer to
+		// reinstall the mount dir and try again as the last resort. This
+		// specifically fixes a situation where /keybase gets created and owned
+		// by root after Keybase app is started, and `kbfs` later fails to
+		// mount because of a permission error.
+		m.Unmount()
+		m.reinstallMountDirForDarwin()
+		m.c, err = fuseMountDir(m.options.MountPoint, m.options.PlatformParams)
+	}
+
 	return err
 }
 
--- a/kbfsfuse/defaults_production.go
+++ b/kbfsfuse/defaults_production.go
@@ -6,4 +6,4 @@
 
 package main
 
-const defaultMountType = "force"
+const defaultMountType = "default"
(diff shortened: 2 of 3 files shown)


strib:
@songgao: Why did the app not do this unmount/mount already on the force-mount crash?  I still don't understand that part.

songgao:
I think restarting `kbfs` process is handled by the `launchd` but not service (which controls the `mountdir` component, which can install or uninstall `/keybase` dir). Also it seems `--uninstall-mountdir` only happens when app exits, since if I have a root-owned `/keybase` before starting the app, `kbfs` process gets stuck in a crash loop too.

strib:
> Also it seems --uninstall-mountdir only happens when app exits

That can't be strictly correct, because then anyone who runs out of battery or has some other sudden crash would be unable to use the keybase mount again, right?

strib:
Also, I'm not sure I'm on-board with merging this during the code freeze.  Seems risky, might want to let it bake a while after the upcoming release.

libfuse/mounter.go
```diff
@@ -20,20 +20,51 @@ type mounter struct {
...
+const darwinInstallerPath = "/Applications/Keybase.app/Contents/Resources/" +
+	"KeybaseInstaller.app/Contents/MacOS/Keybase"
+
+func (m *mounter) reinstallMountDirForDarwin() {
+	args := []string{
+		"--app-path=/Applications/Keybase.app",
+		"--run-mode=prod",
+		"--timeout=60",
+	}
+	exec.Command(darwinInstallerPath,
```
strib: Ignoring err values here -- can we at least print them?

songgao:
> Also, I'm not sure I'm on-board with merging this during the code freeze. Seems risky, might want to let it bake a while after the upcoming release.

Oh I thought we already released. Will revert now.

songgao:
> because then anyone who runs out of battery or has some other sudden crash would be unable to use the keybase mount again, right?

Not really; in a normal setup, that would just leave a `/keybase` owned by user behind. And mount will still succeed. It's only a problem if somehow a `/keybase` owned by root is created, likely by user, or running a third party app (e.g. `kbsecret`) with `sudo`.

libfuse/mounter.go
```diff
@@ -20,20 +20,51 @@ type mounter struct {
 	c       *fuse.Conn
 }
 
+const darwinInstallerPath = "/Applications/Keybase.app/Contents/Resources/" +
+	"KeybaseInstaller.app/Contents/MacOS/Keybase"
+
+func (m *mounter) reinstallMountDirForDarwin() {
```
strib: I'd love to isolate all this into `mounter_osx.go` if possible.

libfuse/mounter.go
```diff
@@ -20,20 +20,51 @@ type mounter struct {
 	c       *fuse.Conn
 }
 
+const darwinInstallerPath = "/Applications/Keybase.app/Contents/Resources/" +
```
strib: Having the app executable path hardcoded here makes me very nervous, because 1) the desktop team can change it any time, and not update this, and 2) it's kind of a separation of powers abuse, right?  kbfsfuse shouldn't really know about the app.  It would seem better to me to either pass in the command for remounting as a CLI flag or an env var, or to just re-build the logic in Go, in `mounter_osx.go`.  It can't be too complicated, right?

songgao: > or to just re-build the logic in Go

I think it involves talking to the helper, which is probably why core does a similar thing too: https://github.com/keybase/client/blob/45834bd25f739a98ff404e61d181d7e393955721/go/install/install_darwin.go#L803

But having a flag sounds good too. I'll try to update.

strib:
Ah I see, fuse just makes it look like its owned by root, but the underlying mountpoint is still owned by the user.  Got it, thanks.

libfuse/mounter.go
strib: Ah.  The core side looks like it's just in the installer though, which is what I'd expect.  I doubt the service itself knows about app-level stuff.

songgao: It get's used lines below: https://github.com/keybase/client/blob/45834bd25f739a98ff404e61d181d7e393955721/go/install/install_darwin.go#L807

Which is the same thing we use here. Perhaps we can move that into a separate package and vendor it here?

strib: Both those links are in the installer, which I don't think it's used by the background Keybase service.  Is it?

songgao: Oh I see what you mean now. I think it is what service uses to call the obj-c written installer to talk to the helper. For example: https://github.com/keybase/client/blob/45834bd25f739a98ff404e61d181d7e393955721/go/keybase/main.go#L188

strib: Oh, hrm.  Yuck.  Thanks for the link.

PR #1296, opened 2017-10-27
default mount type for prod & reinstall mountdir if fail on darwin

songgao:
This supersedes https://github.com/keybase/kbfs/pull/1291 which was reverted.

### DIFF
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -30,13 +30,13 @@ const usageFormatStr = `Usage:
 
 To run against remote KBFS servers:
   kbfsfuse
-    [-runtime-dir=path/to/dir] [-label=label] [-mount-type=` + defaultMountType + `]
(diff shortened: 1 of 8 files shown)


songgao:
@strib have you had a chance to look at this yet?

strib:
Nope.

kbfsfuse/main.go
```diff
@@ -30,13 +30,13 @@ const usageFormatStr = `Usage:
 
 To run against remote KBFS servers:
   kbfsfuse
-    [-runtime-dir=path/to/dir] [-label=label] [-mount-type=force]
+    [-runtime-dir=path/to/dir] [-label=label] [-mount-type=` + defaultMountType + `]
```
strib: Should this be something like "-mount-type=[default|force|none]"?

songgao: sure; will update!

PR #1297, opened 2017-10-30
kbfsgit: add on-demand delta objects, and update go-git vendor for memory fixes

strib:
This adds a custom layer for accessing "delta" git objects, without keeping all the data in memory for the lifetime of the object.  It will be useful once we have git garbage collection, but currently won't be used by `git-remote-keybase` because we set `pack.window = 0` for all kbfs git repos, and so diffs aren't used for pulls at all.

This also vendors in some changes from our fork of `go-git`, which help keep memory stable during delta compression (also not used currently in `git-remote-keybase`).  With these changes I was able to keep memory usage at less than 1.7 GB for the keybase/client repo when calculating diffs, as opposed to 3-4 GB of usage without it.

### DIFF
--- a/kbfsgit/on_demand_storer.go
+++ b/kbfsgit/on_demand_storer.go
@@ -6,7 +6,9 @@ package kbfsgit
 
 import (
 	lru "github.com/hashicorp/golang-lru"
+	"github.com/pkg/errors"
 	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
 	"gopkg.in/src-d/go-git.v4/storage"
 )
 
@@ -23,7 +25,7 @@ var _ storage.Storer = (*onDemandStorer)(nil)
 func newOnDemandStorer(s storage.Storer) (*onDemandStorer, error) {
 	// Track a small number of recent in-memory objects, to improve
 	// performance without impacting memory too much.
-	recentCache, err := lru.New(10)
+	recentCache, err := lru.New(25)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


kbfsgit/on_demand_storer.go
```diff
@@ -23,7 +25,7 @@ var _ storage.Storer = (*onDemandStorer)(nil)
 func newOnDemandStorer(s storage.Storer) (*onDemandStorer, error) {
 	// Track a small number of recent in-memory objects, to improve
 	// performance without impacting memory too much.
-	recentCache, err := lru.New(10)
+	recentCache, err := lru.New(25)
```
jzila: What are the usage characteristics of these objects? Is `lru` helpful? Should we consider storing a `Pool` instead?

jzila: Also: how big are these objects usually, and how much do they vary? How big can we expect this cache to get? Should we use @songgao's memory-limited LRU queue instead of the count-limited one?

strib: LRU is very helpful here because of the way delta compression works.  It first sorts the objects by type and descending size, and then compares each object to a sliding window of previous objects to find a good match.  We set the size of this sliding window to 10, and so I think we were thrashing a bit in this cache because the sliding window, plus the current object, was greater than the cache size.

The sizes are the sizes of the underlying git objects themselves, which vary greatly.  However, since the set in use is based on the sliding window, it seems like that should be the limiting factor.  Eventually we might hit some repo where there's a set of large objects that can overrun memory, but at the limit that could be any two objects, and then the compression algorithm in go-git is worthless.  So I think for now I prefer this more simple thing, until we see a need for something else.

Does that sound reasonable?

jzila: Yeah, awesome explanation @strib! Would you mind putting it into the code as a comment? IMO it'd be very helpful to have in-place if another one of us needs to pick up that part of the code.

PR #1298, opened 2017-10-31
kbfsgit: support "push.gpgSign=if-asked" configuration

strib:
Suggested by celti.

Issue: keybase/client#9309


strib:
> Is the endgame here to eventually support signing?

No idea!  That would be cool though.  Haven't actually thought about it.

PR #1299, opened 2017-10-31
context: Add cleanup function for kbfs socket

jzila:
This first cleans up any existing socket files before trying to bind to it.

I've tested this on my Linux localhost and it works great.

I'd like a review by @oconnor663 and one of the KBFS folks.

### DIFF
--- a/env/context.go
+++ b/env/context.go
@@ -5,6 +5,7 @@
 package env
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


oconnor663:
My plan will be to immediately (like, this afternoon) add in a short-circuit similar to this one: https://github.com/keybase/client/blob/173f21f/go/service/main.go#L721-L725

If that sounds ok to you, this LGTM.

jzila:
That's fine, but I'd guess that it'll never be hit in KBFS.

oconnor663:
That branch is always hit in the systemd configuration we're shooting for. Systemd will provide the listen fd the same way, whether the service is actually started through socket activation, or just with `systemctl start ...`. As long as a `.socket` file exists for the service, systemd will always provide the socket.

I think you're right that we'll probably never sock-activate kbfs. But setting things up this way makes startup a little bit smoother (even though we're robust to it being racy), and it seems to be the way systemd wants it to happen, so I'm running with it :)

jzila:
Sounds good. Please CC me on relevant reviews, I'm interested in following along.

songgao:
With this change, does it mean we are not reusing existing socket file or systemd's socket activation?

oconnor663:
@songgao see https://github.com/keybase/kbfs/pull/1300. In the systemd environment we'll skip this codepath entirely.

PR #1300, opened 2017-10-31
initial systemd support

oconnor663:
r? @jzila 

Dummy commit will need to get revendored after https://github.com/keybase/client/pull/9325 lands. Also I don't think there's any conflict with https://github.com/keybase/kbfs/pull/1299, but I'll test to make sure.

### DIFF
--- a/libfuse/start.go
+++ b/libfuse/start.go
@@ -11,6 +11,7 @@ import (
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/client/go/systemd"
 	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libgit"
 	"github.com/keybase/kbfs/libkbfs"
@@ -112,6 +113,13 @@ func Start(options StartOptions, kbCtx libkbfs.Context) *libfs.Error {
 		return libfs.InitError(err.Error())
 	}
 	defer libkbfs.Shutdown()
+
+	// Report "startup successful" to the supervisor (currently just systemd on
+	// Linux). This isn't necessary for correctness, but it allows commands
+	// like "systemctl start kbfs.service" to report startup errors to the
+	// terminal, by delaying their return until they get this notification
+	// (Type=notify, in systemd lingo).
+	systemd.NotifyStartupFinished()
 
 	if options.SkipMount {
 		log.Debug("Skipping mounting filesystem")
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -57,6 +57,18 @@
 			"revisionTime": "2016-06-30T22:46:53-07:00"
 		},
 		{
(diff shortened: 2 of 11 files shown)


oconnor663:
CI failures because of a vendoring mistake on my part. Rerunning.

oconnor663:
I think the CircleCI failure here will go away after the client PR is in.

libfuse/start.go
```diff
@@ -113,6 +114,13 @@ func Start(options StartOptions, kbCtx libkbfs.Context) *libfs.Error {
 	}
 	defer libkbfs.Shutdown()
 
+	// Report "startup successful" to the supervisor (currently just systemd on
+	// Linux). This isn't necessary for correctness, but it allows commands
+	// like "systemctl start kbfs.service" to report startup errors to the
+	// terminal, by delaying their return until they get this notification
+	// (Type=notify, in systemd lingo).
+	systemd.NotifyStartupFinished()
```
strib: Not sure what the semantics are supposed to be, but this notification is obviously sent before we mount KBFS.  Should t go down below, above `mi.Wait()`?  Or is it just supposed to measure the liveness of the process, and not whether it's been fully initialized?

vendor/vendor.json
```diff
@@ -57,6 +57,18 @@
 			"revisionTime": "2016-06-30T22:46:53-07:00"
 		},
 		{
+			"checksumSHA1": "RBwpnMpfQt7Jo7YWrRph0Vwe+f0=",
+			"path": "github.com/coreos/go-systemd/activation",
+			"revision": "d2196463941895ee908e13531a23a39feb9e1243",
```
strib: Please update the revision with a real client hash after the client side is merged.

libfuse/start.go
oconnor663: Ah, I'll put it as low as possible. I wasn't sure before when that `Wait` call was returning.

vendor/vendor.json
oconnor663: yep, that's the "DUMMY COMMIT" that's going to get replaced.

libfuse/start.go
oconnor663: Hmm, it looks like `startMounting` blocks though.

oconnor663: I'm going to leave this as is, I think. We could move it into `startMounting` right before the call to `Serve`. But then I think we'd also need to keep a call here in `Start` for the `SkipMount` case. I'll stick with the simple thing for now, and we can make more than one call if we decide we need to wait for more mount errors.

oconnor663: (Also it looks like kbfsfuse doesn't actually exit if the mount fails. That's a slightly weird state to report to systemd. Maybe the right thing to do is exit with a standard error code?)

strib: Ah you're right about the blocking, sorry.

There are several schools of thought about what should happen if the mount fails, since we might want to keep running to let chat work.  Still an active topic of debate I think.

PR #1301, opened 2017-11-01
folder_branch_ops: avoid data race in `Lookup()`

strib:
If `Lookup` returns early, it could write to the return param `node`. But the internal goroutine might still be running, which also tried to write to the `node` param.  This leads to a possible (but really rare) data race.

Instead, have the goroutine write to a temp variable, which is only returned when the `Lookup` is successful.

Issue: KBFS-2579

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1838,6 +1838,9 @@ func (fbo *folderBranchOps) Lookup(ctx context.Context, dir Node, name string) (
 		return nil, EntryInfo{}, err
 	}
 
+	// It's racy for the goroutine to write directly to return param
+	// `node`, so use a new param for that.
+	var n Node
 	var de DirEntry
 	err = runUnlessCanceled(ctx, func() error {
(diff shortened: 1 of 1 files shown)


strib:
> How could this be a problem? Whatever that calls Lookup here would assign the returned Node to some variable, and would never get overwritten by the inner goroutine right?

It's not about the caller, it's about the return parameter itself on the stack.  You have two goroutines trying to write to that variable on the stack at the same time, before it gets back to the caller, so it triggers the race detector.

songgao:
Hmmm I'm still not clear; I thought it was `runUnlessCanceled` that caused early-return. Is it not the case? Where did the 2nd routine start?

strib:
There is the goroutine of the original caller, and the goroutine started by `runUnlessCanceled`.  As soon as the ctx is canceled, `runUnlessCanceled` returns with an error, but the goroutine it statred may keep going (because it was waiting on a lock, for example).  Then when that goroutine returns, there can be a race.

PR #1302, opened 2017-11-02
mdserver_memory: check the lock holder and block lock calls

strib:
In order to test locking, we need to track which `MDServerMemory` instance took the lock, and block other instances from taking the lock until the holder releases it (or it expires).

Tests that use this behavior will come in subsequent PRs.

Issue: KBFS-2484

### DIFF
--- a/libkbfs/mdserver_memory.go
+++ b/libkbfs/mdserver_memory.go
@@ -431,6 +431,9 @@ func (md *MDServerMemory) GetRange(ctx context.Context, id tlf.ID,
 		return nil, err
 	}
 
+	// An RPC-based client would receive a throttle message from the
+	// server and retry with backoff, but here we need to implement
+	// the retry logic explicitly.
 	for {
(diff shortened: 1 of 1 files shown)


libkbfs/mdserver_memory.go
```diff
@@ -408,9 +431,24 @@ func (md *MDServerMemory) GetRange(ctx context.Context, id tlf.ID,
...
+	for {
+		rmds, ch, err := md.doGetRange(
+			ctx, id, bid, mStatus, start, stop, lockBeforeGet)
+		if err != nil {
+			return nil, err
+		}
+		if ch == nil {
+			return rmds, err
+		}
+		select {
```
songgao: `mdserver_remote` would get a throttled error here and rely on `rpc` to retry. So maybe include a comment here that we are doing the RPC retry logic here within the handler? Same for `Lock` below.

PR #1303, opened 2017-11-02
kbfsgit: concurrent PackRefs test

strib:
* One where a second client updates a reference during ref-packing.
* Another one where a second client deletes a reference during ref-packing.

This adds a `runner.packRefs()` method for now, that is only invoked by testing.  A later PR will call this as part of a larger GC operation.

(Depends on #1302 and keybase/go-git#13.  Will update the commit hash once the latter is merged.)

Issue: KBFS-2517

### DIFF
--- a/kbfsgit/runner_test.go
+++ b/kbfsgit/runner_test.go
@@ -666,3 +666,210 @@ func TestPushcertOptions(t *testing.T) {
 	checkPushcert("true", "unsupported")
 	checkPushcert("false", "ok")
 }
+
+func TestPackRefsAndOverwritePackedRef(t *testing.T) {
+	ctx, config, tempdir := initConfigForRunner(t)
+	defer libkbfs.CheckConfigAndShutdown(ctx, t, config)
+	defer os.RemoveAll(tempdir)
+
+	git1, err := ioutil.TempDir(os.TempDir(), "kbfsgittest")
+	require.NoError(t, err)
+	defer os.RemoveAll(git1)
+
+	// Make shared repo with 2 branches.
+	makeLocalRepoWithOneFile(t, git1, "foo", "hello", "")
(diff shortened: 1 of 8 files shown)


strib:
> If I understand correctly, this doesn't make sure the second operation waits for the packref to finish. But it seems there's no way to do that unless we add a sleep in test?

Not sure what you mean by "second operation".  Every time the test calls `packRefs()`, it pipes the error into a channel, and there's a select that waits for a receive on that channel.  Did I miss something?

kbfsgit/runner_test.go
```diff
@@ -666,3 +666,210 @@ func TestPushcertOptions(t *testing.T) {
...
+	// Now let the pack-refs finish.
+	close(packUnstall)
+	select {
+	case err := <-packErrCh:
+		require.NoError(t, err)
+	case <-ctx.Done():
+		t.Fatal(ctx.Err())
+	}
+
+	// And the delete should finish right after.
```
songgao: This is what I was talking about. There's no mechanism that guarantees that this happens right *after* . If locking fails and it happened before `<-packErrCh` happened, this test would still pass.

strib: Hmm, if locking fails then `deleteErrCh` should receive a non-nil error and the test would fail.

But there is no guarantee that the delete command will actually block on the lock -- it might get delayed, and only attempt to get the lock after the pack refs is done.  Which would be a less effective test, but yeah it's too much work to tell when a lock attempt is blocking, not really worth it to get that guarantee in my opinion.

songgao: By "locking fails", I meant we accidentally used wrong `LockID` and made it ineffective, not that the locking operations returned error.

But yeah it seems to be a lot more work to know that for sure here. Agreed!

PR #1308, opened 2017-11-03
libgit: add a function for repo garbage collection

strib:
This will eventually be called by a `keybase git gc` command that we will recommend users to run when their repos get too unwieldy.

Unfortunately this drags the go-git dependency into the main kbfsfuse process, but I don't see a way around that.

Depends on #1303 (review that one first!).

Issue: KBFS-2540

### DIFF
--- a/libgit/repo.go
+++ b/libgit/repo.go
@@ -281,10 +281,10 @@ func makeExistingRepoError(
 
 func createNewRepoAndID(
 	ctx context.Context, config libkbfs.Config, tlfHandle *libkbfs.TlfHandle,
-	repoName string, fs *libfs.FS) (ID, error) {
+	repoName string, fs *libfs.FS) (repoID ID, err error) {
 	// TODO: take a global repo lock here to make sure only one
 	// client generates the repo ID.
-	repoID, err := makeRandomID()
+	repoID, err = makeRandomID()
 	if err != nil {
 		return NullID, err
 	}
@@ -296,7 +296,12 @@ func createNewRepoAndID(
 	if err != nil {
 		return NullID, err
 	}
-	defer lockFile.Close()
+	defer func() {
+		closeErr := lockFile.Close()
+		if err == nil {
+			err = closeErr
+		}
+	}()
 
 	f, err := fs.Create(kbfsConfigName)
 	if err != nil && !os.IsExist(err) {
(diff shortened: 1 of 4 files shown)


libgit/repo.go
```diff
@@ -748,3 +751,80 @@ func RenameRepo(
...
+	log.CDebugf(ctx, "Locking for GC")
+	f, err := fs.Create(repoGCLockFileName)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	err = f.Lock()
+	if err != nil {
+		return err
+	}
```
songgao: Don't we need a `f.Unlock()` here? Also it seems if unlock fails because lock expired we need to throw the writes causes by `PackRefs` away right?

libgit/repo.go
```diff
@@ -748,3 +751,80 @@ func RenameRepo(
...
+	// TODO: check object pruning and object re-packing thresholds.
+	return doPackRefs, numLooseRefs, nil
+}
+
+// GCRepo runs garbage collection on the specified repo, if it exceeds
+// any of the thresholds provided in `options`.
+func GCRepo(
+	ctx context.Context, config libkbfs.Config, tlfHandle *libkbfs.TlfHandle,
+	repoName string, uniqID string, options GCOptions) error {
+	log := config.MakeLogger("")
```
songgao: Maybe `log := config.MakeLogger("GITGC")`?

strib: I personally don't think module labels are useful unless we are going to have multiple instances of something, like we have in the case of the dirty block cache.  In this case, the random context debugging label will uniquely identify requests, and I don't see a need to add a module label.  But if you have a good argument for one, I'll add it!

libgit/repo.go
strib: The `defer f.Close()` right above this does the unlock.  But we need to propagate the error back to the caller, I'll fix that.

If the unlock fails, then the journal flush fails, and so once I fix the above issue, this call will return an error.  Then it is up to the caller to delete the journal without flushing it (usually by just returning early and not calling `waitForJournal()`.

songgao: Sounds good. Thanks!

libgit/repo.go
songgao: I don't have a good argument, so this sounds good to me. Thanks!

PR #1316, opened 2017-11-06
libgit: vendor new git gc RPC and implement it

strib:
Issue: KBFS-2540


strib:
(Circle CI will fail until keybase/client#9381 is merged, I'll merge both at the same time.)

PR #1319, opened 2017-11-06
libgit: add loose object pruning to GC

strib:


libgit/repo.go
```diff
@@ -846,17 +875,37 @@ func GCRepo(
...
-	log.CDebugf(ctx, "Packing %s loose refs", numLooseRefs)
-	return storer.PackRefs()
+	if doPackRefs {
+		log.CDebugf(ctx, "Packing %s loose refs", numLooseRefs)
+		err = storage.PackRefs()
+		if err != nil {
+			return err
+		}
+	}
+
```
taruti: Wouldn't it be better to first prune and then pack. And not pack and then prune. (if prune deletes some objects then adding them to a pack is not good). I.e. just change the order of pruning and packing here?

libgit/repo.go
```diff
@@ -846,17 +875,37 @@ func GCRepo(
...
+	if !doPackRefs && !doPruneLoose {
 		log.CDebugf(ctx, "GC no longer needed")
 		return nil
 	}
 
-	// For now just pack refs.  TODO: add object pruning and object
-	// re-packing.
-	log.CDebugf(ctx, "Packing %s loose refs", numLooseRefs)
-	return storer.PackRefs()
+	if doPackRefs {
```
taruti: Wouldn't it be better to first prune loose objects and then pack. I.e. change the order of packing and pruning here.

libgit/repo.go
strib: Above is just packing refs, not objects.  It won't ever delete or unreference any objects.  I don't think the order matters, but please correct me if I'm wrong.

taruti: Sorry for the noice, you are right.

taruti: Sorry for the noise, you are right.

PR #1320, opened 2017-11-07
add proritized round robin address list for bserver and mdserver

songgao:
depends on: https://github.com/keybase/go-framed-msgpack-rpc/pull/126

### DIFF
--- a/libkbfs/init.go
+++ b/libkbfs/init.go
@@ -119,9 +119,14 @@ func defaultBServer(ctx Context) string {
 	case libkb.DevelRunMode:
 		return memoryAddr
 	case libkb.StagingRunMode:
-		return "bserver-0.dev.keybase.io:443"
+		return `
+			bserver-0.dev.keybase.io:443,bserver-1.dev.keybase.io:443;
+			bserver.dev.keybase.io:443`
 	case libkb.ProductionRunMode:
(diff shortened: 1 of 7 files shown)


libkbfs/init.go
```diff
@@ -119,9 +119,14 @@ func defaultBServer(ctx Context) string {
...
 		return memoryAddr
 	case libkb.StagingRunMode:
-		return "bserver-0.dev.keybase.io:443"
+		return `
+			bserver-0.dev.keybase.io:443,bserver-1.dev.keybase.io:443;
+			bserver.dev.keybase.io:443`
 	case libkb.ProductionRunMode:
-		return "bserver.kbfs.keybase.io:443"
+		return `
+			bserver-0.kbfs.keybaseapi.com:443,bserver-1.kbfs.keybaseapi.com:443;
```
strib: Remind me again: what's the point of having new clients connect to both -0 and -1?  If we ever have to blackhole -0, we'd have to blackhole -1 at the same time, right?  What is supposed to be the difference between the two?  Is it just that someday we might want to have two ELBs, and this will help load balance between them?

songgao: Yeah this is in case we'll have two ELBs, and we won't have to update client for that update. I think it'd be useful when we wanna go multi-region, or some A/B testing for server side changes.

PR #1323, opened 2017-11-08
kbfs_ops: some optimizations for single op mode (and some cleanup)

strib:
This PR does a couple different things:

* `folderBranchOps.getMDForReadLocked` doesn't actually require any locks to be taken, so rename it.
* use cached root node in `kbfsOps.getMaybeCreateRootNode` if possible.  `git-remote-keybase` calls this function several times (each time it constructs a `libfs.FS` instance), and there's no reason to check with the servers each time.
* don't fetch unmerged MDs when initializing a TLF in single op mode, because there's basically no way for a TLF to start off as unmerged when in single op mode.

These optimizations bring a simple git push down from 12-14s to 9-10s.

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -396,6 +396,10 @@ func (fs *KBFSOpsStandard) getMDByHandle(ctx context.Context,
 		return rmd, nil
 	}
 
+	// Check for an unmerged MD first, unless we're in single-op
+	// mode.  If this is a single-op, we can skip this check because
+	// there's basically no way for a TLF to start off as unmerged
+	// since single-ops should be using a fresh journal.
 	if fs.config.Mode() != InitSingleOp {
 		_, rmd, err = fs.config.MDOps().GetForHandle(
 			ctx, tlfHandle, kbfsmd.Unmerged, nil)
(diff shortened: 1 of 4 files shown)


strib:
Oops, this breaks some test.  Hold off on reviewing for now.

strib:
Ok, this is ready for a look now.

libkbfs/kbfs_ops.go
```diff
@@ -470,12 +473,41 @@ func (fs *KBFSOpsStandard) getMaybeCreateRootNode(
 		h.GetCanonicalPath(), branch, create)
 	defer func() { fs.deferLog.CDebugf(ctx, "Done: %#v", err) }()
 
+	// Check if we already have the MD cached, before contacting any
+	// servers.
+	fops := func() *folderBranchOps {
+		fs.opsLock.Lock()
+		defer fs.opsLock.Unlock()
+		return fs.opsByFav[h.ToFavorite()]
+	}()
```
jzila: Can you make this a method instead of a closure?

libkbfs/kbfs_ops.go
```diff
@@ -396,9 +396,12 @@ func (fs *KBFSOpsStandard) getMDByHandle(ctx context.Context,
 		return rmd, nil
 	}
 
-	_, rmd, err = fs.config.MDOps().GetForHandle(ctx, tlfHandle, kbfsmd.Unmerged, nil)
-	if err != nil {
-		return ImmutableRootMetadata{}, err
+	if fs.config.Mode() != InitSingleOp {
```
jzila: Please describe in a comment why this check is necessary.

libkbfs/kbfs_ops.go
```diff
@@ -470,12 +473,41 @@ func (fs *KBFSOpsStandard) getMaybeCreateRootNode(
...
+	}
+
 	// Do GetForHandle() unlocked -- no cache lookups, should be fine
 	mdops := fs.config.MDOps()
-	// TODO: only do this the first time, cache the folder ID after that
-	_, md, err := mdops.GetForHandle(ctx, h, kbfsmd.Unmerged, nil)
-	if err != nil {
-		return nil, EntryInfo{}, err
+	var md ImmutableRootMetadata
+	if fs.config.Mode() != InitSingleOp {
```
jzila: Please describe in a comment why this check is necessary.

Issue #1325, opened 2017-11-09
git ls-remote fails for empty repo


SevereOverfl0w:
I believe the cause is empty repo, I get a different result on empty ssh/git repos for example.

```
$ git ls-remote keybase://team/myteam/test-repo
git-remote-keybase error: (1) stat : no such file or directory
git-remote-keybase error: (1) stat : no such file or directory
```

SevereOverfl0w:
This error goes away if I simply make a local commit to the repo, _without_ pushing.

strib:
Thanks for the report.  I can't repro it though. I tried this:

```sh
# Make empty remote repo
$ keybase git create test2
# Make empty local repo
$ mkdir test2
$ cd test2
$ git init
# Try ls-remote.
$ git ls-remote keybase://private/strib/test2
Initializing Keybase... done.
Syncing with Keybase... done.
Syncing encrypted data to Keybase: (100.00%) 9.47/9.47 KB... done.
$ echo $?
0
```

Is this different from the steps you took?  Also, what version of Keybase are you running, there was a similar bug on an older build a while back.  I can look at logs if you feel like doing a `keybase log send`.

SevereOverfl0w:
I created the git repo through the gui.

I did not do mkdir/cd/git init before doing git ls-remote.

ID of log is ef424eddafe991198aa6d31c

SevereOverfl0w:
I'm running on 1.0.34-20171027161021.a0eab0ba7 btw.

strib:
Oh I see, the difference is you're not running it from within a local git repo.  I can track that down.  Thanks!

strib:
#1327 fixes it for me.

strib:
Should be working in the next release.  Thanks!

PR #1327, opened 2017-11-09
git-remote-keybase: don't check the git dir before taking commands

strib:
`git ls-remote` is allowed to be executed outside of a local git directory, so it's wrong to verify the git dir too early.  Instead, rely on `git` itself failing if it executes a command that requires a git dir, outside of a local repo.

Issue: #1325


strib:
> Looks good, but don't we want to verify it elsewhere now?

Looks like the `git` command already verifies it for any command that uses it, before the remote helper is invoked.  This was really just a failsafe.  If somehow the git checks don't catch it, anything that tries to use it will fail in a similar "No such file or directory" manner as the Stat call in the existing check already does.

Didn't feel worth it to try to track down the various places in kbfsgit where it was used to insert checks in multiple places.

PR #1329, opened 2017-11-10
cr: only resurrect deleted, unmerged rm ops

strib:
If a directory was modified and then deleted on the unmerged branch, but modified on the merged branch, we only need to replay the unmerged "remove" operations.  Anything else like setAttrs and syncs are irrelevant, since the nodes were deleted.

In the course of debugging I factored out some logic into its own function.  I didn't end up needing to do that, but I kept it anyway as a nice cleanup.  It's unrelated to the rest of this PR.

Issue: KBFS-1691
Issue: keybase/client#9478

### DIFF
--- a/libkbfs/conflict_resolver.go
+++ b/libkbfs/conflict_resolver.go
@@ -1394,6 +1394,40 @@ func crConflictCheckQuick(unmergedChain, mergedChain *crChain) bool {
 			(unmergedChain.hasSetAttrOp() && mergedChain.hasSetAttrOp()))
 }
 
+func (cr *ConflictResolver) getSingleUnmergedPath(
+	ctx context.Context, unmergedChains *crChains, chain *crChain) (
+	path, error) {
+	// Reuse some code by creating a new chains object
+	// consisting of only this node.
+	newChains := newCRChainsEmpty()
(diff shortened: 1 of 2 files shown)


songgao:
Btw your branch name suggests https://keybase.atlassian.net/browse/KBFS-2507 . Is that related?

strib:
> Btw your branch name suggests https://keybase.atlassian.net/browse/KBFS-2507 . Is that related?

I thought it was, it's a similar crash, but on second look it's in a different, unrelated place.  I just didn't feel like changing the branch name and having to open a new PR.

libkbfs/conflict_resolver.go
```diff
@@ -981,6 +981,16 @@ func (cr *ConflictResolver) resolveMergedPaths(ctx context.Context,
 			"also modified in the merged branch %v",
 			unmergedChain.mostRecent, mergedChain.mostRecent)
 
+		// We know that everything in the directory has been removed,
+		// so only rm ops matter.
+		var newOps []op
```
jzila: s/`newOps`/`rmOps`

strib:
User confirmed this fixed their issue.

PR #1332, opened 2017-11-11
[KBFS-2592] Add more MDServer/doInit race avoidance

jzila:


jzila:
Don't review yet, CI broken.

jzila:
Works now, ready for review.

libkbfs/keybase_service_base.go
```diff
@@ -355,7 +361,8 @@ func (k *KeybaseServiceBase) StartReachability(ctx context.Context) (res keybase
 // CheckReachability implements keybase1.ReachabilityInterface.
 func (k *KeybaseServiceBase) CheckReachability(ctx context.Context) (res keybase1.Reachability, err error) {
 	res.Reachable = keybase1.Reachable_NO
-	if k.config.MDServer().IsConnected() {
+	mdServer := k.config.MDServer()
+	if mdServer != nil && mdServer.IsConnected() {
```
songgao: If `mdServer` is `nil`, might wanna wait for a few seconds and retry here? Otherwise caller might try to reconnect too soon when we are still starting up.

libkbfs/keybase_service_base.go
```diff
@@ -377,7 +384,10 @@ func (k *KeybaseServiceBase) PaperKeyCached(ctx context.Context,
 		}
 		// Ignore any errors for now, we don't want to block this
 		// notification and it's not worth spawning a goroutine for.
-		k.config.MDServer().CheckForRekeys(context.Background())
+		mdServer := k.config.MDServer()
+		if mdServer != nil {
```
songgao: Perhaps retry here as well (maybe in a new goroutine), since otherwise it could delay rekeys?

libkbfs/keybase_service_base.go
```diff
@@ -331,9 +331,12 @@ func (k *KeybaseServiceBase) KeyfamilyChanged(ctx context.Context,
 	k.clearCachedUnverifiedKeys(uid)
 
 	if k.getCachedCurrentSession().UID == uid {
-		// Ignore any errors for now, we don't want to block this
-		// notification and it's not worth spawning a goroutine for.
-		k.config.MDServer().CheckForRekeys(context.Background())
+		mdServer := k.config.MDServer()
+		if mdServer != nil {
```
songgao: Same here.

libkbfs/keybase_service_base.go
jzila: The retries should happen upstream, shouldn't they?

songgao: Yes, that's the reason why I suggested the wait here. If we are still starting up, there's no point for upstream to call reconnect right?

jzila: But we have no idea why `config.MDServer()` is `nil`. Also this is IMO the wrong place for a retry; if `mdServer` is `nil`, then we aren't reachable by definition.

PR #1335, opened 2017-11-14
folder_branch_status: don't hold locks while accessing journal

strib:
Getting the journal status can end up calling into `folderBlockOps` and taking `blockLock`, in order to populate unflushed paths.

However, `folderBranchStatus.rmNode()` can also be called while holding `folderBlockOps.blockLock`, via `folderBlockOps.UpdatePointers()`.  Thus, these two things can interleave and cause a deadlock.

There's no reason to hold the status lock while calling into the journal, so refactor things a bit to avoid this deadlock.

Issue: KBFS-2614

### DIFF
--- a/libkbfs/folder_branch_status.go
+++ b/libkbfs/folder_branch_status.go
@@ -191,7 +191,7 @@ func (fbsk *folderBranchStatusKeeper) getStatusWithoutJournaling(
 
 	var fbs FolderBranchStatus
 
-	var tlfID tlf.ID
+	tlfID := tlf.NullID
 	if fbsk.md != (ImmutableRootMetadata{}) {
 		tlfID = fbsk.md.TlfID()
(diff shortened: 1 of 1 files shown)


libkbfs/folder_branch_status.go
```diff
@@ -181,27 +181,25 @@ func (fbsk *folderBranchStatusKeeper) convertNodesToPathsLocked(
...
+	ctx context.Context) (
+	FolderBranchStatus, <-chan StatusUpdate, tlf.ID, error) {
 	fbsk.dataMutex.Lock()
 	defer fbsk.dataMutex.Unlock()
 	fbsk.updateMutex.Lock()
 	defer fbsk.updateMutex.Unlock()
 
 	var fbs FolderBranchStatus
 
+	var tlfID tlf.ID
```
jzila: Should this be `tlfID := tlf.NullID`?

PR #1342, opened 2017-11-16
[KBFS-2570] Increase bserver and mdserver token expiry times

jzila:
This needs to be vendored into the servers and deployed before any clients start sending such expire times, otherwise clients will get errors.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -201,8 +201,7 @@ helpers.rootLinuxNode(env, {
                             },
                             integrate: {
                                 build([
-                                    // TODO: Switch this back to master before merging.
-                                    job: "/kbfs-server/PR-441",
+                                    job: "/kbfs-server/master",
                                     parameters: [
                                         [$class: 'StringParameterValue',
                                             name: 'kbfsProjectName',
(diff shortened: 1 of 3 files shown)


jzila:
Yep that's what I was talking about in standup. I have to open a vendored PR and trigger that from this to prove that it works.

strib:
Thanks, just making sure!

songgao:
Are 2 hour tokens a problem? I thought the 10s one was the main pain.

jzila:
@songgao: we're still getting 3-4k connections per minute.

songgao:
3-4K per minute doesn't add up when the token is already 2 hours. I think this one is the bottleneck: https://github.com/keybase/client/pull/8626/files

strib:
@songgao: I thought we decided that one has nothing to do with client reconnections.

I think it's probably people stuck in a kbfs mount/crash loop.  I'm good with bumping it up to a day though, even if it doesn't help much, it seems like a reasonable thing to do.

jzila:
Yeah that PR is for mdserver->API server connections.

strib:
> hmm and it doesn't cause client to resend something at all?

Not to my knowledge, it just causes the mdserver to make more API server calls.

jzila:
Current CI run will fail. I'll merge, re-vendor kbfs-server, merge, then rebuild KBFS.

PR #1343, opened 2017-11-16
 libkbfs: add service implicit team methods

strib:
This is PR 1/3 for implicit team handle stuff.

Issue: KBFS-2598

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -48,7 +48,9 @@ type UserInfo struct {
 // about.
 type TeamInfo struct {
 	// Maybe this should be bare string?  The service doesn't give us
-	// a nice type, unfortunately.
+	// a nice type, unfortunately.  Also note that for implicit teams,
+	// this is an auto-generated name that shouldn't be shown to
+	// users.
 	Name         libkb.NormalizedUsername
 	TID          keybase1.TeamID
 	CryptKeys    map[kbfsmd.KeyGen]kbfscrypto.TLFCryptKey
@@ -66,7 +68,7 @@ type TeamInfo struct {
 // resolving/identifying an implicit team.  TeamInfo is used for
 // anything else.
 type ImplicitTeamInfo struct {
-	Name  libkb.NormalizedUsername
+	Name  libkb.NormalizedUsername // The "display" name for the i-team.
 	TID   keybase1.TeamID
 	TlfID tlf.ID
 }
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -62,6 +62,15 @@ type TeamInfo struct {
 	// whether a user was a member given some Merkle seqno?
 }
 
+// ImplicitTeamInfo contains information needed after
+// resolving/identifying an implicit team.  TeamInfo is used for
+// anything else.
+type ImplicitTeamInfo struct {
+	Name  libkb.NormalizedUsername
+	TID   keybase1.TeamID
+	TlfID tlf.ID
+}
+
 // SessionInfo contains all the info about the keybase session that
 // kbfs cares about.
 type SessionInfo struct {
--- a/libkbfs/keybase_daemon_local.go
+++ b/libkbfs/keybase_daemon_local.go
@@ -250,6 +250,29 @@ func (k *KeybaseDaemonLocal) Identify(
 	return k.Resolve(ctx, assertion)
 }
 
+func (k *KeybaseDaemonLocal) resolveForImplicitTeam(
+	ctx context.Context, name string, r []libkb.NormalizedUsername,
+	ur []keybase1.SocialAssertion,
+	resolvedIDs map[libkb.NormalizedUsername]keybase1.UserOrTeamID) (
+	[]libkb.NormalizedUsername, []keybase1.SocialAssertion, error) {
+	id, err := k.assertionToIDLocked(ctx, name)
+	if err == nil {
+		u, err := k.localUsers.getLocalUser(id.AsUserOrBust())
+		if err != nil {
+			return nil, nil, err
+		}
+		r = append(r, u.Name)
+		resolvedIDs[u.Name] = id
+	} else {
+		a, ok := externals.NormalizeSocialAssertion(name)
+		if !ok {
+			return nil, nil, fmt.Errorf("Bad assertion: %s", name)
+		}
+		ur = append(ur, a)
+	}
+	return r, ur, nil
+}
+
 // ResolveIdentifyImplicitTeam implements the KeybaseService interface
 // for KeybaseDaemonLocal.
 func (k *KeybaseDaemonLocal) ResolveIdentifyImplicitTeam(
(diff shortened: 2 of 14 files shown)


libkbfs/keybase_daemon_local.go
```diff
@@ -236,6 +250,139 @@ func (k *KeybaseDaemonLocal) Identify(
...
+
+	// Canonicalize the name.
+	writerNames, readerNames, _, err :=
+		splitAndNormalizeTLFName(assertions, tlfType)
+	if err != nil {
+		return ImplicitTeamInfo{}, err
+	}
+	var writers, readers []libkb.NormalizedUsername
+	var unresolvedWriters, unresolvedReaders []keybase1.SocialAssertion
+	resolve := func(
```
jzila: It seems like the only 2 things needed for this closure are `k` and `ctx`, in which case I think it makes sense to have this be its own helper function. It's possible that the Go compiler these days is good enough to optimize out closures when it can, but I'd prefer to continue separating them out just in case.

libkbfs/keybase_daemon_local.go
```diff
@@ -236,6 +250,139 @@ func (k *KeybaseDaemonLocal) Identify(
...
+	name := tlf.MakeCanonicalName(
+		writers, unresolvedWriters, readers, unresolvedReaders, extensions)
+
+	key := fmt.Sprintf("%s:%s", tlfType.String(), name)
+	tid, ok := k.implicitAsserts[key]
+	if ok {
+		return k.localImplicitTeams[tid], nil
+	}
+
+	if !doIdentifies {
```
jzila: So `doIdentifies` is a hard fail when the team isn't cached?

libkbfs/keybase_daemon_local.go
```diff
@@ -236,6 +250,139 @@ func (k *KeybaseDaemonLocal) Identify(
...
+
+	// Need to make the team info as well, so get the list of user
+	// names and resolve them.
+	asUserName := libkb.NormalizedUsername(name)
+	teams := makeLocalTeams(
+		[]libkb.NormalizedUsername{asUserName}, len(k.localTeams), tlfType)
+	info := teams[0]
+	info.Writers = make(map[keybase1.UID]bool, len(writerNames))
+	for _, w := range writerNames {
+		id, err := k.assertionToIDLocked(ctx, w)
```
jzila: We've already done this in the `resolve` function above, and an error in resolving to either a name or social assertion will have caused this whole function to fail. As a bonus, we also already have the social assertions in `unresolved{Readers,Writers}`. So maybe this loop can iterate over `range resolvedWriters` instead of `writerNames`. Not sure how you'd like to handle the social assertions.

libkbfs/keybase_daemon_local.go
```diff
@@ -236,6 +250,139 @@ func (k *KeybaseDaemonLocal) Identify(
...
+	iteamInfo := ImplicitTeamInfo{
+		// TODO: allow display name to differ?
+		Name: asUserName,
+		TID:  tid,
+	}
+	k.localImplicitTeams[tid] = iteamInfo
+	return iteamInfo, nil
+}
+
+func (k *KeybaseDaemonLocal) addImplicitTeamTlfID(
```
jzila: Where is this used?

libkbfs/keybase_daemon_local.go
strib: This isn't a cache, this is the local/test implementation of the keybase daemon.  I was trying to use `doIdentifies` as a proxy for the "create" parameter we pass to the service, but I think that doesn't work too well yet.  I ended up adding the `addImplicitTeamTlfID` method instead, so I don't think this code is even used anymore.  I'll remove it.

libkbfs/keybase_daemon_local.go
strib: In future PR #1345.

libkbfs/keybase_daemon_local.go
strib: Actually nevermind, I take it back, this code is still used in a future PR.  I'll make things more clear here.

libkbfs/data_types.go
```diff
@@ -62,6 +62,15 @@ type TeamInfo struct {
 	// whether a user was a member given some Merkle seqno?
 }
 
+// ImplicitTeamInfo contains information needed after
+// resolving/identifying an implicit team.  TeamInfo is used for
+// anything else.
```
songgao: Is this gonna be used for single teams as well, or is the plan to use the old code path for single teams since we won't need a resolve for them? If former, it sounds like this could use a different name.

libkbfs/data_types.go
```diff
@@ -62,6 +62,15 @@ type TeamInfo struct {
...
+type ImplicitTeamInfo struct {
+	Name  libkb.NormalizedUsername
```
songgao: This is the team display name right? Perhaps worth a comment since the type says username. Same for TeamInfo.

libkbfs/keybase_daemon_local.go
```diff
@@ -236,6 +250,146 @@ func (k *KeybaseDaemonLocal) Identify(
...
+	// Unresolved users don't need to go in the team info, they're
+	// irrelvant until they're resolved.  TODO: add resolved users
+	// into existing teams they should be on.
+
+	tid = teams[0].TID
+	k.implicitAsserts[key] = tid
+	k.localTeams[tid] = info
+
+	iteamInfo := ImplicitTeamInfo{
+		// TODO: allow display name to differ?
```
songgao: Could you elaborate a bit on this TODO? What would a differed name be and what is it for?

libkbfs/data_types.go
strib: I think it won't be used for single teams.  Though once we start storing the TLF ID in the sigchain for regular teams, it probably makes sense to either add or move the TLF ID over to `TeamInfo`.

Note that the `Name` fields of the two structs are different.  For implicit teams, the `TeamInfo` will contain the auto-generated name, while the `ImplicitTeamInfo` contains the nice "display" name.

libkbfs/keybase_daemon_local.go
strib: There are two things here.  One is that implicit teams have an auto-generated name in `TeamInfo` that something like "_keybase_implicit_team_XXX" (I forget the format exactly).  It's only used internally and shouldn't be shown to users.    I should probably put a different TODO for that up where the `TeamInfo` is constructed.

The second thing is that the "display name" is the _preferred_ canonical name, with the logged-in user listed first, and all the other users in alphabetical order.  The `MakeCanonicalName` above doesn't capture this. I'll update the comment.

libkbfs/data_types.go
songgao: Got you; sounds good. Thanks for clarifying that and adding comment!

libkbfs/keybase_daemon_local.go
songgao: I see. Thanks. How is the auto-generated name relevant here? This one is always display name right?

strib: Yep, I meant the "to differ" part in the comment has two meaning, one of which is that it differs from the implicit team name.  But as I said, I should (and did) fix that above where the team info is made.  I just had it create an implicit team name there.

PR #1344, opened 2017-11-16
libkbfs: add tlfID to TlfHandle, and an ID-getting interface

strib:
This PR doesn't populate the TLF ID yet, only adds it to the data structures and adds the method for getting it.

This is PR 2/3 for the implicit team handle stuff.

Issue: KBFS-2598

### DIFF
--- a/libkbfs/md_ops.go
+++ b/libkbfs/md_ops.go
@@ -303,6 +303,7 @@ func (md *MDOpsStandard) processMetadata(ctx context.Context,
 }
 
 // GetForHandle implements the MDOps interface for MDOpsStandard.
+// TODO: unexport this when no longer part of the MDOps interface.
 func (md *MDOpsStandard) GetForHandle(ctx context.Context, handle *TlfHandle,
 	mStatus kbfsmd.MergeStatus, lockBeforeGet *keybase1.LockID) (
 	id tlf.ID, rmd ImmutableRootMetadata, err error) {
--- a/libkbfs/md_ops.go
+++ b/libkbfs/md_ops.go
@@ -295,9 +295,7 @@ func (md *MDOpsStandard) processMetadata(ctx context.Context,
 	*rmds = RootMetadataSigned{}
 	irmd := MakeImmutableRootMetadata(rmd, key, mdID, localTimestamp, true)
 
-	// Revisions created locally should always override anything else
-	// in the cache.
-	err = md.config.MDCache().Replace(irmd, irmd.BID())
+	err = md.config.MDCache().Put(irmd)
 	if err != nil {
 		return ImmutableRootMetadata{}, err
 	}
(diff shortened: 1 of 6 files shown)


libkbfs/md_ops.go
```diff
@@ -295,9 +295,7 @@ func (md *MDOpsStandard) processMetadata(ctx context.Context,
 	*rmds = RootMetadataSigned{}
 	irmd := MakeImmutableRootMetadata(rmd, key, mdID, localTimestamp, true)
 
-	// Revisions created locally should always override anything else
-	// in the cache.
-	err = md.config.MDCache().Replace(irmd, irmd.BID())
+	err = md.config.MDCache().Put(irmd)
```
strib: This was just something unrelated I noticed.  Fetched MDs should never replace local MDs in the cache.

libkbfs/tlf_handle.go
```diff
@@ -235,7 +248,7 @@ func (h TlfHandle) Extensions() (extensions []tlf.HandleExtension) {
 }
 
 func init() {
-	if reflect.ValueOf(TlfHandle{}).NumField() != 8 {
+	if reflect.ValueOf(TlfHandle{}).NumField() != 9 {
```
jzila: Hmm, what is this check doing? I.e. isn't this tautalogical once we've compiled?

strib: See the text of the panic just below this -- the check was responsible for making me remember I had to update the `Equals` function to check the tlf ID.

libkbfs/md_ops.go
songgao: Could you explain why it should never replace? Is it not only for stuff from server, but also local new MDs?

libkbfs/md_ops.go
```diff
@@ -308,6 +306,27 @@ func (md *MDOpsStandard) processMetadata(ctx context.Context,
 func (md *MDOpsStandard) GetForHandle(ctx context.Context, handle *TlfHandle,
 	mStatus kbfsmd.MergeStatus, lockBeforeGet *keybase1.LockID) (
 	id tlf.ID, rmd ImmutableRootMetadata, err error) {
+	// If we already know the tlf ID, we shouldn't be calling this
+	// function.
```
songgao: If I understand correctly, this would now only be used for classic TLFs, and will be deprecated after we migrate everything to implicit teams? Can we mention it in its comment in MDOps interface?

libkbfs/md_ops.go
strib: `processMetadata` is only used on the get path, not the put path, so it doesn't get called for new local MDs.  The put path uses Replace, as intended.

libkbfs/md_ops.go
strib: In the next PR, I actually remove all uses of this function outside of MDOps implementations.  So I can probably make a PR to unexport it and remove it from the interface entirely.

strib: (I'll add a TODO in this PR for that.)

libkbfs/md_ops.go
songgao: I see; so using `Replace` here would cause overriding MDs when there are new local MDs not flushed to server. Does it sound correct?

PR #1345, opened 2017-11-17
 libkbfs: populate the TlfHandle with a TLF ID, and resolve i-team

strib:
When parsing a TLF handle, we now first check whether there's a matching implicit team with a TLF ID populated.  If so, we use that handle.  Right now this will never happen, since the service doesn't return a TLF ID for any implicit team.

For all other handles, we look up the TLF ID as part of the handle-making process, and put it in the handle.

This adds at least one extra RTT to the servers during handle parsing, which is unfortunate while e.g. listing favorites, but it's necessary for the implicit team design.  This would hurt us badly when listing favorites, but luckily we don't need to know these until someone actually tries to access data within the folder.  So this also adds a method in `libfs` for faking out these time-consuming calls during favorites listing, and then updating the handle later when it is really needed.

PR 3/3 for the implicit team handle stuff.

Issue: KBFS-2598

### DIFF
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -183,7 +183,7 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp 
 	}
 
 	h, err := libkbfs.ParseTlfHandlePreferred(
-		ctx, fl.fs.config.KBPKI(), req.Name, fl.tlfType)
+		ctx, fl.fs.config.KBPKI(), fl.fs.config.MDOps(), req.Name, fl.tlfType)
 	switch err := err.(type) {
 	case nil:
 		// no error
@@ -278,7 +278,7 @@ func (fl *FolderList) Remove(ctx context.Context, req *fuse.RemoveRequest) (err 
 	defer func() { err = fl.fs.processError(ctx, libkbfs.WriteMode, err) }()
 
 	h, err := libkbfs.ParseTlfHandlePreferred(
-		ctx, fl.fs.config.KBPKI(), req.Name, fl.tlfType)
+		ctx, fl.fs.config.KBPKI(), fl.fs.config.MDOps(), req.Name, fl.tlfType)
 
 	switch err := err.(type) {
 	case nil:
--- a/libfuse/folderlist.go
+++ b/libfuse/folderlist.go
@@ -15,6 +15,7 @@ import (
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
 	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libkbfs"
 	"github.com/keybase/kbfs/tlf"
 	"golang.org/x/net/context"
@@ -182,8 +183,8 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp 
 		return nil, fuse.ENOENT
 	}
 
-	h, err := libkbfs.ParseTlfHandlePreferred(
-		ctx, fl.fs.config.KBPKI(), fl.fs.config.MDOps(), req.Name, fl.tlfType)
+	h, err := libfs.ParseTlfHandlePreferredQuick(
+		ctx, fl.fs.config.KBPKI(), req.Name, fl.tlfType)
 	switch err := err.(type) {
 	case nil:
 		// no error
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -34,14 +34,19 @@ type resolvableUser interface {
 	//   - return a non-zero nameIDPair;
 	//   - return a non-zero keybase1.SocialAssertion;
 	//   - return a non-nil error.
-	resolve(context.Context) (nameIDPair, keybase1.SocialAssertion, error)
+	//
+	// The TLF ID may or may not be filled in, depending on whether
+	// the name is backed by an implicit team.
+	resolve(context.Context) (
+		nameIDPair, keybase1.SocialAssertion, tlf.ID, error)
 }
 
 func resolveOneUser(
 	ctx context.Context, user resolvableUser,
 	errCh chan<- error, userInfoResults chan<- nameIDPair,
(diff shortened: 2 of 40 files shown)


strib:
Sorry this is kinda huge.  I did my best to split it up into 3 related PRs (review #1343 and #1344 first), but I couldn't think of a good way to split up this one further.

Suggested reviewing order:
* `tlf/name.go`
* `libkbfs/tlf_handle_resolve.go`
* `libkbfs/tlf_handle_test.go`
* `libkbfs/md_ops.go`
* `libkbfs/kbfs_ops.go`.
* `libfs/tlf_handle_util.go`
* `libfuse/` and `libdokan/`.
* Everything else, which are basically just mechanical param-passing changes.

libfuse/folderlist.go
```diff
@@ -182,7 +183,7 @@ func (fl *FolderList) Lookup(ctx context.Context, req *fuse.LookupRequest, resp
 		return nil, fuse.ENOENT
 	}
 
-	h, err := libkbfs.ParseTlfHandlePreferred(
+	h, err := libfs.ParseTlfHandlePreferredQuick(
```
jzila: What if we had an RPC to resolve many handles in one call, and then cache them in KBFS for later?

libkbfs/tlf_handle_resolve.go
```diff
@@ -121,6 +133,15 @@ func makeTlfHandleHelper(
 			return nil, ctx.Err()
 		}
 	}
+	close(idc)
```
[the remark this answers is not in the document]
songgao: Yeah I was a bit confused too. Perhaps also comment in resolveOneUser that sending to idResults has to be before others.

libkbfs/tlf_handle_resolve.go
```diff
@@ -195,28 +236,32 @@ type resolvableID struct {
...
-func (ruid resolvableID) resolve(ctx context.Context) (nameIDPair, keybase1.SocialAssertion, error) {
+func (ruid resolvableID) resolve(ctx context.Context) (
+	nameIDPair, keybase1.SocialAssertion, tlf.ID, error) {
 	name, err := ruid.nug.GetNormalizedUsername(ctx, ruid.id)
 	if err != nil {
-		return nameIDPair{}, keybase1.SocialAssertion{}, err
+		return nameIDPair{}, keybase1.SocialAssertion{}, tlf.NullID, err
 	}
+	// TODO(KBFS-2621): get the TLF ID and display name for implicit
+	// teams. Blocked on CORE-6623.
```
songgao: In that case, why don't we just call the to-be-added CreateTLF RPC into service here so that at the time the handle is returned it's guaranteed to have a TLF ID?

libkbfs/tlf_handle_resolve.go
```diff
@@ -34,14 +34,19 @@ type resolvableUser interface {
...
+		nameIDPair, keybase1.SocialAssertion, tlf.ID, error)
 }
 
 func resolveOneUser(
 	ctx context.Context, user resolvableUser,
 	errCh chan<- error, userInfoResults chan<- nameIDPair,
-	socialAssertionResults chan<- keybase1.SocialAssertion) {
-	userInfo, socialAssertion, err := user.resolve(ctx)
+	socialAssertionResults chan<- keybase1.SocialAssertion,
+	idResults chan<- tlf.ID) {
```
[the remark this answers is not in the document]
strib: I'll change this to make the channel have only a size-1 buffer, and `resolveOneUser` will error if it can't send anything on it.  Then I'll use an explicit `select`, rather than `range`, to get the value from the channel.

libkbfs/tlf_journal_test.go
```diff
@@ -148,6 +148,10 @@ func (c testTLFJournalConfig) teamMembershipChecker() kbfsmd.TeamMembershipCheck
 	return nil
 }
 
+func (c testTLFJournalConfig) tlfIDGetter() tlfIDGetter {
+	return nil
```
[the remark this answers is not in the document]
strib: Good question.  `nil` means it should skip trying the check altogether -- which is actually what the `libfs` code should be doing.  Null IDs are going to eventually be the signal to generate a TLF ID on the client, so I think `libfs` should be passing in a `nil` instead.  I'll change it.

libfuse/folderlist.go
strib: Sure, if the service gave us something like that it might make things a bit more efficient.  Though if we used it by default on an `ls`, it might be risky for people with tons of favorites who could otherwise use bare `/bin/ls` to avoid the overhead of resolving all their favorites.

libkbfs/tlf_handle_resolve.go
strib: @songgao: I don't understand your comment.  `resolvableID` is used when we start with a iteam ID (say, due to an incoming MD server notification, and we are reconstructing the handle from the MD) and need to resolve it backwards into a handle.  Can you clarify what that has to do with creating a TLF ID?

(It's possible this path won't ever be used for implicit teams, since the one source of incoming notifications without pre-registration I can think of -- rekeys -- won't be relevant.)

songgao: whoops I put the comment in the wrong place when I was moving from my local git here after I got Internet. I'll add the comment in the right place. Sorry!

libkbfs/tlf_handle_resolve.go
```diff
@@ -557,11 +674,16 @@ func parseTlfHandleLoose(
...
-	h, err := makeTlfHandleHelper(ctx, t, writers, readers, extensions)
+	h, err := makeTlfHandleHelper(
+		ctx, t, writers, readers, extensions, idGetter)
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO(KBFS-2621): When we want to start creating new implicit
+	// team TLFs, and h.tlfID is null, return `iteamHandle` instead of
+	// `h`.
```
songgao: In that case, why don't we just call the to-be-added CreateTLF RPC into service here so that at the time the handle is returned it's guaranteed to have a TLF ID?

strib: Yep, that's the plan!  Mentioned in the description for KBFS-2621.

PR #1348, opened 2017-11-18
server side iTeam supporting methods for bare handle and mdv3

songgao:


tlf/handle.go
```diff
@@ -149,6 +149,22 @@ func MakeHandle(
...
+// be either a SingleTeam TLF or a private/public TLF backed by an implicit
+// team.
+func (h Handle) IsBackedByTeam() bool {
+	if len(h.Writers) != 1 ||
+		len(h.Readers) != 0 ||
+		len(h.UnresolvedReaders) != 0 ||
+		len(h.UnresolvedWriters) != 0 {
+		return false
+	}
+	if _, err := h.Writers[0].AsTeam(); err != nil {
```
strib: You can just return `h.Writers[0].IsTeam()` I think.

kbfsmd/root_metadata_v3.go
```diff
@@ -645,11 +645,35 @@ func (md *RootMetadataV3) CheckValidSuccessorForServer(
...
+	if extra != nil {
+		return false
+	}
+	if len(md.WriterMetadata.UnresolvedWriters) != 0 {
+		return false
+	}
+	if len(md.WriterMetadata.Writers) != 1 {
+		return false
+	}
+	if _, err := md.WriterMetadata.Writers[0].AsTeam(); err != nil {
```
strib: You can just return `md.WriterMetadata.Writers[0].IsTeam()` I think.

kbfsmd/root_metadata_v3.go
```diff
@@ -669,16 +693,20 @@ func (md *RootMetadataV3) MakeBareTlfHandle(extra ExtraMetadata) (
...
+				"unexpected md for public TLF; got non-nil extra")
 		}
-
 		writers = md.WriterMetadata.Writers
-		if md.TlfID().Type() == tlf.Public {
-			readers = []keybase1.UserOrTeamID{keybase1.PublicUID.AsUserOrTeam()}
-		}
+		readers = []keybase1.UserOrTeamID{keybase1.PublicUID.AsUserOrTeam()}
+	case tlf.SingleTeam:
+		// We already cover the SingleTeam case at the beginnin with
```
strib: "beginnin" -> "beginning"

kbfsmd/root_metadata_v3.go
```diff
@@ -870,7 +898,12 @@ func (md *RootMetadataV3) IsValidAndSigned(
 	ctx context.Context, codec kbfscodec.Codec,
 	teamMemChecker TeamMembershipChecker, extra ExtraMetadata,
 	writerVerifyingKey kbfscrypto.VerifyingKey) error {
-	if md.TlfID().Type() == tlf.Private {
+	if md.IsBackedByTeam(extra) {
+		return nil
```
strib: This skips a bunch of checks below the `switch` that team TLFs usually undergo.

kbfsmd/root_metadata_v3.go
```diff
@@ -645,11 +645,35 @@ func (md *RootMetadataV3) CheckValidSuccessorForServer(
 	return nil
 }
 
+// IsBackedByTeam returns true if md is for a TLF backed by a team. It could be
+// either a SingleTeam TLF or a private/public TLF backed by an implicit team.
+func (md *RootMetadataV3) IsBackedByTeam(extra ExtraMetadata) bool {
+	if extra != nil {
```
strib: If you keep the `extra`argument (see below), I think this check should go below the team check, and it should return an error if `extra != nil`, because that's not supposed to happen.

kbfsmd/root_metadata_v3.go
```diff
@@ -645,11 +645,35 @@ func (md *RootMetadataV3) CheckValidSuccessorForServer(
...
 func (md *RootMetadataV3) MakeBareTlfHandle(extra ExtraMetadata) (
 	tlf.Handle, error) {
+	if md.IsBackedByTeam(extra) {
+		return tlf.MakeHandle(md.WriterMetadata.Writers,
+			nil, nil, nil, md.TlfHandleExtensions())
+	}
+
 	var writers, readers []keybase1.UserOrTeamID
-	if md.TlfID().Type() == tlf.Private {
+	switch md.TlfID().Type() {
```
strib: My idea was to add a method like this following to the `kbfsmd.RootMetadata` interface:

```go
TypeForKeying() tlf.Type
```

The v3 implementation would be something like:

```go
func (md *RootMetadataV3) TypeForKeying() tlf.Type {
  if md.IsBackedByTeam() {  // Would need to drop the `extra` argument
    return tlf.SingleTeam
  }
  return tlf.Type()
}
```

Then we can just replace `md.TlfID().Type()` with `md.TypeForKeying()` everywhere that an implicit team TLF should be treated as a team TLF.  This would make KBFS-2600 much easier.  What do you think?

songgao: Oh I see; I guess that makes sense, although it's a bit confusing to use `tlf.SingleTeam`. Would it be similar amount of work to have these in `tlf` and use them instead, when what we care is TypeForKeying rather than the actual TLF type?

```
type TypeForKeying Type
const (
    UnknownKeyingType TypeForKeying = iota
    PrivateKeying
    PublicKeying
    TeamKeying
)
```

If not, then I'm OK with your proposed approach too as long as there are comments to remind us of the design!

strib: Sure, I guess adding that new type isn't too much more work.  Sounds good, thanks!

songgao: Cool; will do. thank

kbfsmd/root_metadata_v3.go
```diff
@@ -645,11 +645,35 @@ func (md *RootMetadataV3) CheckValidSuccessorForServer(
 	return nil
 }
 
+// IsBackedByTeam returns true if md is for a TLF backed by a team. It could be
+// either a SingleTeam TLF or a private/public TLF backed by an implicit team.
+func (md *RootMetadataV3) IsBackedByTeam(extra ExtraMetadata) bool {
```
jzila: Please place this above `MakeInitialRootMetadataV3` where it's used.

kbfsmd/root_metadata_v3.go
```diff
@@ -645,11 +645,35 @@ func (md *RootMetadataV3) CheckValidSuccessorForServer(
...
+		return false
+	}
+	return true
+}
+
 // MakeBareTlfHandle implements the RootMetadata interface for RootMetadataV3.
 func (md *RootMetadataV3) MakeBareTlfHandle(extra ExtraMetadata) (
 	tlf.Handle, error) {
+	if md.IsBackedByTeam(extra) {
+		return tlf.MakeHandle(md.WriterMetadata.Writers,
```
jzila: How does this reconcile with the `TlfHandle` work that @strib is doing? Specifically, it seems that the handle produced on the client will contain the correct set of `{resolved,unresolved}{Readers,Writers}`, but the code here just seems to take the implicit team ID from the `md.WriterMetadata.Writers`.

Am I missing something?

songgao: For team backed TLFs, bare handles would be mostly empty. It only contains an (implicit) team ID in the resolved writers list. In fact we use this attribute to determine whether it's a team backed TLF (i.e. SingleTeam or iTeam-backed TLF) or not.

kbfsmd/root_metadata_v3.go
songgao: Hmm that one uses the one defined on `tlf.Handle`. In the new version it's gonna be a different method though.

kbfsmd/root_metadata_v3.go
```diff
@@ -208,12 +208,12 @@ func (extra ExtraMetadataV3) IsReaderKeyBundleNew() bool {
 // must be done separately.
 func MakeInitialRootMetadataV3(tlfID tlf.ID, h tlf.Handle) (
 	*RootMetadataV3, error) {
-	if tlfID.Type() != h.Type() {
+	if h.TypeForKeying() != tlf.TeamKeying && tlfID.Type() != h.Type() {
```
strib: We can also check that if `tlfID.Type() == tlf.SingleTeam` implies `h.Type() == tlf.SingleTeam`, when `h.TypeForKeying() == tlf.TeamKeying`, for completeness.

tlf/id.go
```diff
@@ -88,6 +88,43 @@ func TypeFromFolderType(ft keybase1.FolderType) Type {
...
 }
 
+// KeyingType represents a TLF keying mode. It normally have the same values
+// as Type.
+type KeyingType Type
+
+const (
+	// UnknownKeying is a placeholder type for when TLF keying mode is unknown.
+	UnknownKeying = KeyingType(Unknown)
+	// PrivateKeying specifies the  TLF keying mode used in classic private TLF
```
strib: extra space in "the  TLF" (here and below).

tlf/id.go
```diff
@@ -88,6 +88,43 @@ func TypeFromFolderType(ft keybase1.FolderType) Type {
...
+	// folders.
```
strib: "TLF folders" is redundant, since the "F" stands for "folder".  Can just be "TLFs" (here and below).

tlf/id.go
songgao: In the name of human's curiosity in exploring the space.

kbfsmd/root_metadata_v3.go
```diff
@@ -208,12 +208,16 @@ func (extra ExtraMetadataV3) IsReaderKeyBundleNew() bool {
 // must be done separately.
 func MakeInitialRootMetadataV3(tlfID tlf.ID, h tlf.Handle) (
 	*RootMetadataV3, error) {
-	if tlfID.Type() != h.Type() {
+	switch {
+	case h.TypeForKeying() == tlf.TeamKeying && tlfID.Type() != tlf.SingleTeam:
```
songgao: @strib I assumed this is what you meant; but let me know if I'm wrong!

strib: I think this is wrong, because a private, implicit-teamed-backed folder would fail this test.  What I meant was:

```go
case h.TypeForKeying() == tlf.TeamKeying && tlfID.Type() == tlf.SingleTeam && h.Type() != tlf.SingleTeam:
    fallthrough
```

tlf/id.go
```diff
@@ -95,11 +95,9 @@ type KeyingType Type
 const (
 	// UnknownKeying is a placeholder type for when TLF keying mode is unknown.
 	UnknownKeying = KeyingType(Unknown)
-	// PrivateKeying specifies the  TLF keying mode used in classic private TLF
-	// folders.
+	// PrivateKeying specifies the TLF keying mode used in classic private TLFs.
 	PrivateKeying = KeyingType(Private)
-	// PublicKeying specifies the  TLF keying mode used in classic public TLF
-	// folders.
+	// PublicKeying specifies the  TLF keying mode used in classic public TLFs.
```
strib: The `TeamKeying` line below also has an extra space.

kbfsmd/root_metadata_v3.go
songgao: Ah that's right. Clearly my body's still in west coast time!

PR #1351, opened 2017-11-21
tlf_handle: don't call resolve implicit teams for now

strib:
It has bad performance.


songgao:
(assuming CI passes)

mmaxim:
@cjb This is the PR.

Issue #1352, opened 2017-11-21
`Journal not enabled` error for git-remote-keybase


the-jackalope:
Hi there,

Keybase had been working flawlessly for me until I tried to update one of our Keybase git repos today:
```
$ git pull origin master
Initializing Keybase... done.
Syncing with Keybase... done.
git-remote-keybase error: (1) Journal not enabled for eee714808a2c935b80da9b40a0ebb926
```

I'm not sure what prompted this error, but I am now unable to clone, pull, push, or perform any operations to any of our Keybase git repos! I've tried restarting the Keybase app, restarting my laptop, and re-logging in via the keybase command line tool - the error message persists.

Was wondering if anyone here had any helpful pointers as to what might be going on - thanks!

**OS**: macOS Sierra 10.12.6
**git version**: 2.15.0
**Keybase version**: 1.0.36-20171121022443+587efe057

strib:
Thanks we're looking at it now, sorry.

PR #1353, opened 2017-11-21
journal_md_ops: enable journal in `GetIDForHandle`

strib:
If `journalMDOps` doesn't have access to the handle before the `GetForTLF` call comes in, it won't be able to enable the journal for team folders.  So override `GetIDForHandle` and use it to start up the journal using the handle.

Issue: #1352


strib:
@jzila: I filed KBFS-2626.

PR #1355, opened 2017-11-21
md_ops: use a cache for handle->ID mappings

strib:
`GetIDForHandle` is called frequently by chat now that we need to get the ID when parsing a handle, so cache it so we don't end up having to go to the journal/server to get the ID.

cc: @mmaxim I believe this gets us back to the previous performance levels for everything but the first `GetTLFCryptKeys` call for a given TLF, which involves server RTTs anyway.

Issue: KBFS-2598

### DIFF
--- a/libkbfs/mdcache.go
+++ b/libkbfs/mdcache.go
@@ -144,3 +144,20 @@ func (md *MDCacheStandard) PutIDForHandle(handle *TlfHandle, id tlf.ID) error {
 	md.idLRU.Add(key, id)
 	return nil
 }
+
+// ChangeHandleForID implements the MDCache interface for
+// MDCacheStandard.
+func (md *MDCacheStandard) ChangeHandleForID(
+	oldHandle *TlfHandle, newHandle *TlfHandle) {
+	md.lock.RLock()
+	defer md.lock.RUnlock()
+	oldKey := oldHandle.GetCanonicalPath()
+	tmp, ok := md.idLRU.Get(oldKey)
+	if !ok {
+		return
+	}
+	md.idLRU.Remove(oldKey)
+	newKey := newHandle.GetCanonicalPath()
+	md.idLRU.Add(newKey, tmp)
+	return
+}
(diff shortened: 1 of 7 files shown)


libkbfs/mdcache.go
```diff
@@ -111,3 +117,30 @@ func (md *MDCacheStandard) MarkPutToServer(
...
+	md.lock.RLock()
+	defer md.lock.RUnlock()
+	key := handle.GetCanonicalPath()
+	tmp, ok := md.idLRU.Get(key)
+	if !ok {
+		return tlf.NullID, errors.WithStack(NoSuchTlfIDError{handle})
+	}
+	id, ok := tmp.(tlf.ID)
+	if !ok {
+		return tlf.NullID, errors.Errorf("Bad ID for handle %s", key)
```
jzila: Shouldn't we wipe it from the cache and return a `NoSuchTlfIDError`?

libkbfs/mdcache.go
```diff
@@ -111,3 +117,30 @@ func (md *MDCacheStandard) MarkPutToServer(
...
+
+// GetIDForHandle implements the MDCache interface for
+// MDCacheStandard.
+func (md *MDCacheStandard) GetIDForHandle(handle *TlfHandle) (tlf.ID, error) {
+	md.lock.RLock()
+	defer md.lock.RUnlock()
+	key := handle.GetCanonicalPath()
+	tmp, ok := md.idLRU.Get(key)
+	if !ok {
+		return tlf.NullID, errors.WithStack(NoSuchTlfIDError{handle})
```
songgao: Is a stack here useful for debugging? I wonder how costly `runtime.Callers` is, since we'd get a cache miss here at beginning, for every TLF we try to get the ID for.

songgao:
Also I tried this locally, and from service logs the latency for getting crypt keys mostly stays under 10ms for me. It occasionally bumps above 10ms but normally doesn't exceed 20ms.

libkbfs/mdcache.go
strib: I was being consistent with the other Get method in this class.  Also, if there's a type error like this I'd rather know about it than hide it, when we're talking about the client side.  I'd be happy to panic here if you prefer.

libkbfs/mdcache.go
strib: Ok, I'll remove it.

PR #1356, opened 2017-11-22
libgit: construct the git Config before getting the TLF handle

strib:
Otherwise, a journal will be created for the TLF under the normal Config, and later journal calls on the git Config will fail due to a missing handle.


jzila:
@strib: ok TestGitTeams passes, 🚢 it

PR #1357, opened 2017-11-22
libkbfs: add RootMetadata support for implicit teams

strib:
This PR adds MD support for implicit teams, including initializing them, keying them like regular teams, and creating them from a bare MD/handle.  The latter requires passing an explicit type into `MakeTlfHandle`, because the bare handle is ambiguous.

It also adds DSL tests for implicit teams.

This PR assumes that public-implicit-team TLFs will be able to fetch "keys" from the service, just like with any other team.  I still need to confirm that with CORE.

Issue: KBFS-2600

### DIFF
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -303,9 +303,9 @@ func (rsa resolvableSocialAssertion) resolve(ctx context.Context) (
 }
 
 // MakeTlfHandle creates a TlfHandle from the given tlf.Handle and the
-// given normalizedUsernameGetter (which is usually a KBPKI). Note
-// that `t` could be different from `bareHandle.Type()`, if this is an
-// implicit team TLF.
+// given normalizedUsernameGetter (which is usually a KBPKI). `t` is
+// the `tlf.Type` of the new handle.  (Note this could be different
+// from `bareHandle.Type()`, if this is an implicit team TLF.)
 func MakeTlfHandle(
 	ctx context.Context, bareHandle tlf.Handle, t tlf.Type,
 	nug normalizedUsernameGetter, idGetter tlfIDGetter) (*TlfHandle, error) {
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -253,12 +253,18 @@ func makeTlfHandleHelper(
 		h.tlfID = tlfID
 	}
 
+	if h.tlfID != tlf.NullID && h.tlfID.Type() != t {
+		return nil, fmt.Errorf("ID type=%s doesn't match expected type=%s",
+			h.tlfID.Type(), t)
+	}
+
 	return h, nil
 }
 
 type resolvableID struct {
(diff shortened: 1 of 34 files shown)


libkbfs/tlf_handle_resolve.go
```diff
@@ -268,7 +274,21 @@ func (ruid resolvableID) resolve(ctx context.Context) (
...
 	}
 	// TODO(KBFS-2621): get the TLF ID and display name for implicit
-	// teams. Blocked on CORE-6623.
+	// teams properly, using a proper service RPC rather than hacking
+	// the tlfIDGetter like this. Blocked on CORE-6623.
+	if doResolveImplicit(ctx) && ruid.id.IsTeamOrSubteam() &&
+		ruid.idGetter != nil {
+		// Passing in a nil handle will only work if we have a
+		// constIDGetter.  See the above TODO.
+		id, err := ruid.idGetter.GetIDForHandle(ctx, nil)
```
strib: This is a huge hack that I'll fix once CORE-6623 is done.  Seems to work for now though.

libkbfs/tlf_handle_resolve.go
```diff
@@ -283,13 +303,15 @@ func (rsa resolvableSocialAssertion) resolve(ctx context.Context) (
 }
 
 // MakeTlfHandle creates a TlfHandle from the given tlf.Handle and the
-// given normalizedUsernameGetter (which is usually a KBPKI).
+// given normalizedUsernameGetter (which is usually a KBPKI). Note
+// that `t` could be different from `bareHandle.Type()`, if this is an
+// implicit team TLF.
```
songgao: Could you explain what `t` is here? It seems to me it should have been `tlf.KeyingType` but I think I understood it wrong.

strib: It simply replaces `bareHandle.Type()` in the `makeTlfHandleHelper()` call below.  That type is unreliable because it would report any team-backed TLF as `tlf.SingleTeam`, since `tlf.Handle` doesn't have it's own type field.

It's used to set `tlfType` of the new `libkbfs.TlfHandle`, so it's definitely not supposed to be the keying type.

I'll update this comment to explain that a bit.

songgao: Thanks for the explanation. I'm still a bit confused though. Why does `bareHandle.Type()` report iTeam TLFs as `tlf.SingleTeam`? Shouldn't that be what `TypeForKeying()` does?

strib: A `tlf.Handle` does not have a separate type field, and so it infers the type by whether or not the first writer is a team.  It's basically the same thing that `tlf.Handle.TypeForKeying()` does (except with the Keying types).  But for `Type()`, this isn't necessarily correct, because an i-team TLF which should either have a type of Private or Public would have a team as the first writer, and so would return `tlf.SingleTeam` as the type incorrectly.

songgao: Ah that's right. Thanks!

PR #1358, opened 2017-11-23
libgit: get ready for object-repacking during GC

strib:
It's turned off for now by always passing in `MaxObjectPacks=-1`, but once more re-packing testing has been done, we'll flip it on.

Issue: KBFS-2627


songgao:
image

PR #1362, opened 2017-11-29
kbfsmd: add a new ImplicitTeamsVer (MDv4!)

strib:
This adds a new metadata version (4) that is data-compatible with v3
and uses the same struct.  It only uses the new version for
i-team-backed TLFs, otherwise it continues to use v3.  For now, prod
still uses v3 no matter what, until we get a little more testing done.

Issue: KBFS-2621

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -94,7 +94,7 @@ type EncryptedTLFCryptKeyClientAndEphemeral struct {
 }
 
 const (
-	defaultClientMetadataVer kbfsmd.MetadataVer = kbfsmd.SegregatedKeyBundlesVer
+	defaultClientMetadataVer kbfsmd.MetadataVer = kbfsmd.ImplicitTeamsVer
 )
 
 // DataVer is the type of a version for marshalled KBFS data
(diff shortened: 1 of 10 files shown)


libkbfs/data_types.go
```diff
@@ -94,7 +94,7 @@ type EncryptedTLFCryptKeyClientAndEphemeral struct {
 }
 
 const (
-	defaultClientMetadataVer kbfsmd.MetadataVer = kbfsmd.SegregatedKeyBundlesVer
+	defaultClientMetadataVer kbfsmd.MetadataVer = kbfsmd.ImplicitTeamsVer
```
songgao: This looked excitingly scary at first, but it seems this is only used in tests. Perhaps we should rename it to have a `ForTest` suffix?

strib: You're talking about `defaultClientMetadataVer`?  In practice yeah it's used just in tests, but it's actually the default in `ConfigLocal`, if no other code calls `config.SetMetadataVersion()` after construction.  It happens that the libkbfs init code always calls it, but I'm not sure that all other users would.

songgao: I see; LGTM then!

PR #1363, opened 2017-11-29
[KBFS-2588] prefetcher: Add test for simple prefetching and fix a race.

jzila:
OK I added a bunch of tests as part of finding the repro.

This happened because of a bug that occurred when a block in the middle of the prefetch tree was evicted and got its prefetch canceled. The next time such a block was re-requested, it would get added back into the tree but its children still thought they knew about it, so the prefetcher used the wrong value in accounting the size of the subtree for the new block. I've fixed this accounting by tracking whether a block is newly being added to the tree, in which case it'll always use its children's block values.

### DIFF
--- a/libkbfs/block_retrieval_queue.go
+++ b/libkbfs/block_retrieval_queue.go
@@ -351,6 +351,7 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context,
 			}
 		}
 		br.reqMtx.Lock()
+		defer br.reqMtx.Unlock()
 		br.requests = append(br.requests, &blockRetrievalRequest{
 			block:  block,
 			doneCh: ch,
@@ -358,22 +359,23 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context,
 		if lifetime > br.cacheLifetime {
 			br.cacheLifetime = lifetime
 		}
-		br.reqMtx.Unlock()
-		// If the new request priority is higher, elevate the retrieval in the
-		// queue.  Skip this if the request is no longer in the queue (which
-		// means it's actively being processed).
 		oldPriority := br.priority
-		if br.index != -1 && priority > oldPriority {
+		if priority > oldPriority {
 			br.priority = priority
-			heap.Fix(brq.heap, br.index)
-			if oldPriority < defaultOnDemandRequestPriority &&
-				priority >= defaultOnDemandRequestPriority {
-				// We've crossed the priority threshold for prefetch workers,
-				// so we now need an on-demand worker to pick up the request.
-				// This means that we might have up to two workers "activated"
-				// per request. However, they won't leak because if a worker
-				// sees an empty queue, it continues merrily along.
-				brq.notifyWorker(priority)
+			// If the new request priority is higher, elevate the retrieval in the
+			// queue.  Skip this if the request is no longer in the queue (which
+			// means it's actively being processed).
+			if br.index != -1 {
+				heap.Fix(brq.heap, br.index)
+				if oldPriority < defaultOnDemandRequestPriority &&
+					priority >= defaultOnDemandRequestPriority {
+					// We've crossed the priority threshold for prefetch workers,
+					// so we now need an on-demand worker to pick up the request.
+					// This means that we might have up to two workers "activated"
+					// per request. However, they won't leak because if a worker
+					// sees an empty queue, it continues merrily along.
+					brq.notifyWorker(priority)
+				}
 			}
 		}
 		return ch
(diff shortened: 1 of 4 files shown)


libkbfs/block_retrieval_queue.go
```diff
@@ -351,29 +351,31 @@ func (brq *blockRetrievalQueue) Request(ctx context.Context,
 			}
 		}
 		br.reqMtx.Lock()
+		defer br.reqMtx.Unlock()
```
jzila: I needed to move this so that the assignment of `br.priority` is protected by the lock.

jzila:
FWIW, 990c980 reproduces the failure, and 6da4268 fixes it.

libkbfs/block_retrieval_queue.go
strib: It's a bit confusing that this is in a `for` loop, though it seems kosher since we never `continue` after this, only `return`.  Still, it might be more clear to define `br` outside of the `for` loop, then break it after the `if` statement above, so that the locking happens after the loop.

libkbfs/prefetcher.go
```diff
@@ -272,23 +273,36 @@ func (p *blockPrefetcher) recordPrefetchParent(childBlockID kbfsblock.ID,
...
+	if !isParentWaiting {
+		p.log.CDebugf(pre.ctx, "prefetcher doesn't know about parent block "+
+			"%s for child block %s", parentBlockID, childBlockID)
+		panic("prefetcher doesn't know about parent block when trying to " +
+			"record parent-child relationship")
+	}
+	if !pre.parents[parentBlockID] || isParentNew {
+		// The new parent needs its subtree block count increased.  This
+		// happens when either the child didn't know about the parent, or the
+		// parent was newly created.
```
strib: Please elaborate here why both of these cases could happen, since I don't think it's obvious.

libkbfs/prefetcher.go
```diff
@@ -272,23 +273,36 @@ func (p *blockPrefetcher) recordPrefetchParent(childBlockID kbfsblock.ID,
...
 		pre.parents[parentBlockID] = true
 		return pre.subtreeBlockCount, needNewFetch
 	}
-	return 0, needNewFetch
+	// If we return 0, then a prefetch was waiting, and the parent->child
+	// relationship was already known. Thus `needNewFetch` is necessarily
```
strib: Should we panic here if `needNewFetch == true`?

libkbfs/prefetcher.go
```diff
@@ -561,7 +592,7 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 			//
 			// `numBlocks` now represents only the number of blocks to add
 			// to the tree from `pre` to its roots, inclusive.
-			numBlocks, isTail, err := p.handlePrefetch(pre)
+			numBlocks, isTail, err := p.handlePrefetch(pre, isParentNew)
```
strib: Looks like `isParentNew` is always the same as `isPrefetchWaiting`.  It's a little weird to have a variable about a "parent" above, since there's no notion of parents/children in this function, we're just processing a single block.

libkbfs/prefetcher_test.go
```diff
@@ -1066,9 +1077,401 @@ func TestSyncBlockCacheWithPrefetcher(t *testing.T) {
...
+		TransientEntry)
+	testPrefetcherCheckGet(t, config.BlockCache(), bPtr, b, FinishedPrefetch,
+		TransientEntry)
+
+	// Then we wait for the pending prefetches to complete.
+	waitForPrefetchOrBust(t, q.Prefetcher().Shutdown())
+}
+
+func TestPrefetcherUnsyncedPrefetchRootEvictedCanceled(t *testing.T) {
+	t.Log("Regression test for KBFS-2588: when a prefetched block has " +
```
strib: This summary is exactly the same as for the test above (except no "partial"), can you differentiate it somehow?

libkbfs/prefetcher_test.go
```diff
@@ -1066,9 +1077,401 @@ func TestSyncBlockCacheWithPrefetcher(t *testing.T) {
...
+	q.Prefetcher().CancelPrefetch(aPtr.ID)
+	// Notify sync channel for the cancelation.
+	notifySyncCh(t, prefetchSyncCh)
+
+	t.Log("Set the metadata of the root block in the disk cache to " +
+		"NoPrefetch, simulating an eviction and a BlockServer.Get.")
+	err = dbc.UpdateMetadata(ctx, rootPtr.ID, NoPrefetch)
+	require.NoError(t, err)
+
+	t.Log("Evict the root block from the block cache.")
```
strib: The test name contains "ChildEvicted", but we're evicting the root here.  I assume that's intentional, but maybe the test name needs to be clarified?  (Maybe "Evicted" should just be removed from the name?)

PR #1365, opened 2017-11-30
key_manager: use a dummy crypt key for public i-team TLFs

strib:
Miles let me know that random users won't be able to read the dummy crypt key from a public i-team's sigchain.


strib:
> Will this affect existing public TLFs?

Shouldn't.  It's the same key either way.

PR #1366, opened 2017-11-30
init: Add environment variable switch for journal.

jzila:


jzila:
This is so we can turn off the journal for keybase.pub using an environment variable.

libkbfs/init.go
```diff
@@ -169,6 +169,10 @@ func defaultLogPath(ctx Context) string {
 
 // DefaultInitParams returns default init params
 func DefaultInitParams(ctx Context) InitParams {
+	journalEnv := os.Getenv("KBFS_ENABLE_JOURNAL")
```
songgao: Would a KBFS_DISABLE_JOURNAL make more sense?

strib: Agreed that DISABLE would be less confusing in the code.  Or "KBFS_DEFAULT_ENABLE_JOURNAL_VALUE"?

PR #1369, opened 2017-12-01
change some Debug to Info and resolve a TODO

songgao:
so that we can get these logs without turning on debug logging for KBFS on mobile.

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -320,7 +320,11 @@ func (md *MDServerRemote) RefreshAuthToken(ctx context.Context) {
 	default:
 		md.log.CInfof(ctx,
 			"MDServerRemote: error refreshing auth token: %v", err)
-		md.reconnect()
+		err = md.reconnect()
+		if err != nil {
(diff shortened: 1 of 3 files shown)


patrickxb:
I don't know how frequently these get called, but looks fine to me.

songgao:
@patrickxb They should only happen around reconnect, so I don't think it's gonna be too much.

libkbfs/mdserver_remote.go
```diff
@@ -313,16 +313,14 @@ func (md *MDServerRemote) RefreshAuthToken(ctx context.Context) {
...
+		md.log.CInfof(ctx,
 			"MDServerRemote: no session available, connection remains anonymous")
 	default:
-		md.log.CDebugf(ctx,
+		md.log.CInfof(ctx,
 			"MDServerRemote: error refreshing auth token: %v", err)
-		// TODO: once KBFS-1982 is merged, an unknown error here
-		// should just cause a complete disconnect, and we can let the
-		// rpc connection do the retry.
+		md.reconnect()
```
strib: This returns an error, maybe log it at warning level?

PR #1371, opened 2017-12-01
disable block retrieval on mobile

songgao:


songgao:
I don't think there's a decision yet, so I'm not gonna merge that one (https://github.com/keybase/client/pull/9752) yet. Although my test in simulator suggests it connects to real bserver for git stuff.

PR #1372, opened 2017-12-03
add a dialer timeout

songgao:
Depends on: https://github.com/keybase/go-framed-msgpack-rpc/pull/128

Will update vendor to master once that is in.


songgao:
CI fails because of underlying `rpc` struct update. Need a client one merged first after `rpc` update is in.

songgao:
Hmm good call will do.

PR #1373, opened 2017-12-04
libkbfs: make the client create an implicit team TLF ID if the server refuses to

strib:
If the mdserver returns a null TLF ID when responding to a get-by-handle request for a non-existent TLF, the client will now create an implicit team and deterministically generate a TLF ID and store it in that team's sigchain (via the service).

This mode is now enabled in the DSL implicit team tests.

Issue: KBFS-2621

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -329,10 +329,11 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
 	return ops
 }
 
-// createTlfIDIfNeeded creates a TLF ID for a handle that doesn't have
-// one yet.  If it returns a `nil` error, it may have modified `h` to
-// include the new TLF ID.
-func (fs *KBFSOpsStandard) createTlfIDIfNeeded(
+// createAndStoreTlfIDIfNeeded creates a TLF ID for a team-backed
+// handle that doesn't have one yet, and associates it in the service
+// with the team.  If it returns a `nil` error, it may have modified
+// `h` to include the new TLF ID.
+func (fs *KBFSOpsStandard) createAndStoreTlfIDIfNeeded(
 	ctx context.Context, h *TlfHandle) error {
 	if h.tlfID != tlf.NullID {
 		return nil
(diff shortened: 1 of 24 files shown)


libkbfs/kbfs_ops.go
```diff
@@ -329,6 +329,44 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
 	return ops
 }
 
+// createTlfIDIfNeeded creates a TLF ID for a handle that doesn't have
+// one yet.  If it returns a `nil` error, it may have modified `h` to
+// include the new TLF ID.
+func (fs *KBFSOpsStandard) createTlfIDIfNeeded(
```
songgao: maybe name it `createAndStoreTlfIDIfNeeded`?

tlf/id.go
```diff
@@ -257,3 +257,45 @@ func MakeRandomID(t Type) (ID, error) {
 	}
 	return id, nil
 }
+
+// MakeIDFromTeam makes a deterministic TLF ID from a team ID and an epoch
+// representing how many times a new TLF has been needed for this
+// team.  Returns NullID on failure.
+func MakeIDFromTeam(t Type, tid keybase1.TeamID, epoch byte) (ID, error) {
```
songgao: I have a DeJa vu of writing this in `keybase1` and was gonna suggest removing it from `keybase1`, but I can't find it there. So maybe I didn't ...

tlf/id.go
```diff
@@ -257,3 +257,45 @@ func MakeRandomID(t Type) (ID, error) {
 	}
 	return id, nil
 }
+
+// MakeIDFromTeam makes a deterministic TLF ID from a team ID and an epoch
+// representing how many times a new TLF has been needed for this
+// team.  Returns NullID on failure.
```
jzila: What's the scenario where tracking this is necessary?

strib: Tracking what, the epoch?  It's for if we ever have to manually reset an implicit team TLF (due to a KBFS bug, say), without wanting to reset the entire implicit team.

jzila: Makes sense, thanks.

PR #1374, opened 2017-12-05
revendor go-framed-msgpack-rpc

oconnor663:
r? @songgao @jzila 

There are upstream API changes that we're vendoring in here, so I think CI will break for this PR and for https://github.com/keybase/client/pull/9720 until both of them land at the same time.


oconnor663:
(We ended up splitting out the client side revendor into https://github.com/keybase/client/pull/9787, to pick up some Info->Debug changes.)

PR #1375, opened 2017-12-05
update vendor and force debug logging for rpc package in KBFS

songgao:
Now that everything in `rpc` is back to Debug level again, we won't get any `rpc` logging on mobile. This forces the Debug logging in `rpc` package on mobile.

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -97,7 +97,7 @@ type ConfigLocal struct {
 	kbpki            KBPKI
 	renamer          ConflictRenamer
 	registry         metrics.Registry
-	loggerFn         func(prefix string) logger.Logger
+	loggerFn         func(prefix string, overrideEnableDebug bool) logger.Logger
 	noBGFlush        bool // logic opposite so the default value is the common setting
 	rwpWaitTime      time.Duration
 	diskLimiter      DiskLimiter
@@ -369,7 +369,8 @@ func getDefaultCleanBlockCacheCapacity() uint64 {
 //
 // TODO: Now that NewConfigLocal takes loggerFn, add more default
 // components.
-func NewConfigLocal(mode InitMode, loggerFn func(module string) logger.Logger,
+func NewConfigLocal(mode InitMode,
+	loggerFn func(module string, overrideEnableDebug bool) logger.Logger,
 	storageRoot string, diskCacheMode DiskCacheMode, kbCtx Context) *ConfigLocal {
 	config := &ConfigLocal{
 		loggerFn:      loggerFn,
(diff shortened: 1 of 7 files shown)


libkbfs/config_local.go
```diff
@@ -97,7 +97,7 @@ type ConfigLocal struct {
 	kbpki            KBPKI
 	renamer          ConflictRenamer
 	registry         metrics.Registry
-	loggerFn         func(prefix string) logger.Logger
+	loggerFn         func(prefix string, overrideEnableDebug bool) logger.Logger
```
strib: Why is there an "enable" in the param name?  Couldn't it just be `overrideDebug`?  Here are everywhere else.

libkbfs/interfaces.go
```diff
@@ -28,6 +28,10 @@ type logMaker interface {
 	MakeLogger(module string) logger.Logger
 }
 
+type debugForcedLogMaker interface {
```
strib: This doesn't seem to be used anywhere, probably doesn't need to be a new interface.

libkbfs/config_local.go
songgao: Because setting it to false doesn't disable debug if it was already enabled through `params` [here](https://github.com/keybase/kbfs/pull/1375/files#diff-2bd2766c2ea1726e9c09c18beb70ee12R576). It could be a `overrideDebug *bool` though, which if non-nil overrides `params.Debug`. WDYT?

libkbfs/interfaces.go
songgao: Sure; I'll put it into `logMaker` then.

libkbfs/config_local.go
strib: I think we have different definitions for the word "override" or something.  If a boolean called "overrideDebug" is false, I would assume that it doesn't change anything about the debug value, it just leaves it alone.  That's what this does, right?

songgao: Oh I see; that makes sense. Will update the word then. Thanks!

jzila: so make it `ensureDebugEnabled`

libkbfs/interfaces.go
songgao: Ah it looks like `testLogMaker` maps to `logMaker` too. I'll just put it in `Config` I guess.

songgao:
>  However the subject says that this updates vendor and I don't think it does.

I didn't realize Jack already updated; will update the commit message then.

libkbfs/config_local.go
songgao: @jzila Thanks; `overrideDebug` actually sounds good to me now; but let me know if it bothers you and I'll change it into `ensureDebugEnabled`

PR #1377, opened 2017-12-05
rename ServerErrorRequiredLockIsNotHeld to ServerErrorLockConflict

songgao:
It's easiest if we return the same error for 1) git lock missing; and 2) migration lock is blocking rekey, since we wouldn't need to lookup DB to determine what's prevent the row to be inserted (in the server PR I'm about to put up). So this PR just renames the error into a more generic one.

Since the status code is preserved, I don't *think* this breaks anything. The error's `Name` field is changed, but it seems it's not used anywhere in Go code. But please let me know if you think otherwise.

### DIFF
--- a/kbfsmd/server_errors.go
+++ b/kbfsmd/server_errors.go
@@ -49,11 +49,11 @@ const (
 	// to indicate that a reader has requested to read a TLF ID that
 	// has been finalized, which isn't allowed.
 	StatusCodeServerErrorCannotReadFinalizedTLF = 2812
-	// StatusCodeServerErrorRequiredLockIsNotHeld is the error code returned by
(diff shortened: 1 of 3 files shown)


kbfsmd/server_errors.go
```diff
@@ -49,11 +49,11 @@ const (
 	// to indicate that a reader has requested to read a TLF ID that
 	// has been finalized, which isn't allowed.
 	StatusCodeServerErrorCannotReadFinalizedTLF = 2812
-	// StatusCodeServerErrorRequiredLockIsNotHeld is the error code returned by
+	// StatusCodeServerErrorLockConflict is the error code returned by
```
strib: This description needs to be made more generic too, right?

PR #1378, opened 2017-12-06
journal_server: check for writership before creating journal

strib:
So we don't create worthless on-disk journals for public TLFs, etc.

Issue: KBFS-2643

### DIFF
--- a/libkbfs/journal_server_test.go
+++ b/libkbfs/journal_server_test.go
@@ -748,7 +748,7 @@ func TestJournalServerEnableAuto(t *testing.T) {
 	require.Len(t, tlfIDs, 1)
 }
 
-func TestJournalServerPublicTLF(t *testing.T) {
+func TestJournalServerReaderTLFs(t *testing.T) {
 	tempdir, ctx, cancel, config, _, jServer := setupJournalServerTest(t)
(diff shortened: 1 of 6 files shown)


libkbfs/journal_server_test.go
```diff
@@ -747,6 +748,45 @@ func TestJournalServerEnableAuto(t *testing.T) {
...
+		ctx, config.KBPKI(), nil, "test_user2", tlf.Public)
+	require.NoError(t, err)
+
+	// Access someone else's public TLF, which should NOT create a
+	// journal automatically.
+	_, err = config.MDOps().GetIDForHandle(ctx, h)
+	require.NoError(t, err)
+
+	status, tlfIDs = jServer.Status(ctx)
+	require.True(t, status.EnableAuto)
```
jzila: What is `EnableAuto`?

strib: `EnableAuto` means that journals are created automatically when you access them (now the default for everyone, but it didn't used to be).

jzila: All right, so this check is essentially to make sure that the `EnableAuto=false` switch isn't the cause `status.JournalCount==0`.

strib: Right.

PR #1380, opened 2017-12-07
revert forcing debug logging

songgao:
The LogLevel is global rather than component-wide, so there's no way to turn on debug only for a particular logger.

https://github.com/keybase/client/blob/773a0c3acab154cfa5e2a6f426401f6b32c5b838/go/logger/standard.go#L312


songgao:
Thanks for the thumb. I don't know! Currently I'm looking at `zap` and possibilities of swapping that in for the logger.

PR #1384, opened 2017-12-07
kbfsfuse: update the md version for the startup script

strib:
Used in docker tests.


jzila:
You can specify this in the docker tests themselves via `docker-compose.yml`.

strib:
Cool, thanks. Agreed we could do that, but it's not done that way yet, and this also needs to be fixed regardless, so I'll merge when CI passes.

PR #1385, opened 2017-12-08
kbfsgit: use handle type/name to check if tlf is managed by app

strib:
Previously we were using the number of users in the handle, which breaks for implicit teams.  So instead use the type, and for private TLFs, use the full name and see if multiple users are listed.

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -253,6 +253,25 @@ func (r *runner) printDoneOrErr(
 	}
 }
 
+func (r *runner) isManagedByApp() bool {
+	switch r.h.Type() {
+	case tlf.Public:
+		// Public TLFs are never managed by the app.
+		return false
+	case tlf.SingleTeam:
+		// Single-team TLFs are always managed by the app.
+		return true
+	case tlf.Private:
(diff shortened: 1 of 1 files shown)


kbfsgit/runner.go
```diff
@@ -253,6 +253,25 @@ func (r *runner) printDoneOrErr(
...
+func (r *runner) isManagedByApp() bool {
+	switch r.h.Type() {
+	case tlf.Public:
+		// Public TLFs are never managed by the app.
+		return false
+	case tlf.SingleTeam:
+		// Single-team TLFs are always managed by the app.
+		return true
+	case tlf.Private:
+		// Only single-user private TLFs are managed by the app.  So
```
jzila: What about single-user private TLFs that haven't yet been transitioned to implicit teams?

strib: What's the concern?  This function will correctly return `true` in that case as well, right?

strib: Yes, single-user private repos are managed by the app/CLI (i.e., the "personal" section of the GUI), along with team repos.

jzila: Ah, I see now what is meant by "managed". Thanks!

PR #1386, opened 2017-12-08
 tlf_handle_resolve: support for resolving implicit team IDs

strib:
Previously we were only resolving from a TLF name -> implicit team ID.  But KBFS also needs to do the reverse, and the service just added an RPC to make it possible, so now we can make use of it.

Issue: KBFS-2621

### DIFF
--- a/libkbfs/kbpki_client.go
+++ b/libkbfs/kbpki_client.go
@@ -65,6 +65,25 @@ func (k *KBPKIClient) ResolveImplicitTeam(
 		ctx, assertions, suffix, tlfType, false, "")
 }
 
+// ResolveImplicitTeamByID implements the KBPKI interface for KBPKIClient.
+func (k *KBPKIClient) ResolveImplicitTeamByID(
+	ctx context.Context, teamID keybase1.TeamID, tlfType tlf.Type) (
+	ImplicitTeamInfo, error) {
+	name, err := k.serviceOwner.KeybaseService().ResolveImplicitTeamByID(
+		ctx, teamID)
+	if err != nil {
+		return ImplicitTeamInfo{}, err
+	}
+
+	assertions, suffix, err := tlf.SplitExtension(name)
+	if err != nil {
+		return ImplicitTeamInfo{}, err
+	}
+
+	return k.serviceOwner.KeybaseService().ResolveIdentifyImplicitTeam(
+		ctx, assertions, suffix, tlfType, false, "")
+}
+
 // IdentifyImplicitTeam identifies (and creates if necessary) the
 // given implicit team.
 func (k *KBPKIClient) IdentifyImplicitTeam(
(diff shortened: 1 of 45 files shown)


strib:
(I chickened out on enabled mdv4 yet.  Needs more testing.  Going to make another PR for a better error in this version once mdv4 is turned on at the server.)

libkbfs/kbpki_client.go
```diff
@@ -65,6 +65,25 @@ func (k *KBPKIClient) ResolveImplicitTeam(
...
+		return ImplicitTeamInfo{}, err
+	}
+
+	assertions, suffix, err := tlf.SplitExtension(name)
+	if err != nil {
+		return ImplicitTeamInfo{}, err
+	}
+
+	return k.serviceOwner.KeybaseService().ResolveIdentifyImplicitTeam(
+		ctx, assertions, suffix, tlfType, false, "")
```
jzila: Could we get a unified call that does these two things in one RPC?

libkbfs/keybase_daemon_local.go
```diff
@@ -239,12 +239,11 @@ func (k *KeybaseDaemonLocal) Resolve(ctx context.Context, assertion string) (
...
 
-	// TODO(KBFS-2621): Resolve shouldn't work for implicit teams, but
-	// until CORE-6623 is done, this is required.
-	iti, err := k.localImplicitTeams.getLocalImplicitTeam(id.AsTeamOrBust())
-	if err == nil {
-		// An implicit team exists, so use the display name.
-		return iti.Name, id, nil
+	_, ok := k.localImplicitTeams[id.AsTeamOrBust()]
+	if ok {
+		// An implicit team exists, so Resolve shouldn't work.
```
jzila: Meaning we should be using `ResolveImplicitTeamByID`?

libkbfs/kbpki_client.go
strib: No, this is the way CORE wanted to do it.  It's ok.

libkbfs/keybase_daemon_local.go
strib: Yes, will clarify.

libkbfs/tlf_handle_resolve.go
```diff
@@ -262,33 +262,33 @@ func makeTlfHandleHelper(
...
+		iteamInfo, err := ruid.resolver.ResolveImplicitTeamByID(
+			ctx, ruid.id.AsTeamOrBust(), ruid.tlfType)
 		if err == nil {
 			return nameIDPair{
-				name: name,
-				id:   ruid.id,
-			}, keybase1.SocialAssertion{}, id, err
+				name: iteamInfo.Name,
+				id:   iteamInfo.TID.AsUserOrTeam(),
+			}, keybase1.SocialAssertion{}, iteamInfo.TlfID, nil
```
songgao: Do we want to verify `ruid.id == itemInfo.TID.AsUserOrTeam()` here?

PR #1388, opened 2017-12-09
Keybase Pages prototype

songgao:
This PR adds a prototype for Keybase Pages. It's now hardcoded to work with the *staging* Let's Encrypt environment, meaning the certs are not trusted by browsers, but we get higher rate limits.

To test it, you need configure two things on your domain name:
1. An A record that points to the EC2 public IP I'm running this on. For example: `kbp-test2.song A 1m 174.129.95.214`
2. A TXT record to configure how content should be served. The TLF should be readable by kbpbot. For example: `kbp-test2.song TXT 1m "kbp=/keybase/public/songgao/kbp-test2"`

I vendored [`zap`](https://github.com/uber-go/zap) in for logging. It's a pretty nice alternative to `logrus` with way better performance and nicer interface, so I thought we could try it with Keybase Pages. But let me know if any of you have a strong opinion against it. The development logger looks like this:
```
2017-12-09T00:37:36.976Z        INFO    libpages/root.go:66     root.MakeFS     {"type": "kbfs", "tlf_type": "public", "tlf": "songgao", "path": "kbp-test2"}
```
while the prod logger will be fully json like on our server side.


songgao:
I'm realizing this won't work for multi-instance deployments, as when ACME client (`kbp`) requests certificate, ACME server (Let's Encrypt) need to verify domain ownership, and there's no way to have the ELB route that traffic to the original instance that initiated the cert request.

I think we'll need to either:

1) aggregate HTTP-hosted well-known assets used by ACME client between all instances, or
2) deterministically delegate cert requesting and management to specific instances based on domain name, and proxy HTTP requests to the one that's responsible for the domain. HTTPS can still be served from any instance, and they don't have a cert, they'd just request it from the cert-owning instance.

I'm leaning towards second approach as we can use the same infrastructure for cert distribution and requests. Of course we always have the option to use ACM in our deployment, but keep the single-instance setup for "community" version. But that makes it hard for other people to deploy larger-scale `kbp`.

Let me know what you guys think!

songgao:
Oh I was being inaccurate about `autocert` using HTTP (:80) for domain verification. It actually uses TLS SNI for verification. In other words, this only listens on :443. And that means the two approaches I suggested above are not applicable to `autocert` per se.

libpages/root.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// MakeFS makes a http.FileSystem from *r, to be used by http package to serve
+// through HTTP.
+func (r *Root) MakeFS(ctx context.Context, log *zap.Logger,
+	kbfsConfig libkbfs.Config) (hfs http.FileSystem, err error) {
+	defer func() {
+		if err == nil {
+			log.Info("root.MakeFS",
+				zap.String("type", r.Type.String()),
```
strib: Might be nice to construct a slice of these zap strings outside the `if` statement, to avoid duplication.

libpages/root.go
```diff
@@ -0,0 +1,164 @@
...
+	}()
+	switch r.Type {
+	case KBFSRoot:
+		tlfHandle, err := libkbfs.GetHandleFromFolderNameAndType(
+			ctx, kbfsConfig.KBPKI(), kbfsConfig.MDOps(), r.TlfNameUnparsed, r.TlfType)
+		if err != nil {
+			return nil, err
+		}
+		rawFS, err := libfs.NewFS(context.Background(), kbfsConfig,
+			tlfHandle, r.PathUnparsed, "TODO", keybase1.MDPriorityNormal)
```
strib: Don't need "TODO" here, and just use "" if you want.  The uniq ID is only used for writes (e.g., create temp file names) and isn't needed in a read-only system.

libpages/root.go
```diff
@@ -0,0 +1,164 @@
...
+		return rawFS.ToHTTPFileSystem(), nil
+	case GitRoot:
+		// TODO: implment git root
+		return nil, errors.New("unimplemented")
+	default:
+		return nil, ErrInvalidKeybasePagesRecord{}
+	}
+}
+
+const gitPrefix = "git-keybase://"
```
strib: Is the "git-" prefix necessary?

libpages/root.go
```diff
@@ -0,0 +1,164 @@
...
+			setRootTlfNameAndPath(&root, str[len(teamPrefix):])
+			return root, nil
+		default:
+			return Root{}, ErrInvalidKeybasePagesRecord{}
+		}
+
+	case strings.HasPrefix(str, kbfsPrefix):
+		str = str[len(kbfsPrefix):]
+		root := Root{Type: KBFSRoot}
+		switch {
```
strib: Seems like this `switch` is identical to the one above.  Refactor a bit to de-duplicate?

libpages/server.go
```diff
@@ -0,0 +1,206 @@
...
+	if !ok {
+		fs, err := root.MakeFS(ctx, s.config.Logger, s.kbfsConfig)
+		if err != nil {
+			return nil, err
+		}
+		s.fsCache.Add(root, fs)
+		return fs, nil
+	}
+	fs, ok := fsCached.(http.FileSystem)
+	if !ok {
```
strib: Seems like this should log a pretty nasty warning, as it would be very unexpected.

libpages/server.go
```diff
@@ -0,0 +1,206 @@
...
+	fsCache *lru.Cache
+
+	whiteList     map[string]bool
+	whiteListOnce sync.Once
+}
+
+func (s *Server) getFS(
+	ctx context.Context, root Root) (http.FileSystem, error) {
+	fsCached, ok := s.fsCache.Get(root)
+	if !ok {
```
strib: I'd prefer fewer indents and code duplication with something like:

```go
if ok {
    if fs, ok := fsCached.(http.FileSystem); ok {
        return fs, nil
    } else {
        // Log warning.
    }
}

fs, err := root.MakeFS(ctx, s.config.Logger, s.kbfsConfig)
...
```

libpages/server.go
```diff
@@ -0,0 +1,206 @@
...
+
+// Error implements the error interface.
+func (ErrDomainNotAllowedInWhitelist) Error() string {
+	return "a whitelist is configured and the given domain is not in the list"
+}
+
+func (s *Server) allowedByDomainWhiteList(domain string) bool {
+	s.whiteListOnce.Do(func() {
+		if len(s.config.DomainWhitelist) > 0 {
+			s.whiteList = make(map[string]bool)
```
strib: Can be `make(map[string]bool), len(s.config.DomainWhitelist))` for more efficiency.

libpages/server.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	httpServer := http.Server{
+		Handler: server,
+	}
+
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(
+			context.Background(), gracefulShutdownTimeout)
```
strib: I don't really understand how this graceful shutdown works.  How will the caller (in this case, the `main` function) know that shutdown is complete?  Wouldn't we need to return a channel that gets closed or something?

libpages/server.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// ServeHTTP implements the http.Handler interface.
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	root, err := LoadRootFromDNS(s.config.Logger, r.Host)
+	if err != nil {
+		s.handleError(w, err)
+		return
+	}
+	fs, err := s.getFS(r.Context(), root)
```
strib: We should probably a new debug tag into the ctx for each of `ServeHTTP` call, e.g. with `libkbfs.CtxWithRandomIDReplayable`.

libfs/httpfs.go
```diff
@@ -0,0 +1,184 @@
...
+		return os.ErrPermission
+	default:
+		return err
+	}
+}
+func (d *dir) Readdir(count int) (fis []os.FileInfo, err error) {
+	d.fs.log.CDebugf(d.fs.ctx, "ReadDir %s", count)
+	defer func() {
+		d.fs.deferLog.CDebugf(d.fs.ctx, "ReadDir done: %+v", err)
+		err = translateErr(err)
```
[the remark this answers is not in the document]
songgao: Ha there's already a `translateErr` and that's why it compiled lol. I guess I should merge the latter into former.

libpages/dns.go
```diff
@@ -0,0 +1,92 @@
...
+//
+// NOTE that a TXT record that's long than 256 characters can optionally be
+// split to multiple quoted strings. https://github.com/golang/go/issues/10482
+//
+// Examples:
+//
+// static.keybase.io TXT "kbp=/keybase/team/keybase.bots/static.keybase.io"
+// meatball.gao.io   TXT "kbp=/keybase/public/songgao/meatball/"
+// song.gao.io       TXT "kbp=/keybase/private/songgao,kb_bot/blah"
+// blah.strib.io     TXT "kbp=/keybase/private/strib#kb_bot/blahblahb" "lah/blah/"
```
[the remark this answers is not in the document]
songgao: Good question! I didn't know so I looked up, and it seems the RFC didn't this is illegal: https://tools.ietf.org/html/rfc1035#section-3.3.14 I tried with the Go client and it worked too. So I think we are good here. I'll clarify it in comment.

libpages/dns.go
```diff
@@ -0,0 +1,92 @@
...
+			)
+		}
+	}()
+
+	txtRecords, err := net.LookupTXT(domain)
+	if err != nil {
+		return Root{}, err
+	}
+
+	for _, r := range txtRecords {
```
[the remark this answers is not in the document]
songgao: Not a TODO; Go's DNS client already does that! Each entry in txtRecords here is actually a concatenated complete TXT record. If length is larger than 1, it means multiple TXT records are configured.

libpages/root.go
```diff
@@ -0,0 +1,164 @@
...
+	PathUnparsed    string
+}
+
+// MakeFS makes a http.FileSystem from *r, to be used by http package to serve
+// through HTTP.
+func (r *Root) MakeFS(ctx context.Context, log *zap.Logger,
+	kbfsConfig libkbfs.Config) (hfs http.FileSystem, err error) {
+	defer func() {
+		if err == nil {
+			log.Info("root.MakeFS",
```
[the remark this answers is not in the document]
songgao: Good call; will do!

> if you are going to continue using a separate logger here

Personally I think it'd be nice if we could switch everything to `zap`. It's structured and has better performance. But that's a different discussion :)

libpages/dns.go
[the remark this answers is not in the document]
strib: Cool.  So probably no need to have the description above about 256 char limits at all then, it just confused me.

libpages/root.go
songgao: Not really, but when configuring DNS, it'd be nice to have `git` showing somewhere. Here's why I think so: In context of `git` commands, `keybase://` makes it clear that it's a keybase remote. But when configuring a `kbp` root, we don't need to make it clear we are working with `keybase`; that's redundant with `kbp`. Instead, I think we need to differentiate `git` from a normal KBFS path. Does it make sense?

strib: Then how about just `git://` instead of `git-keybase://`?

libpages/server.go
songgao: The `ListenAndServe` function blocks until everything is finished. If `ctx` is canceled in `kbp` package, the goroutine (main) will block on this function until everything is shutdown, or `gracefulShutdownTimeout` has passed, whichever happens first.

libpages/root.go
strib: Or maybe we can follow git's lead and have `git@keybase:private/strib.git` or whatever, which allows for future protocol expansion.

songgao: Sounds good; do we wanna do `git://keybase/private/strib,kbpbot/bigface` or `git://private/strib,kbpbot/bigface`?

libpages/dns.go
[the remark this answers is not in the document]
songgao: Oh cool; will do!

libpages/server.go
strib: I guess you mean `httpServer.Serve()` will block until everything is finished?  There's no scenario where it returns early without `ctx` being canceled, but where a `Shutdown` call is still needed?

libpages/root.go
strib: I think I like the format `git@keybase:private/strib,kbpbot/bigface` best, but that's just one vote.

songgao: Sounds good! (I missed your last comment when submitting my last comment)

libpages/server.go
songgao: Yes, `httpServer.Serve()` is where it blocks.

> There's no scenario where it returns early without ctx being canceled, but where a Shutdown call is still needed?

Hmm perhaps we should cancel the ctx in a defered call in `ListenAndServe` function too?

libfs/httpfs.go
```diff
@@ -0,0 +1,184 @@
...
+				offset:   0,
+			},
+			ei: ei,
+		}, nil
+	}
+	return fileOrDir{
+		dir: &dir{
+			fs:      hfs.fs,
+			dirname: n.GetBasename(),
+			node:    n,
```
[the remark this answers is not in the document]
songgao: Saving the `Node` saves a couple `KBFSOps.Lookup()`s when caller needs to read from it in the future right? Is there any downside for saving the node?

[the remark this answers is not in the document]
strib: I just don't like the duplicated code here, I'd rather optimize later if we think it's needed.  If you're really worried about it, you can probably `chmod` the `FS` into the directory itself and then just `fs.ReadDir("")`.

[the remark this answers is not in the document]
strib: (By "here", I meant in the read dir function.)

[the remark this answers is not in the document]
songgao: Did you mean `chroot` rather than `chmod`? That would create a new `FS` which has its own overhead. Would you mind if I split `fs.Readdir` into two small functions, and call them in `fs.Readdir` and `httpfs.go`? This we we don't have duplicate code.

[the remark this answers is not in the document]
songgao: Actually just one new function that gets dir children from a `Node`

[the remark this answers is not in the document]
strib: Oops yeah I meant `chroot`.  Refactoring `ReadDir` is fine with me if you prefer that.

PR #1390, opened 2017-12-09
[KBFS-2650] Remove prefetches for updated pointers

jzila:
This will address when non-tail blocks are removed, but I'm not sure about tail blocks.


jzila:
Please hold off reviewing; this AppVeyor failure is worrying.

jzila:
I need to combine this with #1391

PR #1391, opened 2017-12-09
[KBFS-2650, KBFS-2651] Don't store actual block data in the prefetcher

jzila:
* We no longer store actual block data in the prefetcher so it doesn't run out of memory. Instead, we store an empty block of the correct type.
  * To facilitate this, I made a new API for `BlockRetriever` that allows the prefetcher to fetch a block on-demand without triggering prefetches when the block is available in one of the caches.
* Added an API to remove blocks from the prefetcher without automatically canceling their parent blocks.
  * Consume that API in `updatePointer`, to remove old blocks still being prefetched.
  * In-flight prefetches get automatically canceled by the prefetch context being closed.

### DIFF
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -3375,7 +3375,7 @@ func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, n
 				lifetime)
 		}
 		// remove any prefetches for the old pointer from the prefetcher.
-		fbo.config.BlockOps().Prefetcher().RemovePrefetch(oldPtr.ID)
+		fbo.config.BlockOps().Prefetcher().CancelPrefetch(oldPtr.ID)
 	}
 }
 
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -3374,6 +3374,8 @@ func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, n
 				updatePointerPrefetchPriority, kmd, newPtr, block.NewEmpty(),
 				lifetime)
 		}
+		// remove any prefetches for the old pointer from the prefetcher.
+		fbo.config.BlockOps().Prefetcher().RemovePrefetch(oldPtr.ID)
 	}
 }
 
--- a/libkbfs/folder_block_ops.go
+++ b/libkbfs/folder_block_ops.go
@@ -3374,7 +3374,7 @@ func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, n
 				updatePointerPrefetchPriority, kmd, newPtr, block.NewEmpty(),
 				lifetime)
 		}
-		// remove any prefetches for the old pointer from the prefetcher.
+		// Cancel any prefetches for the old pointer from the prefetcher.
 		fbo.config.BlockOps().Prefetcher().CancelPrefetch(oldPtr.ID)
 	}
 }
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4663,6 +4663,10 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 			return nil
 		}
 	}
+	// Remove any unreferenced blocks from the prefetcher.
+	for _, ptr := range op.Unrefs() {
+		fbo.config.BlockOps.Prefetcher().CancelPrefetch(ptr.ID)
+	}
 
 	fbo.observers.batchChanges(ctx, changes)
 	return nil
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4663,9 +4663,9 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 			return nil
 		}
 	}
-	// Remove any unreferenced blocks from the prefetcher.
+	// Cancel any block prefetches for unreferenced blocks.
 	for _, ptr := range op.Unrefs() {
-		fbo.config.BlockOps.Prefetcher().CancelPrefetch(ptr.ID)
+		fbo.config.BlockOps().Prefetcher().CancelPrefetch(ptr.ID)
 	}
 
 	fbo.observers.batchChanges(ctx, changes)
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -4456,6 +4456,11 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 		return err
 	}
 
+	// Cancel any block prefetches for unreferenced blocks.
+	for _, ptr := range op.Unrefs() {
+		fbo.config.BlockOps().Prefetcher().CancelPrefetch(ptr.ID)
+	}
+
 	var changes []NodeChange
 	switch realOp := op.(type) {
 	default:
@@ -4662,10 +4667,6 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 		if len(changes) == 0 {
 			return nil
 		}
-	}
-	// Cancel any block prefetches for unreferenced blocks.
-	for _, ptr := range op.Unrefs() {
-		fbo.config.BlockOps().Prefetcher().CancelPrefetch(ptr.ID)
 	}
 
 	fbo.observers.batchChanges(ctx, changes)
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -2015,6 +2015,10 @@ type BlockRetriever interface {
 	// Request retrieves blocks asynchronously.
 	Request(ctx context.Context, priority int, kmd KeyMetadata,
 		ptr BlockPointer, block Block, lifetime BlockCacheLifetime) <-chan error
+	// Request retrieves blocks asynchronously, but doesn't trigger a prefetch
+	// unless the block had to be retrieved from the server.
+	RequestNoPrefetch(ctx context.Context, priority int, kmd KeyMetadata,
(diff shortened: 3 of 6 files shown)


libkbfs/folder_block_ops.go
```diff
@@ -3374,6 +3374,8 @@ func (fbo *folderBlockOps) updatePointer(kmd KeyMetadata, oldPtr BlockPointer, n
 				updatePointerPrefetchPriority, kmd, newPtr, block.NewEmpty(),
 				lifetime)
 		}
+		// remove any prefetches for the old pointer from the prefetcher.
+		fbo.config.BlockOps().Prefetcher().RemovePrefetch(oldPtr.ID)
```
strib: This seems ok, but don't we also need to call `RemovePrefetch` for blocks that have been removed?  That is the case that seems to leak blocks the worst.   So like when `folderBranchOps.notifyOneOpLocked` processes an `rmOp` or a `resolutionOp`, it should call this for each of the block IDs in `op.Unrefs()`, I think?

But I don't totally understand how the leak happened, so maybe I'm missing something.

jzila: I think you are correct. Should it be on `rmOp` or `gcOp`, do you think?

strib: `rmOp` and `resolutionOp` both contain unreferenced, but archived, blocks in their `op.Unrefs()` list.  `gcOp.Unrefs()` contains blocks that are being deleted entirely.

And actually, `renameOp` can also contain `Unrefs()`, because you can rename one file over another one, deleting the latter.

So maybe the right thing to do is not be dependent on any operation type, but just do it for all `op.Unrefs()`?

jzila: Will do. Some follow-up questions:
1. Should we then remove the `RemovePrefetch` call in `updatePointer`?
2. At what point does the root block get updated? After the `rmOp` or before?
3. Should I just be using `CancelPrefetch` in all these instances? Removing a block is a content-addressable operation...so any parent block IDs will change.

strib: > Should we then remove the RemovePrefetch call in updatePointer?

It still seems ok to me to keep that call.  There's no point in prefetching a block that won't be looked at anymore.  I just don't think would fix the underlying cause of the leak I was able to repro by removing a file.

> At what point does the root block get updated? After the rmOp or before?

If there was batching enabled, the root block will generally be updated while processing the `resolutionOp`, which is usually the first op in the MD update.  If for some reason there was no batching, then the root block gets updated in every operation.  (I.e., the MD update would only have one op, which could be an `rmOp`.)

> Should I just be using CancelPrefetch in all these instances? Removing a block is a content-addressable operation...so any parent block IDs will change.

I actually don't have a good grasp on the difference between remove and cancel.  If a block's removed, I guess we want any outstanding prefetches for that block canceled, and all state associated with it removed from memory?  We should make whatever calls get us that.

jzila: OK thanks. I'll add it to all the appropriate `{resolution,rm}Op`s, and leave it in `updatePointer`.

> I actually don't have a good grasp on the difference between remove and cancel.

`RemovePrefetch` walks up the tree, subtracting the removed block's count from its parents, up to all roots of the tree. It's akin to a completion, except that it doesn't update the state of the block in the caches.

`CancelPrefetch` walks up the tree, but instead of only subtracting the removed block's count, it removes _every_ block it encounters, all the way to any roots that are connected in the current tree. I built it that way because of the content addressing I mentioned: removing any block effectively invalidates all its parents, because they would get new parents. I'm wondering if `RemoveBlock` is even a useful abstraction.

Actually now that I think about it further, once we've canceled the context for a prefetch, then the next worker that picks up that block will `CancelPrefetch` on it anyway. So we may as well just call that.

jzila: Scratch that, it will call `CancelPrefetch`, but we will have already removed the block in order to have called the `prefetch.cancel()`, so that cancelation will be a no-op.

jzila:
@strib: I changed some things, including using `CancelPrefetch` instead of `RemovePrefetch`, which is now no longer used.

I added a `CancelPrefetch` call to `updatePointer`, and for all `Unrefs()` in `notifyOneOpLocked()`.

Note that the disk cache still only removes items on a `gcOp`, which I think is the right approach: if a prefetch request is currently being serviced when its `CancelPrefetch` is received, then we still want it to come from the disk cache rather than be retrieved from the server.

strib:
Thanks, I'll take another look.  Can you also please write a little something either here or in the JIRA issue about why this solves the leak?

strib:
> Note that the disk cache still only removes items on a gcOp, which I think is the right approach: if a prefetch request is currently being serviced when its CancelPrefetch is received, then we still want it to come from the disk cache rather than be retrieved from the server.

Am good with this, it will be especially useful once we have histories turned on.

libkbfs/interfaces.go
```diff
@@ -1246,6 +1246,9 @@ type Prefetcher interface {
 	// CancelPrefetch notifies the prefetcher that a prefetch should be
 	// canceled.
 	CancelPrefetch(kbfsblock.ID)
+	// RemovePrefetch notifies the prefetcher that a prefetch should be
```
strib: If you're going to keep this function (even though it's not used anymore), maybe explain here the difference between remove and cancel, like you did in the comments?

libkbfs/folder_branch_ops.go
```diff
@@ -4663,6 +4663,10 @@ func (fbo *folderBranchOps) notifyOneOpLocked(ctx context.Context,
 			return nil
 		}
 	}
+	// Cancel any block prefetches for unreferenced blocks.
+	for _, ptr := range op.Unrefs() {
```
strib: There's a few `return nil` cases above this -- we should probably move this above the big switch.

jzila:
For unsynced folders (and synced folders that run out of disk cache space), the prefetcher can have block subtrees sitting in memory that never get prefetched. This led to 2 leaks:
1. We were storing a pointer to a real block in the prefetcher tree, for each block. So for the stuck blocks, they were taking up unnecessary RAM. In fact, they were completely defeating the cache retention algorithm of the in-memory cache, since the garbage collector always had at least the prefetcher's reference to the block.
   * Fix: we don't actually need the block content except when we're determining this block's children for future prefetching. But we do still need its type, so that we can instantiate it for retrieval. Thus, instead of storing the block content, we now store `block.NewEmpty()` when we create a prefetch request. We also now have a method in the block retriever that will allow fetching such blocks without causing prefetch requests, which would have caused a redundant O(n^2) explosion in the number of prefetch requests.
2. When block refs are removed, either through `updatePointer` or through an `rmOp`/`resolutionOp`/`renameOp`, we were not removing them from the prefetcher. Since we were storing actual blocks, this was a mortal sin, since we were eternally caching removed blocks. With the fix for Leak Number 1, it's nowhere near as bad as it was before, but we still don't need to keep those references anyway.
   * Now, we `CancelPrefetch` any time a block is removed, which allows the prefetcher to forever forget that block, and its whole parent tree (which is possible because block IDs are content addressed).

strib:
Thanks!  Though I still haven't quite grasped one thing:

> When block refs are removed, either through updatePointer or through an rmOp/resolutionOp/renameOp, we were not removing them from the prefetcher. Since we were storing actual blocks, this was a mortal sin, since we were eternally caching removed blocks.

Why weren't they getting removed in the normal course of prefetching?  If the prefetch got kicked off after a few minutes, but the block failed to fetch because it had already been deleted from the bserver, why wasn't this good enough to delete the whole entry from the prefetch map?

jzila:
> why wasn't this good enough to delete the whole entry from the prefetch map?

If a fetch fails, it does notify the prefetcher, and the prefetch would be canceled appropriately.

The gap lies in the fact that the prefetcher tracks in-progress prefetches for blocks that have only been _triggered_ by the prefetcher. So for unsynced folders, the tree sits in the prefetcher, waiting for someone to traverse the next step of the tree. It does this so that it can correctly represent if a prefetch becomes Finished:

Let's say we have a tree:
```
a -> {b -> c}
```
These are the steps that the prefetcher will take after we fetch `a`:
```
a:1 -> {b:1}
```
`b` is returned by the server, but the priority of the request is below `lowestTriggerPrefetchPriority`, and so we short circuit, allowing the prefetcher to wait for a later operation involving that block:
https://github.com/keybase/kbfs/blob/131aaeceee9e3c7cd6e4af744eca0a1f562a7676/libkbfs/prefetcher.go#L529-L533

We don't even look at the block for evaluating the prefetch, because the prefetcher only cares about higher priority blocks.

Later when something or someone requests `b`, the prefetcher updates as follows:
```
a:1 -> {b:1 -> {c:1}}
```
This involves decrementing `b` and `a`, because we're now actually evaluating `b` for prefetch, and incrementing them both again to prepare for `c`.

When `c` eventually is fetched with higher than `lowestTriggerPrefetchPriority`, the whole tree completes and everyone gets flagged as `FinishedPrefetch`.

strib:
Ah ok, thanks!  I guess I was missing that `b` stays in the `prefetch` map even if no child prefetches are scheduled for it.  Makes sense now.

libkbfs/interfaces.go
```diff
@@ -2012,6 +2012,10 @@ type BlockRetriever interface {
 	// Request retrieves blocks asynchronously.
 	Request(ctx context.Context, priority int, kmd KeyMetadata,
 		ptr BlockPointer, block Block, lifetime BlockCacheLifetime) <-chan error
+	// RequestNoPrefetch retrieves blocks asynchronously, but doesn't trigger a
```
songgao: Does this mean it would get the block, but won't fetch its subtree?

jzila: Yeah, basically it's a way for the prefetcher to use the cache just like everything else does, instead of caching its own blocks. If it used the regular `Request`, then every prefetch would trigger another prefetch for the block itself, recursively, forever.

PR #1393, opened 2017-12-14
fs_notifications: Fix potential panic due to closed channel

jzila:
Fixes https://keybase.atlassian.net/browse/KBFS-2461

### DIFF
--- a/libfs/fs_notifications.go
+++ b/libfs/fs_notifications.go
@@ -71,17 +71,14 @@ func (f *FSNotifications) processNotifications(ctx context.Context) {
 // QueueNotification queues a notification, which must be
 // goroutine-safe.
 func (f *FSNotifications) QueueNotification(fn func()) {
-	channel := func() channels.Channel {
-		f.notificationMutex.RLock()
-		defer f.notificationMutex.RUnlock()
-		return f.notifications
-	}()
-	if channel == nil {
+	f.notificationMutex.RLock()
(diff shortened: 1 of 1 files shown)


libfs/fs_notifications.go
```diff
@@ -71,17 +71,14 @@ func (f *FSNotifications) processNotifications(ctx context.Context) {
...
-	if channel == nil {
+	f.notificationMutex.RLock()
+	defer f.notificationMutex.RUnlock()
+	if f.notifications == nil {
 		f.log.Warning("Ignoring notification, no available channel")
 		return
 	}
 	f.notificationGroup.Add(1)
-	channel.In() <- fn
+	f.notifications.In() <- fn
```
strib: This is probably fine since an infinite channel send should always be fast.  But for the sake of argument here's another option. You could have a `sync.WaitGroup` that is added to at the beginning of this function, and `Done` is called in a `defer`.  Then `processNotifications()` would wait on that group in between setting the channel to nil, and closing the channel.  That way there would be less time with the lock held, in case the send blocks.

But am fine with this way too if you prefer it.

jzila: It's a read lock though, and it's only competing with a write lock that occurs on context cancelation. It doesn't seem harmful to me.

strib: Fair point!

PR #1394, opened 2017-12-14
implement StartMigration rpc

songgao:
Related: https://github.com/keybase/kbfs-server/pull/462

Also I'm realizing I haven't done anything in mdserver_remote.go to support iTeam migration yet. I'll create a ticket and work  on it later.


libkbfs/keybase_service_base.go
```diff
@@ -1040,6 +1041,22 @@ func (k *KeybaseServiceBase) TeamDeleted(ctx context.Context,
...
 
+// StartMigration implements keybase1.ImplicitTeamMigrationInterface for
+// KeybaseServiceBase.
+func (k *KeybaseServiceBase) StartMigration(ctx context.Context,
+	folder keybase1.Folder) (err error) {
+	mdServer := k.config.MDServer()
+	if mdServer == nil {
+		return errors.New("no mdserver")
+	}
+	fav := NewFavoriteFromFolder(folder)
```
strib: Why make a favorite?  You can just use `folder.Name` and `folder.Type` in the call below directly.

libkbfs/mdserver_memory.go
```diff
@@ -88,6 +88,7 @@ type mdServerMemShared struct {
 	// tracks expire time and holder
 	lockIDs              map[mdLockMemKey]mdLockMemVal
 	implicitTeamsEnabled bool
+	iTeamMigrationLocks  map[tlf.ID]struct{}
```
strib: Slightly prefer a `map[tlf.ID]bool` instead so you can just `return md.iTeamMigrationLocks[id]` below.

libkbfs/mdserver_memory.go
```diff
@@ -465,6 +467,19 @@ func (md *MDServerMemory) GetRange(ctx context.Context, id tlf.ID,
 	}
 }
 
+func (md *MDServerMemory) iTeamMigrationRemoveLock(id tlf.ID) {
```
strib: I'm not sure, but it seems likely that these two functions will be called with `md.lock` already held.  But we can fix that when that code is written, I guess.

libkbfs/keybase_service_base.go
songgao: It's to reuse `keybase1.FolderType`->`tlf.Type` conversion. Would you prefer if I do a switch here?

libkbfs/mdserver_memory.go
songgao: Hmmm OK!

libkbfs/mdserver_memory.go
songgao: Yeah; good point. I'll fix the in the follow up ticket which will port the migration checking into `MDServerMemory`

libkbfs/keybase_service_base.go
strib: Ah I see. Ok.  I guess that makes sense.  Can you please add a comment about it though?

libkbfs/interfaces.go
```diff
@@ -1388,6 +1388,11 @@ type MDServer interface {
 	// expired, this is a no-op.
 	ReleaseLock(ctx context.Context, tlfID tlf.ID, lockID keybase1.LockID) error
 
+	// StartImplicitTeamMigration tells mdserver to put a implicit team
+	// migration lock on id, which prevents any rekey MD writes from going
```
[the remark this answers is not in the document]
songgao: Those still go in. I'll update the comment to mention it.

PR #1396, opened 2017-12-14
[EXPERIMENT] vendor: Switch to go dep

jzila:
Just testing...will delete the branch if the experiment fails.


jzila:
OK, I've learned several things:
1. It won't pull forked dependencies properly. You have to use the downstream fork in your Go `import`.
2. It's bigger for some unknown reason.
3. You need to `dep init` and then `dep prune` because who knows? After that `dep ensure` should just make everything work.

jzila:
Nope.

songgao:
Ugh; thanks for trying!

PR #1397, opened 2017-12-18
block_retrieval_queue_test: Fix a race in the test verification code...

jzila:
Fix a race in the test verification code, where the prefetcher would sometimes race to get more blocks.

Our asserts check internal values, not under lock, under the assumption that these tests run everything themselves with no background work. So the fix is simply to deactivate the prefetcher and make this assumption true again.

### DIFF
--- a/libkbfs/block_retrieval_queue_test.go
+++ b/libkbfs/block_retrieval_queue_test.go
@@ -71,9 +71,15 @@ func makeKMD() KeyMetadata {
 	return emptyKeyMetadata{tlf.FakeID(0, tlf.Private), 1}
 }
 
+func initBlockRetrievalQueueTest(t *testing.T) *blockRetrievalQueue {
+	q := newBlockRetrievalQueue(0, 0, newTestBlockRetrievalConfig(t, nil, nil))
+	<-q.TogglePrefetcher(false, nil)
+	return q
+}
+
 func TestBlockRetrievalQueueBasic(t *testing.T) {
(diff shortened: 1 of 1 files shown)


libkbfs/block_retrieval_queue_test.go
```diff
@@ -71,9 +71,15 @@ func makeKMD() KeyMetadata {
 	return emptyKeyMetadata{tlf.FakeID(0, tlf.Private), 1}
 }
 
+func initBlockRetrievalQueueTest(t *testing.T) *blockRetrievalQueue {
+	q := newBlockRetrievalQueue(0, 0, newTestBlockRetrievalConfig(t, nil, nil))
+	<-q.TogglePrefetcher(false, nil)
```
strib: It feels like the test should be waiting for the prefetcher to shut down, but I don't have a great reason why other than that it might be confusing for debugging if there are still goroutines from old prefetchers lying.  But I guess since `TogglePrefetcher(false)` doesn't actually set the prefetcher to nil, the queue shutdown method will still end up waiting for the shutdown to finish.  If that sounds right, nevermind me.

jzila: The channel returned by `TogglePrefetcher` is the channel from the previous prefetcher's `Shutdown`, so that's exactly what we're doing here.

strib: D'oh!  I didn't notice the `<-` somehow.  Nevermind.

PR #1398, opened 2017-12-19
implement kbp Config

songgao:
Also I created https://keybase.atlassian.net/browse/KBFS-2683 for tracking the CLI tool that edits the config file.


kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+//   1) Call path.Clean;
+//   2) Remove leading and trailing "/";
+//   3) Split the rest.
+// If thePath is root, nil elems is returned. Otherwise, elems is guaranteed to
+// be non-empty.
+//
+// If split2 is true, at most elems has at most 2 elements and the second one
+// would include everything other than the first element in the path;
+// otherwise, the entire thePath is split by "/" till the end of the world.
+func cleanPath(thePath string, split2 bool) (elems []string) {
```
strib: "thePath" -> "p" would be better in my opinion.  ("path" would be better but it conflicts.  I just really don't like putting "the" in var names.)

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// cleanPath cleans thePath in following order:
+//   1) Call path.Clean;
+//   2) Remove leading and trailing "/";
+//   3) Split the rest.
+// If thePath is root, nil elems is returned. Otherwise, elems is guaranteed to
+// be non-empty.
+//
+// If split2 is true, at most elems has at most 2 elements and the second one
```
strib: At most, have just one "at most".

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	if thePath == "/" || thePath == "." {
+		// Examples of thePath (before clean) that lead us here:
+		//   ""
+		//   "/"
+		//   "."
+		//   "/.."
+		//   "/."
+		return nil
+	}
+	if strings.HasPrefix(thePath, "/") {
```
strib: You can just use `strings.Trim` to strip the prefix and suffix in one call, I think.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	if split2 {
+		return strings.SplitN(thePath, "/", 2)
+	}
+	return strings.Split(thePath, "/")
+}
+
+// getAccessControl gets the corresponding accessControlV1 for thePath.
+// If a specifically defined exists, it's returned. Otherwise the default one
+// is returned.
+func (m *aclCheckerV1) getAccessControl(thePath string) (aci *accessControlV1) {
```
strib: Why "m"?

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	elems := cleanPath(thePath, true)
+	if elems == nil || m.children == nil {
+		return m.aci
+	}
+	// elems can't be empty.
```
strib: Technically it still could be empty if `cleanPath` has a bug:  https://play.golang.org/p/SsOPVoFvy6

Maybe better to replace `elems == nil` with `len(elems) == 0` above?

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	//   "a/b"
+	//   "a/b/c"
+	if split2 {
+		return strings.SplitN(thePath, "/", 2)
+	}
+	return strings.Split(thePath, "/")
+}
+
+// getAccessControl gets the corresponding accessControlV1 for thePath.
+// If a specifically defined exists, it's returned. Otherwise the default one
```
strib: A specifically defined "one", maybe?

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+// is returned.
+func (m *aclCheckerV1) getAccessControl(thePath string) (aci *accessControlV1) {
+	elems := cleanPath(thePath, true)
+	if elems == nil || m.children == nil {
+		return m.aci
+	}
+	// elems can't be empty.
+	if subChecker, ok := m.children[elems[0]]; ok {
+		if len(elems) > 1 {
+			return subChecker.getAccessControl(elems[1])
```
strib: It's intended that someone could read "a/b/c" without having read or list permissions on "a" or "a/b", right?  Just making sure.  I don't remember the HTTP ACL model very well.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (m *aclCheckerV1) getPermissions(
+	thePath string, username *string) (permissions permissionsV1) {
+	aci := m.getAccessControl(thePath)
+	permissions = aci.anonymous
+	if aci.whitelistAdditional == nil || username == nil {
+		return permissions
+	}
+	if perms, ok := aci.whitelistAdditional[*username]; ok {
+		permissions.read = perms.read || permissions.read
```
strib: This makes sense, but I wonder if there's every any reason to deny a specific user read access, while allowing it for anonymous.  Probably not, but just thought I'd mention it so we remember to document it.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	root := &aclCheckerV1{aci: defaultACI}
+	if acl == nil {
+		return root, nil
+	}
+	// path -> *AccessControlV1
+	cleaned := make(map[string]*AccessControlV1)
+
+	// Make sure there's no duplicate paths.
+	for p := range acl {
```
strib: `for p, ac := range acl` to avoid writing the next line.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	for p, ac := range cleaned {
+		aci, err := ac.makeAccessControlV1Internal(users)
+		if err != nil {
+			return nil, err
+		}
+
+		elems := cleanPath(p, false)
+		if elems == nil {
```
strib: `len(elems) == 0`

kbp/config/config_v1.go
```diff
@@ -0,0 +1,119 @@
...
+
+// DefaultV1 returns a default V1 config, which allows anonymous read to
+// everything.
+func DefaultV1() *V1 {
+	return &V1{
+		Common: Common{
+			Version: Version1Str,
+		},
+		DefaultACL: AccessControlV1{
+			AnonymousPermissions: "read",
```
strib: Why isn't "read,list" the right anonymous permissions?  We don't want to support the `keybase.pub/chris`-style navigation by default?  (If not, why do we even have a "list" permision?)

kbp/config/config_v1_test.go
```diff
@@ -0,0 +1,171 @@
...
+				PasswordSalt:      "tlas",
+				PasswordSHA512Hex: MakeSaltedSHA512PasswordHex("54321", "tlas"),
+			},
+		},
+		DefaultACL: AccessControlV1{
+			AnonymousPermissions: "read",
+		},
+		ACLs: map[string]AccessControlV1{
+			"/alice-and-bob": AccessControlV1{
+				AnonymousPermissions: "",
```
strib: Can you make one of these not specify an explicit `AnonymousPermissions: ""`?  Should behave the same way, but I wouldn't expect users to actually do this most of the time, so it's a good thing to check.

kbp/config/errors.go
```diff
@@ -0,0 +1,46 @@
...
+type ErrDuplicateAccessControlPath struct {
+	cleanedPath string
+}
+
+// Error implements the error interface.
+func (e ErrDuplicateAccessControlPath) Error() string {
+	return "duplicate access control for " + e.cleanedPath
+}
+
+// ErrInvalidVersion is returned when Version filed of the config is invalid.
```
strib: "filed" -> "field"

kbp/config/config_v1_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+func TestConfigV1Full(t *testing.T) {
+	config := V1{
+		Common: Common{
+			Version: Version1Str,
+		},
+		Users: map[string]UserV1{
+			"alice": UserV1{
+				PasswordSalt:      "salt",
+				PasswordSHA512Hex: MakeSaltedSHA512PasswordHex("12345", "salt"),
```
strib: I don't see any `Authenticate()` tests, would be nice to have one.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+		default:
+			return permissionsV1{}, ErrInvalidPermissions{
+				permissions: permsStr}
+		}
+	}
+	return perms, nil
+}
+
+func (ac *AccessControlV1) makeAccessControlV1Internal(users map[string]UserV1) (
+	aci *accessControlV1, err error) {
```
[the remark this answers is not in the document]
songgao: Oh I used to have an "accessControlV1Internal" and `i` is for "internal". But I forgot to update the names after renaming it to `accessControlV1`.

kbp/config/acl_checker.go
songgao: another left-over from a different renaming!

kbp/config/acl_checker.go
songgao: Oh I meant cleanPath shouldn't generate empty `elems` because `strings.Split` never returns an empty slice unless the separator is empty. Will clarify the comment.

kbp/config/acl_checker.go
songgao: > It's intended that someone could read "a/b/c" without having read or list permissions on "a" or "a/b", right?

Yes that's correct. 

> I don't remember the HTTP ACL model very well.

Oh is there such a standard? I should look it up!

kbp/config/acl_checker.go
songgao: Good call for documenting it. Yeah I don't think denying a user permission while still allowing for anonymous is useful. One can always access the same thing in an Incognito window or something and it'd be the same. For now when everything is static anyways.

kbp/config/acl_checker.go
songgao: I used to do that, but then I learnt the hard way that ac works like a reference here (i.e. the address of `ac` never changes), and the content gets overriden every time! And since we take the address of `ac` below, it wouldn't work: https://play.golang.org/p/gP6GzzSnZSu

kbp/config/acl_checker.go
strib: Thanks.  Though I still think the `len()` check would be safer above.

kbp/config/acl_checker.go
strib: Ah ok, makes sense.  Comment below?

kbp/config/config_v1.go
songgao: > by default

I'm OK with turning that on by default.

> If not, why do we even have a "list" permision?

Hmm I'm not sure I understand the logic here. Why would we not have a "list" permission if we don't have that on by default? User might want to turn on listing only for a particular subdir right?

kbp/config/config_v1_test.go
songgao: Oh that's right. Good catch!

kbp/dns.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD
 // license that can be found in the LICENSE file.
 
-package libpages
+package kbp
```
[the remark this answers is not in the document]
songgao: Hmm I don't actually remember what I was thinking, but your argument makes sense. So I'll change the package name back!

kbp/config/config_v1.go
strib: It seems like if we were to implement directory listings, we could replace `keybase.pub/chris` with a kbpbot.  But we can't do that if the default permissions do not include listing, because it would require each user to add a config file to their public directory to get the same permissions we have today.

If we are not going to implement directory listings, then there's no need for a "list" permission at all.

So I don't understand what middle ground you're targeting with these defaults.

kbp/config/acl_checker.go
songgao: Sure! will do

kbp/config/config.go
```diff
@@ -0,0 +1,86 @@
...
+	Authenticate(username, password string) bool
+	GetPermissionsForAnonymous(thePath string) (read, list bool, err error)
+	GetPermissionsForUsername(
+		thePath, username string) (read, list bool, err error)
+}
+
+// ParseConfig parses a configu from reader.
+func ParseConfig(reader io.Reader) (config Config, err error) {
+	buf := &bytes.Buffer{}
+	// todo close
```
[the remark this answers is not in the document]
songgao: I was worried about closing the buffer but it seems it should be automatically recycled.

kbp/config/config_v1.go
songgao: Oh I see; that makes sense. I missed the point where public TLF backed sites don't have config files. The middle ground I'm targeting is `strib.io/photos/secret` where `photos` can be listed, but `secret` cannot so you can have `strib.io/photos/secret/blahblah-secret-url/`.

Go's `FileServer` implementation already supports listing, so we don't really need to implement anything for it, assuming we are happy with what Go provides. If we are ever gonna implement keybase.pub/chris styled listing, it would be somewhat different I think, since it'd require integrating with other parts of the API server like login, etc.

strib: > The middle ground I'm targeting is strib.io/photos/secret where photos can be listed, but secret cannot so you can have strib.io/photos/secret/blahblah-secret-url/.

But why is it the right idea to require the user to override the defaults for `photos`, but not for `photos/secret`?  Wouldn't they already have to set an explicit permissions in `photos/secret`, in order to override the `photos` permissions (which, in your case, would have to have been added in order to allow for listing)?

> If we are ever gonna implement keybase.pub/chris styled listing, it would be somewhat different I think, since it'd require integrating with other parts of the API server like login, etc.

I don't follow, what about keybase.pub listing requires login?  I thought it was just a formatting difference.

songgao: > Wouldn't they already have to set an explicit permissions in photos/secret, in order to override the photos permissions

Not as in this implementation, where `photos/secret` permissions is completely separate from `photos`, since unspecified paths just use the default ACL. But I'm realizing that's a bad idea, since user might just accidentally put in directories in `photos` and they would get automatic defaultACL permission even if `photos` has a more restrictive one.

I think the right idea is to inherit parent permission instead of a "default" one. The `Config.DefaultACL` can then go away. And if "/" doesn't appear in the ACLs field, we'd construct using `DefaultV1` for "/". How does that sound?

> I don't follow, what about keybase.pub listing requires login? I thought it was just a formatting difference.

Ah you're right! Somehow I thought keybase.pub populates usernames like on keybase.io. But it doesn't, so I agree it'd just be formatting difference now.

strib: > I think the right idea is to inherit parent permission instead of a "default" one. The Config.DefaultACL can then go away.

Yes, that sounds right.  That's how I read the code but I messed up.  A directories permissions should definitely apply by default to all its subdirectories as well.

songgao: Oh actually you didn't read it wrong. If user defines permission for `/` only, then anything under that automatically inherits it. But it's an issue if user only defines `/` and `/a/b`, but not `/a`, in which case `/a` would get the default ACL rather than the one from `/`. It's more like a bug here :)

jzila: I'm not sure I agree that a parent's permissions could be automatically inherited by children. IMO the default permissions should be "nothing", and you have to explicitly express paths that are allowed to be read and/or listed.

Then again, `.htaccess` has `Allow,Deny`, and `Deny,Allow` to allow you to dictate those semantics explicitly. The question remains whether we want our default to be more or less permissive.

kbp/config/config_v1.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (c *V1) initACLChecker() {
+	c.aclChecker, c.aclCheckerInitErr = makeACLCheckerV1(c.DefaultACL, c.ACLs, c.Users)
+}
+
+// GetPermissionsForAnonymous implements the Config interface.
+func (c *V1) GetPermissionsForAnonymous(thePath string) (
+	read, list bool, err error) {
+	c.initOnce.Do(c.initACLChecker)
```
jzila: Why don't we just init this when we instantiate it in `DefaultV1`?

kbp/config/config_v1.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	perms := c.aclChecker.getPermissions(thePath, nil)
+	return perms.read, perms.list, nil
+}
+
+// GetPermissionsForUsername implements the Config interface.
+func (c *V1) GetPermissionsForUsername(thePath, username string) (
+	read, list bool, err error) {
+	c.initOnce.Do(c.initACLChecker)
```
jzila: Same as above, why don't we init when we instantiate?

.gitignore
```diff
@@ -8,7 +8,8 @@ kbfsfuse/kbfs_md/
 kbfstool/kbfstool
 kbfsfuse/revision
 kbfsgit/git-remote-keybase/git-remote-keybase
-kbp/kbp
+kbp/kbp/kbp
```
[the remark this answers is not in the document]
songgao: Yeah good call; in the latest iteration I had the binary as `kbpagesd`, and the library `libpages`

kbp/config/config_v1.go
songgao: Because I wanted to make `Config` with only public fields populated is valid and usable. But I'm OK with calling `c.initOnce.Do` in functions that construct this object; will do that!

kbp/config/config_v1.go
```diff
@@ -0,0 +1,119 @@
...
+
+import (
+	"crypto/sha512"
+	"encoding/hex"
+	"sync"
+)
+
+// MakeSaltedSHA512PasswordHex takes concatenated salt and password, calculate
+// SHA512 of it, and return the result encoded in hex.
+func MakeSaltedSHA512PasswordHex(password string, salt string) string {
```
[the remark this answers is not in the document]
songgao: Hrm I guess by the time you saw this I hadn't pushed that commit yet.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+		elems := cleanPath(p, false)
+		if elems == nil {
+			root.aci = aci
+			continue
+		}
+
+		m := root
+		// Construct aclCheckerV1 objects along the path and populate
+		// defaultACI.
+		for i := range elems {
```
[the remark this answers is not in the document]
songgao: I did the `for i := range` intentionally to avoid a copy for each element in `elems`. Let me know if you still want it changed into `for i, elem := range` style!

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	return m.aci
+}
+
+func (m *aclCheckerV1) getPermissions(
+	thePath string, username *string) (permissions permissionsV1) {
+	aci := m.getAccessControl(thePath)
+	permissions = aci.anonymous
+	if aci.whitelistAdditional == nil || username == nil {
```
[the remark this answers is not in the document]
songgao: It's documented in [its public counterpart](https://github.com/keybase/kbfs/pull/1398/files#diff-6be0e14f53f0e6495db249a4ac6c0c24R24). I'll add comment to point there.

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	return strings.Split(thePath, "/")
+}
+
+// getAccessControl gets the corresponding accessControlV1 for thePath.
+// If a specifically defined exists, it's returned. Otherwise the default one
+// is returned.
+func (m *aclCheckerV1) getAccessControl(thePath string) (aci *accessControlV1) {
+	elems := cleanPath(thePath, true)
+	if elems == nil || m.children == nil {
```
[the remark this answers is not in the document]
songgao: The `cleanPath` function returns a `nil` `elems` if it's root: https://github.com/keybase/kbfs/pull/1398/files#diff-a27f7e9f1f279699781f88dbfdc37c6cR99 . Although the latest iteration changes it into `len(elems)==0`

kbp/config/acl_checker.go
```diff
@@ -0,0 +1,198 @@
...
+			perms.list = true
+		default:
+			return permissionsV1{}, ErrInvalidPermissions{
+				permissions: permsStr}
+		}
+	}
+	return perms, nil
+}
+
+func (ac *AccessControlV1) makeAccessControlV1Internal(users map[string]UserV1) (
```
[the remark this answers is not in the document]
songgao: There's no permission nesting on this type; I'll change it into a function to avoid confusions.

kbp/config/config_v1.go
songgao: > I'm not sure I agree that a parent's permissions could be automatically inherited by children.

More precisely, it's the parent's ACL that's inherited by children. In other words, if `/a/b` has an ACL that disallows anything from anonymous, `/a/b/c` automatically gets that too, even if `/a` has a more permissive ACL. This is so that if you copy a site consisting of a bunch of directories and sub directories into a place, you don't have to define an ACL for each of the subdirectories.

kbp/config/acl_checker.go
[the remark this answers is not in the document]
songgao: (talked on kbc and John made me realize it's not a string copy)

PR #1400, opened 2017-12-20
libgit: implement a Clone function

strib:
This function "clones" a repo from a billy filesystem, into another billy filesystem.  However, it doesn't make a true git clone, for two reasons.

1. This will be used to clone KBFS repos, and go-git doesn't natively understand the keybase:// protocol.  There might be a way around this with the `InstallProtocol` go-git method, but I haven't investigated that yet and it would be some extra work.
2. These clones are intended for autogit, which will display them as read-only directories.  They do not need to be real git repos, and so there's no need to copy object files or much else in the .git repo. All we really need is an index, to compute the differences.  Making a full git clone, or even a "shallow" git clone, would be wasteful in that it would read and write way more data than we really need.

To make this happen, we open the existing bare repo, but fake out the storage layer in such a way that it stores its index, and any reference modifications, in a .git/ directory in the worktree, but everything else goes through the main bare repo's .git directory.  Then we can just use a simple `Reset` call to set up the cloned files.

Issue: KBFS-2672

### DIFF
--- a/libgit/worktree.go
+++ b/libgit/worktree.go
@@ -14,6 +14,18 @@ import (
 	"gopkg.in/src-d/go-git.v4/storage/filesystem"
 )
 
+// repoFromStorageAndWorktree returns a repo instance where the dotgit
+// storage layer points to the bare repo represented by `repoFS`, but
+// the worktree storage layer uses `worktreeFS`.  This is useful for
+// checking out the git repo without having to read and copy the
+// complete git history.  The index and the HEAD commit of the
+// worktree are stored in `worktreeFS/.git`.  The caller should make
+// no modifying calls into the repo, other than checkout/reset-related
+// calls.
+//
+// For the convenience of the caller, it also returns `currentHead`,
+// which is the current commit at the HEAD of `branch`, according to
+// the bare repo represented by `repoFS`.
 func repoFromStorageAndWorktree(
 	repoFS billy.Filesystem, worktreeFS billy.Filesystem,
 	branch plumbing.ReferenceName) (
--- a/libgit/worktree.go
+++ b/libgit/worktree.go
@@ -1,0 +1,101 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import (
+	"context"
+
+	billy "gopkg.in/src-d/go-billy.v4"
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+)
+
+func repoFromStorageAndWorktree(
+	repoFS billy.Filesystem, worktreeFS billy.Filesystem,
+	branch plumbing.ReferenceName) (
+	repo *gogit.Repository, currentHead plumbing.Hash, err error) {
+	repoStorer, err := filesystem.NewStorage(repoFS)
+	if err != nil {
+		return nil, plumbing.ZeroHash, err
+	}
+
+	// Wrap it in an on-demand storer, so we don't try to read all the
+	// objects of big repos into memory at once.
+	storage, err := NewOnDemandStorer(repoStorer)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


libgit/worktree.go
```diff
@@ -0,0 +1,101 @@
...
+	"context"
+
+	billy "gopkg.in/src-d/go-billy.v4"
+	gogit "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage/filesystem"
+)
+
+func repoFromStorageAndWorktree(
```
jzila: document please.

libgit/worktree.go
```diff
@@ -0,0 +1,101 @@
...
+//    git clone, would be wasteful in that it would read and write way
+//    more data than we really need.
+//
+// The resulting clone in `worktreeFS` is therefore not a functional
+// git repo.  The caller should only interact with `worktreeFS` in a
+// read-only way, and should not attempt any git operations on it.
+func Clone(
+	ctx context.Context, repoFS billy.Filesystem, worktreeFS billy.Filesystem,
+	branch plumbing.ReferenceName) error {
+	repo, currentHead, err := repoFromStorageAndWorktree(
```
jzila: `currentHead` is the remote's `HEAD` for the branch indicated, right?

libgit/worktree_storer.go
```diff
@@ -0,0 +1,72 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/index"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage"
+)
+
+// worktreeStorer fake out the storage layer in such a way that it
+// stores its index, and any reference modifications, in a .git/
+// directory in the worktree, but everything else goes through the
+// main bare repo's .git directory.  This is useful for making a quick
```
jzila: Can you give an example? I'm having trouble visualizing what this is describing.

libgit/worktree.go
strib: Yes, where "remote" == the bare repo represented by `repoFS`.

libgit/worktree_storer.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/format/index"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage"
+)
+
+// worktreeStorer fake out the storage layer in such a way that it
```
songgao: s/fake/fakes?

libgit/worktree_storer.go
```diff
@@ -0,0 +1,88 @@
...
+// stores its index, and any reference modifications, in a .git/
+// directory in the worktree, but everything else goes through the
+// main bare repo's .git directory.  This is useful for making a quick
+// checkout of a git repo in a new directory, without making a full
+// git clone. For example, consider the following instance:
+//
+// * `storage.Storer` points to a bare repo stored in KBFS, say
+//   `/keybase/team/keybase/.kbfs_git/secrets/.git`.
```
songgao: Wouldn't it be `/keybase/team/keybase/.kbfs_git/secrets` instead?

libgit/worktree_storer.go
```diff
@@ -0,0 +1,88 @@
...
+// * `delta` is the same as above, but cast as a `DeltaObjectStorer`.
+// * `wtDotgit` can be pointed anywhere else in KBFS where you want to
+//   expose a checked-out version of the keybase secrets repo, say
+//   `/keybase/private/strib/.kbfs_autogit/team/keybase/secrets/`
```
songgao: In the `repoFromStorageAndWorktree` [above](https://github.com/keybase/kbfs/pull/1400/files#diff-e3df4596a1bff828fb4e958580cd23b3R54), `wtDotgit` is the ".git" subdir of the worktree. So it looks like this should've been `/keybase/private/strib/.kbfs_autogit/team/keybase/secrets/.git`?

PR #1403, opened 2017-12-22
vendor: update keybase1 protocol to get new saltpack behavior

strib:
cc: @maxtaco 

Issue: CORE-6789


strib:
@maxtaco @oconnor663: saltpack can start using the new saltpack identify behavior now.

PR #1404, opened 2017-12-22
implement access control using HTTP basic auth

songgao:
Also made some changes to the Config package to support `realm`, which is needed for HTTP Basic Authentication, as an identifier for resources sharing the same credentials.

I updated the staging instance and did some simple test and it seems to have worked well. The config I'm using for kbp-test2.song.gao.io for now is:

```json
{
  "version": "v1",
  "users": {
    "alice": "$2a$10$XDG6k7.k3942gm4pY42oJeJl//HeiNr5X9GgF9AzonNzyKuhFq6ra",
    "bob": "$2a$10$KBzG6TTGkCRr7ZEm6Z7RVOICrOfunmyiLZC6X4uLkEgx5yAWPBAnu"
  },
  "acls": {
    "/alice-and-bob": {
      "whitelist_additional_permissions": {
        "alice": "read,list",
        "bob": "read"
      },
      "anonymous_permissions": ""
    },
    "/bob": {
      "whitelist_additional_permissions": {
        "bob": "read,list"
      },
      "anonymous_permissions": ""
    },
    "/bob/dir/deep-dir/deep-deep-dir": {
      "whitelist_additional_permissions": null,
      "anonymous_permissions": ""
    },
    "/public": {
      "whitelist_additional_permissions": null,
      "anonymous_permissions": "read,list"
    },
    "/public/not-really": {
      "whitelist_additional_permissions": {
        "alice": "read,list"
      },
      "anonymous_permissions": ""
    }
  }
}
```

where `alice` and `bob`'s passwords are gloriously "12345" and "54321".

### DIFF
--- a/libpages/config/acl_checker_v1.go
+++ b/libpages/config/acl_checker_v1.go
@@ -44,8 +44,8 @@ type accessControlV1 struct {
 	// for more details. It's a map of username -> permissionsV1.
 	whitelistAdditional map[string]permissionsV1
 	anonymous           permissionsV1
-	// p stores the path (from Config declaration) that a *accessControlV1
-	// object is constructed for. When a *aclCheckerV1 is picked for a path,
+	// p stores the path (from Config declaration) that an *accessControlV1
+	// object is constructed for. When an *aclCheckerV1 is picked for a path,
 	// the p field can be used as a realm for HTTP Basic Authentication.
 	p string
 }
--- a/libpages/config/acl_checker_v1.go
+++ b/libpages/config/acl_checker_v1.go
@@ -54,7 +54,7 @@ type accessControlV1 struct {
 // *AccessControlV1. The users map is used to check if every username defined
 // in WhitelistAdditionalPermissions is defined.
 func makeAccessControlV1Internal(
-	a *AccessControlV1, users map[string][]byte, p string) (
+	a *AccessControlV1, users map[string]string, p string) (
 	ac *accessControlV1, err error) {
 	if a == nil {
 		return nil, errors.New("nil AccessControlV1")
@@ -194,7 +194,7 @@ func (c *aclCheckerV1) getPermissions(p string, username *string) (
 // recursively constructs nested *aclCheckerV1 so that each defined path has a
 // corresponding checker, and all intermediate nodes have a checker populated.
 func makeACLCheckerV1(acl map[string]AccessControlV1,
-	users map[string][]byte) (*aclCheckerV1, error) {
+	users map[string]string) (*aclCheckerV1, error) {
 	root := &aclCheckerV1{ac: defaultAccessControlV1InternalForRoot()}
 	if acl == nil {
 		return root, nil
--- a/libpages/config/acl_checker_v1.go
+++ b/libpages/config/acl_checker_v1.go
@@ -44,17 +44,22 @@ type accessControlV1 struct {
 	// for more details. It's a map of username -> permissionsV1.
 	whitelistAdditional map[string]permissionsV1
 	anonymous           permissionsV1
+	// p stores the path (from Config declaration) that a *accessControlV1
+	// object is constructed for. When a *aclCheckerV1 is picked for a path,
+	// the p field can be used as a realm for HTTP Basic Authentication.
+	p string
 }
 
 // makeAccessControlV1Internal makes an *accessControlV1 out of an
 // *AccessControlV1. The users map is used to check if every username defined
 // in WhitelistAdditionalPermissions is defined.
-func makeAccessControlV1Internal(a *AccessControlV1, users map[string][]byte) (
+func makeAccessControlV1Internal(
+	a *AccessControlV1, users map[string][]byte, p string) (
 	ac *accessControlV1, err error) {
 	if a == nil {
 		return nil, errors.New("nil AccessControlV1")
 	}
-	ac = &accessControlV1{}
+	ac = &accessControlV1{p: p}
 	ac.anonymous, err = parsePermissionsV1(a.AnonymousPermissions)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


libpages/site.go
```diff
@@ -0,0 +1,77 @@
...
+	"time"
+
+	"github.com/keybase/kbfs/libfs"
+	"github.com/keybase/kbfs/libpages/config"
+)
+
+const configCacheTime = 16 * time.Second
+
+type site struct {
+	// fs should never once constructed.
```
strib: "never once"?

libpages/site.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (s *site) getCachedConfig() (cfg config.Config, expiresAt time.Time) {
+	s.cachedConfigLock.RLock()
+	defer s.cachedConfigLock.RUnlock()
+	return s.cachedConfig, s.cachedConfigExpiresAt
+}
+
+func (s *site) fetchConfigAndRefreshCache() (cfg config.Config, err error) {
+	// Take the lock early to block other reads, since otherwise they would
```
strib: Shouldn't you check the expiration time again after taking the lock, to prevent those unnecessary reads?

libpages/site.go
```diff
@@ -0,0 +1,77 @@
...
+		cfg, err = config.ParseConfig(f)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, err
+	}
+
+	s.cachedConfig = cfg
+	s.cachedConfigExpiresAt = time.Now()
```
strib: Add `configCacheTime`?

libpages/server.go
```diff
@@ -121,12 +137,64 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			r.RequestURI)
+	}
 	if err != nil {
 		s.handleError(w, err)
 		return
 	}
-	http.FileServer(fs).ServeHTTP(w, r)
+
+	if !canRead {
+		s.handleNeedAuthentication(w, r, realm)
```
strib: If it's a directory with no index.html file, shouldn't we allow listing if `canList == true`?  Or do you intend for reading to be a requirement for listing?

songgao: I thought `read` as a prerequisite for `list`, as it'd be like reading a directory. But I guess there's a use case here where one can have `list` on `/foo` which has a bunch of sub-directories, only some of which have `read` for a particular user. Will update to apply your logic!

libpages/site.go
songgao: Whoa how did that happen. Should've been "fs should never be changed once it's constructed".

libpages/site.go
songgao: Whooops. Thanks

libpages/config/acl_checker_v1.go
```diff
@@ -44,17 +44,22 @@ type accessControlV1 struct {
 	// for more details. It's a map of username -> permissionsV1.
 	whitelistAdditional map[string]permissionsV1
 	anonymous           permissionsV1
+	// p stores the path (from Config declaration) that a *accessControlV1
```
jzila: s/a/an, here and on the line below.

libpages/server.go
```diff
@@ -104,13 +106,57 @@ func (a adaptedLogger) Warning(format string, args ...interface{}) {
...
+	w.Header().Set("WWW-Authenticate", fmt.Sprintf("Basic realm=%s", realm))
+	w.WriteHeader(http.StatusUnauthorized)
+}
+
+func (s *Server) isDirWithNoIndexHTML(
+	st *site, requestURI string) (bool, error) {
+	fi, err := st.fs.Stat(strings.Trim(path.Clean(requestURI), "/"))
+	switch {
+	case os.IsNotExist(err):
+		// It doesn't exist! So jsut let the http package handle it.
```
jzila: s/jsut/just

libpages/server.go
```diff
@@ -104,13 +106,57 @@ func (a adaptedLogger) Warning(format string, args ...interface{}) {
...
+
 // ServeHTTP implements the http.Handler interface.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	s.config.Logger.Info("ServeHTTP",
 		zap.String("host", r.Host),
 		zap.String("urk", r.RequestURI),
 		zap.String("proto", r.Proto),
 	)
+
+	if path.Clean(strings.ToLower(r.RequestURI)) == kbpConfigPath {
```
jzila: won't `r.RequestURI` include the domain info??

libpages/server.go
```diff
@@ -121,12 +167,56 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 		s.handleError(w, err)
 		return
 	}
-	http.FileServer(fs).ServeHTTP(w, r)
+
+	// Check if it's a directory containing no index.html before letting
+	// http.FileServer handle it.  This permission check should ideally
+	// happen inside the http package, but unfortunately there isn't a
+	// way today.
+	isListing, err := s.isDirWithNoIndexHTML(st, r.RequestURI)
```
jzila: As above, won't `r.RequestURI` include the domain info?

songgao:
> Also: did we actually decide on zap, or was that stealthily added in?

Nobody opposed in the original PR, so I was assuming you guys were OK with that! But happy to discuss and remove it later if needed!

I personally think zap is great and I'd be glad if we ever switch kbfs-server into it as well!

libpages/server.go
songgao: No I think it's the path under the domain. So for `https://keybase.io/jzila`, `RequestURI` would be `/jzila`

songgao: For example, for a request to `https://kbp-test.song.gao.io/`, this is what we get in logs:
```
2017-12-23T02:13:48.226Z        INFO    libpages/server.go:147  ServeHTTP       {"host": "kbp-test.song.gao.io", "urk": "/", "proto": "HTTP/2.0"}
```

jzila: The RFC says it can be the absolute URI:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2

jzila: Seems like what you want is `r.URL.Path`

songgao: Oh good catch. And it seems the RequestURI can include parameters as well. Will do; thanks!

PR #1406, opened 2017-12-23
libgit: implement AutogitManager

strib:
This manager can clone and pull source git repos into a destination folder, potentially across different TLFs.  New requests for an operation in a destination folder are blocked by any ongoing requests for the same folder, and multiple outstanding requests for the same destination folder get rolled up into one.

Also fixes some issues in libfs.FS to match what gogit is expecting out of a billy.Filesystem.

The new test is pretty basic. I hope to expand them more, after integrating this into the filesystem view.  But if there's anything you think is critical that should be tested now but is not, let me know.

Issue: KBFS-2674


libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+		am.lock.Lock()
+		defer am.lock.Unlock()
+		if req, ok := am.resetsInQueue[dp]; ok {
+			return req.doneCh
+		}
+		am.resetsInQueue[dp] = req
+		return nil
+	}()
+	if doneCh != nil {
+		return doneCh, nil
```
songgao: I guess this is where "multiple outstanding requests for the same destination folder get rolled up into one" comes from. But what if `req` is different from what's in `am.resetsInQueue` already? For example it could be about "resetting" a different repo within the same `dstPath`, or "resetting" the same repo to a different ref right?

libgit/init.go
```diff
@@ -109,3 +110,56 @@ func Init(ctx context.Context, gitKBFSParams libkbfs.InitParams,
...
+
+func (ksp keybaseServicePassthrough) NewCrypto(
+	_ libkbfs.Config, _ libkbfs.InitParams, _ libkbfs.Context,
+	_ logger.Logger) (libkbfs.Crypto, error) {
+	return ksp.config.Crypto(), nil
+}
+
+var _ libkbfs.KeybaseServiceCn = keybaseServicePassthrough{}
+
+func getNewConfig(
```
jzila: I think you mean `gitNewConfig` 😆

libfs/fs.go
```diff
@@ -305,7 +305,7 @@ func (fs *FS) lookupOrCreateEntry(
 	filename string, flag int, perm os.FileMode) (
 	n libkbfs.Node, ei libkbfs.EntryInfo, err error) {
 	// Shortcut the case where there's nothing to look up.
-	if filename == "" || filename == "/" {
+	if filename == "" || filename == "/" || filename == "." {
 		return fs.root, fs.rootInfo, nil
 	}
 	filename = strings.TrimPrefix(filename, "/")
```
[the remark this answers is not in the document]
strib: I don't understand why that matters for this check -- this is checking the whole file name, not the first character.  In any case, callers are supposed to build paths with `Join`, which cleans them already.

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+	// And a dst parent checkout FS, which better already exist.
+	dstFS, err := libfs.NewFS(
+		ctx, gitConfig, req.dstTLF, req.dstDir, uniqID,
+		keybase1.MDPriorityNormal)
+	if err != nil {
+		return err
+	}
+
+	// Take the lock for the dst repo while resetting.
+	lockFile, err := dstFS.Create(autogitLockName(req.srcRepo))
```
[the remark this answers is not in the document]
strib: Yeah exactly, I didn't feel like making a way to specify the destination repo name -- right now it's always the same as the source repo name.  This lock does implicitly include the whole repo path (`dstPath`+`srcRepo`), since it includes the lock namespace for the `FS`.

libgit/autogit_manager.go
strib: Yeah, it's a bug that `resetReq.dstPath()` doesn't include the repo name.  Good catch, I'll fix.

libfs/fs.go
```diff
@@ -475,6 +475,11 @@ func (fs *FS) Rename(oldpath, newpath string) (err error) {
 		err = translateErr(err)
 	}()
 
+	err = fs.mkdirAll(path.Dir(newpath), 0755)
```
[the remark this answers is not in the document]
strib: You would think so, but apparently not in billy-land.  go-git relies on auto-dir creation for both regular file opens and for renames.

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+		am.deferLog.CDebugf(ctx, "Reset request completed: %+v", err)
+	}()
+
+	// Make a new single-op config for processing this request.
+	ctx, gitConfig, tempDir, err := am.getNewConfig(ctx)
+	if err != nil {
+		return err
+	}
+	// Cloned/pull data should be regular data blocks.
+	gitConfig.SetDefaultBlockType(keybase1.BlockType_DATA)
```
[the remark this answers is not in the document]
strib: Because `libgit.Init()` sets the default block type to GIT, so we need to set it back to DATA here so the checkouts count towards regular quota.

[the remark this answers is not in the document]
jzila: Perfect, thanks.

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+				"%s/%s to %s/%s",
+				req.srcTLF.GetCanonicalPath(), req.srcRepo,
+				req.dstTLF.GetCanonicalPath(), req.dstDir)
+			<-waitCh
+			am.log.CDebugf(ctx, "Done waiting")
+		}
+
+		_ = am.doReset(ctx, req)
+		am.clearFromInProgress(req)
+		close(req.doneCh)
```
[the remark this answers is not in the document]
strib: > Actually thinking about it a bit more, maybe you can close the channel first. Then if another worker races to get the in-progress request, it'll get the already-closed channel, rather than creating a new request.

A race on the in-progress channel doesn't result in a new request.  It only applies to the `for` loop right above here, which waits for any in-progress request to finish before starting the next one.  (The next one has to be done, because it must have been started after the current reset started, and therefore might refer to a new head.)  You might be confusing it with the entry in `resetsInProgress`, which has already been cleared above.

I think it's somewhat irrelevant whether it's cleared from in-progress first or closed first.  I'll clarify in a comment.

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+// however.
+//
+// If the cloned directory doesn't exist yet, this function creates a
+// placeholder "CLONING" file to let users know the clone is happening
+// in the background.
+//
+// Note that if there's already data in `dstDir/srcRepo`, this
+// tramples it, so the caller must ensure that they are requesting a
+// clone from the same repo/branch as the existing repo that might
+// have already been there.
```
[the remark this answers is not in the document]
strib: I'm relying on the correctness of go-git's `Reset` implementation, but in theory it should work exactly the same as regular git's reset, right?

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+//
+// It returns a channel that, when closed, indicates the pull request
+// has finished (though not necessarily successfully).  The caller may
+// have to sync from the server to ensure they are see the changes,
+// however.
+//
+// Note that this tramples any data that was previously in
+// `dstDir/srcRepo`, so the caller must ensure that they are
+// requesting a pull from the correct repo/branch as the existing repo
+// that was already there.
```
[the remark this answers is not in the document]
strib: I'm not planning to expose branches at first to the user, but I'm thinking the UI will have non-master branch names as part of the pathname somewhere.  So you'd never let a user switch a branch on a given checkout, it would just do a whole new checkout.

That said, if somehow a caller passed in a different branch with the same `dstDir`, it should work fine -- go-git's `Reset` should just take care of it.  The branch name just gets resolved to a commit hash, and that's what the reset uses.

libgit/init.go
strib: No, I don't think I meant that.  "git" is already implied in the library name.  If you don't like the "get", I can drop it and just name the function `newConfig()`.

jzila: I know, I was kidding :)

strib:
> Also, could you explain why "an operation in a destination folder are blocked by any ongoing requests for the same folder", i.e., why is it per-dstPath rather than per-repo? It seems to me we different repos being checked-out to the same dstPath should be allowed.

@songgao: Is this clear now?  The dest path was supposed to contain the repo name, along with the full path in this dest TLF.

libgit/init.go
jzila: The emoji doesn't come across in email unfortunately.

strib: Oops sorry, I'm not well versed in emoji.  :older_man:

libfs/fs.go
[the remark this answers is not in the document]
songgao: Oh I see. Makes sense!

libgit/autogit_manager.go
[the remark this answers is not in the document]
songgao: Sounds good to me! Would it ever happen that we are checking out repos from different TLFs into same `dstFS`, or is "checkout" of repos from a different TLF always be under a full path (e.g. `.kb_git/team/keybase/repo.git`)? If latter, then it seems we may never need a way to make `dstRepo` different from `srcRepo` in the future.

songgao:
> @songgao: Is this clear now? The dest path was supposed to contain the repo name, along with the full path in this dest TLF.

Yup it totally makes sense now. Thanks!

libgit/autogit_manager.go
```diff
@@ -0,0 +1,404 @@
...
+	srcRepo    string
+	branchName string
+	dstTLF     *libkbfs.TlfHandle
+	dstDir     string
+	doneCh     chan struct{}
+}
+
+func (r resetReq) dstPath() string {
+	return path.Clean(path.Join(
+		r.dstTLF.GetCanonicalPath(), r.dstDir, r.srcRepo, r.branchName))
```
songgao: As you mentioned in [another comment](https://github.com/keybase/kbfs/pull/1406/files#r158985366), `path.Join` already calls `path.Clean` before returning: https://golang.org/src/path/path.go?s=4034:4066#L145

libgit/autogit_manager.go
```diff
@@ -0,0 +1,404 @@
...
+type resetReq struct {
+	srcTLF     *libkbfs.TlfHandle
+	srcRepo    string
+	branchName string
+	dstTLF     *libkbfs.TlfHandle
+	dstDir     string
+	doneCh     chan struct{}
+}
+
+func (r resetReq) dstPath() string {
```
songgao: Is it planed to use `resetReq.dstPath()` for something other than keys in `resetsInQueue` and `resetsInProgress`? If it's not gonna be used as paths, would it be clearer to rename this into something like `resetReq.id()`?

libgit/autogit_manager.go
```diff
@@ -0,0 +1,404 @@
...
+func autogitLockName(srcRepo string) string {
+	return fmt.Sprintf(".%s.lock", srcRepo)
+}
+
+type getNewConfigFn func(context.Context) (
+	context.Context, libkbfs.Config, string, error)
+
+// AutogitManager can clone and pull source git repos into a
+// destination folder, potentially across different TLFs.  New
+// requests for an operation in a destination folder are blocked by
```
songgao: So "destination folder" should be "destination repo+branch" here per `resetReq.dstPath()` implementation.

songgao: Also, this means reset requests for the same repo with different branch names won't block each other or get rolled into one. Is that intended? Does it affect user-facing "checked-out" branch, or is this just to get all necessary objects written?

libgit/autogit_manager.go
```diff
@@ -0,0 +1,397 @@
...
+	if doneCh != nil {
+		return doneCh, nil
+	}
+	select {
+	case am.resetQueue <- req:
+		am.log.CDebugf(ctx, "Queued new reset request for %s", dp)
+	case <-ctx.Done():
+		// We've already promised to queue this, and may have turned
+		// away other requests for it already, so we better queue it.
+		go func() { am.resetQueue <- req }()
```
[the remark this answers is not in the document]
songgao: So many `interface{}` in that `channels` package! This is a rare situation where I wish Go had template/generics :)

libgit/autogit_manager.go
strib: The intent is that the caller will use any non-master branch name in the `dstDir` passed to `Clone`/`Pull`.  I guess I should just take the branch name out of the id.  I'll write a comment to clarify.

libgit/autogit_manager.go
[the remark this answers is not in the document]
strib: Yeah, the caller should always do checkouts under a full, unique path.  Not enforced in this interface, but the next layer up will do it.

libgit/autogit_manager.go
```diff
@@ -307,6 +306,10 @@ func (am *AutogitManager) makeCloningFile(
 // tramples it, so the caller must ensure that they are requesting a
 // clone from the same repo/branch as the existing repo that might
 // have already been there.
+//
+// If the caller specifies a non-master `branchName`, they should make
```
songgao: Hrm how did this line not starting with "Clone ..." pass the linter? Maybe we aren't linting this package.

songgao:
Thanks for clarifying in comment!

libgit/autogit_manager.go
strib: It's not the first line in the comment.  Only the first paragraph matters,

songgao: Oh I see; I totally missed the previous lines..

PR #1407, opened 2017-12-23
quick fix to make Pages work for prod Let's Encrypt

songgao:


libpages/server.go
```diff
@@ -298,8 +298,8 @@ func ListenAndServe(ctx context.Context,
 		HostPolicy: server.allowConnectionTo,
 	}
 
-	if config.UseDiskCacheForDev {
-		manager.Cache = autocert.DirCache("./kbp-cert-cache-dev")
+	if config.UseDiskCertCache {
+		manager.Cache = autocert.DirCache("./kbp-cert-cache")
```
strib: Should this use a different directory if `!fProd`, so the certs don't get mixed up if you switch between modes on the same instance?

songgao: I figured we'd almost never use the same instance for testing and prod, so I just removed the `-dev` suffix. But having different suffix sounds good with me too; will change!

kbpagesd/main.go
```diff
@@ -23,7 +23,7 @@ var (
 
 func init() {
 	flag.BoolVar(&fProd, "prod", false, "disable development mode")
```
jzila: Should this be true now by default?

songgao: I'd prefer having to explicitly set it. Same concept as requiring to set `KEYBASE_RUN_MODE` for non-dev environments. WDYT?

jzila: Sounds good.

PR #1411, opened 2017-12-29
[KBFS-2682] kbfsgit: Support loading commits from refspecs.

jzila:


jzila:
@strib: when you have a moment, I'd love if you could take a look to validate my approach here.

I'm thinking of doing some kind of backward traversal to get parent commits, but my issue is to figure out what the pre-fetch state is, so I know at which commits to stop. WDYT?

kbfsgit/runner.go
```diff
@@ -1501,17 +1503,49 @@ func (r *runner) pushSome(
...
 	return results, nil
 }
 
+func (r *runner) refsToCommits(ctx context.Context,
+	storer gogitstor.EncodedObjectStorer,
+	refs []*plumbing.Reference) ([]*gogitobj.Commit, error) {
+	erroredRefs := make(map[string]error, 0, len(refs))
+	commits := make([]*gogitobj.Commit, 0, len(refs))
+	for _, ref := range refs {
+		commit, err := gogitobj.GetCommit(storer, ref.Hash())
```
strib: To get the parent commits here, you can use `gogit.Repository.Log()` for the local `Repository` you need to create, and pass it the commit hash.  You'll get back an iterator that you can follow as long as desired.

kbfsgit/runner.go
```diff
@@ -1501,17 +1503,49 @@ func (r *runner) pushSome(
...
+		// TODO (KBFS-2682): also load parent refs.
+		refs := make([]*plumbing.Reference, 0, len(refspecs))
+
 		// All non-delete refspecs in the batch get the same error.
 		for _, refspec := range refspecs {
 			refStr := refspec.String()
 			start := strings.Index(refStr, ":") + 1
 			dst := refStr[start:]
 			results[dst] = err
+			ref, refErr := repo.Storer.Reference(plumbing.ReferenceName(dst))
```
strib: `repo.Storer` will give you the commits that are already in KBFS, not the ones being pushed from the local repo.  I think you'll need to make a new `Repository` object using a `billy.osfs` storage layer pointing to the local repo, to find the references being pushed.

jzila: At this point they should already be in KBFS, since we've already done the `FetchContext`. Am I missing something?

jzila: I'll take your suggestion anyway and move the delta derivation to before the `FetchContext`, since otherwise calculating the delta is harder. But I still want to make sure that my understanding is correct.

strib: Ah I see, yeah I missed that it was after the FetchContext.  You're right.  But yeah, you should do it before, so you can figure out what changed.

jzila: Thanks! I hope to have a final review ready for you tomorrow.

jzila:
I believe this is ready for review. Obviously it's not connected yet, but will be once we have the Core API.

kbfsgit/runner.go
```diff
@@ -1462,13 +1524,22 @@ func (r *runner) pushSome(
 			}
 			results[dst] = err
 		} else {
+			refName := plumbing.ReferenceName(refspec.Src())
```
strib: Have you tested this with wildcard pushes?  I don't actually know whether wildcard refspecs can come through the git-remote-helper interface, or if they get expanded elsewhere.  If the latter, can we error if `IsWildcard` here?  If the former, we probably need to find some gogit utility for expanding the wildcards.

kbfsgit/runner.go
```diff
@@ -1423,26 +1430,81 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
 		fmt.Sprintf("Preparing and %s packed refs", verb), "pushprefs")
 }
 
+// parentCommitsForRef returns a map of refs with a list of commits for each
+// ref, newest first. It only includes commits that exist in `localStorer` but
+// not in `remoteStorer`.
+func (r *runner) parentCommitsForRef(ctx context.Context,
+	localStorer gogitstor.Storer, remoteStorer gogitstor.Storer,
+	refs refLookupByName) (commitsByRefName, error) {
```
strib: Why is `refs` a map?  The key isn't used that I see.  Could it just be a slice, or is there some uniqueness property that this is important for achieving?

kbfsgit/runner_test.go
```diff
@@ -1004,3 +1016,57 @@ func TestRunnerWithKBFSReset(t *testing.T) {
...
+func TestRunnerHandlePushBatch(t *testing.T) {
+	ctx, config, tempdir := initConfigForRunner(t)
+	defer libkbfs.CheckConfigAndShutdown(ctx, t, config)
+	defer os.RemoveAll(tempdir)
+
+	git, err := ioutil.TempDir(os.TempDir(), "kbfsgittest")
+	require.NoError(t, err)
+	defer os.RemoveAll(git)
+
+	// Setup the repo
```
strib: End with punctuation.

kbfsgit/runner_test.go
```diff
@@ -1004,3 +1016,57 @@ func TestRunnerWithKBFSReset(t *testing.T) {
...
+	h, err := libkbfs.ParseTlfHandle(
+		ctx, config.KBPKI(), config.MDOps(), "user1", tlf.Private)
+	require.NoError(t, err)
+	_, err = libgit.CreateRepoAndID(ctx, config, h, "test")
+	require.NoError(t, err)
+
+	t.Log("Make a new local repo with one commit, and push it. " +
+		"We expect this to return no commits, since it should push the " +
```
strib: Was this explicitly decided on?  Seems like it should contain the most recent N commits being pushed, in my opinion.  Might be worth adding this to `pushAll` as well.

kbfsgit/runner_test.go
```diff
@@ -1004,3 +1016,57 @@ func TestRunnerWithKBFSReset(t *testing.T) {
...
+	makeLocalRepoWithOneFileCustomCommitMsg(t, git, "foo", "hello", "", "one")
+	commits := testHandlePushBatch(t, ctx, config, git,
+		"refs/heads/master:refs/heads/master", "user1")
+	require.Empty(t, commits)
+
+	addOneFileToRepoCustomCommitMsg(t, git, "foo2", "hello2", "two")
+	commits = testHandlePushBatch(t, ctx, config, git,
+		"refs/heads/master:refs/heads/master", "user1")
+	require.Len(t, commits, 1)
+	require.Len(t, commits["refs/heads/master"], 1)
```
strib: It would be great if this test checked the commit messages too, otherwise there's no reason for those `CustomCommitMsg` methods.

kbfsgit/runner_test.go
jzila: No, this is until I get a response on how we want `pushAll` to behave.

kbfsgit/runner.go
jzila: 1. Uniqueness to ref name.
2. Easier lookup if we ever need to.
3. We need the ref name anyway, and it seems intuitive to me that I use that as the key. On the ticket we're discussing mapping commits to the branch name.

kbfsgit/runner.go
jzila: I haven't. Should we error, or should we just add the refspec and ignore the commits?

jzila:
@strib: here's a link to the implemented feedback, and the glue that ties everything together: https://github.com/keybase/kbfs/pull/1411/files/df80624ac77a0b5f4d089caaa3ea6c7912d64065..c0edf97f58bb60fcd56275639413c48b1ef05e1e

I'm still figuring out how to deal with `pushAll`, dependent on whether we get refspecs passed in or not.

kbfsgit/runner.go
strib: We should test and see what happens when there's a wildcard on the command -- what does the remote helper see?  If it also sees a wildcard refspec, I think we have to support commits for it, it would be pretty common.

jzila: OK I'll test.

kbfsgit/runner_test.go
strib: Note that `pushSome` can still be what's used for the initial repo creation (if not all of the local branches are being pushed), so the remote repo could be empty in either case, for what it's worth.

jzila: That's fine, the code should work either way.

kbfsgit/runner.go
strib: Ah, the same src refspec can be used to push into multiple remote branches, so the uniqueness _is_ required I guess.  Carry on.

libgit/repo.go
```diff
@@ -181,7 +182,11 @@ func CleanOldDeletedReposTimeLimited(
 // UpdateRepoMD lets the Keybase service know that a repo's MD has
 // been updated.
 func UpdateRepoMD(ctx context.Context, config libkbfs.Config,
-	tlfHandle *libkbfs.TlfHandle, fs billy.Filesystem) error {
+	tlfHandle *libkbfs.TlfHandle, fs billy.Filesystem,
+	commitsByRef map[plumbing.ReferenceName][]*object.Commit) error {
+	if len(commitsByRef) == 0 {
+		commitsByRef = map[plumbing.ReferenceName][]*object.Commit{"": nil}
```
strib: Would be nice to have a quick comment here explaining what this is doing.

kbfsgit/runner.go
```diff
@@ -1423,26 +1430,81 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
...
+		walker := gogitobj.NewCommitPreorderIter(commit, haves, nil)
+		toVisit := maxCommitsToVisitPerRef
+		err = walker.ForEach(func(c *gogitobj.Commit) error {
+			haves[c.Hash] = true
+			toVisit--
+			// If toVisit starts out at 0 (indicating there is no
+			// max), then it will be negative here and we won't stop
+			// early.
+			hasEncodedObjectErr := remoteStorer.HasEncodedObject(c.Hash)
+			if toVisit == 0 || hasEncodedObjectErr == nil {
```
songgao: I slightly prefer `if toVisit == 0 || remoteStorer.HasEncodedObject(c.Hash) == nil` here but I'm OK with this too.

jzila: I had it like that first and found myself having to mentally detangle the cases...separated out it seemed easier to reason through.

jzila: The issue is that `HasEncodedObject` is terribly named: a name like that should return a `bool`, and it's confusing that it doesn't.

songgao: Oh I see. 

> a name like that should return a bool

Yeah agree; that was why I didn't realize this was what deals with the case where a commit is found in remote branch, and my motivation of making this comment.

But the three points above are helpful enough so carry on!

kbfsgit/runner.go
jzila: It looks like wildcards are ignored:
```
Johns-iMac.local 18:57:41 zipdroid (jzila/test-2) ➜ git push origin jzila/*
Initializing Keybase... done.
Syncing with Keybase... done.
Everything up-to-date
```
```
Johns-iMac.local 18:59:48 zipdroid (jzila/test-2) ➜ gp
Initializing Keybase... done.
Syncing with Keybase... done.
Counting and decrypting: 3 objects... done.
Preparing and encrypting: (100.00%) 3/3 objects... done.
Indexing hashes: (100.00%) 3/3 objects... done.
Indexing CRCs: (100.00%) 3/3 objects... done.
Indexing offsets: (100.00%) 3/3 objects... done.
Syncing encrypted data to Keybase: (100.00%) 41.92/41.92 KB... done.
Syncing encrypted data to Keybase: (100.00%) 18.72/18.72 KB... done.
To keybase://private/jzila/zipdroid.git
 * [new branch]      jzila/test-2 -> jzila/test-2
Branch 'jzila/test-2' set up to track remote branch 'jzila/test-2' from 'origin'.
```

jzila: Actually I was using incorrect syntax. Looks like git expands the wildcard itself:
```
Johns-iMac.local 19:18:57 zipdroid (john) ➜ git push origin refs/heads/jzila/*
Initializing Keybase... done.
Syncing with Keybase... done.
Syncing encrypted data to Keybase: (100.00%) 20.39/20.39 KB... done.
Syncing encrypted data to Keybase: (100.00%) 18.72/18.72 KB... done.
To keybase://private/jzila/zipdroid.git
 * [new branch]      jzila/test-1 -> jzila/test-1
 * [new branch]      jzila/test-2 -> jzila/test-2
```

Which shows up in the logs as:
```
Johns-iMac.local 19:19:57 zipdroid (john) ➜ tail -f ~/Library/Logs/keybase.git.log | grep 'runner.go'
...
2018-01-03T19:20:10.163428-08:00 ▶ [DEBU git runner.go:1773] 3c8 Received command: push refs/heads/jzila/test-1:refs/heads/jzila/test-1
2018-01-03T19:20:10.163505-08:00 ▶ [DEBU git runner.go:1773] 3c9 Received command: push refs/heads/jzila/test-2:refs/heads/jzila/test-2
2018-01-03T19:20:10.163538-08:00 ▶ [DEBU git runner.go:1773] 3ca Received command:
2018-01-03T19:20:10.163595-08:00 ▶ [DEBU git runner.go:1704] 3cb Processing push batch [tags:GITID=Iavjs5KrdgI1yB_UwiXDmA,GITCMDID=GZFZBS8EIFbk1KRbpDmmyA]
```

jzila:
@strib I moved the commit fetching code to before the `pushSome`/`pushAll` branch, so now it always happens.

It looks like wildcards are expanded before the commands are sent to the helper. Shall I ship?

PR #1412, opened 2017-12-29
redirect HTTP to HTTPS and logging tweaks

songgao:


libpages/server.go
```diff
@@ -271,10 +272,61 @@ const (
...
+			zap.String("Opaque", req.URL.Opaque))
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	http.Redirect(w, req, "https://"+req.Host+req.URL.RequestURI(),
+		// Using 307 (Temporary Redirect) here, 1) instead of 302 to require
+		// same method to be used; and 2) instead of 308/301 to avoid pollute
+		// visitors' browser cache in case kbp user wants to use the domain in
+		// a way that needs HTTP not to be redirected to HTTPS.
+		http.StatusTemporaryRedirect)
```
jzila: https://support.google.com/webmasters/answer/6073543?hl=en

Standard is to use 301 redirects.

libpages/server.go
```diff
@@ -332,8 +369,18 @@ func ListenAndServe(ctx context.Context,
 		shutdownCtx, cancel := context.WithTimeout(
 			context.Background(), gracefulShutdownTimeout)
 		defer cancel()
-		httpServer.Shutdown(shutdownCtx)
+		httpsServer.Shutdown(shutdownCtx)
+		httpRedirectServer.Shutdown(shutdownCtx)
```
jzila: Why aren't these the same server? Seems wasteful to have 2 sets of listeners for it.

libpages/server.go
songgao: Thanks for the link! First that's not a standard, but a best-practice guide. Second, this is different in the sense that we are not redirecting our own website; we are redirecting user's domains. 301 saves one round-trip, but when the cache pollution is a problem for the domain owner, there's pretty much nothing you can do.

I think it might make sense in the future to switch to 308 (just like 301) if opted-in through `.kbp_config`. But that's slightly more work. WDYT? If that sounds good to you, I'll create a ticket.

libpages/server.go
songgao: > Why aren't these the same server? 

They are not same server because the two servers handle requests in different ways. I don't see how different handlers on same path can be put into same `http.Server`. Or did you mean something different?

> Seems wasteful to have 2 sets of listeners for it.

We need a different listener for each port we are listening on right?

jzila: Argh brain fart, I forgot that they need to be on separate ports.

libpages/server.go
songgao: Sounds good; thanks! FWIW it'd only be an extra roundtrip for every first visit. Subsequent requests in the same tab would go through HTTPs as long as no `http://` absolute address is in the HTML.

libpages/server.go
songgao: Must have been too much holiday food :D

jzila: Also: I'm used to having a server of some kind proxying requests, so the application port is usually arbitrary.

PR #1414, opened 2018-01-03
kbfs_ops_concur_test: disable prefetch during sensitive tests

strib:
These tests measure precisely the order in which MDs are read, and the prefetcher running async in the background can cause flakes.


jzila:
Great catch!

PR #1415, opened 2018-01-03
node_cache: add Node wrappers for custom behavior

strib:
This PR adds a way to "wrap" Nodes as they get created by the `NodeCache`.  The idea is that this will allow us to add custom behavior to each Node, which can be controlled from outside of the `libkbfs` package.  The caller can add a root wrapper function that can transform a root node whenever one is created.  Then, that wrapper can choose to transform any child nodes that are created, via the `WrapChild` function.  And so on, down the tree of nodes, as needed.

This will be used by the autogit implementation in the `libgit` package, to control Node properties that will be added in future PRs: read-only-ness, and hooks for reads, updates and lookup misses which will automatically kick off clones and pulls.  (See KBFS-2654 for the design.)

Issue: KBFS-2675

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -124,6 +124,9 @@ type Node interface {
 	// GetBasename returns the current basename of the node, or ""
 	// if the node has been unlinked.
 	GetBasename() string
+	// WrapChild returns a wrapped version of `child`, if desired, to
+	// add custom behavior to the child node.
+	WrapChild(child Node) Node
 }
 
 // KBFSOps handles all file system operations.  Expands all indirect
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -125,8 +125,14 @@ type Node interface {
 	// if the node has been unlinked.
 	GetBasename() string
 	// WrapChild returns a wrapped version of `child`, if desired, to
-	// add custom behavior to the child node.
+	// add custom behavior to the child node.  If the Node instance
+	// receiving this call is itself a wrapped node, it should call
+	// `WrapChild(child)` on its internal wrapped node as well, in
+	// case there are multiple wrapping layers available.
 	WrapChild(child Node) Node
+	// Unwrap returns the initial, unwrapped Node that was used to
+	// create this Node.
+	Unwrap() Node
 }
 
 // KBFSOps handles all file system operations.  Expands all indirect
@@ -1926,6 +1932,9 @@ type NodeCache interface {
 	PathFromNode(node Node) path
 	// AllNodes returns the complete set of nodes currently in the cache.
 	AllNodes() []Node
+	// AddRootWrapper adds a new wrapper function that will be applied
+	// whenever a root Node is created.
+	AddRootWrapper(func(Node) Node)
 }
 
 // fileBlockDeepCopier fetches a file block, makes a deep copy of it
--- a/libkbfs/node_cache_test.go
+++ b/libkbfs/node_cache_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/tlf"
+	"github.com/stretchr/testify/require"
 )
 
 func setupNodeCache(t *testing.T, id tlf.ID, branch BranchName, flat bool) (
@@ -112,8 +113,8 @@ func simulateGC(ncs *nodeCacheStandard, liveList []Node) {
 
 		// Everything referenced as a parent is live.
 		for _, e := range ncs.nodes {
-			p := e.core.parent
-			if p != nil {
+			if e.core.parent != nil {
+				p := e.core.parent.Unwrap().(*nodeStandard)
 				liveSet[p.core] = true
 			}
 		}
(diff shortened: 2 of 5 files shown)


libkbfs/node_cache_test.go
```diff
@@ -472,3 +473,42 @@ func TestNodeCacheGCReal(t *testing.T) {
 	// Make sure childNode2 isn't GCed until after this point.
 	func(interface{}) {}(childNode2)
 }
+
+type wrappedTestNode struct {
+	Node
+	wrapCalled bool
```
songgao: Make it `wrapChildCalled`?

libkbfs/node_cache_test.go
```diff
@@ -472,3 +473,42 @@ func TestNodeCacheGCReal(t *testing.T) {
...
+	ncs.AddRootWrapper(rw2)
+
+	rootPtr := BlockPointer{ID: kbfsblock.FakeID(0)}
+	rootName := "root"
+	rootNode, err := ncs.GetOrCreate(rootPtr, rootName, nil)
+	require.NoError(t, err)
+
+	childPtr := BlockPointer{ID: kbfsblock.FakeID(1)}
+	childName := "child1"
+	_, err = ncs.GetOrCreate(childPtr, childName, rootNode)
```
songgao: To confirm my understanding, this is where the `WrapChild` gets called on `rootNode`, with the child node passed in as argument, correct?

strib: Yes, correct.

libkbfs/interfaces.go
```diff
@@ -124,6 +124,15 @@ type Node interface {
 	// GetBasename returns the current basename of the node, or ""
 	// if the node has been unlinked.
 	GetBasename() string
+	// WrapChild returns a wrapped version of `child`, if desired, to
+	// add custom behavior to the child node.  If the Node instance
+	// receiving this call is itself a wrapped node, it should call
+	// `WrapChild(child)` on its internal wrapped node as well, in
+	// case there are multiple wrapping layers available.
```
jzila: I don't understand this last sentence. Shouldn't this `WrapChild` apply all necessary wrappers?

strib: I'm thinking of a world where there could be multiple independent packages, all outside of `libkbfs`, that want to apply their own `Node` wrappers.  We have each of these register their own root node wrappers, but we don't know what those root node wrappers will do.  So when `WrapChild` is called, the implementing `Node` instance won't know about all the other wrappers out there; it doesn't seem to be practical to have a central registry, outside of the ones for the root nodes.  So I thought it would be easiest to require a pseudo-recursive call to make sure all wrappers get applied, without tracking anything centrally.  If you have a better idea, I'd be open to it, this is definitely a little kludgy.

jzila: Here's how I understood it:
* cache has a bunch of root node wrappers, `w1`, `w2`.
* `n1` is cached, with children `n1n1`, `n1n2`.
* Cache has `w1(w2(n1))=n1w`.
* `n1n1` is cached. Cache finds the parent `n1`, and calls `n1.WrapChild`, which passes `n1`'s existing wrappers onto `n1n1`, creating `w1(w2(n1n1))`.

So my confusion rests in understanding what the above sentence is saying: are you saying that if `n1n1` is already wrapped, you need `n1.WrapChild` to be called on `n1n1.Unwrap()`? If so, why?

strib: > n1n1 is cached. Cache finds the parent n1w, and calls n1w.WrapChild(n1n1), which passes n1w's existing wrappers onto n1n1, creating w1(w2(n1n1)).

`n1w` doesn't have a set of "existing wrappers".  It's just an implementation of `Node`, which may have another embedded `Node` within it (or equivalently, a pointer to a different `Node`).  Let me define some vars:

* `n1n1` == NodeA, of type `nodeStandard`
* `w2(n1n1)` == NodeA2, of some type produced by root node wrapper `w2`.  It contains an embedded `Node`, set to NodeA.
* `w1(w2(n1n1))` == NodeA1, of some type produced by root node wrapper `w1`.  It contains an embedded `Node`, set to NodeA2. 

Later, if the user wants to add a directory entry under the directory represented by NodeA, they will pass their copy of NodeA2 into e.g. the `KBFSOps.CreateDir()` function.  But NodeA2 has no idea about the full set of available wrappers in the system; all it has available is the embedded pointer to NodeA1.  So to make sure the child is wrapped in all available ways, it should call `NodeA1.WrapChild(node)`, before applying its own wrapper.  Similarly, NodeA1 should call `NodeA.WrapChild(node)` before applying _its_ own wrapper.

Does that help?

jzila: > But NodeA2 has no idea about the full set of available wrappers in the system; all it has available is the embedded pointer to NodeA1.

But NodeA1 actually has an embedded NodeA2...do you mean NodeA?

I'm trying to reason about this in terms of the tree of nodes. Specifically, the important questions are:
* What is wrapped at which point with which wrappers
* Where the wrappers are "stored" (or not stored)
* What is traversed in order to do the wrapping and unwrapping.

Here is my revised understanding based on your description: (root wrapper `w1`), `a` has a wrapper of `w2`, and `a`'s child `aa` is a bare `*nodeStandard`. Then, `aa` has a child of `aaa`, which is again just a `*nodeStandard`.

1. Insert the wrapped `w2(a)`: wrap it with the root wrapper of `w1`:
* root: `w1(w2(a))`
2. Insert `aa`: call `a.WrapChild`: `w2(a)`, and wrap it with the root `w1` wrapper:
* root: `w1(w2(a))`
  * `w1(w2(aa))`
3. Insert `aaa`: call the no-op `aa.WrapChild` finally adding the root wrapper `w1`: `w1(aaa)`:
* root: `w1(a)`
  * `w1(w2(aa))`
    * `w1(aaa)`

It looks like we lost a wrapper at `aaa`. Is that correct?

strib: > But NodeA1 actually has an embedded NodeA2...do you mean NodeA?

Oops, I meant to say NodeA1 has an embedded NodeA2.  Edited.

> It looks like we lost a wrapper at aaa. Is that correct?

`w1` and `w2` are only _root_ wrapper functions, and are only used to wrap the root node.  So it doesn't technically make sense to write `w1(w2(aa))`.  Let's call the type returned by `w1` as `T1`, and the type returned by `w2` as `T2`, and let's say `w1` was added first and then `w2`, and lets say calls to `WrapChild` want to propagate its own type to the child (not required).

This is what should happen.
1. Instantiating the root node: The node cache creates a node `a` of type `*nodeStandard` and wraps it in succession with root wrappers `w1` then `w2`, inserting `T2(T1(a))` into the cache and returning it to the caller.
2. Creating/looking up a child directory `aa` under `a`: The node cache is given `T2(T1(a))`, makes a new node aa of type `*nodeCacheStandard`, and calls `T2.WrapChild(aa)`.  The node cache can't call any other wrappers here, because this isn't the root node.  It's up to `T2.WrapChild()` to call `T1.WrapChild()` on its embedded `Node` before wrapping it, and then `T1.WrapChild` calls `nodeStandard.WrapChild`.  Ultimately, this produces `T2(T1(aa))`.
3. Creating/looking up a child `aaa` under `aa`: Exactly the same thing happens here, ultimately producing `T2(T1(aaa))`.

jzila: I only have one root wrapper `w1` in the simplified example.

Using types instead, let's see if we can revisit my latest example:

1. Insert the wrapped `T2(a)`: wrap it with the root wrapper of `w1`: `T1(T2(a))`:
* root: `T1(T2(a))`
2. Insert `aa`: call `a.WrapChild`, which may or may not recurse to its child: `T1(T2(aa))` or `T1(aa)`
* root: `T1(T2(a))`
  * `T1(T2(aa))` or `T1(aa)`
3. Insert `aaa`: call `aa.WrapChild`: `T1(T2(aaa))` or `T1(aaa)`
* root: `T1(T2(a))`
  * `T1(T2(aa))` or `T1(aa)`
    * `T1(T2(aaa))` or `T1(aaa)`

So it's up to the types to determine whether to make its wrapping inheritable by children.

I _think_ that my understanding is now correct, and that the behavior is correct, but please let me know if I'm still missing something.

strib: Yes, that sounds right.

Do you recommend any changes, either to the scheme or the documentation?

jzila: Yeah, I'd find it easier to follow if this comment read as follows:
> WrapChild returns a wrapped version of `child`, if desired, to add custom behavior to the child node. An implementation that wraps another Node (`inner`) must first call `inner.WrapChild` before performing its own wrapping operation, to ensure that all wrapping is preserved and that it happens in the correct order.

PR #1416, opened 2018-01-03
add kbpagesconfig CLI for editing .kbp_config file

songgao:
This should provide an interface for everything in the current `.kbp_config` file. Here's how it looks like in action: https://asciinema.org/a/e3OOpm6OU93Hq7Hd0GifFBdBm

The huge diff is because this pulls in `spf13/cobra` for CLI flags/args parsing, and `keybase/client/go/minterm` for terminal prompts, along with their dependencies.

Also cc @malgorithms How does it look from UX perspectives?


songgao:
CI error is because of missing vendor. We are using both our own fork and original version of `golang.org/x/crypto`. I'm gonna wait until https://github.com/keybase/client/pull/10095 is resolved before fixing the vendor here.

strib:
Can you make a case for adding this cobra dependency, rather than sticking with the already existing `keybase/cli` package?

Also, from your video it's not clear where the `.kbp_config` file is written to.  How do you specify the TLF?  Is it just written to `.`?

songgao:
> Also, from your video it's not clear where the .kbp_config file is written to. How do you specify the TLF? Is it just written to .?

Yeah it's by default in `./.kbp_config`, and can be changed with the `-p, --path` flag. For example, `kbpagesconfig --path /tmp/a/.kbp_config user add alice`. Although I'm realizing the `--path` flag should be able to take a directory as well, so user can do `kbpagesconfig --path /keybase/private/blah,kbpbot user add alice`.

> Can you make a case for adding this cobra dependency, rather than sticking with the already existing keybase/cli package?

`keybase/cli` is based on `urfave/cli`, and it was chosen before `cobra` was around. The latter has better support for subcommands and flags. We looked at different CLI libraries when exploring auto-completion, and `cobra` seemed to be the most feature-rich: https://keybase.atlassian.net/wiki/spaces/GEN/pages/36274200/Exploring+Auto-completion#cobra

songgao:
Also, `keybase/cli` is not vendored in this repo either, so either way we are introducing a new vendor :)

strib:
> keybase/cli is based on urfave/cli, and it was chosen before cobra was around. The latter has better support for subcommands and flags. We looked at different CLI libraries when exploring auto-completion, and cobra seemed to be the most feature-rich: https://keybase.atlassian.net/wiki/spaces/GEN/pages/36274200/Exploring+Auto-completion#cobra
>
> Also, keybase/cli is not vendored in this repo either, so either way we are introducing a new vendor :)

I'm not really concerned by the number of lines being vendored, but more about the mental vendoring of adding a new API for a maintainer to learn.  Seems like we shouldn't mix different technologies in different repos unless we have a good reason to, otherwise it becomes harder for people to jump around our codebase. So are there specific features of cobra that you're using here, that we can't support in the `cli` package?

songgao:
> So are there specific features of cobra that you're using here, that we can't support in the cli package?

I didn't try implementing this with `keybase/cli`, so I don't really know for sure. My guess is `keybase/cli` can do the same thing, probably the API is not as good though. If you prefer, I can try to switch to it. Let me know!

strib:
I'd vote for either switching to `cli`, or switching `cli` to cobra.  I guess you'll pick the former.

Note this is partly selfish since I don't want to have to learn a new API during the review, but I think it's representative of problems future code maintainers/readers will have as well.

songgao:
I'd be happy if we can switch `cli` to `cobra`. But that's probably at least 6 points of work. Will do the former then! 

It does seem like `cobra` is becoming dominant, so in benefit of community contribution or code readers, it might be a good idea to switch to it in whole at some point in the future.

maxtaco:
i second not including YACLAPL

(yet another command line arg parsing library)

songgao:
> Although I'm realizing the --path flag should be able to take a directory as well, so user can do kbpagesconfig --path /keybase/private/blah,kbpbot user add alice.

I ended up just switching the flag from file path to dir path, rather than supporting both.

Ready for review again!

.gitignore
```diff
@@ -9,6 +9,7 @@ kbfstool/kbfstool
 kbfsfuse/revision
 kbfsgit/git-remote-keybase/git-remote-keybase
 kbpagesd/kbpagesd
+kbpagesconfig/kbpagesconfig
```
strib: We should build both `kbpagesd` and `kbpagesconfig` in `appveyor.yml`, since I think it only builds/installs specific directories.

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
+package main
```
strib: Copyright notice (2018!).

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
```
strib: 2018!

kbpagesconfig/main.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
```
strib: 2018!

kbpagesconfig/user.go
```diff
@@ -0,0 +1,71 @@
+package main
```
strib: Copyright notice (2018!).

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+		if err := editor.confirmAndWrite(); err != nil {
+			fmt.Fprintf(os.Stderr, "writing new config error: %v\n", err)
+			os.Exit(1)
+		}
+	},
+}
+
+var aclSetAdditionalCmd = cli.Command{
+	Name: "set-additional",
+	Usage: "set permission(s) <username> is granted on the " +
```
strib: I'm having trouble parsing this sentence.  Maybe it should be something like: "set a given user's permissions on the given path(s), overriding the permissions for an anonymous user".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		if err := editor.confirmAndWrite(); err != nil {
+			fmt.Fprintf(os.Stderr, "writing new config error: %v\n", err)
+			os.Exit(1)
+		}
+	},
+}
+
+var aclCheckCmd = cli.Command{
+	Name:         "check",
```
strib: To me, "check" implies that it will be returning true or false.  I'd call it "get" (or maybe "list" if "list" wasn't already a permission type).

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/keybase/client/go/minterm"
+	"github.com/keybase/kbfs/libpages/config"
+	"github.com/sergi/go-diff/diffmatchpatch"
+	"golang.org/x/crypto/bcrypt"
+)
+
+const configFileName = ".kbp_config"
```
strib: Can we export the duplicate value from `libpages` and use it here?

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
...
+	if cfg, err = config.ParseConfig(io.TeeReader(from, buf)); err != nil {
+		return nil, "", err
+	}
+	if err = from.Close(); err != nil {
+		return nil, "", err
+	}
+	return cfg, buf.String(), nil
+}
+
+// initConfig reads in config file and ENV variables if set.
```
strib: Wrong function name? Also, what ENV variables?

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
...
+
+const configFileName = ".kbp_config"
+
+var term *minterm.MinTerm
+
+func init() {
+	var err error
+	if term, err = minterm.New(); err != nil {
+		fmt.Fprintf(os.Stderr, "opening terminal error: %s\n", err)
+		os.Exit(1)
```
strib: It would be nice to have some tests for this editor stuff, but this might prevent it?  Any way around it?  We can probably pass the term in from elsewhere, maybe as an interface, and a test could provide one that always says yes or no.  Then `minterm.New()` would be done by the CLI code that calls into here.

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
...
+func (e *kbpConfigEditor) removeUser(username string) {
+	delete(e.kbpConfig.Users, username)
+}
+
+func (e *kbpConfigEditor) setAnonymousPermission(
+	permsStr string, pathStr string) error {
+	if e.kbpConfig.ACLs == nil {
+		e.kbpConfig.ACLs = make(map[string]config.AccessControlV1)
+	}
+	pathACL := e.kbpConfig.ACLs[pathStr] // struct
```
strib: What does the comment mean?

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,206 @@
...
+
+func (e *kbpConfigEditor) setAdditionalPermission(
+	username string, permsStr string, pathStr string) error {
+	if e.kbpConfig.ACLs == nil {
+		e.kbpConfig.ACLs = make(map[string]config.AccessControlV1)
+	}
+	pathACL := e.kbpConfig.ACLs[pathStr] // struct
+	if pathACL.WhitelistAdditionalPermissions == nil {
+		// If permsStr is empty, we'd leave an empty permission entry behind.
+		// But that's OK. If user really wants it gone, they can use the
```
strib: It doesn't change any behavior, right?  Things are still correct, there's just an unnecessary map entry?

kbpagesconfig/user.go
```diff
@@ -0,0 +1,71 @@
...
+	ArgumentHelp: "add <username> [username ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 1 {
+			fmt.Fprintf(os.Stderr, "empty username\n")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/user.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/keybase/cli"
+)
+
+var userAddCmd = cli.Command{
+	Name:         "add",
+	Usage:        "add new user(s) to config",
+	ArgumentHelp: "add <username> [username ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 1 {
+			fmt.Fprintf(os.Stderr, "empty username\n")
```
strib: `Fprintln`?

kbpagesconfig/user.go
```diff
@@ -0,0 +1,71 @@
...
+	},
+}
+
+var userRemoveCmd = cli.Command{
+	Name:         "remove",
+	Usage:        "remove user(s) from config",
+	ArgumentHelp: "remove <username> [username ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 1 {
+			fmt.Fprintf(os.Stderr, "empty username\n")
```
strib: `Fprintln`?

kbpagesconfig/user.go
```diff
@@ -0,0 +1,71 @@
...
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+	ArgumentHelp: "set-anonymous <read|list|read,list> <path> [path ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 2 {
+			fmt.Fprintln(os.Stderr, "need at least 2 args")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+	ArgumentHelp: "clear <path> [path ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 1 {
+			fmt.Fprintln(os.Stderr, "need at least 1 arg")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+	ArgumentHelp: "set-additional <username> <read|list|read,list> <path> [path ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 3 {
+			fmt.Fprintln(os.Stderr, "need at least 3 args")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+	ArgumentHelp: "remove <username> <path> [path ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 2 {
+			fmt.Fprintln(os.Stderr, "need at least 2 args")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+	ArgumentHelp: "check <username> <path> [path ...]",
+	Action: func(c *cli.Context) {
+		if len(c.Args()) < 2 {
+			fmt.Fprintln(os.Stderr, "need at least 2 args")
+			os.Exit(1)
+		}
+		editor, err := newKBPConfigEditor(c.GlobalString("dir"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr,
+				"creating config editor error: %v", err)
```
strib: No "\n".

kbpagesconfig/acl.go
songgao: It doesn't override though; it's additional on top of anonymous ones. Would "set additional permission(s) that <username> is granted on top of anonymous ones on the given path(s) "

kbpagesconfig/editor.go
songgao: This package (intentionally) doesn't import `libpages`. But I guess we could move that to `config` package? The config itself doesn't care about what the file is called though. Let me know!

kbpagesconfig/editor.go
songgao: lol it's generated comment from `cobra`. Sorry I missed it.

kbpagesconfig/editor.go
songgao: It means pathACL is a struct which is pass-by-value. Seems like it's causing confusion more than helpful. Will remove.

kbpagesconfig/editor.go
songgao: Yes that's correct. I'll add this in the comment.

kbpagesconfig/acl.go
strib: "is granted" -> "are granted", but otherwise yes that's better thanks.

kbpagesconfig/editor.go
strib: Sure, `config` package seems reasonable even if it's not used there.

.gitignore
songgao: Good point; I'll add the tests to Jenkins as well.

libpages/response_writer_wrappers.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Write implements the http.ResponseWriter interface.
+func (w responseWriterWrapper) Write(data []byte) (int, error) {
+	if w.writeOverrides == nil || w.writtenCode == 0 {
+		return w.w.Write(data)
+	}
+	override := w.writeOverrides[w.writtenCode]
+	if override == nil {
+		return w.w.Write(data)
```
jzila: That's what I was saying in standup today, that in non-overridden cases there won't be any additional writes--it's just a passthrough.

libpages/server_test.go
```diff
@@ -0,0 +1,10 @@
+package libpages
+
+import "testing"
+
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+func TestStub(t *testing.T) {
```
jzila: ??

libpages/server_test.go
```diff
@@ -0,0 +1,10 @@
+package libpages
+
+import "testing"
+
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
```
jzila: Copyright notice should be at the top.

libpages/response_writer_wrappers.go
```diff
@@ -0,0 +1,44 @@
...
+var _ http.ResponseWriter = responseWriterWrapper{}
+
+type writeOverride interface {
+	Bytes(req *http.Request) []byte
+}
+
+// responseWriterWrapper wraps a http.ResponseWriter, allowing overriding
+// response for selected status codes. This is mainly for custom error pages.
+//
+// This type is not goroutine safe.
```
jzila: I don't see this used anywhere yet. Where/how are you planning on using it?

kbpagesconfig/acl.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		if err := editor.confirmAndWrite(); err != nil {
+			fmt.Fprintf(os.Stderr, "writing new config error: %v\n", err)
+			os.Exit(1)
+		}
+	},
+}
+
+var aclSetAdditionalCmd = cli.Command{
+	Name: "set-additional",
```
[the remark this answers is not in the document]
songgao: `set-additional` sets additional permissions that a user gets, on top of anonymous ones, so need to make it clear. I guess we could have another layer of subcommands though:

```
kbpagesconfig acl set default <read|list|read,list> <path> [path ...]
kbpagesconfig acl set additional <username> <read|list|read,list> <path> [path ...]
```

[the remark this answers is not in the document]
jzila: Yeah that sounds great.

kbpagesconfig/editor.go
```diff
@@ -0,0 +1,210 @@
...
+	if strings.ToLower(input) != "y" {
+		return fmt.Errorf("write not confirmed")
+	}
+
+	// write the new config to kbpConfigPath
+	f, err := os.Create(e.kbpConfigPath)
+	if err != nil {
+		return fmt.Errorf(
+			"opening file [%s] error: %v", e.kbpConfigPath, err)
+	}
```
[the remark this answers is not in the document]
songgao: Good catch; I still want to return the error from `f.Close()` though, so will keep the original close but add the deferred call here.

libpages/response_writer_wrappers.go
songgao: Yup; as I said in kbchat, I realized later that `WriteHeader` can only happen before `Write` is called, so my concern beforehand was a pointless :)

libpages/server_test.go
songgao: I'm planning to add test for this package later, and wanted to add all tests to CI in this PR. Then I incorrectly thought there had to be test files for `go test` to exit with code 0; just realized no test files is fine too, so will just remove this file.

libpages/response_writer_wrappers.go
songgao: This was a half done file and accidentally committed; sorry! Will remove from this branch.

PR #1418, opened 2018-01-03
libkbfs: add a way to mark nodes as read-only

strib:
This will be used by autogit in a future PR, to make everything under the .kbfs_autogit directory read-only unless it's being accessed by the autogit manager itself.

Also add a global list of root node wrappers in the Config object, that are applied automatically to every newly-created node cache.

Issue: KBFS-2676

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -105,6 +105,7 @@ type ConfigLocal struct {
 	defaultBlockType keybase1.BlockType
 	kbfsService      *KBFSService
 	kbCtx            Context
+	rootNodeWrappers []func(Node) Node
 
 	maxNameBytes  uint32
 	maxDirBytes   uint64
@@ -1458,4 +1459,20 @@ func (c *ConfigLocal) SetKBFSService(k *KBFSService) {
 		c.kbfsService.Shutdown()
 	}
 	c.kbfsService = k
+}
+
+// RootNodeWrappers implements the Config interface for ConfigLocal.
+func (c *ConfigLocal) RootNodeWrappers() []func(Node) Node {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	ret := make([]func(Node) Node, len(c.rootNodeWrappers))
+	copy(ret, c.rootNodeWrappers)
+	return ret
+}
+
+// AddRootNodeWrapper implements the Config interface for ConfigLocal.
+func (c *ConfigLocal) AddRootNodeWrapper(f func(Node) Node) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.rootNodeWrappers = append(c.rootNodeWrappers, f)
 }
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -1465,9 +1465,7 @@ func (c *ConfigLocal) SetKBFSService(k *KBFSService) {
 func (c *ConfigLocal) RootNodeWrappers() []func(Node) Node {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	ret := make([]func(Node) Node, len(c.rootNodeWrappers))
-	copy(ret, c.rootNodeWrappers)
-	return ret
+	return c.rootNodeWrappers[:]
 }
 
 // AddRootNodeWrapper implements the Config interface for ConfigLocal.
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -358,6 +358,9 @@ func newFolderBranchOps(ctx context.Context, config Config, fb FolderBranch,
 		// ensure that the user doesn't try any data reads or writes.
 	} else {
 		nodeCache = newNodeCacheStandard(fb)
+		for _, f := range config.RootNodeWrappers() {
+			nodeCache.AddRootWrapper(f)
+		}
 	}
 
 	// make logger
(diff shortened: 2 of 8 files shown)


jzila:
Please wait to merge until we've resolved the conversation on #1415

strib:
> So in git, will we have a root wrapper that returns either the node itself or a wrapped one that overrides the Readonly() method based on path of the node, or is there gonna be a way to inject a wrapper to a particular node?

Yeah my plan is for `WrapChild()` of the root node to wrap the `.kbfs_autogit` child in an instance that sets Readonly to true, and sets Readonly to true for all its children.  Though it will be overridable with a value in the `Context`.

> Please wait to merge until we've resolved the conversation on #1415

Sure, though I'll be making any changes from that discussion in a different PR, should be independent of this one.

songgao:
> Yeah my plan is for WrapChild() of the root node

By "`WrapChild()` of the root node", did you mean a root wrapper injected through `config.AddRootNodeWrapper()` will cause `libkbfs` to produce a root node that has a `WrapChild()` with that behavior, if it's the root node of the TLF? Otherwise I think the implementation has to be in `libkbfs` rather than a `git` package.

strib:
> By "WrapChild() of the root node", did you mean a root wrapper injected through config.AddRootNodeWrapper() will cause libkbfs to produce a root node that has a WrapChild() with that behavior, if it's the root node of the TLF?

Yes, sorry for being imprecise.

libkbfs/config_local.go
```diff
@@ -1459,3 +1460,19 @@ func (c *ConfigLocal) SetKBFSService(k *KBFSService) {
...
 	c.kbfsService = k
 }
+
+// RootNodeWrappers implements the Config interface for ConfigLocal.
+func (c *ConfigLocal) RootNodeWrappers() []func(Node) Node {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	ret := make([]func(Node) Node, len(c.rootNodeWrappers))
+	copy(ret, c.rootNodeWrappers)
+	return ret
```
jzila: I wonder if you can instead do `return c.rootNodeWrappers[:]`, since the underlying `func(Node) Node` elements aren't going to change.

libkbfs/config_local.go
```diff
@@ -1459,3 +1460,19 @@ func (c *ConfigLocal) SetKBFSService(k *KBFSService) {
...
+}
+
+// AddRootNodeWrapper implements the Config interface for ConfigLocal.
+func (c *ConfigLocal) AddRootNodeWrapper(f func(Node) Node) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.rootNodeWrappers = append(c.rootNodeWrappers, f)
```
jzila: Shouldn't you also update the node cache's root node wrappers?

libkbfs/folder_branch_ops.go
```diff
@@ -3449,7 +3470,11 @@ func (fbo *folderBranchOps) Rename(
 			getNodeIDStr(newParent), newName, err)
 	}()
 
-	err = fbo.checkNode(newParent)
+	err = fbo.checkNodeForWrite(ctx, oldParent)
+	if err != nil {
+		return err
+	}
+	err = fbo.checkNodeForWrite(ctx, newParent)
```
jzila: I'm wondering if you need a lock between these two.

libkbfs/kbfs_ops_test.go
```diff
@@ -3610,3 +3610,39 @@ func TestKBFSOpsBasicTeamTLF(t *testing.T) {
...
+
+func TestKBFSOpsReadonlyNodes(t *testing.T) {
+	config, _, ctx, cancel := kbfsOpsInitNoMocks(t, "test_user")
+	defer kbfsTestShutdownNoMocks(t, config, ctx, cancel)
+
+	config.AddRootNodeWrapper(func(root Node) Node {
+		return wrappedReadonlyNode{root}
+	})
+
+	// Not read-only, should work.
```
jzila: `t.Log` the comments please.

libkbfs/config_local.go
strib: Sounds good, will do.

libkbfs/kbfs_ops_test.go
strib: These ones aren't log-worthy in my opinion, I respectfully decline to log them.

libkbfs/folder_branch_ops.go
strib: I don't believe any lock is needed here.  The `Readonly()` method only takes a context as input, so any locking in FBO should be independent of what it decides.  Let me know if I'm missing something though.

jzila: Can the checks be invalidated before the lock is taken?

strib: I guess one could build a `Node` implementation that does so, but even if you did, taking a lock in FBO wouldn't be able to prevent it in all cases, only in the case where the catalysis for change was coming through FBO itself somehow.  You can imagine systems like that, but it doesn't seem worth engineering a fix for, since it wouldn't fix the bigger problem anyway.

I could add a second check behind all the locks if you feel strongly about it, it's probably fairly cheap, just a bit of extra code.

jzila: Naw, I think it's extremely low risk that this would ever come up. I just saw 2 operations on mutable data not under lock and red flags went off.

strib: Yeah good instinct.  Though in this case, a race wouldn't be just limited to this double-check, but to any check that happens before the lock (i.e., all of them).  But yeah, agreed it's not worth worrying about now.

libkbfs/config_local.go
strib: Oops, I realized I put my comment above on the wrong comment.  It belonged on the thread above about returning a copy of `c.rootNodeWrappers`.

As for updating the node cache's root node wrappers -- that's hard to do, because there's not just one node cache, there is one for each FBO.  That's why I commented the interface definition for `AddRootNodeWrapper` like this:

```go
// This will only apply to TLFs that are first accessed after `AddRootNodeWrapper` is called.
```

Seems like it's not worth supporting something more complicated that adds them after startup, but let me know if you see a need for that.

jzila: I guess with that comment it's fine.

PR #1420, opened 2018-01-04
libgit: allow auto-creation of basic autogit directory structure

strib:
This PR does a few things:
* `folderBranchOps` has a check for disallowed prefixes when creating new directory entries, currently just ".kbfs" is disallowed.  We had an override for SingleOp mode to allow `.kbfs_git` to be auto-created, but now that we need a second name to be auto-created, I've changed this to be overridable with a context value rather than using the mode.
* The `Node` interface now has a `ShouldCreateMissedLookup` function, which allows `Node` implementations to cause the auto-creation of a directory entry when a lookup misses.
* `kbfsfuse` and `kbfsdokan` now enable several layers of the basic autogit directory structure via different `Node` implementations.  For now, it autocreates up to the TLF name, e.g. `.kbfs_autogit/private/strib`.  A future PR will add repo clones/pull under that directory structure.  All nodes under `.kbfs_autogit` are read-only when accessed via regular file system calls.

Issue: KBFS-2677

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -1846,6 +1846,8 @@ func (r *runner) processCommands(ctx context.Context) (err error) {
 	reader := bufio.NewReader(r.input)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+	// Allow the creation of .kbfs_git within KBFS.
+	ctx = context.WithValue(ctx, libkbfs.CtxAllowNameKey, kbfsRepoDir)
 
 	// Process the commands with a separate queue in a separate
 	// goroutine, so we can exit as soon as EOF is received
--- a/libgit/autogit_node_wrappers.go
+++ b/libgit/autogit_node_wrappers.go
@@ -23,7 +23,7 @@ import (
 // * `readonlyNode` is always marked as read-only, unless
 //   `ctxReadWriteKey` has a non-nil value in the context.
 // * `autogitRootNode` allows the auto-creation of subdirectories
-//   representing TLF types, e.g.. .kbfs_autogit/private or
+//   representing TLF types, e.g. .kbfs_autogit/private or
 //   .kbfs_autogit/public.  It wraps child nodes two ways, as both a
 //   `readonlyNode`, and an `tlfTypeNode`.
 // * `tlfTypeNode` allows the auto-creation of subdirectories
@@ -62,7 +62,7 @@ func (ttn tlfTypeNode) ShouldCreateMissedLookup(
 	_, err := libkbfs.ParseTlfHandle(
 		ctx, ttn.am.config.KBPKI(), ttn.am.config.MDOps(), name, ttn.tlfType)
 
-	ctx = context.WithValue(ctx, ctxReadWriteKey, 1)
+	ctx = context.WithValue(ctx, ctxReadWriteKey, struct{}{})
 	switch e := errors.Cause(err).(type) {
 	case nil:
 		return true, ctx, libkbfs.Dir, ""
@@ -91,7 +91,7 @@ func (arn autogitRootNode) ShouldCreateMissedLookup(
 	bool, context.Context, libkbfs.EntryType, string) {
 	switch name {
 	case public, private, team:
-		ctx = context.WithValue(ctx, ctxReadWriteKey, 1)
+		ctx = context.WithValue(ctx, ctxReadWriteKey, struct{}{})
 		return true, ctx, libkbfs.Dir, ""
 	default:
 		return arn.Node.ShouldCreateMissedLookup(ctx, name)
@@ -112,7 +112,11 @@ func (arn autogitRootNode) WrapChild(child libkbfs.Node) libkbfs.Node {
 	default:
 		return child
 	}
-	return &tlfTypeNode{child, arn.am, tlfType}
+	return &tlfTypeNode{
+		Node:    child,
+		am:      arn.am,
+		tlfType: tlfType,
+	}
 }
 
 // readonlyNode is a read-only node by default, unless `ctxReadWriteKey`
@@ -147,7 +151,7 @@ var _ libkbfs.Node = (*rootNode)(nil)
 func (rn rootNode) ShouldCreateMissedLookup(ctx context.Context, name string) (
 	bool, context.Context, libkbfs.EntryType, string) {
 	if name == autogitRoot {
-		ctx = context.WithValue(ctx, ctxReadWriteKey, 1)
+		ctx = context.WithValue(ctx, ctxReadWriteKey, struct{}{})
 		ctx = context.WithValue(ctx, libkbfs.CtxAllowNameKey, autogitRoot)
 		return true, ctx, libkbfs.Dir, ""
 	}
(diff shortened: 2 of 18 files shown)


strib:
Needs keybase/client#10115 to pass circleci.

strib:
FYI just added a simple commit to make `readonlyNode`s wrap their own children.

libkbfs/interfaces.go
```diff
@@ -132,6 +132,18 @@ type Node interface {
 	// `inner.Readonly()` if it decides not to return `true` on its
 	// own.
 	Readonly(ctx context.Context) bool
+	// ShouldCreateMissedLookup is called for Nodes representing
+	// directories, whenever `name` is looked up but is not found in
+	// the directory.  If the Node decides a new entry should be
+	// created matching this lookup, it should return `true` as well
+	// as a context to use for the creation, the type of the new entry
+	// and the symbolic link contents is the entry is a Sym; the
```
songgao: s/is/if?

libkbfs/interfaces.go
```diff
@@ -132,6 +132,18 @@ type Node interface {
...
+	// caller should then create this entry.  Otherwise it should
+	// return false.  An implementation that wraps another `Node`
+	// (`inner`) must return `inner.ShouldCreateMissedLookup()` if it
+	// decides not to return `true` on its own.
+	ShouldCreateMissedLookup(ctx context.Context, name string) (
+		bool, context.Context, EntryType, string)
```
songgao: I assume the last returned argument `string` is the content of the symlink target. Perhaps use named return type here to make the meaning clear from method signature?

libgit/autogit_node_wrappers.go
```diff
@@ -0,0 +1,174 @@
...
+	case public:
+		tlfType = tlf.Public
+	case private:
+		tlfType = tlf.Private
+	case team:
+		tlfType = tlf.SingleTeam
+	default:
+		return child
+	}
+	return &tlfTypeNode{child, arn.am, tlfType}
```
songgao: Would you mind using named arguments here? Same for below.

kbfsgit/runner.go
```diff
@@ -1750,6 +1750,8 @@ func (r *runner) processCommands(ctx context.Context) (err error) {
 	reader := bufio.NewReader(r.input)
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
+	// Allow the creation of .kbfs_git within KBFS.
```
jzila: So the autogit checkout will be at e.g. `/keybase/private/jzila/.kbfs_autogit/team/keybase.devops/.kbfs_git/secrets.git`, where this key allows that `.kbfs_git` to be created? Why not just skip the `.kbfs_git` and go straight to the git repos, since that can be implicit?

libgit/autogit_node_wrappers.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+var _ libkbfs.Node = (*rootNode)(nil)
+
+// ShouldCreateMissedLookup implements the Node interface for
+// rootNode.
+func (rn rootNode) ShouldCreateMissedLookup(ctx context.Context, name string) (
+	bool, context.Context, libkbfs.EntryType, string) {
+	if name == autogitRoot {
+		ctx = context.WithValue(ctx, ctxReadWriteKey, 1)
```
songgao: It just needs to have the key/value in context right? Perhaps make the value type `struct{}` to avoid confusion (where one may think the value matters). Same for above where this ctx key is used.

libkbfs/folder_branch_ops.go
```diff
@@ -2487,14 +2524,26 @@ func (fbo *folderBranchOps) finalizeGCOp(ctx context.Context, gco *GCOp) (
...
-		return nil
-	}
+// CtxAllowNameKeyType is the type for a context allowable name override key.
+type CtxAllowNameKeyType int
+
+const (
+	// CtxAllowNameKey can be used to set a value in a context, and
+	// that value will be treated as an allowable directory entry
+	// name, even if it also matches a disallowed prefix.
+	CtxAllowNameKey CtxAllowNameKeyType = iota
```
songgao: Can we mention in comment that the value to this key has to be `string` typed? Otherwise it may panic.

libgit/autogit_node_wrappers.go
```diff
@@ -0,0 +1,174 @@
...
+//
+// * `rootWrapper.wrap()` is installed as a root node wrapper, and wraps
+//   the root node for each TLF in a `rootNode` instance.
+// * `rootNode` allows .kbfs_autogit to be auto-created when it is
+//   looked up, and wraps it two ways, as both a `readonlyNode`, and
+//   an `autogitRootNode`.
+// * `readonlyNode` is always marked as read-only, unless
+//   `ctxReadWriteKey` has a non-nil value in the context.
+// * `autogitRootNode` allows the auto-creation of subdirectories
+//   representing TLF types, e.g.. .kbfs_autogit/private or
```
jzila: nittiest of nits: `e.g..` should be `e.g.`

libkbfs/folder_branch_ops.go
```diff
@@ -2487,14 +2524,26 @@ func (fbo *folderBranchOps) finalizeGCOp(ctx context.Context, gco *GCOp) (
...
+)
+
+func checkDisallowedPrefixes(ctx context.Context, name string) error {
 	for _, prefix := range disallowedPrefixes {
 		if strings.HasPrefix(name, prefix) {
+			if allowedName := ctx.Value(CtxAllowNameKey); allowedName != nil {
```
songgao: Any reason why this `if` block isn't at the beginning of the function?

libgit/autogit_node_wrappers.go
```diff
@@ -0,0 +1,174 @@
...
+	bool, context.Context, libkbfs.EntryType, string) {
+	_, err := libkbfs.ParseTlfHandle(
+		ctx, ttn.am.config.KBPKI(), ttn.am.config.MDOps(), name, ttn.tlfType)
+
+	ctx = context.WithValue(ctx, ctxReadWriteKey, 1)
+	switch e := errors.Cause(err).(type) {
+	case nil:
+		return true, ctx, libkbfs.Dir, ""
+	case libkbfs.TlfNameNotCanonical:
+		return true, ctx, libkbfs.Sym, e.NameToTry
```
songgao: This would cause the creation of a symlink for the TLF if user-provided TLF name is not canonical right? Seems like it would be very common. Have you considered just disallowing accessing autogit from non-canonical paths?

kbfsgit/runner.go
strib: No, the `.kbfs_git` is not present in the autogit path.  The line here is adding permission for the kbfsgit remote helper, not for anything to do with autogit.  It's needed because I changed the way the overrides work, so it can't be implicit here with the mode anymore, like it used to be.

libgit/autogit_node_wrappers.go
strib: Yes that's right.  I thought this would be way more user-friendly, since it's just like the way /keybase works (except the symlink is permanently to a directory rather than being ephemeral).  Seems pretty useful to me.  What do you think?

libkbfs/folder_branch_ops.go
strib: Yes, I wanted to avoid doing the context lookup and string match for the common case of the name NOT matching a disallowed prefix.

libgit/autogit_node_wrappers.go
songgao: I was a bit worried if this would end up with a lot of symlinks in `.kbfs_autogit`. But on the other hand now thinking about it more, I suppose user would barely do a `ls` on `.kbfs_autogit` dir, and in GUI we can just filter out the symlinks. So I'm OK with this.

libkbfs/folder_branch_ops.go
songgao: Oh good point. Carry on!

PR #1424, opened 2018-01-06
libgit: update autogit dir when corresponding repo is updated

strib:
This depends on #1423 and that one should be reviewed before this one.

Issue: KBFS-2678

### DIFF
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -218,11 +218,26 @@ func (f *File) Unlock() (err error) {
 		return f.fs.config.MDServer().ReleaseLock(f.fs.ctx,
 			f.fs.root.GetFolderBranch().Tlf, f.getLockID())
 	}
-	return jServer.FinishSingleOp(f.fs.ctx,
+	err = jServer.FinishSingleOp(f.fs.ctx,
 		f.fs.root.GetFolderBranch().Tlf, &keybase1.LockContext{
 			RequireLockID:       f.getLockID(),
 			ReleaseAfterSuccess: true,
 		}, f.fs.priority)
+	if err != nil {
+		return err
+	}
+
+	if f.fs.config.Mode() != libkbfs.InitSingleOp {
+		f.fs.log.CDebugf(f.fs.ctx, "Releasing the lock")
+
+		// Need to explicitly release the lock from the server.
+		err = f.fs.config.MDServer().ReleaseLock(
+			f.fs.ctx, f.fs.root.GetFolderBranch().Tlf, f.getLockID())
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // Truncate implements the billy.File interface for File.
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -229,3 +229,8 @@ func (f *File) Unlock() (err error) {
 func (f *File) Truncate(size int64) error {
 	return f.fs.config.KBFSOps().Truncate(f.fs.ctx, f.node, uint64(size))
 }
+
+// GetNode returns the libkbfs.Node associated with this file.
+func (f *File) GetNode() libkbfs.Node {
+	return f.node
+}
--- a/libfs/file.go
+++ b/libfs/file.go
@@ -230,7 +230,14 @@ func (f *File) Unlock() (err error) {
 	if f.fs.config.Mode() != libkbfs.InitSingleOp {
 		f.fs.log.CDebugf(f.fs.ctx, "Releasing the lock")
 
-		// Need to explicitly release the lock from the server.
+		// Need to explicitly release the lock from the server. If
+		// single-op mode isn't enabled, then the journal will be
+		// flushing on its own without waiting for the call to
+		// `FinishSingleOp`. That means the journal can already be
+		// completely flushed by the time `FinishSingleOp` is called,
+		// and it will be a no-op. It won't have made any call to the
+		// server to release the lock, so we have to do it explicitly
+		// here.
 		err = f.fs.config.MDServer().ReleaseLock(
 			f.fs.ctx, f.fs.root.GetFolderBranch().Tlf, f.getLockID())
 		if err != nil {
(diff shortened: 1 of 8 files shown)


libgit/autogit_node_wrappers.go
```diff
@@ -82,12 +86,19 @@ var _ libkbfs.Node = (*repoNode)(nil)
...
 		Node:     n,
 		am:       am,
 		h:        h,
 		repoName: repoName,
 	}
+	// We can't rely on a particular repo node being passed back into
+	// libkbfs by callers, since they may not keep a reference to it
+	// after looking it up, and `WrapChild` makes a new `repoNode` for
+	// each call to it, even for the same underlying NodeID.  So we
+	// keep the populated state in the AutogitManager.
```
jzila: What about if the user's client hasn't yet created a `folderBranchOps` for that TLF?

libgit/autogit_node_wrappers.go
```diff
@@ -165,6 +198,25 @@ func (rn *repoNode) finishPopulate(populated bool) {
...
+	if err != nil {
+		rn.am.log.CDebugf(ctx, "Error getting handle: %+v", err)
+		return
+	}
+
+	dstDir := rn.dstDir()
+	rn.am.log.CDebugf(
+		ctx, "Repo %s/%s/%s updated", h.GetCanonicalPath(), dstDir, rn.repoName)
+	_, err = rn.am.Pull(
+		context.Background(), rn.h, rn.repoName, "master", h, dstDir)
```
jzila: We should make custom contexts for this with tags.

libfs/file.go
```diff
@@ -218,14 +218,34 @@ func (f *File) Unlock() (err error) {
...
 			ReleaseAfterSuccess: true,
 		}, f.fs.priority)
+	if err != nil {
+		return err
+	}
+
+	if f.fs.config.Mode() != libkbfs.InitSingleOp {
+		f.fs.log.CDebugf(f.fs.ctx, "Releasing the lock")
+
+		// Need to explicitly release the lock from the server.
```
jzila: why?

libgit/autogit_manager.go
```diff
@@ -68,10 +69,18 @@ type AutogitManager struct {
 	resetQueue     channels.Channel
 	queueDoneCh    chan struct{}
 	getNewConfig   getNewConfigFn
+	resetsWG       kbfssync.RepeatedWaitGroup
+	updatingWG     kbfssync.RepeatedWaitGroup
```
jzila: Is this only used for testing right now? I don't see a `{resets,updating}WG.Wait` anywhere in non-test code.

libgit/autogit_manager.go
```diff
@@ -409,6 +423,75 @@ func (am *AutogitManager) Pull(
...
+	ctx context.Context, node libkbfs.Node) {
+	rn, ok := am.repoNodesForWatchedIDs[node.GetID()]
+	if !ok {
+		return
+	}
+
+	am.updatingWG.Add(1)
+	go func() {
+		defer am.updatingWG.Done()
+		rn.updated(context.Background())
```
jzila: This should have debug tags.

libgit/autogit_node_wrappers.go
jzila: Actually, you pass in a `context.Background()` in the one place where it's called, so maybe the tags can come from there, and this `Pull` uses `ctx`.

libfs/file.go
strib: Because if single-op mode isn't enabled, then the journal will be flushing on its own without waiting for the call to `FinishSingleOp`.  That means the journal can already be completely flushed by the time `FinishSingleOp` is called, and it will be a no-op.  It won't have made any call to the server to release the lock, so we have to do it explicitly here.

Perhaps there's a way to structure `FinishSingleOp` so that it can know whether the lock was released as part of the journal flush or not, and it can release the lock explicitly if so.  I don't think that will be trivial, but I can look into it if you all want.  Or I can just expand the comment here.  What do you think?

libgit/autogit_manager.go
strib: Correct.  They might be useful in `Shutdown` too but I haven't really thought it through yet.

libfs/file.go
jzila: I'm fine with an expanded comment (which can just be your first paragraph here).

libgit/autogit_node_wrappers.go
strib: Which TLF are you asking about, the source git repo TLF or the destination checkout TLF?  The `repoNode` here is from the destination TLF, and will only exist if the callbacks have been triggered by the corresponding `folderBranchOps` instance.

If you're talking about the source repo TLF, that's orthogonal to this `isRepoNodePopulated` call, but in general it will get created by the autogit code as soon as it tries to read the repo to service the `Reset` call. No user action is needed.

libgit/autogit_node_wrappers.go
strib: `updated()` is called from a goroutine though, so I didn't want to pass that same `ctx` into a new call (since it might be canceled, etc.).  Maybe it's better to make a new ctx with new tags at the caller and pass it / use it here.  That seems cleaner, I'll do that.

strib: (Ah, now I see that's probably what you were suggested with the very next comment.)

libgit/autogit_node_wrappers.go
jzila: I'm just trying to understand the flow of what happens to an auto-checkout in dstTlf when srcTlf/fooRepo receives a push. And which client does the work.

strib: When a client instance first reads the autogit dir in the dest, it necessarily ends up reading from the source repo TLF.  When it does that, it registers for updates with that TLF's `folderBranchOps`, and handles the updates subsequently.  (If multiple client instances have read it, the reset will be done by whoever first takes the server-side lock.)

The client that wrote to the repo doesn't do anything in terms of updating the autogit repo.

jzila: OK so if no clients have a `folderBranchOps` on the src (and thus have registered for updates), the auto-checkout doesn't get updated on a push. So then when the next read attempt happens, all that scaffolding gets set up, the client does a `Pull`, and also registers for updates.

Does that sound right?

strib: Yep, exactly.

PR #1427, opened 2018-01-09
git root for keybase pages

songgao:


libpages/dns.go
```diff
@@ -63,14 +63,14 @@ const kbpRecordPrefix = "_keybase_pages."
 // _keybase_pages.meatball.gao.io   TXT "kbp=/keybase/public/songgao/meatball/"
 // _keybase_pages.song.gao.io       TXT "kbp=/keybase/private/songgao,kb_bot/blah"
 // _keybase_pages.blah.strib.io     TXT "kbp=/keybase/private/strib#kb_bot/blahblahb" "lah/blah/"
-// _keybase_pages.kbp.jzila.com     TXT "kbp=git-keybase://private/jzila,kb_bot/kbp.git"
+// _keybase_pages.kbp.jzila.com     TXT "kbp=git@keybase:private/jzila,kb_bot/kbp.git"
```
jzila: I don't love this syntax; it doesn't match anything we're currently doing and it's not obvious.
Why isn't this `/keybase/private/jzila,kb_bot/.kbfs_autogit/public/jzila/kbp.git`?

libpages/server.go
```diff
@@ -143,21 +150,79 @@ func (s *Server) isDirWithNoIndexHTML(
...
+}
+
+// TODO: replace this with something nicer when fancy error pages and landing
+// pages are ready.
+var cloningLandingPage = []byte(`
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<meta http-equiv="refresh" content="5">
```
jzila: Maybe give it a `<title>CLONING</title>` too.

libpages/server.go
```diff
@@ -143,21 +150,79 @@ func (s *Server) isDirWithNoIndexHTML(
...
+const gitRootInitialTimeout = time.Second
+
+func (s *Server) shouldShowCloningLandingPage(st *site) (bool, error) {
+	if st.root.Type != GitRoot {
+		// CLONING file only matters for git roots.
+		return false, nil
+	}
+	newCtx, err := libkbfs.NewContextWithCancellationDelayer(
+		libkbfs.CtxWithRandomIDReplayable(context.Background(),
+			ctxIDKey, ctxOpID, nil))
```
jzila: This will leak a goroutine. `libkbfs.NewContextWithCancellationDelayer` must take a cancellable context, which itself is at some point canceled. I'd suggest you make the below timeout context first (and `defer` its `cancel`), and then pass that into the `CtxWithRandomIDReplayable` above.

libpages/dns.go
songgao: The checkout is now in `kb_bot` rather than `jzila,kb_bot`, but we can switch to the latter if desired I guess. One advantage for the `git@keybase` syntax though is it doesn't require user to remember the autogit path, and is shorter to type.

@strib any thoughts?

strib: Yeah I don't think we should expose the autogit path at this interface.  I suggested the `git@` syntax because it matches git remote syntax, and gives us the option in the future of adding other types of protocols.  I think we should hide any autogit implementation details away from the DNS TXT records.

jzila: Then why is the `@keybase` necessary? Can it just be `kbp=git:private/jzila,kbp_bot/foo.git`?

libpages/server.go
songgao: Good catch; I'm realizing this is an issue in MakeFS as well. Will fix that too. Thanks

libpages/dns.go
strib: It's not necessary, it just matches well-known syntax like git's.  I dunno, it just seems like it plays more nicely -- maybe other providers would pop up that serve keybase pages and use the TXT field in the same way.

songgao: I like `git@keybase:` slightly better over `git:` too, so will go with that for now. John let me know if you have a strong preference/reason for `git:`.

jzila: Nope, I've been democratically overruled and defer to the will of The People.

libpages/server.go
```diff
@@ -143,21 +164,81 @@ func (s *Server) isDirWithNoIndexHTML(
...
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="UTF-8">
+		<meta http-equiv="refresh" content="5">
+		<title>CLONING</title>
+	</head>
+	<body>
+		Keybase Pages server is cloning your site.
+		This page will refresh in 5 second ...
```
strib: second -> seconds, also no space before "..."

libpages/server.go
```diff
@@ -56,31 +58,50 @@ func (s *Server) getSite(ctx context.Context, root Root) (st *site, err error) {
 			return s, nil
 		}
 		s.config.Logger.Error("nasty entry in s.siteCache",
-			zap.String("type", reflect.TypeOf(siteCached).String()))
+			zap.String("reflect_type", reflect.TypeOf(siteCached).String()))
 	}
-	fs, err := root.MakeFS(ctx, s.config.Logger, s.kbfsConfig)
+	fs, fsShutdown, err := root.MakeFS(ctx, s.config.Logger, s.kbfsConfig)
```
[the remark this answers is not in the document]
songgao: You mean the line immediately below? The `cancel` for the context made in `MakeFS` is never returned with a non-nil error, i.e. `fsShutdown` is always `nil` if `err != nil`. If a non-nil error is returned from `MakeFS`, `cancel` would have been called before returning to here. So it's called in all routines.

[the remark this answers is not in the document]
strib: Ah right, good point.  Panics (and then panic recovery, leading to a goroutine) are still possible in the final few lines of the function, but maybe that's overkill.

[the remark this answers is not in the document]
songgao: Oh I see; good point. I guess we could have a bool variable indicating whether `st` has been added to `siteCache`, and call `shutdown` if it hasn't.

PR #1428, opened 2018-01-09
kbp: add interface for stats reporting and a stathat reporter

songgao:


libpages/server.go
```diff
@@ -33,20 +33,38 @@ type ServerConfig struct {
 	UseStaging       bool
 	Logger           *zap.Logger
 	UseDiskCertCache bool
+	StatsReporter    StatsReporter
+
+	whiteList     map[string]bool
+	whiteListOnce sync.Once
```
jzila: please place this before what it's guarding (`whitelist`)

libpages/server.go
```diff
@@ -33,20 +33,38 @@ type ServerConfig struct {
 	UseStaging       bool
 	Logger           *zap.Logger
 	UseDiskCertCache bool
+	StatsReporter    StatsReporter
+
+	whiteList     map[string]bool
```
jzila: `DomainWhitelist` but `whiteList`: I think both can be lowercase `l`.

libpages/server.go
```diff
@@ -33,20 +33,38 @@ type ServerConfig struct {
...
+	whiteListOnce sync.Once
+}
+
+func (c *ServerConfig) allowedByDomainWhiteList(domain string) bool {
+	c.whiteListOnce.Do(func() {
+		if len(c.DomainWhitelist) > 0 {
+			c.whiteList = make(map[string]bool, len(c.DomainWhitelist))
```
jzila: Now that `whitelist` is in this config, why is this separation from `DomainWhitelist` necessary?

songgao: The DomainWhitelist is a slice, while whitelist is a map.

kbpagesd/stathat.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
```
strib: 2018!

libpages/stats.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2017 Keybase Inc. All rights reserved.
```
strib: 2018!

libpages/server.go
```diff
@@ -145,6 +167,18 @@ func (s *Server) isDirWithNoIndexHTML(
 
 // ServeHTTP implements the http.Handler interface.
 func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var statInfo *ServedRequestInfoForStats
+	statInfo = &ServedRequestInfoForStats{
+		Proto: r.Proto,
+		Host:  r.Host,
+		// If we don't end up responding with 200, this field will be
+		// overwritten when writting the header.
```
strib: writting -> writing

songgao:
Chris mentioned yesterday we should have a stat for # of active sites. So I added an in-memory tracking for that. Without a centralized DB or caching layer it won't work for multi-instances but this should be good enough for now. Let me know if any of you feel differently!

Also would be nice if at least one of you can take a look at the added activity-based stats. Thanks!

kbpagesd/stathat.go
```diff
@@ -25,29 +31,90 @@ type stathatReporter struct {
...
+	statNameHourlyActiveTlfs  string
 }
 
 var _ libpages.StatsReporter = (*stathatReporter)(nil)
 
 const stathatReportInterval = time.Second * 10
+const activeChSize = 1024
+
+// NOTE that this won't work for a multi-instance deployment. But before we
+// start to horizontal scale, this should be good enough.
```
strib: horizontal -> horizontally

kbpagesd/stathat.go
```diff
@@ -25,29 +31,90 @@ type stathatReporter struct {
...
+	statNameHourlyActiveHosts string
+	statNameHourlyActiveTlfs  string
 }
 
 var _ libpages.StatsReporter = (*stathatReporter)(nil)
 
 const stathatReportInterval = time.Second * 10
+const activeChSize = 1024
+
+// NOTE that this won't work for a multi-instance deployment. But before we
```
strib: Also, it doesn't work after restarts.

songgao:
I was gonna keep this until we need to scale horizontally. But good point that deployments and crashes would break this badly.

I was hesitating to have a DB for kbp, since it makes community deployment harder. But it seems some sort of central persistent storage is inevitable. Perhaps we could store this in KBFS as well?

kbpagesd/stathat.go
```diff
@@ -0,0 +1,145 @@
...
+		case <-hourlyTicker.C:
+			s.reporter.PostEZCount(
+				s.statNameHourlyActiveHosts, s.ezKey, len(s.hourlyActiveHosts))
+			s.reporter.PostEZCount(
+				s.statNameHourlyActiveTlfs, s.ezKey, len(s.hourlyActiveTlfs))
+			s.hourlyActiveHosts = make(map[string]struct{})
+			s.hourlyActiveTlfs = make(map[tlf.ID]struct{})
+
+		case <-dailyTicker.C:
+			s.reporter.PostEZCount(
```
songgao: Oh and I'm realizing these are supposed to be values rather than counts. 😓

songgao:
Actually I think storing in KBFS is a good idea. We can have a folder for TLF IDs and a folder for host names, with files inside with file names of TLF IDs and host names respectively. On each request, we’d just update mtime for corresponding files. IIRC mtime won’t cause conflicts, so we get horizontal scalability here for free. When posting stats we can just do a GetDirChildren call and go through all file entries.

How does that sound? @strib @jzila

strib:
Sounds pretty unscalable in terms of number of TLFs/hosts, but if you promise to replace it with a proper DB by the time we get to ~1000 entries or so, it's fine with me.

songgao:
> Sounds pretty unscalable in terms of number of TLFs/hosts

Is this because each update to mtime requires updating the parent directory node? What if we have two or three levels of dirs and divide TLF ID by prefixes, and host name by hash prefixes?

strib:
I'm not sure if splaying makes it better or worse actually, because then you have more blocks to fetch whenever you're iterating.  Just in general file systems make very bad DBs I think.  I'm worried about both the block sizes on puts and the work it takes to iterate.  Also, mtime updates do cause conflicts, but that's probably not a huge deal because you can just treat it as first-writer-wins in this case and ignore/delete the conflict copies automatically.  You can probably even just use symlinks that go nowhere, rather than real files, to avoid making any file blocks.

But I still think it should eventually be replaced by a real DB, which can be queried more easily than a file system, for richer stats and more personalized info for users on a future dashboard or something.

songgao:
Hmm OK I think I'm convinced that we should still have a proper DB for this. Thanks for the thoughts!

jzila:
Sorry I meant to comment on this earlier. +1 to strib's points: we don't want to take another dependency on KBFS if we can avoid it, and we want something that has much better scaling guarantees than we can currently provide for a single TLF.

songgao:
I think I'll have a active-stats storer and implement a mtime-based one (using local FS rather than KBFS) for a starter since it's simpler, and we can switch to a proper DB when we reach 500 sites.

songgao:
@strib @jzila Sorry for getting so much stuff into the same PR, but please take a look at the last commit when you get a chance; thanks!!!

kbpagesd/main.go
```diff
@@ -1,13 +1,12 @@
 // Copyright 2017 Keybase Inc. All rights reserved.
-// Use of this source code is governed by a BSD
-// license that can be found in the LICENSE file.
-
+// Use of this source code is governed by a BSD license that can be found in the LICENSE file.
```
strib: Intended?

libpages/stathat.go
```diff
@@ -0,0 +1,141 @@
...
+	if len(s.activityStats.Durations) == 0 || s.activityStats.Interval == 0 {
+		return
+	}
+
+	durations := make([]time.Duration, len(s.activityStats.Durations))
+	copy(durations, s.activityStats.Durations)
+	statNamesHosts := make([]string, 0, len(durations))
+	statNamesTlfs := make([]string, 0, len(durations))
+	for _, d := range durations {
+		statNamesHosts = append(statNamesHosts, s.statPrefixActiveHosts+"("+d.String()+")")
```
strib: The string for `time.Hour * 24 * 7` is "168h0m0s".  We might want to pass in a map with a more friendly label.

kbpagesd/main.go
```diff
@@ -61,6 +60,8 @@ func newLogger(isCLI bool) (*zap.Logger, error) {
 }
 
 const autoGitNumWorkers = 10
+const activityStatsReportInterval = time.Minute
```
strib: I think 5 min is probably fine.

libpages/stats.go
```diff
@@ -4,8 +4,55 @@
...
+	"os"
+	"path/filepath"
+	"sort"
+	"time"
+
+	"github.com/keybase/kbfs/tlf"
+	"go.uber.org/zap"
+)
+
+// ActivesGetter holds necessary data to generate number of active Tlfs or
```
strib: In comments let's use "TLFs" instead of "Tlfs".

kbpagesd/main.go
songgao: Hmm how did this  happen .... Thanks

PR #1429, opened 2018-01-09
libgit: use node ID updates to figure out when a repo was updated

strib:
This PR lets TLF observers see the complete set of nodes that have been updated.  Previously, the observer could only see nodes that had operations done directly on them (like a write, create, etc).  Now the observer can also see nodes that had their underlying blocks updated, even if that didn't result in user-visible changes (e.g., `mkdir a/b/c` affects the underlying data of node `a`).

Then we use that in autogit to watch the entire top directory for a repo to see when something changed, rather than watching just the branch reference node.  That's because the branch ref could be in `packed-refs`, and moreover it could jump back and forth for the lifetime of the repo.

The downside is now we will check for updates whenever _anything_ changes in the git repo directory, which will result in some unneeded reset requests.  But we can optimize those pretty well in future PRs I think.

Issue: KBFS-2678

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -1697,9 +1697,14 @@ type Observer interface {
 	// updated locally, but not yet saved at the server.
 	LocalChange(ctx context.Context, node Node, write WriteRange)
 	// BatchChanges announces that the nodes have all been updated
-	// together atomically.  Each NodeChange in changes affects the
-	// same top-level folder and branch.
-	BatchChanges(ctx context.Context, changes []NodeChange)
+	// together atomically.  Each NodeChange in `changes` affects the
+	// same top-level folder and branch. `allAffectedNodeIDs` is a
+	// list of all the nodes that had their underlying data changed,
+	// even if it wasn't an user-visible change (e.g., if a
+	// subdirectory was updated, the directory block for the TLF root
+	// is updated but that wouldn't be visible to a user).
+	BatchChanges(ctx context.Context, changes []NodeChange,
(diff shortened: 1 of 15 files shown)


libkbfs/interfaces.go
```diff
@@ -1697,9 +1697,14 @@ type Observer interface {
...
 	// BatchChanges announces that the nodes have all been updated
-	// together atomically.  Each NodeChange in changes affects the
-	// same top-level folder and branch.
-	BatchChanges(ctx context.Context, changes []NodeChange)
+	// together atomically.  Each NodeChange in `changes` affects the
+	// same top-level folder and branch. `allAffectedNodeIDs` is a
+	// list of all the nodes that had their underlying data changed,
+	// even if it wasn't an user-visible change (e.g., if a
+	// subdirectory was updated, the directory block for the TLF root
+	// is updated but that wouldn't be visible to a user).
```
jzila: What happens if nodes throughout a hierarchy are modified (as they would be)? It looks like we're going to `Reset` once for each `NodeID` affected.

strib: We're only registering the Node corresponding to the top directory of the repo (i.e., `.kbfs_git/reponame`), not any of the child nodes.  git updates happen in single-op batches, so we'll only get one `BatchChanges` callback per update where that node's ID has changed.  Any changes on nodes below should also be included in the same batch, but aren't registered with `AutogitManager`, so they won't cause a `Reset`.

jzila: Got it, thanks.

PR #1430, opened 2018-01-10
libfs: use `Node.Readonly()` to determine FS permissions

strib:
If a node is read-only, it shouldn't look writable to the OS.

Issue: KBFS-2678

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -463,7 +463,7 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 	return &FileInfo{
 		fs:   fs,
 		ei:   ei,
-		name: n.GetBasename(),
+		node: n,
 	}, nil
 }
 
@@ -547,10 +547,15 @@ func (fs *FS) readDir(n libkbfs.Node) (fis []os.FileInfo, err error) {
 
 	fis = make([]os.FileInfo, 0, len(children))
 	for name, ei := range children {
+		child, _, err := fs.config.KBFSOps().Lookup(fs.ctx, n, name)
+		if err != nil {
+			return nil, err
+		}
+
 		fis = append(fis, &FileInfo{
 			fs:   fs,
 			ei:   ei,
-			name: name,
+			node: child,
 		})
 	}
 	return fis, nil
@@ -594,7 +599,7 @@ func (fs *FS) Lstat(filename string) (fi os.FileInfo, err error) {
 		return nil, err
 	}
 
-	_, ei, err := fs.config.KBFSOps().Lookup(fs.ctx, n, base)
+	n, ei, err := fs.config.KBFSOps().Lookup(fs.ctx, n, base)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


libfuse/symlink.go
```diff
@@ -41,8 +41,8 @@ func (s *Symlink) Attr(ctx context.Context, a *fuse.Attr) (err error) {
 		return err
 	}
 
-	s.parent.folder.fillAttrWithUIDAndWritePerm(ctx, &de, a)
-	a.Mode = os.ModeSymlink | 0777
+	s.parent.folder.fillAttrWithUIDAndWritePerm(ctx, s.parent.node, &de, a)
+	a.Mode = os.ModeSymlink | a.Mode | 0400
```
songgao: Did you intend to clear the execution bit as well?

strib: Oh, I just forgot that they were supposed to be executable.  I'll fix it, thanks.

libfs/fs.go
```diff
@@ -547,10 +547,15 @@ func (fs *FS) readDir(n libkbfs.Node) (fis []os.FileInfo, err error) {
 
 	fis = make([]os.FileInfo, 0, len(children))
 	for name, ei := range children {
+		child, _, err := fs.config.KBFSOps().Lookup(fs.ctx, n, name)
```
jzila: This means that we're now explicitly doing an on-demand lookup for any `readDir`. It'll deepen the prefetch tree, which may or may not be what we want. I'm not necessarily expressing an opinion on whether this is correct, I just wanted to make sure it was considered.

strib: I did consider that and frowned as I did so.  But I don't see a better way to get the proper permissions for all the children.  Basically, the billy interface diverges a bit from the normal FUSE interface in that it asks for the permissions up-front on a readdir, wherease FUSE does not.  But most of the time, the OS will end up doing a stat on all the children anyway, so I think in the end it works out to be the same thing.

It does mean some extra prefetching for some go-git use cases that might not have happened before.  I don't think `ReadDir` is used extensively, so I think it's ok.

jzila: You could pass the request priority into the `Lookup`, to prevent it from triggering prefetches.

strib: Yeah, that's an option if it becomes an issue.  I don't love exposing prefetching options outside of the libkbfs package, so I vote for tackling this later if it sucks.

jzila: Sounds good to me!

jzila:
One more question: should all nodes for TLFs where you are a reader but not a writer be wrapped with `readonlyNode`? It'd be a good way to automatically implement that.

strib:
> One more question: should all nodes for TLFs where you are a reader but not a writer be wrapped with readonlyNode? It'd be a good way to automatically implement that.

That's a good idea.  We'd probably want to refactor a bit to move `readonlyNode` into `libkbfs` and export it. It also gets a little tricky because currently the root node wrap method doesn't take a context (and doesn't have one available where it's called inside the node cache), and figuring out whether someone's a writer might involve server communication (to look up team membership, for example). So some thing might need to be re-worked, so I'm inclined to delay that to a new ticket.  (Might make a good starter ticket if someone new starts on the project!)

PR #1432, opened 2018-01-10
tlf: Add DeepEqual verifying method.

jzila:
I need this to be able to check whether two handles are properly equal.

### DIFF
--- a/tlf/handle.go
+++ b/tlf/handle.go
@@ -12,9 +12,11 @@ import (
 	"github.com/keybase/client/go/protocol/keybase1"
 )
 
-// Handle uniquely identifies top-level folders by readers and
-// writers.
+// Handle uniquely identified top-level folders by readers and writers.
 //
+// NOTE: if you change this type, ensure you update the `NumField` check in
+// `MakeHandle`, and that the new fields are correctly checked for equality in
+// `DeepEqual`.
 // TODO: Have separate types for writers vs. readers.
 type Handle struct {
 	Writers           []keybase1.UserOrTeamID    `codec:"w,omitempty"`
@@ -23,6 +25,9 @@ type Handle struct {
 	UnresolvedReaders []keybase1.SocialAssertion `codec:"ur,omitempty"`
 	ConflictInfo      *HandleExtension           `codec:"ci,omitempty"`
 	FinalizedInfo     *HandleExtension           `codec:"fi,omitempty"`
+
+	// caching field to track whether we've sorted the slices.
+	sorted bool `codec:"-"`
 }
 
 // errNoWriters is the error returned by MakeHandle if it is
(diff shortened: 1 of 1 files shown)


songgao:
Is there a way to avoid using `reflect` here? We know all the fields and it seems we can just compare one-by-one.

jzila:
Sigh, okaaaay. I was trying to avoid writing that much boilerplate.

songgao:
Understood; sorry! I guess if we do this we should also add comment in the type definition so we don't forget updating the `DeepEqual` function if the type ever gets updated.

jzila:
@songgao: good call. We have to do that anyway because of the sorts. And I could actually cache the sorting.

strib:
Could add a static check like in `libkbfs.TlfHandle`: https://github.com/keybase/kbfs/blob/57d1e1d69b19b651ef466dd8761b1389bd51485e/libkbfs/tlf_handle.go#L271-L278

jzila:
OK ready with feedback implemented.

tlf/handle.go
```diff
@@ -77,6 +83,12 @@ func MakeHandle(
 	writers, readers []keybase1.UserOrTeamID,
 	unresolvedWriters, unresolvedReaders []keybase1.SocialAssertion,
 	extensions []HandleExtension) (Handle, error) {
+	if reflect.ValueOf(Handle{}).NumField() != 7 {
```
strib: Can we do this in an `init()` instead, so it doesn't happen on every call?

tlf/handle.go
```diff
@@ -326,3 +338,71 @@ func (h Handle) IsFinal() bool {
...
+	if len(h.UnresolvedReaders) != len(other.UnresolvedReaders) {
+		return false
+	}
+
+	if !h.sorted {
+		sort.Sort(UIDList(h.Writers))
+		sort.Sort(UIDList(h.Readers))
+		sort.Sort(SocialAssertionList(h.UnresolvedWriters))
+		sort.Sort(SocialAssertionList(h.UnresolvedReaders))
+		h.sorted = true
```
strib: This doesn't have a pointer receiver, so isn't setting this caching field pretty useless?

tlf/handle.go
```diff
@@ -326,3 +338,71 @@ func (h Handle) IsFinal() bool {
...
+	}
+	if !other.sorted {
+		sort.Sort(UIDList(other.Writers))
+		sort.Sort(UIDList(other.Readers))
+		sort.Sort(SocialAssertionList(other.UnresolvedWriters))
+		sort.Sort(SocialAssertionList(other.UnresolvedReaders))
+		other.sorted = true
```
strib: Same here.

tlf/handle.go
jzila: Excellent point. Will fix.

tlf/handle.go
```diff
@@ -326,3 +345,72 @@ func (h Handle) IsFinal() bool {
...
+		sort.Sort(UIDList(h.Readers))
+		sort.Sort(SocialAssertionList(h.UnresolvedWriters))
+		sort.Sort(SocialAssertionList(h.UnresolvedReaders))
+		h.sorted = true
+	}
+	if !other.sorted {
+		sort.Sort(UIDList(other.Writers))
+		sort.Sort(UIDList(other.Readers))
+		sort.Sort(SocialAssertionList(other.UnresolvedWriters))
+		sort.Sort(SocialAssertionList(other.UnresolvedReaders))
```
songgao: Is it intended not to set `sorted` for `other`?

jzila: Yeah, I want to be able to accept a non pointer comparer. Also, it seems wrong to mutate the comparison target.

PR #1433, opened 2018-01-10
folder_branch_ops: use `constIDGetter` for `ResolvesTo()`

strib:
If we use `MDOps()` on a folder that's transitioning to be a conflict folder, we could end up looking up a now fully-resolved handle that matches the handle of the winning, non-conflict folder, and getting back the wrong TLF ID.

Also use one in `tlfJournal` because it's safer and faster, and why not?

Issue: KBFS-2708


strib:
This passed the kbfs-server tests ~6 times in a row, so I think it fixes the `TestSBSConflicts` problem.  Still no idea why this just started happening, since the code hasn't changed for weeks; I guess some subtle timing race started becoming more common.

cc: @mmaxim

PR #1437, opened 2018-01-10
libgit: prevent goroutine leaks by canceling contexts

strib:
Pointed out by jzila on #1427.

### DIFF
--- a/libgit/rpc.go
+++ b/libgit/rpc.go
@@ -136,6 +136,8 @@ func (rh *RPCHandler) CreateRepo(
 		rh.log.CDebugf(ctx, "Done creating repo: %+v", err)
 	}()
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	ctx, gitConfig, tlfHandle, tempDir, err := rh.getHandleAndConfig(
 		ctx, arg.Folder)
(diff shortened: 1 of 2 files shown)


libgit/rpc.go
```diff
@@ -170,8 +172,10 @@ func (rh *RPCHandler) scheduleCleaning(folder keybase1.Folder) {
 	time.AfterFunc(minDeletedAgeForCleaning+1*time.Second, func() {
 		log := rh.config.MakeLogger("")
 
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
 		ctx, gitConfig, tlfHandle, tempDir, err := rh.getHandleAndConfig(
-			context.Background(), folder)
+			ctx, folder)
```
jzila: It's a better pattern to use the `context.WithCancel`, so we should switch anyway, but I don't see the leak here.

strib: Eventually `getHandleAndConfig` calls `libgit.Init()`, which calls `libkbfs.NewContextWithCancellationDelayer()`.

jzila: Fair enough. I guess the contract of any ctx is that it will eventually be canceled, so we should just always be doing this.

PR #1442, opened 2018-01-12
libgit: don't lock during the whole reset, just make a file

strib:
Resets can take way longer than the server lock expiration time.  So just lock while creating a "working" file, and set the mtime on that file to our best estimate of a common time derived based on our time offset with the mdserver we're connected to.

Then resets can take up to an hour, and when they're done the worker can delete the file and write any errors into a new "lasterr" file.

### DIFF
--- a/libgit/autogit_manager.go
+++ b/libgit/autogit_manager.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
 
 	"github.com/eapache/channels"
 	"github.com/keybase/client/go/logger"
@@ -40,6 +41,8 @@ const (
 
 	// Debug tag ID for an individual autogit operation
 	ctxOpID = "AGM"
+
+	workTimeLimit = 1 * time.Hour
 )
 
 type ctxTagKey int
@@ -50,6 +53,14 @@ const (
 
 func autogitLockName(srcRepo string) string {
 	return fmt.Sprintf(".autogit_%s.lock", srcRepo)
+}
+
+func autogitWorkingName(srcRepo string) string {
+	return fmt.Sprintf(".autogit_%s.working", srcRepo)
+}
+
+func autogitLastErrName(srcRepo string) string {
+	return fmt.Sprintf(".autogit_%s.lasterr", srcRepo)
 }
 
 type getNewConfigFn func(context.Context) (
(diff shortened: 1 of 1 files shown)


libgit/autogit_manager.go
```diff
@@ -119,6 +130,144 @@ func (am *AutogitManager) getNewConfigDefault(ctx context.Context) (
...
+	lockFile, err := dstFS.Create(autogitLockName(repo))
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		// Because we took the lock, this Close will sync/flush the
+		// whole journal.
+		closeErr := lockFile.Close()
+		if err == nil {
+			err = closeErr
```
jzila: log the error otherwise please

libgit/autogit_manager.go
```diff
@@ -119,6 +130,144 @@ func (am *AutogitManager) getNewConfigDefault(ctx context.Context) (
...
+	err error) {
+	am.log.CDebugf(ctx, "Completing work on %s, workErr=%+v", repo, workErr)
+	defer func() {
+		am.deferLog.CDebugf(ctx, "Work done completed: %+v", err)
+	}()
+
+	// Take the lock for the dst repo while checking the work time.
+	lockFile, err := dstFS.Create(autogitLockName(repo))
+	if err != nil {
+		return err
```
jzila: In case something else is _trying_ to lock at the same time, maybe this should retry with a backoff.

strib: The `Create` won't fail in that case, but the `Lock` below will block indefinitely.  @jzila does that cover the case you're thinking of?

jzila: Yep, thanks.

libgit/autogit_manager.go
```diff
@@ -119,6 +130,148 @@ func (am *AutogitManager) getNewConfigDefault(ctx context.Context) (
...
+	ctx context.Context, dstFS *libfs.FS, repo string) (
+	canWork bool, err error) {
+	am.log.CDebugf(ctx, "Checking if we can work on %s", repo)
+	defer func() {
+		am.deferLog.CDebugf(ctx, "Work check completed: canWork=%t, %+v",
+			canWork, err)
+	}()
+
+	// Take the lock for the dst repo while checking the work time.
+	lockFile, err := dstFS.Create(autogitLockName(repo))
```
songgao: I suppose this doesn't have the same problem as in https://github.com/keybase/kbfs/pull/1446 where `Create` doesn't error if dst file exists, because we are relying on the `lockFile.Lock()` for synchronization.

libgit/autogit_manager.go
```diff
@@ -119,6 +130,148 @@ func (am *AutogitManager) getNewConfigDefault(ctx context.Context) (
...
+		return false, err
+	}
+
+	// See if someone else is already working on this repo by seeing
+	// if the timestamp (converted to "common" time) is within the
+	// expected time limit for a worker.
+	workingFileName := autogitWorkingName(repo)
+	fi, err := dstFS.Stat(workingFileName)
+	currCommonTime := am.commonTime(ctx)
+	if err != nil && !os.IsNotExist(err) {
```
songgao: I have slight preference for using a switch here, but not a big deal.

libgit/autogit_manager.go
strib: Yep, exactly.

PR #1444, opened 2018-01-12
KBFS-2702 Hide some OS X files on Windows

taruti:


libdokan/dir.go
```diff
@@ -17,6 +17,8 @@ import (
 	"golang.org/x/net/context"
 )
 
+const HiddenFilePrefix = `._`
```
jzila: This needs a comment to pass lint.

jzila:
@taruti: is there something blocking this PR?

PR #1445, opened 2018-01-12
kbp: update vendor to support http-01

songgao:
Also remove the option to disable HTTP->HTTPS redirection since we'd
need to listen on :80 for http-01 anyway, and there's not much use for
disabling the redirection.

### DIFF
--- a/libpages/server.go
+++ b/libpages/server.go
@@ -496,9 +496,12 @@ func ListenAndServe(ctx context.Context,
 	httpServer := http.Server{
 		Addr: ":80",
 		// Enable http-01 by calling the HTTPHandler method, and set the
-		// fallback HTTP handler to nil, which means we'll get a handler that
-		// redirects all HTTP traffic that's not for ACME domain verification
-		// to HTTPS.
+		// fallback HTTP handler to nil. As described in the autocert doc
+		// (https://github.com/golang/crypto/blob/13931e22f9e72ea58bb73048bc752b48c6d4d4ac/acme/autocert/autocert.go#L248-L251),
+		// this means for requests not for ACME domain verification, a default
+		// fallback handler is used, which redirects all HTTP traffic using GET
+		// and HEAD to HTTPS using 302 Found, and responds with 400 Bad Request
+		// for requests with other methods.
 		Handler:           manager.HTTPHandler(nil),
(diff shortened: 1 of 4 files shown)


libpages/server.go
```diff
@@ -519,9 +493,13 @@ func ListenAndServe(ctx context.Context,
...
-	httpRedirectServer := http.Server{
-		Addr:              ":80",
-		Handler:           http.HandlerFunc(server.redirectHandlerFunc),
+	httpServer := http.Server{
+		Addr: ":80",
+		// Enable http-01 by calling the HTTPHandler method, and set the
+		// fallback HTTP handler to nil, which means we'll get a handler that
+		// redirects all HTTP traffic that's not for ACME domain verification
+		// to HTTPS.
+		Handler:           manager.HTTPHandler(nil),
```
jzila: Does this need to be shutdown?

libpages/server.go
```diff
@@ -519,9 +496,13 @@ func ListenAndServe(ctx context.Context,
...
 	}
 
-	httpRedirectServer := http.Server{
-		Addr:              ":80",
-		Handler:           http.HandlerFunc(server.redirectHandlerFunc),
+	httpServer := http.Server{
+		Addr: ":80",
+		// Enable http-01 by calling the HTTPHandler method, and set the
+		// fallback HTTP handler to nil, which means we'll get a handler that
+		// redirects all HTTP traffic that's not for ACME domain verification
```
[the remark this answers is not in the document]
songgao: 302 for GET/HEAD and 400 for other methods (since many user-agents always respond to a 302 with a GET request, so 302 only [predictably](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) works with GET/HEAD). ([source](https://github.com/golang/crypto/blob/13931e22f9e72ea58bb73048bc752b48c6d4d4ac/acme/autocert/autocert.go#L288-L295) and [comment](https://github.com/golang/crypto/blob/13931e22f9e72ea58bb73048bc752b48c6d4d4ac/acme/autocert/autocert.go#L248-L251).

I'll amend the comment.

libpages/server.go
songgao: It's a handler, so no. But the server that listens on `:80` is shutdown in the deferred function.

PR #1448, opened 2018-01-13
libgit: fix autogit updating when source repo is read-only

strib:
When go-git calls `OpenFile` on an object file, they expect to get back a not-exists error if the object doesn't exist.  However, if the object's parent directory didn't exist, we were trying to create it automatically (which go-git _also_ expects to happen, *if* the file is successfully created).  If the source repo is read-only (say, someone else's public directory), this doesn't work because `OpenFile` returned a `WriteAccessError` instead of a non-exists error.

This commit fixes that issue by returning a not-exists error in that case.  It also splits up the autogit update tests to have a separate read-only test, and makes that read-only test repack the objects and delete the object directories in order to repro the problem.  The test fails without this fix.

### DIFF
--- a/libgit/autogit_node_wrappers_test.go
+++ b/libgit/autogit_node_wrappers_test.go
@@ -6,6 +6,7 @@ package libgit
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"testing"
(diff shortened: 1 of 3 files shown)


libgit/autogit_node_wrappers_test.go
```diff
@@ -59,6 +60,41 @@ func TestAutogitNodeWrappers(t *testing.T) {
...
+	f, err := rootFS.Open(
+		fmt.Sprintf(".kbfs_autogit/%s/user1/test/foo", tlfType))
+	require.NoError(t, err)
+	defer f.Close()
+	data, err := ioutil.ReadAll(f)
+	require.NoError(t, err)
+	require.Equal(t, "hello", string(data))
+}
+
+func checkAutogit2(t *testing.T, rootFS *libfs.FS, tlfType string) {
```
jzila: more descriptive name please.

PR #1450, opened 2018-01-15
add a couple HTTP response headers

songgao:


libpages/server.go
```diff
@@ -287,6 +287,27 @@ func (s *Server) logRequest(sri *ServedRequestInfo, requestPath string) {
...
 }
 
+func (s *Server) setCommonResponseHeaders(w http.ResponseWriter) {
+	// Since http.FileServer already sets MIME type properly, disable MIME type
+	// sniffing on browser-side. This would prevent e.g. an attack where a
+	// malicious html file with .jpg suffix being executed by browser without
+	// site visitors' awareness. References:
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
+	// https://helmetjs.github.io/docs/dont-sniff-mimetype/
+	w.Header().Set("X-Content-Type-Options", "nosniff")
```
songgao: I'm actually leaning towards not having this header, for three reasons:

1) Keybase Pages doesn't have visitor-uploaded content; all content is provided by the user who hosts with us. This means an evil user Evan has to already be hosting with us in order to launch an attack using this, on his own site. For example, Evan hosts an html under jpg, and sends a link to Alice, in which case we are hosting hostile content from Evan to Alice, when Evan controls the domain. This is pretty rare, and arguably as a hosting provider, we shouldn't be worried about this kind of attack.

2) This attack in example didn't work in my Chrome without the added `nosniff` header, so modern browsers might have already been doing the right thing.

3) As mentioned on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options) A naive implementation of `nosniff` may break image rendering, and it was fixed in standard [only recently](https://github.com/whatwg/fetch/issues/395), so enabling this header globally for all returned content may still break some use cases.

That being said, keybase.pub has this header returned for both webpage and images. So removing this would be a change.

cc @malgorithms

libpages/server.go
```diff
@@ -287,6 +287,27 @@ func (s *Server) logRequest(sri *ServedRequestInfo, requestPath string) {
...
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
+	// https://blog.innerht.ml/the-misunderstood-x-xss-protection/
+	w.Header().Set("X-XSS-Protection", "1; mode=block")
+	// Only allow HTTPS on this domain, and make this policy it expire in a
+	// week. This means if user decides to migrate off Keybase Pages, there's a
+	// 1-week gap before they can use HTTP again. Note that we don't use the
+	// 'preload' directive, for the same reason we use 302 instead of 301 for
+	// HTTP->HTTPS redirection. Reference: https://hstspreload.org/#opt-in
+	w.Header().Set("Strict-Transport-Security", "max-age=604800; includeSubDomains")
+	// TODO: allow user to opt-in some directives of Content-Security-Policy?
```
songgao: CSP is present on a neocities site as following:

```
content-security-policy: upgrade-insecure-requests; default-src 'unsafe-inline' 'unsafe-eval' 'self' data: blob: *
```

But it seems these all should be opt-in by user, perhaps through the .kbp_config, thus the TODO. But let me know if anybody feels differently.

libpages/server.go
malgorithms: cool, your call on this since I'm not sure.

libpages/server.go
```diff
@@ -287,6 +287,27 @@ func (s *Server) logRequest(sri *ServedRequestInfo, requestPath string) {
...
+	// malicious html file with .jpg suffix being executed by browser without
+	// site visitors' awareness. References:
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
+	// https://helmetjs.github.io/docs/dont-sniff-mimetype/
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	// Enforce XSS protection. References:
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection
+	// https://blog.innerht.ml/the-misunderstood-x-xss-protection/
+	w.Header().Set("X-XSS-Protection", "1; mode=block")
+	// Only allow HTTPS on this domain, and make this policy it expire in a
```
strib: "policy it" -> "policy"?

songgao:
@jzila Is there a specific CORS header you are thinking of? If I understand correctly, they enable CORS in specific scenarios rather than disabling it. Browsers already block cross-origin by default.

PR #1452, opened 2018-01-15
kbfsgit: avoid infinite loops when a "." entry is present

strib:
Early repos may have been created before the check for "." was done on create, and go-git could have caused it to be created.  Then when cloning, the recursive byte count function would enter an infinite loop.  So just ignore "." during any recursion.

Issue: keybase/client#10189


strib:
> What about ../<self>?

That would get scrubbed away by `path.Clean()` I believe (which is required for path inputs into libfs), whereas "." gets left alone.

jzila:
Cool, thanks.

PR #1453, opened 2018-01-15
libfs: explicitly pass name to FileInfo, so that symlinks work

strib:
Issue: KBFS-2714

### DIFF
--- a/libfs/file_info.go
+++ b/libfs/file_info.go
@@ -17,13 +17,14 @@ type FileInfo struct {
 	fs   *FS
 	ei   libkbfs.EntryInfo
 	node libkbfs.Node
+	name string
 }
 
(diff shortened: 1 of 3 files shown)


libfs/file_info.go
```diff
@@ -17,13 +17,14 @@ type FileInfo struct {
...
 	node libkbfs.Node
+	name string
 }
 
 var _ os.FileInfo = (*FileInfo)(nil)
 
 // Name implements the os.FileInfo interface for FileInfo.
 func (fi *FileInfo) Name() string {
-	return fi.node.GetBasename()
+	return fi.name
```
songgao: Is this because links don't have a node (i.e. `fi.node == nil`)?

strib: Yep!

PR #1454, opened 2018-01-15
libgit: allow autogit removal via rmdir

strib:
This PR allows you to remove autogit repo directories via `rmdir` in the filesystem.  It launches the removal in the background, and blocks for 10s waiting for it to finish, returning with no error if it doesn't finish in time.

It adds another function (`RemoveDir`) to the node interface to allow for custom deletions.

Issue: KBFS-2681

### DIFF
--- a/libgit/autogit_manager.go
+++ b/libgit/autogit_manager.go
@@ -453,6 +453,26 @@ func (am *AutogitManager) queueReset(ctx context.Context, req resetReq) (
 	return req.doneCh, nil
 }
 
+func (am *AutogitManager) removeLock(
+	ctx context.Context, gitConfig libkbfs.Config, fs *libfs.FS, repo string) (
+	err error) {
+	err = fs.Remove(autogitLockName(repo))
+	if err != nil {
+		return err
+	}
+	err = fs.SyncAll()
+	if err != nil {
+		return err
+	}
+	jServer, err := libkbfs.GetJournalServer(gitConfig)
+	if err != nil {
+		return err
+	}
+	return jServer.FinishSingleOp(
+		ctx, fs.RootNode().GetFolderBranch().Tlf, nil,
+		keybase1.MDPriorityNormal)
+}
+
 func (am *AutogitManager) doDelete(req deleteReq) (err error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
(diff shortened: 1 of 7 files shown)


libgit/autogit_node_wrappers.go
```diff
@@ -275,13 +282,53 @@ func (tn tlfNode) ShouldCreateMissedLookup(
...
+	_, _, err = GetRepoAndID(ctx, tn.am.config, tn.h, name, "")
+	if err != nil {
+		return false, err
+	}
+
+	if normalizeRepoName(name) != name {
+		// Can't remove a symlink as if it were a directory.
+		return false, nil
+	}
+	ctx = context.WithValue(ctx, ctxSkipRemoveDirKey, struct{}{})
```
songgao: Could you explain how `ctxSkipRemoveDirKey` is used and why it's needed here? It seems `tn.am.Delete` only utilizes timeout/cancellation on the `ctx`, and I don't see a recursive call into this function. So in what cases can the check on this key at the beginning return?

libgit/autogit_manager.go
```diff
@@ -508,6 +528,13 @@ func (am *AutogitManager) doDelete(req deleteReq) (err error) {
 		if err == nil {
 			err = workDoneErr
 		}
+		// Remove the lock file.  This happens outside of the main
+		// deletion single-op, and so won't appear strictly atomically
+		// with the rest of the delete.
```
songgao: Hmm I'm bit confused. It still creates the lock file through `canWorkOnRepo`, and does an unlock through `workDoneOnRepo`. Could you explain what you meant by not "strictly atomically"?

libgit/autogit_node_wrappers.go
```diff
@@ -36,7 +36,8 @@ import (
 // * `tlfNode` allows auto-creation of subdirectories representing
 //   valid repository checkouts of the corresponding TLF, e.g.
 //   `.kbfs_autogit/private/chris/dotfiles`.  It wraps child nodes in
-//   two ways, as both a `readonlyNode` and a `repoNode`.
+//   two ways, as both a `readonlyNode` and a `repoNode`.  It allows
+//   repo directories to be removed via `RemoveDir`.
```
jzila: Might want to add that this only happens if you have write permissions to the folder.

libgit/autogit_node_wrappers.go
strib: Oops, you're right, this isn't actually needed.  I'll remove it.  Thanks!

libgit/autogit_manager.go
strib: The call to `workDoneOnRepo` unlocks the file, which causes a `FinishSingleOp` call.  So the `Remove` call below happens _after_ that first single operation finishes, and is part of a new operation.  That means the lock will be removed in a separate MD update from the rest of the deletion.

songgao: Oh I see; makes sense. Thanks!

PR #1456, opened 2018-01-18
kbp: fix team git site root

songgao:
Tested on staging and made sure it works for both team folder and team git.

### DIFF
--- a/libpages/root.go
+++ b/libpages/root.go
@@ -49,6 +49,19 @@ func (t RootType) String() string {
 	}
 }
 
+func tlfTypeToPath(t tlf.Type) string {
+	switch t {
+	case tlf.Private:
+		return string(libkbfs.PrivatePathType)
+	case tlf.Public:
+		return string(libkbfs.PublicPathType)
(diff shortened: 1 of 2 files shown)


libpages/root.go
```diff
@@ -49,6 +49,19 @@ func (t RootType) String() string {
 	}
 }
 
+func tlfTypeToPath(t tlf.Type) string {
+	switch t {
+	case tlf.Private:
+		return string(libkbfs.PrivatePathType)
```
strib: Technically the "right" variables to use would be the ones in `libgit/autogit_node_wrappers.go", which define how the tlf types map to subdirs for autogit specifically.  (They would have to be exported.)

But this is fine too I guess, just a little dirtier.

songgao: Ah I only saw the ones in `libkbfs`. I'll export it from `libgit`!

songgao:
@jzila `tlf.Type.String()` returns `singleTeam` for `tlf.SingleTeam`

jzila:
Got it, thanks.

PR #1457, opened 2018-01-19
Exit kbfs when service shuts down, even with no fs mount

zanderz:
so it can be more like other platforms, for cleaner shutdown and updating.
The separate NotifyServiceHandler seemed necessary because there already was a Shutdown() (no argument) satisfying some other interface.
Related to DESKTOP-5956, which will be blocked until this is merged.
Tested in a few production builds today.
@taruti @maxtaco @jzila

### DIFF
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -435,6 +435,7 @@ func (s NotifyServiceHandler) Shutdown(_ context.Context) error {
 	return nil
 }
 
+// NewNotifyServiceHandler makes a new NotifyServiceHandler
 func NewNotifyServiceHandler(config Config, log logger.Logger) keybase1.NotifyServiceInterface {
 	s := NotifyServiceHandler{config: config, log: log}
 	return s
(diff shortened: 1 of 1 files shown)


zanderz:
I realize this will close KBFS in prod mode every time the service closes, not just when `keybase ctl stop --kill-kbfs` is issued. If we don't like that, we could forward an argument through the rpc layers to distinguish that case, but I think macos works like this anyway.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -403,9 +413,31 @@ func (k *KeybaseDaemonRPC) Shutdown() {
...
+func (s notifyServiceHandler) Shutdown(_ context.Context) error {
+	s.log.Warning("NotifyService: Shutdown")
+	if runtime.GOOS == "windows" {
+		os.Exit(0)
+	}
+	return nil
+}
+
+// newNotifyServiceHandler makes a new NotifyServiceHandler
+func newNotifyServiceHandler(config Config, log logger.Logger) keybase1.NotifyServiceInterface {
```
taruti: Is there a reason we use this by value and not by reference here and in the methods?

zanderz: No, will change

zanderz:
yes, see `watchdog2`. The gui starts the service via the watchdog, and so does the system startup shortcut.

zanderz:
That counts as a thumb I believe?

taruti:
LGTM

PR #1459, opened 2018-01-19
populate basename to fix mime type

songgao:
This is a bug reported in keybasefriends, where we are reporting `text/plain` for css resources.

The `name` field here is used eventually in `.Name()` which is used by `http` package to determine the mime type. I forgot to populate it when implementing the wrapper for `http.FileSystem` in the first place and missed this bug in testing because of my browser cache ...

Tested on staging to make sure it fixes the mime bug.


songgao:
Whoa that was fast!

songgao:
Oh I thought it was my browser cache; no worries!

PR #1461, opened 2018-01-19
libfuse: avoid doing full TLF loads when `.localized` is requested

strib:
In particular, specify a set of directory entry names that should NOT cause a TLF to be fully loaded and identified on a lookup.  If the directory is not yet loaded and one of these names are looked-up, then ENOENT will be returned automatically.  This is to avoid unnecessary loading and tracker popups when listing the folder list directories. For example, when macOS finder opens `/keybase/private`, it looks up `.localized` in every TLF subdirectory to see if it should translate the TLF folder name or not, which can cause a tracker popup storm

Issue: KBFS-2649

### DIFF
--- a/libfuse/tlf.go
+++ b/libfuse/tlf.go
@@ -152,7 +152,7 @@ func (tlf *TLF) Attr(ctx context.Context, a *fuse.Attr) error {
 	return dir.Attr(ctx, a)
 }
 
-// tlfLoadavoidingLookupNames specifies a set of directory entry names
+// tlfLoadAvoidingLookupNames specifies a set of directory entry names
 // that should NOT cause a TLF to be fully loaded and identified on a
 // lookup.  If the directory is not yet loaded and one of these names
 // are looked-up, then ENOENT will be returned automatically.  This is
@@ -161,13 +161,13 @@ func (tlf *TLF) Attr(ctx context.Context, a *fuse.Attr) error {
 // `/keybase/private`, it looks up `.localized` in every TLF
 // subdirectory to see if it should translate the TLF folder name or
 // not, which can cause a tracker popup storm (see KBFS-2649).
-var tlfLoadavoidingLookupNames = map[string]bool{
+var tlfLoadAvoidingLookupNames = map[string]bool{
 	".localized": true,
 }
 
 // Lookup implements the fs.NodeRequestLookuper interface for TLF.
 func (tlf *TLF) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.LookupResponse) (fs.Node, error) {
-	if tlfLoadavoidingLookupNames[req.Name] {
+	if tlfLoadAvoidingLookupNames[req.Name] {
 		dir := tlf.getStoredDir()
 		if dir == nil {
 			tlf.log().CDebugf(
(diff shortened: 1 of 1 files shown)


libfuse/tlf.go
```diff
@@ -152,8 +152,31 @@ func (tlf *TLF) Attr(ctx context.Context, a *fuse.Attr) error {
...
+// tlfLoadavoidingLookupNames specifies a set of directory entry names
+// that should NOT cause a TLF to be fully loaded and identified on a
+// lookup.  If the directory is not yet loaded and one of these names
+// are looked-up, then ENOENT will be returned automatically.  This is
+// to avoid unnecessary loading and tracker popups when listing the
+// folder list directories.  For example, when macOS finder opens
+// `/keybase/private`, it looks up `.localized` in every TLF
+// subdirectory to see if it should translate the TLF folder name or
+// not, which can cause a tracker popup storm (see KBFS-2649).
+var tlfLoadavoidingLookupNames = map[string]bool{
```
jzila: s/avoiding/Avoiding

libfuse/tlf.go
```diff
@@ -152,8 +152,31 @@ func (tlf *TLF) Attr(ctx context.Context, a *fuse.Attr) error {
...
+}
+
 // Lookup implements the fs.NodeRequestLookuper interface for TLF.
 func (tlf *TLF) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.LookupResponse) (fs.Node, error) {
+	if tlfLoadavoidingLookupNames[req.Name] {
+		dir := tlf.getStoredDir()
+		if dir == nil {
+			tlf.log().CDebugf(
+				ctx, "Avoiding TLF loading for name %s", req.Name)
+			return nil, fuse.ENOENT
```
jzila: Does this risk caching the `ENOENT` and not showing it when you actually want it?

strib: As far as I know the kernel doesn't cache errors.

libfuse/tlf.go
strib: I did this on purpose because I wanted to evoke "load-avoiding lookup" rather than "load avoiding lookup" (which is ambiguous), and it seemed like the right way to do it.  But I guess I don't feel too strongly about.

jzila: it's pretty hard to read though.

libfuse/tlf.go
```diff
@@ -152,8 +152,31 @@ func (tlf *TLF) Attr(ctx context.Context, a *fuse.Attr) error {
...
 // Lookup implements the fs.NodeRequestLookuper interface for TLF.
 func (tlf *TLF) Lookup(ctx context.Context, req *fuse.LookupRequest, resp *fuse.LookupResponse) (fs.Node, error) {
+	if tlfLoadAvoidingLookupNames[req.Name] {
+		dir := tlf.getStoredDir()
+		if dir == nil {
+			tlf.log().CDebugf(
+				ctx, "Avoiding TLF loading for name %s", req.Name)
+			return nil, fuse.ENOENT
+		}
+		dir.Lookup(ctx, req, resp)
```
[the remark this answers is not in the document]
strib: This has been a test, and you finally passed!  https://github.com/keybase/kbfs/pull/1894

PR #1462, opened 2018-01-21
kbp: 403 instead of 401 when appropriate; add a simple test; status code logging/stat fix

songgao:
Sorry for adding out-of-sprint stuff for review! I've made a few small changes for keybase pages over the weekend, mainly:

1. When `.kbp_config` doesn't have more permissive permissions defined for a given path, it's not possible to get the required permission even with authentication. So return 403 rather than 401. ([more](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) on MDN)

2. I noticed a bug where we were logging status code 200 and reporting 200 to stathat for non-200 responses generated by `http.FileServer`. This PR introduces a simple wrapper around `http.ResponseWriter` to make sure we are getting correct status code for logging and stats.

3. Added boilerplates for server.go tests.

It might be easier to look at individual commits rather than the whole thing together.

Thanks!!

### DIFF
--- a/libpages/server.go
+++ b/libpages/server.go
@@ -113,13 +113,18 @@ func (s *Server) handleErrorAndPopulateSRI(
 	// TODO: have a nicer error page for configuration errors?
 	switch err.(type) {
 	case nil:
-	case ErrKeybasePagesRecordNotFound:
+	case ErrKeybasePagesRecordNotFound, ErrDomainNotAllowedInWhitelist:
 		sri.HTTPStatus = http.StatusServiceUnavailable
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 		return
 	case ErrKeybasePagesRecordTooMany, ErrInvalidKeybasePagesRecord:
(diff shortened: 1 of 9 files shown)


strib:
@songgao do you want to fix the compile before we review, or should we review as is?

songgao:
Sorry about that! Let's see if this one passes ... it shouldn't matter much for the change though.

songgao:
CI passed!

libpages/server.go
```diff
@@ -286,7 +319,6 @@ func (s *Server) logRequest(sri *ServedRequestInfo, requestPath string) {
 		zap.Bool("invalid_config", sri.InvalidConfig),
 	)
 }
-
```
strib: Function separator line now missing?

songgao: Thanks; I got it back from lost-n-found.

songgao:
@jzila yup that's correct! Sorry I didn't make it clear in the PR description.

PR #1464, opened 2018-01-23
Vendor team notification protocol change, placeholder

patrickxb:
needed for https://github.com/keybase/client/pull/10271 to pass circle ci


strib:
(Ideally you could quickly push the client-side first, and then update this one to revendor to the real git commit ID.  But if that's too much trouble, it's ok.)

PR #1465, opened 2018-01-23
[KBFS-2730] Pass branch deletions to the GitLocalMetadata RPC

jzila:
cc @patrickxb 
cc @keybase/kbfs-hackers

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -1439,20 +1439,20 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
 // not in `remoteStorer`.
 func (r *runner) parentCommitsForRef(ctx context.Context,
 	localStorer gogitstor.Storer, remoteStorer gogitstor.Storer,
-	refs map[string]bool) (libgit.CommitsByRefName, error) {
+	refs map[gogitcfg.RefSpec]bool) (libgit.RefDataByName, error) {
 
-	commitsByRef := make(libgit.CommitsByRefName, len(refs))
+	commitsByRef := make(libgit.RefDataByName, len(refs))
 	haves := make(map[plumbing.Hash]bool)
 
-	for refName := range refs {
-		ref, err := localStorer.Reference(plumbing.ReferenceName(refName))
+	for refspec := range refs {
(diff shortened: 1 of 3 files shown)


jzila:
Please hold off reviewing, I thought I'd fixed tests but looks like I was running the previous test binary.

jzila:
OK ready for review again.

kbfsgit/runner.go
```diff
@@ -1439,20 +1439,26 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
...
+	refs map[gogitcfg.RefSpec]bool) (libgit.RefDataByName, error) {
 
-	commitsByRef := make(libgit.CommitsByRefName, len(refs))
+	commitsByRef := make(libgit.RefDataByName, len(refs))
 	haves := make(map[plumbing.Hash]bool)
 
-	for refName := range refs {
-		ref, err := localStorer.Reference(plumbing.ReferenceName(refName))
+	for refspec := range refs {
+		if refspec.IsDelete() {
```
strib: It would be pretty weird for `pushAll` to have a delete refspec, but I guess it's not impossible.  Anyway, this code isn't too intrusive I guess.

libgit/repo.go
```diff
@@ -238,6 +244,7 @@ func UpdateRepoMD(ctx context.Context, config libkbfs.Config,
 			RefName:              string(refName),
 			Commits:              kbCommits,
 			MoreCommitsAvailable: hasMoreCommits,
+			IsDelete:             refData.IsDelete,
```
strib: If there are no commits for this ref (and there shouldn't be for deleted refs, right?)  then this metadata won't ever be appended, so it seems like the delete wouldn't get sent at all, right?

jzila: I don't follow...we're iterating over `refData` here, and we haven't used `range refData.Commits` to skip out of this completely. A ref with no commits will still send an `UpdateRepoMD` (as indeed we do with repo creations).

strib: Oh sorry, it looked like this was in the `range refData.Commits` loop.  I misread the github tabbing.

PR #1469, opened 2018-01-26
simplefs: fix CopyRecursive-to-local by not trying to open dirs

strib:
[Note that this PR is mostly just me getting my feet wet for SimpleFS -- for KBFS-2661 this code will be largely rewritten, but I thought it would be nice to have a working test in place before that happens.]

Issue: KBFS-2737

### DIFF
--- a/simplefs/simplefs_test.go
+++ b/simplefs/simplefs_test.go
@@ -202,6 +202,75 @@ func TestCopyToLocal(t *testing.T) {
 	assert.True(t, exists, "File copy destination must exist")
 }
 
+func TestCopyRecursive(t *testing.T) {
+	ctx := context.Background()
+	sfs := newSimpleFS(libkbfs.MakeTestConfigOrBust(t, "jdoe"))
+	defer closeSimpleFS(ctx, t, sfs)
+
+	// make a temp local dest directory + files we will clean up later
+	tempdir, err := ioutil.TempDir("", "simpleFstest")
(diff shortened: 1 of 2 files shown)


simplefs/simplefs_test.go
```diff
@@ -202,6 +202,75 @@ func TestCopyToLocal(t *testing.T) {
...
+		string(readRemoteFile(ctx, t, sfs, pathAppend(path2, "test2.txt"))))
+
+	// Copy it back.
+	tempdir2, err := ioutil.TempDir("", "simpleFstest")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempdir2)
+	path3 := keybase1.NewPathWithLocal(filepath.Join(tempdir2, "testdir"))
+	opid2, err := sfs.SimpleFSMakeOpid(ctx)
+	require.NoError(t, err)
+	err = sfs.SimpleFSCopyRecursive(ctx, keybase1.SimpleFSCopyRecursiveArg{
```
jzila: Is this the part that was failing before?

strib: Yeah.

> LGTM, but it's not obvious to me what the scenario was that caused the crash. Was it that OpenFile(path.Local() failed because the directory didn't exist?

No, it failed because `OpenFile` doesn't work on a directory.

jzila: D'oh!

PR #1471, opened 2018-01-29
libpages: Add a new possible record prefix

jzila:
This is to get around domain registrars that disallow underscores in the middle of the name.

### DIFF
--- a/libpages/dns.go
+++ b/libpages/dns.go
@@ -52,7 +52,7 @@ func (ErrKeybasePagesRecordTooMany) Error() string {
 // locate the root of these Keybase pages. We have 2 records since some
 // registrars don't support underscores in the middle of a domain. This order
 // must remain fixed since it reflects the order the strings are evaluated in.
-var kbpRecordPrefixes []string = []string{"_keybase_pages.", "_keybasepages."}
+var kbpRecordPrefixes = []string{"_keybase_pages.", "_keybasepages."}
 
 // LoadRoot loads the root path configured for domain from DNS, with following
 // steps:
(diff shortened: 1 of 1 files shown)


libpages/dns.go
```diff
@@ -93,7 +93,14 @@ func (l DNSRootLoader) LoadRoot(domain string) (root Root, err error) {
 		}
 	}()
 
-	txtRecords, err := net.LookupTXT(kbpRecordPrefix + domain)
+	// Check all possible kbp record prefixes.
+	for _, kbpRecordPrefix := range kbpRecordPrefixes[:] {
+		var txtRecords []string
```
songgao: This declaration needs to be outside the for loop I think?

libpages/dns.go
```diff
@@ -48,7 +48,7 @@ func (ErrKeybasePagesRecordTooMany) Error() string {
 	return "more than 1 TXT record are found for " + keybasePagesPrefix
 }
 
-const kbpRecordPrefix = "_keybase_pages."
+const kbpRecordPrefixes = [2]string{"_keybase_pages.", "_keybasepages."}
```
songgao: Would you mind adding a comment here explaining it’s important not to change the ordering of the two prefixes?

libpages/dns.go
```diff
@@ -48,7 +48,7 @@ func (ErrKeybasePagesRecordTooMany) Error() string {
...
 
 // LoadRoot loads the root path configured for domain from DNS, with following
```
songgao: Could you updated this comment block as well to mention there are two possible TXT record prefixes?

PR #1478, opened 2018-02-06
libkbfs: misc fixes needed for implicit teams

strib:
* Get TLF IDs for regular, non-implicit teams from the sigchain via the service.
* Always use a number at the end of a conflict suffix, i.e. "(conflicted copy 2016-03-14 #1)" instead of "(conflicted copy 2016-03-14)", because that's how the API server does it now.
* Fix some handle resolution issues for team-backed handles.

Issue: KBFS-2652

### DIFF
--- a/tlf/handle_extension.go
+++ b/tlf/handle_extension.go
@@ -126,7 +126,7 @@ type HandleExtension struct {
 func (e HandleExtension) String() string {
 	date := time.Unix(e.Date, 0).UTC().Format(handleExtensionDateFormat)
 	var num string
-	if e.Number > 1 {
+	if e.Number > 0 {
 		num = " #"
 		num += strconv.FormatUint(uint64(e.Number), 10)
 	}
@@ -187,7 +187,7 @@ func parseHandleExtension(fields []string) (*HandleExtension, error) {
 		if err != nil {
 			return nil, err
 		}
-		if num < 2 {
+		if num < 1 {
 			return nil, errHandleExtensionInvalidNumber
 		}
 	}
--- a/tlf/handle_extension.go
+++ b/tlf/handle_extension.go
@@ -123,14 +123,26 @@ type HandleExtension struct {
 
 // String implements the fmt.Stringer interface for HandleExtension.
 // Ex: "(conflicted copy 2016-05-09 #2)"
-func (e HandleExtension) String() string {
+func (e HandleExtension) string(isBackedByTeam bool) string {
 	date := time.Unix(e.Date, 0).UTC().Format(handleExtensionDateFormat)
 	var num string
-	if e.Number > 0 {
+	minNumberSuffixToShow := uint16(2)
+	if isBackedByTeam {
+		// When a TLF is backed by an implicit team, it should always
+		// use the "#1" suffix, unlike for older TLFs.
+		minNumberSuffixToShow = 1
+	}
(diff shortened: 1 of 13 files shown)


strib:
Ok, this passes the docker tests now and is (still) ready to review!

tlf/handle_extension.go
```diff
@@ -123,16 +123,28 @@ type HandleExtension struct {
...
 		num = " #"
 		num += strconv.FormatUint(uint64(e.Number), 10)
 	}
 	return fmt.Sprintf(handleExtensionFormat, e.Type.String(e.Username), date, num)
 }
 
+// String implements the fmt.Stringer interface for HandleExtension.
+// Ex: "(conflicted copy 2016-05-09 #2)"
+func (e HandleExtension) String() string {
+	return e.string(false)
```
jzila: I wonder if this should be the default. Can we get rid of this interface implementation?

tlf/handle_extension.go
```diff
@@ -123,16 +123,28 @@ type HandleExtension struct {
 
 // String implements the fmt.Stringer interface for HandleExtension.
 // Ex: "(conflicted copy 2016-05-09 #2)"
-func (e HandleExtension) String() string {
+func (e HandleExtension) string(isBackedByTeam bool) string {
```
jzila: Don't forget to vendor this into `kbfs-server`

tlf/handle_extension.go
strib: Sorry, I don't follow the suggestion.  You want to get rid of `String()` entirely?  I'd rather not do that, since it'll break any log message that prints the extension directly, and we won't know that it's broken.

Or do you want to flip this to `e.string(true)`?  That might make sense once we turn on implicit teams on the server, but even then there will probably be a long time when the majority of the folders are not team-backed, at least until migration happens.

jzila: Alternatively: can we store the `isBackedByTeam` state in the object itself? It still won't address when it's loaded from the database, but at least when it's created on the fly it should show the right thing.

strib: I looked into this a bit -- there are a few places where we parse a string to create the handle extension, before we know anything about the backing TLF ID (e.g., when first parsing a folder name passed in on the command line).  So there's no way to know if it's back by a team at that point.

In general, whether to show the "#1" or not really only matters when you're comparing the string against something else for an equality check, which is what happens in the chat inbox.  Whether we show it in logs, or in the folder list or whatever, really doesn't matter.  So in my view, the default doesn't matter, I'm only trying to fix compatibility issues with the chat inbox.

jzila: All right, sounds good.

tlf/handle_extension.go
```diff
@@ -123,16 +123,28 @@ type HandleExtension struct {
...
 	date := time.Unix(e.Date, 0).UTC().Format(handleExtensionDateFormat)
 	var num string
-	if e.Number > 1 {
+	minNumberSuffixToShow := uint16(2)
+	if isBackedByTeam {
+		// When a TLF is backed by an implicit team, it should always
+		// use the "#1" suffix, unlike for older TLFs.
+		minNumberSuffixToShow = 1
```
songgao: Is this the number that gets increased when TLF resets?

strib: Yeah, within the same day.

PR #1479, opened 2018-02-11
kbp: add mime types and blacklist

songgao:
Re mime types: I added common mime types from nginx to Go code directly rather than relying on e.g. `/etc/mime.types` which Go's `mime` package also supports, to ease the deployment and avoid depending on the configuration file. Let me know if any of you feel strongly against that.


libpages/mime/builder/index.js
```diff
@@ -0,0 +1,135 @@
...
+
+  // Silence all warnings
+  Promise.config({
+    warnings: false
+  })
+
+  return Promise
+}
+
+function printGo(types) {
```
strib: Maybe add a comment into the file about this being autogenerated and shouldn't be edited by hand?

libpages/mime/builder/yarn.lock
```diff
@@ -0,0 +1,71 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
strib: Checked in on purpose?

libpages/server.go
```diff
@@ -29,31 +30,70 @@ import (
...
-	domainWhitelistOnce sync.Once
-	domainWhitelist     map[string]bool
+	domainListsOnce sync.Once
+	domainWhitelist map[string]bool
+	domainBlacklist []string
 }
 
-func (c *ServerConfig) allowedByDomainWhiteList(domain string) bool {
-	c.domainWhitelistOnce.Do(func() {
+// ErrDomainBlockedInBlacklist is returned when the server is configured
```
strib: This comment seems wrong, looks copy-pasted from the whitelist one.

libpages/server.go
```diff
@@ -29,31 +30,70 @@ import (
...
 	})
-	if c.domainWhitelist != nil {
-		return c.domainWhitelist[domain]
+
+	for _, blocked := range c.domainBlacklist {
+		if strings.HasSuffix(domain, blocked) {
+			return ErrDomainBlockedInBlacklist{}
+		}
+	}
+	if c.domainWhitelist != nil && !c.domainWhitelist[domain] {
```
strib: Maybe change `c.domainWhitelist != nil` to `len(c.domainWhitelist) > 0`?  At first I thought it was an unnecessary nil check, but then I realized you were trying to check if the whitelist was configured at all. My suggestion might be a little clearer for that, unless you want to support the case where an empty whitelist blocks EVERYTHING!  (Which is fine too, but might need a comment here.)

libpages/mime/builder/yarn.lock
songgao: Checked in with my consciousness. I think this locks down the node modules versions. Since frontend checks this file in, I figured that might be a good practice.

libpages/mime/Makefile
```diff
@@ -0,0 +1,10 @@
+.phony: default clean
+
+default: init.go clean
```
jzila: I don't think the default should be to clean.

libpages/server.go
```diff
@@ -29,31 +30,70 @@ import (
 
 // ServerConfig holds configuration parameters for Server.
 type ServerConfig struct {
-	DomainWhitelist  []string
+	// If DomainWhitelist is non-nil and non-empty, only domains in the
+	// whitelist are served and others are blocked.
```
jzila: Why do we have a whitelist? I don't see it used anywhere.

kbpagesd/main.go
```diff
@@ -36,6 +38,8 @@ func init() {
 		"stathat EZ key for reporting stats to stathat; empty disables stathat")
 	flag.StringVar(&fStathatPrefix, "stathat-prefix", "kbp -",
 		"prefix to stathat statnames")
+	flag.StringVar(&fBlacklist, "blacklist", "",
+		"a comma-separated list of domains to block")
```
jzila: Not for this PR, but once we scale horizontally, then we'll need a way to make this blacklist scalable and modifiable in-place.

libpages/server.go
songgao: It was in the initial version where I was imagining someone who host themselves might only want it to serve their own sites. We don't use that in `kbpagesd` yet, but I suppose when we make the "host yourself" doc we'll want that.

kbpagesd/main.go
songgao: Good point!

songgao: I'll add a TODO

PR #1484, opened 2018-02-20
lazy loading site root to avoid caching obsolete Node

songgao:
This fixes a bug that `e38383` reported in keybasefriends. In short, if the directory corresponding to the site root is moved or removed, kbpagesd kept track of the old `Node` that doesn't represent the real site root anymore, thus serves outdated data.

I tried to use the `libkbfs.Observer` but it seemed fairly complicated since I'd need to keep track of all intermediate nodes too. So this PR just changes the `Site` to cache a `*libfs.FS` of either the TLF, or the `<tlf>` dir in autogit, instead of the actual site root. The server does a `chroot` on each request. This way it always gets the newest site root. I don't think this handles TLF resets yet, which is tracked in a different ticket (KBFS-2657).

Let me know if you guys have better ideas!

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -714,8 +714,8 @@ func (fs *FS) Chtimes(name string, atime time.Time, mtime time.Time) (
 	return fs.config.KBFSOps().SetMtime(fs.ctx, n, &mtime)
 }
 
-// Chrute returns a *FS whose root is p.
-func (fs *FS) Chrute(p string) (newFS *FS, err error) {
+// ChrootAsLibFS returns a *FS whose root is p.
+func (fs *FS) ChrootAsLibFS(p string) (newFS *FS, err error) {
 	fs.log.CDebugf(fs.ctx, "Chroot %s", p)
 	defer func() {
 		fs.deferLog.CDebugf(fs.ctx, "Chroot done: %+v", err)
@@ -753,7 +753,7 @@ func (fs *FS) Chrute(p string) (newFS *FS, err error) {
 
 // Chroot implements the billy.Filesystem interface for FS.
 func (fs *FS) Chroot(p string) (newFS billy.Filesystem, err error) {
-	return fs.Chrute(p)
+	return fs.ChrootAsLibFS(p)
 }
 
 // Root implements the billy.Filesystem interface for FS.
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -714,8 +714,8 @@ func (fs *FS) Chtimes(name string, atime time.Time, mtime time.Time) (
 	return fs.config.KBFSOps().SetMtime(fs.ctx, n, &mtime)
 }
 
-// Chroot implements the billy.Filesystem interface for FS.
-func (fs *FS) Chroot(p string) (newFS billy.Filesystem, err error) {
+// Chrute returns a *FS whose root is p.
+func (fs *FS) Chrute(p string) (newFS *FS, err error) {
 	fs.log.CDebugf(fs.ctx, "Chroot %s", p)
 	defer func() {
 		fs.deferLog.CDebugf(fs.ctx, "Chroot done: %+v", err)
@@ -725,7 +725,7 @@ func (fs *FS) Chroot(p string) (newFS billy.Filesystem, err error) {
 	// lookupOrCreateEntry doesn't handle "..", so we don't have to
 	// worry about someone trying to break out of the jail since this
 	// lookup will fail.
-	n, _, err := fs.lookupOrCreateEntry(p, os.O_RDONLY, 0)
+	n, ei, err := fs.lookupOrCreateEntry(p, os.O_RDONLY, 0)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


songgao:
I was being inaccurate about what's being cached, so I edited the PR description to fix it.

libfs/fs.go
```diff
@@ -714,8 +714,8 @@ func (fs *FS) Chtimes(name string, atime time.Time, mtime time.Time) (
 	return fs.config.KBFSOps().SetMtime(fs.ctx, n, &mtime)
 }
 
-// Chroot implements the billy.Filesystem interface for FS.
-func (fs *FS) Chroot(p string) (newFS billy.Filesystem, err error) {
+// Chrute returns a *FS whose root is p.
```
strib: I don't really get this function name.  Maybe `ChrootAsLibFS()`?

libpages/root.go
```diff
@@ -68,11 +68,26 @@ type Root struct {
...
+}
+
+// Use returns a *libfs.FS to use.
+func (fs CacheableFS) Use() (*libfs.FS, error) {
+	return fs.fs.Chrute(fs.subdir)
+}
+
+// MakeFS makes a CacheableFS from *r, which can be adapted to a http.FileSystem
 // (through ToHTTPFileSystem) to be used by http package to serve through HTTP.
+// Caller must call use() to get a usable FS.
```
strib: `use()` -> `Use()`?

libpages/root.go
```diff
@@ -92,25 +107,28 @@ func (r *Root) MakeFS(
...
-		fs, err = libfs.NewFS(fsCtx, kbfsConfig,
-			tlfHandle, r.PathUnparsed, "", keybase1.MDPriorityNormal)
+		tlfFS, err := libfs.NewFS(fsCtx, kbfsConfig,
+			tlfHandle, "", "", keybase1.MDPriorityNormal)
 		if err != nil {
-			return nil, tlf.ID{}, nil, err
+			return CacheableFS{}, tlf.ID{}, nil, err
 		}
-		return fs, tlfHandle.TlfID(), cancel, nil
+		return CacheableFS{
```
strib: Is it worth trying to do a `Use()` here, just to make sure `r.PathUnparsed` is valid before we return (and presumably cache) the object?

libfs/fs.go
jzila: Cute name, but we shouldn't use cute names for exports.
Here are some alternatives:
1. You can use `Chroot` and allow the caller to type assert `newFS.(*libfs.FS)`.
2. You can edit `Chroot` to return the concrete type instead.
3. You can name `Chrute` => `ChrootFS`.

libpages/root.go
```diff
@@ -68,11 +68,26 @@ type Root struct {
 	PathUnparsed    string
 }
 
-// MakeFS makes a *libfs.FS from *r, which can be adapted to a http.FileSystem
+// CacheableFS is a wrapper around a *libfs.FS and a subdir. Use use() to get a
```
jzila: "Use \`Use()\`"

libpages/root.go
```diff
@@ -68,11 +68,26 @@ type Root struct {
...
+// *libfs.FS that roots at subdir. This essentially delays "cd"ing into subdir,
+// and is useful for caching a *libfs.FS object without the downside of caching
+// a libkbfs.Node that can be obsolete when it's renamed or removed.
+type CacheableFS struct {
```
jzila: Does this need to be exported?

songgao: Yeah because `MakeFS` is exported and golint would complain otherwise.

libfs/fs.go
songgao: I don't think 1 and 2 would work since it needs to satisfy the billy interface. Will go with rename path.

libpages/root.go
jzila: Cool thx

libfs/fs.go
jzila: I don't understand... if `Chroot` can passthrough to `ChrootLibFS`, then a type assert should work.

songgao: The `Chroot` in `billy.FileSystem` expects a return type of `billy.FileSystem`. This is the same issue when migrating to stdlib's `context`. See: https://play.golang.org/p/XeWVEJCyM9p

jzila: I know, the recommendation was to leave `Chroot` alone, and then assert its return type in the `libpages.CacheableFS.Use()` caller.

songgao: Oh I see; I misunderstood.

jzila: Not going to do that? It seems better--no abstraction leakage in the API.

songgao: Oh I thought you were OK with renaming too. I just didn't want to have that type assertion; seemed unnecessary and it casts an assumption about the underlying type that's not guaranteed from method definition, which means if we change it later it'd break in `libpages`.

jzila: Yeah the tradeoffs seem to be:
* `ChrootLibFS`:
  * (+) Validates typing at compile time
  * (-) Leaks the `*FS` concrete type
  * (-) `FS` is no longer just an implementation of `billy.Filesystem`
* `Chroot` + type assert:
  * (-) Doesn't validate at compile time, can fail at runtime
  * (+) Preserves the `billy.Filesystem` abstraction
  * (+) Doesn't create a new API to support solely for a custom downstream use case

I guess it doesn't really matter.

songgao: Nice summary! Just one thing:

> (-) FS is no longer just an implementation of billy.Filesystem

Unfortunately we already have `ToHTTPFileSystem` on `*libfs.FS` (also for the purpose of typing), so this one's already broken.

PR #1485, opened 2018-02-20
show non-existent TLF as empty dir in simplefs

songgao:
This is KBFS side of [KBFS-2781](https://keybase.atlassian.net/browse/KBFS-2781) where we return an empty result for Readdir on non-existent TLFs. A separate client side PR will fix displaying empty folders.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -147,37 +147,37 @@ func remoteTlfAndPath(path keybase1.Path) (
 }
 
 func (k *SimpleFS) getFS(ctx context.Context, path keybase1.Path) (
-	fs billy.Filesystem, finalElem string, exitEarly bool, err error) {
+	fs billy.Filesystem, finalElem string, err error) {
 	pt, err := path.PathType()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


simplefs/simplefs.go
```diff
@@ -147,33 +147,37 @@ func remoteTlfAndPath(path keybase1.Path) (
 }
 
 func (k *SimpleFS) getFS(ctx context.Context, path keybase1.Path) (
-	fs billy.Filesystem, finalElem string, err error) {
+	fs billy.Filesystem, finalElem string, exitEarly bool, err error) {
```
jzila: Instead of this new boolean flag, why not just log and reset the error?

songgao: Caller would otherwise use the returned `fs` to `Stat` and `Readdir`, which in this case would be a `nil`. I guess we could just return early in caller if `fs == nil`, but I feel having a flag is better since it explicitly says the caller should exit early.

jzila: I'd rather have a special error type to check against.

PR #1486, opened 2018-02-21
[KBFS-2747] Quick and dirty 'fixes' to unblock mobile

jzila:
I don't expect these to be the final fixes.

In this PR:
* Enable the dirty block cache for `InitMinimal`
* Enable the node cache for `InitMinimal`
* Enable the metrics registry for `InitMinimal`, because why not.

We have a lot of unguarded access for the dirty block cache and the node cache. I suspect that we need the node cache because of `GetOrCreateRootNode`, but the dirty block cache can probably get a no-op version of itself that does nothing.

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -401,7 +401,7 @@ func NewConfigLocal(mode InitMode, loggerFn func(module string) logger.Logger,
 
 	// Don't bother creating the registry if UseNilMetrics is set, or
 	// if we're in minimal mode.
-	if !metrics.UseNilMetrics && config.Mode() != InitMinimal {
+	if !metrics.UseNilMetrics {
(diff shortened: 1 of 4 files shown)


jzila:
@strib: I'm happy to defer to any and all recommendations you give. For now I'm testing with this because it's needed to make the native app not crash. But I don't expect to merge this at all.

strib:
Ah cool, thought you were looking for a review to merge.  These seem totally fine for testing a branch, and are a good starting point for doing a proper KBFS-2744 someday.

jzila:
Yeah part of the reason I tagged you for review is to surface the parts that are actually crashing on mobile, so you know where to go with KBFS-2744.

jzila:
Not needed any more since the constrained build mode has been introduced.

PR #1489, opened 2018-02-23
Jenkins changes

songgao:
It's happened twice that `libpages.test` just times out. It runs within 0.01s on my laptop, so not sure how it could happen.

I reduced the `test.timout` to 10s, but not sure if it helps with this particular case since the test on Jenkins just runs over 1m30s while the test it self is run with a 30s timeout. But the retry should hopefully get it eventually to pass.

Update: added a new commit to build test binaries first before running all tests in parallel

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -365,19 +365,19 @@ def runNixTest(prefix) {
     tests[prefix+'libpages'] = {
         dir('libpages') {
             sh 'go test -race -c'
-            sh './libpages.test -test.timeout 10s'
+            sh './libpages.test -test.timeout 30s'
         }
     }
     tests[prefix+'libpages_config'] = {
         dir('libpages/config') {
             sh 'go test -race -c'
-            sh './config.test -test.timeout 10s'
+            sh './config.test -test.timeout 30s'
         }
     }
     tests[prefix+'kbpagesconfig'] = {
         dir('kbpagesconfig') {
             sh 'go test -race -c'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -268,97 +268,138 @@ def runNixTest(prefix) {
...
+        sh 'go test -i'
+        sh 'go test -race -c'
+    }
+    dir('libkbfs') {
+        sh 'go test -i'
+        sh 'go test -race -c'
+    }
+    dir('libfuse') {
+        sh 'go test -i'
+        sh 'go test -race -c'
```
jzila: no race here

Jenkinsfile
```diff
@@ -268,97 +268,138 @@ def runNixTest(prefix) {
...
+    }
+    dir('simplefs') {
+        sh 'go test -i'
+        sh 'go test -race -c'
```
jzila: no race here

jzila:
Looks like we've severely slowed down tests :(

let's try a `go test -i ./...` at the root first, then remove the individual ones.

jzila:
@songgao: I meant my "slowed down tests" comment to block merging.

songgao:
Ooops; sorry I'll revert?

jzila:
Wait, I'm still investigating.

jzila:
OK you don't need to revert, but I think we'll need another PR.

PR #1495, opened 2018-02-28
use MinCost for bcrypt in kbpagesconfig test as well

songgao:
This decreases the time for running `kbpagesconfig` tests on my laptop from 0.39s to 0.01s. So hopefully it'll help with the CI flake.

### DIFF
--- a/kbpagesconfig/editor.go
+++ b/kbpagesconfig/editor.go
@@ -29,6 +29,7 @@ type kbpConfigEditor struct {
 	kbpConfig         *config.V1
 	originalConfigStr string
 	prompter          prompter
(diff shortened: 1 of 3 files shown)


songgao:
You mean `DefaultCost` might be too slow for production as well? I was thinking about that too. I did some simple test and it looks like `bcrypt.CompareHashAndPassword` and `bcrypt.GenerateFromPassword` are equally slow (about 0.1s each). So this could be an DoS surface. Not sure what the right answer is here yet though.

songgao:
I'm gonna merge this for now as it affects CI. But I think we should discuss the performance thing.

PR #1496, opened 2018-02-28
redirector: a Go implementation for the root redirector

strib:
This implementation of the root redirector will likely be used for macOS, at least until we figure out the kbfuse+Rust situation.

Note that I needed some new bazil.org/fuse options, so I forked it to github.comkeybase/fuse and added two new features:
* A NoBrowse() option that keeps the redirector from showing up in the Finder (instead, the specific user's mount will show up).
* Adding the requesting user's UID to the context for every request, since `ReadDirAll` needs to know whether the requesting user has a valid mountpoint or not.  The bazil.org/fuse interface for `ReadDirAll` doesn't pass this down, so putting it in the context seemed like the easiest option for now.

Issue: KBFS-2774


redirector/go/main.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+func main() {
+	currUser, err := user.Current()
+	if err != nil {
+		panic(err)
+	}
+	if currUser.Uid != "0" {
+		runtime.LockOSThread()
+		_, _, errNo := syscall.Syscall(syscall.SYS_SETUID, 0, 0, 0)
```
oconnor663: Help me understand what happens if we don't make this call? My Rust redirector doesn't do any SUID step, but if I `chmod 4755` the binary, then I'm able to point it at a regular `0x755 root:root` directory and mount with no problems.

strib: @oconnor663: without this call, if you do `./redirector` as a regular user (without using `sudo`), you get an error.  I.e.:

```sh
$ go build
$ cp go /tmp/redirector
$ cd /tmp/
$ mkdir keybase
$ sudo chown root:root keybase
$ sudo chmod 755 keybase
$ sudo chown root:root redirector
$ sudo chmod 4755 redirector
$ ./redirector /tmp/keybase 
2018/02/28 13:15:29 mount helper error: fusermount: user has no write access to mountpoint /tmp/keybase
Mount error, exiting cleanly: fusermount: exit status 1
```

oconnor663: Oof, why does this work with a Rust binary but not a Go binary? I just rechecked that it does. (I wasn't able to use `/tmp` for everything though, because it's a `nosuid` mount on my system. Hopefully that doesn't matter here.)

oconnor663: Another more trivial Go program seems to work just fine with the suid flag. (I just had it create a file in the current directory. Works in root-owned directories without `sudo`.)

strib: Say what?  Yeah you're right, I just tried your branch and it worked for me (in `/tmp/`).  Stumped.

strib: I guess maybe it has something to do with the difference between linking into the .so file and calling an external mount program?  Maybe the latter doesn't get run as root unless you officially set the uid?

oconnor663: Just tried it with a toy example, and you're right about what happens to child processes with and without setuid. So I believe this theory. Interesting.

oconnor663: More detail: It looks like a setuid binary does _not_ set the actual `uid`, but rather sets the "effective" uid (that is, the value of `geteuid()`). The latter is used to check file access permissions, but it's not inherited by children.

oconnor663: (This is filthy and Unix should be ashamed :-D )

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+			continue
+		}
+		fuseMountPoints = append(fuseMountPoints, v.Path)
+	}
+
+	if len(fuseMountPoints) == 0 {
+		return "", fuse.ENOENT
+	}
+	if len(fuseMountPoints) == 1 {
+		return fuseMountPoints[0], nil
```
jzila: I'm slightly worried about this short circuit: if we somehow have the redirector running but not KBFS, and we have exactly one FUSE mount that isn't KBFS, we'll return it here.

IMO it'd be better to go through the loop below even in the `len(fuseMountPoints) == 1` case.

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		// Double-check that this is a real KBFS mount.
+		if _, err := os.Stat(filepath.Join(mp, ".kbfs_error")); err != nil {
+			continue
+		}
+		return mp, nil
+	}
+
+	// Give up and return the first one.
+	return fuseMountPoints[0], nil
```
jzila: Shouldn't we return `"", fuse.ENOENT`? We've established at this point that none of the FUSE mounts are KBFS.

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	resp.EntryValid = 0
+	switch req.Name {
+	case "private", "public", "team", ".kbfs_error", ".kbfs_metrics",
+		".kbfs_profiles", ".kbfs_reset_caches", ".kbfs_status",
+		"kbfs.error.txt", "kbfs.nologin.txt", ".kbfs_enable_auto_journals",
+		".kbfs_disable_auto_journals", ".kbfs_enable_block_prefetching",
+		".kbfs_disable_block_prefetching", ".kbfs_enable_debug_server",
+		".kbfs_disable_debug_server":
```
jzila: Might be nice to export this set in one of our libraries.

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	c, err := fuse.Mount(os.Args[1], options...)
+	if err != nil {
+		fmt.Printf("Mount error, exiting cleanly: %+v\n", err)
+		os.Exit(0)
+	}
+
+	interruptChan := make(chan os.Signal, 1)
+	signal.Notify(interruptChan, os.Interrupt)
```
jzila: Can we do this on a `SIGTERM` as well?

redirector/go/main.go
oconnor663: Ok what if instead of calling `syscall.Syscall` (which as I understand it is the main thing preventing us from running this implementation on Linux), what if we used some CGO like this?

```go
/*
#include <sys/types.h>
#include <unistd.h>
#include <errno.h>

int setuid_root() {
	int ret = setuid(0);
	if (ret == 0) {
		return 0;
	} else {
		return errno;
	}
}
*/
import "C"

func setUidRoot() error {
	ret := C.setuid_root()
	if ret != 0 {
		return fmt.Errorf("setuid(0) returned non-zero ERRNO: %d", ret)
	}
	return nil
}
```

If that worked, maybe we'd be fine running the Go redirector on both Linux and Mac. I don't think the security situation is really much different between the two languages.

oconnor663:
Added a comment above about CGO.

redirector/go/main.go
strib: For the record, we tried the cgo solution for `keybase-mount-helper` and it apparently didn't work: https://github.com/keybase/client/issues/10487#issuecomment-369709787

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+	// that has "keybase" in the path.
+	sort.Strings(fuseMountPoints)
+	for _, mp := range fuseMountPoints {
+		// TODO: a better regexp that will rule out keybase.staging if
+		// we're in prod mode, etc.
+		if !strings.Contains(mp, "keybase") {
+			continue
+		}
+		// Double-check that this is a real KBFS mount.
+		if _, err := os.Stat(filepath.Join(mp, ".kbfs_error")); err != nil {
```
songgao: Curious why not `.kbfs_status`? Is this one cheaper?

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+
+func (r *root) getCachedMountpoint(uid uint32) string {
+	r.lock.RLock()
+	defer r.lock.RUnlock()
+	entry, ok := r.mountpointCache[uid]
+	if !ok {
+		return ""
+	}
+	now := time.Now()
+	if now.Sub(entry.time) > mountpointTimeout {
```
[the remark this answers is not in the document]
strib: In Go, `time.Sub()` uses the monotonic clock, so wallclock and timezone changes aren't relevant.  From https://golang.org/pkg/time/#hdr-Monotonic_Clocks:
> If Times t and u both contain monotonic clock readings, the operations t.After(u), t.Before(u), t.Equal(u), and t.Sub(u) are carried out using the monotonic clock readings alone, ignoring the wall clock readings.

redirector/go/main.go
songgao: SIGTERM is actually probably more often since that's the default for `kill`/`killall`

redirector/go/main.go
strib: My thought on this, and the one above, is that we're not actually sure it's not KBFS, we just know it wasn't mounted with "keybase" in the name.  I was trying not to make that a requirement.  But you're right, it was kinda half-assed.  We can make the naming an official requirement.

redirector/go/main.go
strib: I didn't want to suck `libfs` and `libkbfs` into this binary, and I didn't feel like making a new one-file library for this.  But if you feel strongly about it, let me know.

vendor/github.com/keybase/gomounts/darwin.go
```diff
@@ -0,0 +1,47 @@
...
+
+func getMountedVolumes() ([]Volume, error) {
+	// getmntinfo is non-reentrant
+	mtx.Lock()
+	defer mtx.Unlock()
+
+	result := make([]Volume, 0)
+
+	var mntbuf *C.struct_statfs
+	count := int(C.getmntinfo(&mntbuf, C.MNT_NOWAIT))
```
akalin-keybase: this code is buggy:

- `getmntinfo` return *0* on an error and sets `errno`.
- would feel much better using the thread-safe version `getmntinfo_r_np` that doesn't shrae a buffer.
- the code casting the array to a slice below is kinda scary -- https://golang.org/pkg/reflect/#SliceHeader says `Data` might still be garbage collected? would much rather just iterate manually somehow.

redirector/go/main.go
strib: Ah, I realize I left the `.kbfs_error` check in this version.  That is wrong; the root user doesn't actually have access to check that in someone else's mount.  So we are forced to go completely from the name.

jzila: Really? The root user doesn't have universal access?

jzila: Is this a mount namespace thing?

strib: You need to give `allow_root` as a mount option, which we don't do, so root can't, say, index your files.

jzila: We could make a redirector for the redirector, which redirects to a mount that root has access to but maps 1:1 to user mounts. Redirects all the way down!

</joking>

vendor/github.com/keybase/gomounts/linux.go
```diff
@@ -0,0 +1,45 @@
...
+	defer C.endmntent(file)
+	var ent *C.struct_mntent
+
+	mntopt := C.CString("user_id")
+	for ent = C.getmntent(file); ent != nil; ent = C.getmntent(file) {
+		mntType := C.GoString(ent.mnt_type)
+		uidstr := ""
+		if substr := C.GoString(C.hasmntopt(ent, mntopt)); len(substr) > 0 {
+			commasplit := strings.SplitN(substr, ",", 2)
+			equalsplit := strings.SplitN(commasplit[0], "=", 2)
```
akalin-keybase: this seems a bit brittle to me, i'd have to test to see if it handles edge cases ok

vendor/github.com/keybase/gomounts/linux.go
```diff
@@ -0,0 +1,45 @@
...
+	defer C.free(unsafe.Pointer(cmode))
+	var file *C.FILE = C.setmntent(cpath, cmode)
+	if file == nil {
+		return nil, errors.New("Unable to open /proc/mounts")
+	}
+	defer C.endmntent(file)
+	var ent *C.struct_mntent
+
+	mntopt := C.CString("user_id")
+	for ent = C.getmntent(file); ent != nil; ent = C.getmntent(file) {
```
akalin-keybase: would prefer `getmntent_r` here, too

vendor/github.com/keybase/gomounts/windows.go
```diff
@@ -0,0 +1,40 @@
...
+import "C"
+
+import (
+	"unsafe"
+)
+
+// Windows implementation
+func getMountedVolumes() ([]Volume, error) {
+	result := make([]Volume, 0)
+	var buf [256]C.char
```
akalin-keybase: should be `MAX_PATH + 1`, which is 261

vendor/github.com/keybase/gomounts/windows.go
```diff
@@ -0,0 +1,40 @@
...
+
+	drives := uint32(C.GetLogicalDrives())
```
akalin-keybase: hmm, isn't it possible for mounts to be places other than `<LETTER>:\`? @zanderz do you know?

akalin-keybase: Not that it matters too much, since we won't be running this on windows

vendor/github.com/keybase/gomounts/linux.go
akalin-keybase: alternatively, we could just look at what `setmntopt` does and mimick its behavior.

redirector/go/main.go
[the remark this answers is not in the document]
jzila: Oh right I forgot they made that change!

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+		// Don't bother deleting the entry, since the caller should
+		// just overwrite it.
+		return ""
+	}
+	return entry.mountpoint
+}
+
+func (r *root) findKBFSMount(ctx context.Context) (
+	mountpoint string, err error) {
+	uid := ctx.Value(fs.CtxHeaderUIDKey).(uint32)
```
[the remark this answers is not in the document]
strib: I actually wanted a crash here if it's not set.  It means it's compiled with a wrong version of bazil and we'd have big problems. I'll add a comment.

redirector/go/main.go
```diff
@@ -0,0 +1,266 @@
...
+	// Don't let the root see anything here; we don't want a symlink
+	// loop back to this mount.
```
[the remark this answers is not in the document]
strib: > Not sure if making a root-started KBFS work with redirector is a goal though.

Not a goal.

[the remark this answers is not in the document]
strib: (The keybase service refuses to even run as root.)

redirector/go/main.go
strib: Yeah much cheaper, but anyway, I removed this.

vendor/github.com/keybase/gomounts/linux.go
strib: @akalin-keybase what is `setmntopt`?  Google shows nothing.

I think this code is pretty straightforward.  `hasmntopt` returns a substring starting at "user_id=", possibly with comma-separated options after it.  So we want the string between the first equal sign and the first comma.

Unless you have specific concerns, I'm not going to spend any more time on this part.

akalin-keybase: sorry, meant `setmntent`

redirector/go/main.go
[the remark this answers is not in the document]
songgao: Got it. Comment is good enough for me!

vendor/github.com/keybase/gomounts/linux.go
strib: I've been struggling with converting this code to `getmntent_r` for a couple hours now, and have failed.  Basically it is very not obvious how to pass a `struct mntent *` to an stack-allocated `C.struct_mntent` into C-land, without rewriting the entire function is C.

Instead, I'm going to give up and just protect all calls to `GetMountedVolumes()` with a global lock in `main.go`.

redirector/go/main.go
```diff
@@ -56,14 +56,35 @@ type cacheEntry struct {
...
+	case "staging":
+		runmodeStr = "keybase.staging"
+	case "devel":
+		runmodeStr = "keybase.devel"
+	}
+
+	// Find mountpoints like "/home/user/.local/share/keybase/fs", or
+	// "/home/user/keybase", and make sure it doesn't match mounts for
+	// another run mode, say "/home/user/keybase.staging".
+	rx, err := regexp.Compile(fmt.Sprintf("/%s/|%s$", runmodeStr))
```
jzila: This `Sprintf` takes 2 strings, but only passes in one.

jzila: Additionally, the runmode string can have a `.` in it, but in regex land that means any character.

strib:
> I'm not a fan of using a regex in a setuid root application. Regexes are notorious sources of security holes and bugs. We should use them sparingly. Can you use strings.IndexOf instead?

Ok done.  I think the Go regexp is pretty safe and the new code is harder to understand though.

vendor/github.com/keybase/gomounts/linux.go
strib: I don't think `setmntent` does anything with options, it just opens a file: https://github.com/seL4/libmuslc/blob/master/src/linux/mntent.c#L8-L10

I'm going to leave this as is for now until we discover an issue with it.

jzila:
Thanks! I agree that the check is a bit harder to read but I still think it's lower risk.

strib:
Ok I think I've addressed all the comments I'm qualified to address.  I'll merge when CI passes, but if there are follow-on comments I'll fix them in a separate PR.  Thanks all!

PR #1501, opened 2018-03-05
redirector: restrict mountpoints to /keybase*

strib:
Since this is a suid binary, allowing arbitrary mountpoints is dangerous.

Suggested by oconnor663.

### DIFF
--- a/redirector/go/main.go
+++ b/redirector/go/main.go
@@ -245,14 +245,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Restrict the mountpoint to paths starting with "/keybase" (just
-	// check a prefix, to include things like "/keybase.staging").
+	// Restrict the mountpoint to paths starting with "/keybase".
 	// Since this is a suid binary, it is dangerous to allow arbitrary
 	// mountpoints.  TODO: Read a redirector mountpoint from a
 	// root-owned config file.
-	if !strings.HasPrefix(os.Args[1], "/keybase") {
+	r := newRoot()
+	if os.Args[1] != fmt.Sprintf("/%s", r.runmodeStr) {
 		fmt.Fprintf(os.Stderr, "ERROR: The redirector may only mount at "+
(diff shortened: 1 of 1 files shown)


strib:
> It might be worth being exceptionally conservative and hardcoding the list of keybase/keybase.staging/keybase.devel. We can always change our mind later, right?

Ok done.

> Do you know what happens if the mount target is a symlink? If basil-fuse is willing to traverse a symlink to mount, we might want to be extra conservative and try to prevent it from doing that?

When I try this, I get a nonempty-mount error:

```sh
$ mkdir /tmp/blah
$ sudo ln -s /tmp/blah /keybase.staging
$ KEYBASE_RUN_MODE=staging ./redirector /keybase.staging
2018/03/05 14:01:56 mount helper error: fusermount: mountpoint is not empty
2018/03/05 14:01:56 mount helper error: fusermount: if you are sure this is safe, use the 'nonempty' mount option
Mount error, exiting cleanly: fusermount: exit status 1
``` 

So it seems like it should be safe.

redirector/go/main.go
```diff
@@ -240,6 +240,22 @@ func (r *root) Lookup(
...
+		fmt.Fprintf(os.Stderr, "Usage: %s <mountpoint>\n", os.Args[0])
+		os.Exit(1)
+	}
+
+	// Restrict the mountpoint to paths starting with "/keybase".
+	// Since this is a suid binary, it is dangerous to allow arbitrary
+	// mountpoints.  TODO: Read a redirector mountpoint from a
+	// root-owned config file.
+	r := newRoot()
+	if os.Args[1] != fmt.Sprintf("/%s", r.runmodeStr) {
```
jzila: Maybe this should be case insensitive.

strib: @oconnor663 wanted this to be as restrictive as possible.  We can always change it later if we need   to.

jzila: Understood, thanks.

PR #1502, opened 2018-03-06
redirector: turn off local caches on osx

strib:
Otherwise, the kernel will cache symlinks for a long time.

### DIFF
--- a/redirector/go/main.go
+++ b/redirector/go/main.go
@@ -276,6 +276,8 @@ func main() {
 		options = append(options, fuse.OSXFUSELocations(kbfusePath))
 		options = append(options, fuse.VolumeName("keybase-redirector"))
 		options = append(options, fuse.NoBrowse())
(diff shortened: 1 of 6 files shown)


strib:
> I thought this could was controlled by the Valid field. Is this approach just a simpler way to disable caching globally for the mount, or is it a different thing?

Yeah I wish that worked, but my testing on macOS shows that it doesn't work for symlinks.  Not sure why.

songgao:
Got it; thanks for the background.

On the other hand, now that we have this, do we still have to override `Valid` field to `0` or does it not matter anymore? (not proposing a change; just making sure)

strib:
> On the other hand, now that we have this, do we still have to override Valid field to 0 or does it not matter anymore? (not proposing a change; just making sure)

We wouldn't need it for osx probably, but we still want it for linux (which doesn't have the `nolocalcaches` option, as far as I know).

PR #1503, opened 2018-03-06
(3-liner) enable simplefs in kbpagesd

songgao:


songgao:
@jzila OK with number-of-log-files change too?

jzila:
Sure

PR #1506, opened 2018-03-07
redirector: allow new "fancy" mount names

strib:
This will let us redirect to something like `/Volumes/strib's Keybase` on macOS.

I also moved the redirector from `redirector/go/main.go` to just `redirector/main.go`, since the Rust impl will probably never be checked in.

Issue: KBFS-2814


redirector/main.go
```diff
@@ -176,10 +181,15 @@ func (r *root) findKBFSMount(ctx context.Context) (
...
+		str := r.runmodeStr
+		if i < 0 {
+			i = strings.Index(mp, r.runmodeStrFancy)
+			str = r.runmodeStrFancy
+		}
 		if i < 0 {
 			continue
 		}
-		if len(mp) > i+len(r.runmodeStr) && mp[i+len(r.runmodeStr)] != '/' {
+		if len(mp) > i+len(str) && mp[i+len(str)] != '/' {
```
songgao: The second condition is specifically for `/home/user/.local/share/keybase/fs` I suppose?

strib: Yes, or just generally to be flexible.

strib:
> For matching mount points, I probably have missed or didn't pay attention to some discussion on this, but would having a list of suffixes and iterating on them using HasSuffix work? That seems to be easier to read. I'm OK with this too though.

That doesn't work because of the `~/.local/share/keybase/fs` and `/run/user/$UID/keybase/kbfs` cases.

songgao:
> That doesn't work because of the ~/.local/share/keybase/fs and /run/user/$UID/keybase/kbfs cases.

Hmm why doesn't it work? In the approach I was imagining we'd have a list of suffixes, e.g.:

```
suffixes := []string{
    "/keybase/fs"
    "/keybase/kbfs"
    "/keybase"
}
```

And we'd just iterate through them. I don't think this'd be slower. Am I missing something?

songgao:
> so it would be more likely to match user-created mountdirs that they put into config.json.

Oh I see. I didn’t think of custom mounts. Makes sense. Thanks!

PR #1508, opened 2018-03-08
libkbfs: introduce a new InitMode interface, and a new "constrained" mode for mobile KBFS

strib:
This cleans up all the scattered mode cruft into an interface, and puts all the behaviors of each mode into a single struct for easier reference and maintenance.

It also adds a new "constrained" mode, with behavior matching what jzila wrote in #1486.

Issue: KBFS-2744

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -664,11 +664,6 @@ type RekeyResult struct {
 type InitModeType int
 
 const (
-	// InitModeMask masks out mode flags.
-	InitModeMask InitModeType = 0xffff
-	// InitTest is a mode flag that represents whether we're running in a test.
-	InitTest InitModeType = 1 << 16
-
 	// InitDefault is the normal mode for when KBFS data will be read
 	// and written.
 	InitDefault InitModeType = iota
@@ -685,18 +680,8 @@ const (
 	InitConstrained
 )
 
-// Mode returns the mode absent any mode flags.
-func (im InitModeType) Mode() InitModeType {
-	return im & InitModeMask
-}
-
-// HasFlags returns whether all the specified flags are set.
-func (im InitModeType) HasFlags(flags InitModeType) bool {
-	return im&flags > 0
-}
-
 func (im InitModeType) String() string {
-	switch im.Mode() {
+	switch im {
 	case InitDefault:
 		return InitDefaultString
 	case InitMinimal:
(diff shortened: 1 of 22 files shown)


libkbfs/modes.go
```diff
@@ -0,0 +1,310 @@
...
+	case InitConstrained:
+		return modeConstrained{modeDefault{t}}
+	default:
+		panic(fmt.Sprintf("Unknown mode: %s", t))
+	}
+}
+
+// Default mode:
+
+type modeDefault struct {
```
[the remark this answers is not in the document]
songgao: I like the current approach, and agree it's cleaner than having separate fields.

libkbfs/modes.go
```diff
@@ -0,0 +1,310 @@
...
+type modeConstrained struct {
+	InitMode
+}
+
+func (mc modeConstrained) Type() InitModeType {
+	return InitConstrained
+}
+
+func (mc modeConstrained) BlockWorkers() int {
+	return 1
```
songgao: I know this is from @jzila 's temporary PR, but now that we are formalize it, do we maybe want to use a larger number here? Especially considering mobile connections probably have higher latency.

libkbfs/modes.go
```diff
@@ -0,0 +1,310 @@
...
+
+package libkbfs
+
+import (
+	"fmt"
+
+	"github.com/keybase/client/go/protocol/keybase1"
+)
+
+// NewInitModeFromType returns an InitMode object corresponding to the
```
songgao: Did you intend to have this at beginning of the file?

libkbfs/data_types.go
```diff
@@ -680,26 +680,31 @@ const (
...
 
 // Mode returns the mode absent any mode flags.
-func (im InitMode) Mode() InitMode {
+func (im InitModeType) Mode() InitModeType {
 	return im & InitModeMask
 }
 
 // HasFlags returns whether all the specified flags are set.
-func (im InitMode) HasFlags(flags InitMode) bool {
+func (im InitModeType) HasFlags(flags InitModeType) bool {
```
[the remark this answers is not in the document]
jzila: I think there are still useful methods like `String()` and `Mode()` to pull out the flags. But I agree that `HasFlags` can go away.

libkbfs/modes.go
jzila: We can start with 1 and change it later to optimize.

libkbfs/modes.go
strib: Yeah, I kind of consider it a constructor, which I usually put at the top for easy access.  But I could be convinced otherwise.

PR #1509, opened 2018-03-10
don't error on InputCanceledError

songgao:
Otherwise we'd fall into this case here: https://github.com/keybase/kbfs/blob/2cc8f8832ed876f51044ddefe7edb5e57bef5432/libkbfs/rekey_fsm.go#L386-L395 and schedule another rekey in `rekeyRecheckInterval` which is 30s. With the added error check in this PR, we'd continue with a rekey scheduled in `r.fsm.fbo.config.RekeyWithPromptWaitTime()` which is 10min for now.


songgao:
This error was introduced 2 years ago I think, so it could be we just missed it in the first place.

PR #1514, opened 2018-03-15
update rpc vendor; fixes keybase/client#10172

songgao:
Upstream PRs:

https://github.com/keybase/go-framed-msgpack-rpc/pull/136

https://github.com/keybase/go-framed-msgpack-rpc/pull/137


songgao:
Ugh it looks like this might have broke some SBS tests. I'll look tomorrow.

jzila:
@songgao: seems like we should vendor this into client as well.

songgao:
@jzila I have a PR: https://github.com/keybase/client/pull/10906 that errored on CI as well. I haven't looked into the CI errors, but I'm not sure if merging them both would make the error go away, as it's not a build error.

PR #1515, opened 2018-03-15
redirector: handle mountpoints like  `/Volumes/Keybase (user)`

strib:
So instead of checking that "Keybase" is necessarily the last thing in the path, maybe sure it's at least followed by a space.

Issue: KBFS-2820

### DIFF
--- a/redirector/main.go
+++ b/redirector/main.go
@@ -189,7 +189,8 @@ func (r *root) findKBFSMount(ctx context.Context) (
 		if i < 0 {
 			continue
 		}
-		if len(mp) > i+len(str) && mp[i+len(str)] != '/' {
+		if len(mp) > i+len(str) && mp[i+len(str)] != '/' &&
+			mp[i+len(str)] != ' ' {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


redirector/main.go
```diff
@@ -189,7 +189,8 @@ func (r *root) findKBFSMount(ctx context.Context) (
 		if i < 0 {
 			continue
 		}
-		if len(mp) > i+len(str) && mp[i+len(str)] != '/' {
+		if len(mp) > i+len(str) && mp[i+len(str)] != '/' &&
+			mp[i+len(str)] != ' ' {
```
jzila: It seems like the required conditions for such a mount point are:
* Has the `keybase` string, adjusted for runMode and
* Is equal to the length of the runmode string or
  * or the `keybase` string ends in `'/'`
  * or the `keybase` string is followed by a `' '`

This seems fine, but maybe the condition should be explicit rather than negative.

strib: That seems like the right summary.  You're suggested I should invert the logic for the `if` statement and `return` from in there, instead of `continue`?  I have no preference.

jzila: Yep that's what I'm suggesting.

strib: Ok done.

PR #1516, opened 2018-03-16
 simplefs: report detailed progress for async operations

strib:
`SimpleFSCheck` now reports more details about each async op:

* The time it "started", which is actually the time we finished calculating how much data we will be dealing with.
* An estimate for when we think the operation will finish, based on how much data has been read so far.
* The total number of bytes we will need to transfer for the operation, plus the number read so far and the number written so far (these might be different for copies).
* The total number of files we will need to transfer for the operation, plus the number read so far and the number written so far (these might be different for copies).

Depends on keybase/client#10940, should be merged together.  Will break CircleCI until then.

Issue: KBFS-2799


simplefs/simplefs.go
```diff
@@ -430,6 +523,37 @@ func (k *SimpleFS) SimpleFSReadList(_ context.Context, opid keybase1.OpID) (keyb
 	return lr, nil
 }
 
+func recursiveByteAndFileCount(fs billy.Filesystem) (
```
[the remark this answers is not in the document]
jzila: It's still not easy. As I mentioned yesterday, we'll need to crawl the full subtree to get folder size, so it's not a cheap operation. When you're copying something recursively, then you already have to do that so it's not as big a deal.

We'll need a ticket to cache a `RecursiveSize` field in `DirEntry`.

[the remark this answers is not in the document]
songgao: We don't have to do that automatically. We can only do it as user clicks a button or something, so it's not like it slows down the app unnecessarily.

> We'll need a ticket to cache a RecursiveSize field in DirEntry.

For discussion purposes and not near future planning: do you mean cache it in the directory block? Do we update it as we update anything under the directory, or only lazily update it as it's requested?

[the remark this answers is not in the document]
jzila: I mean `DirEntry`, but directory block would work too (and would be less disruptive to folder capacity). Yes, we update these totals as we crawl the tree on any writes, since we have to crawl the tree at that point anyway.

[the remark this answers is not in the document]
songgao: I see. That makes sense. Sounds like a good idea!

strib:
@jzila: we decided to merge this before the end of the day, so we can monitor what happens with any circle CI / build failures.  But I'll be happy to address any feedback in a separate PR, if you have any.

simplefs/simplefs.go
```diff
@@ -950,15 +1173,42 @@ func (k *SimpleFS) SimpleFSCancel(_ context.Context, opid keybase1.OpID) error {
...
+		progress := p.progress
+		if progress.BytesTotal > 0 {
+			n = progress.BytesRead
+			d = progress.BytesTotal
+		} else if p.progress.FilesTotal > 0 {
+			n = progress.FilesRead
+			d = progress.FilesTotal
+		}
+		if n > 0 && d > 0 && !progress.Start.IsZero() &&
+			progress.EndEstimate.IsZero() {
```
songgao: Perhaps we should recompute `EndEstimate` on every `SimpleFSCheck` call instead of only first time?

simplefs/simplefs.go
```diff
@@ -950,15 +1173,42 @@ func (k *SimpleFS) SimpleFSCancel(_ context.Context, opid keybase1.OpID) error {
 // SimpleFSCheck - Check progress of pending operation
 // Progress variable is still TBD.
 // Return errNoResult if no operation found.
-func (k *SimpleFS) SimpleFSCheck(_ context.Context, opid keybase1.OpID) (keybase1.Progress, error) {
+func (k *SimpleFS) SimpleFSCheck(_ context.Context, opid keybase1.OpID) (keybase1.OpProgress, error) {
 	k.lock.RLock()
 	defer k.lock.RUnlock()
```
jzila: Won't this read lock cause the read/write loops to block on updating the progress? We could slow down transfers a bit. I'm wondering if a channel approach would be better.

simplefs/simplefs_test.go
```diff
@@ -423,3 +429,183 @@ func readRemoteFile(ctx context.Context, t *testing.T, sfs *SimpleFS, path keyba
...
+	require.NoError(t, err)
+	checkPendingOp(
+		ctx, t, sfs, opid, keybase1.AsyncOps_COPY, path1, path2, true)
+
+	t.Log("Wait for the first mkdir")
+	select {
+	case <-waitCh:
+	case <-ctx.Done():
+		t.Fatal(ctx.Err())
+	}
```
jzila: Can you make a helper function for these `select`s?

simplefs/simplefs.go
strib: @songgao: we do recompute it every time, until the operation ends.  `progress` is a copy of `p.progress`, which isn't a pointer, but just a regular `keybase1.OpProgress`.

simplefs/simplefs.go
```diff
@@ -430,6 +519,37 @@ func (k *SimpleFS) SimpleFSReadList(_ context.Context, opid keybase1.OpID) (keyb
...
+	for _, fi := range fileInfos {
+		if fi.IsDir() {
+			if fi.Name() == "." {
+				continue
+			}
+			chrootFS, err := fs.Chroot(fi.Name())
+			if err != nil {
+				return 0, 0, err
+			}
+			var chrootBytes int64
```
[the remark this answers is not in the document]
strib: I have no idea, I'll fix it.  This was adapted from git code, so it probably made sense before I changed it around a bit.

simplefs/simplefs.go
strib: I can't imagine it would matter to perceived performance, given the speed of KBFS uploads/downloads and the fact that this will be called maybe a few times a second at most and doesn't block on anything network-related, but if it becomes a problem, sure we can make a goroutine-based channel thingy.  I think that's overkill for now though.

simplefs/simplefs.go
songgao: I see; makes sense. It must be the bug I had earlier in my local branch that made me think this was not recomputed.

Why the `progress.EndEstimate.IsZero()` check though?

strib: That's checking if the operation has ended, and if `EndEstimate` was set in the actual map.

songgao: Oh I see this is for when after `doneOp` is called. NVM. Thanks!

songgao: Ugh! GitHub again failed to show me the new comment before I submitted my comment.

PR #1518, opened 2018-03-16
simplefs: return the last writer of a directory entry

strib:
Get this info by using the `os.FileInfo.Sys()` generic interface, and introducing a new type of interface that the `Sys()` return value can implement.  That way we don't try to get the last writer when using a local disk file system.

Note: this won't return the last writer for symlink directory entries. I assume that's ok, but if not we can figure out a different way of doing this.

This will break CircleCI until keybase/client#10948 is merged.

Issue: KBFS-2836

### DIFF
--- a/libfs/file_info.go
+++ b/libfs/file_info.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/kbfs/libkbfs"
 )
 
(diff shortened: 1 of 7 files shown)


libfs/file_info.go
```diff
@@ -65,7 +66,51 @@ func (fi *FileInfo) IsDir() bool {
...
+		// TODO: if we want symlink last writers, we'll need to add a
+		// new interface to KBFSOps to get them.
+		return keybase1.User{}, nil
+	}
+	md, err := fis.fi.fs.config.KBFSOps().GetNodeMetadata(
+		fis.fi.fs.ctx, fis.fi.node)
+	if err != nil {
+		return keybase1.User{}, err
+	}
+	lastWriterName := md.LastWriterUnverified
```
songgao: Is there a way to get verified writer, or would that be too much plumbing?

strib: > Is there a way to get verified writer, or would that be too much plumbing?

It would require doing something with the Merkle tree and the MD history that we haven't fully figured out yet.  There's no easy way to map a particular file back to its last-modified MD revision.

PR #1523, opened 2018-03-19
simplefs: support new debug-dumping RPC

strib:
Uses a new instance of the impatient debug dumper, without any background goroutines, to make this happen.

Issue: KBFS-2746

### DIFF
--- a/libkbfs/impatient_debug_dumper.go
+++ b/libkbfs/impatient_debug_dumper.go
@@ -119,7 +119,20 @@ func NewImpatientDebugDumper(config Config, dumpIn time.Duration) *ImpatientDebu
 	return d
 }
 
-func (d *ImpatientDebugDumper) dump(tracker *ctxTimeTracker) {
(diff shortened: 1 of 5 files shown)


libkbfs/impatient_debug_dumper.go
```diff
@@ -119,7 +119,20 @@ func NewImpatientDebugDumper(config Config, dumpIn time.Duration) *ImpatientDebu
 	return d
 }
 
-func (d *ImpatientDebugDumper) dump(tracker *ctxTimeTracker) {
+// NewImpatientDebugDumperForForcedDumps creates a new
+// *ImpatientDebugDumper, just as above, though without launching any
+// background goroutines or allowing the user to begin any
+// time-tracked tasks
```
songgao: Missing period. Also, with this dumper, is calling `ForceDump` the only way to trigger a dump? If so would you mind adding that to comment?

PR #1526, opened 2018-03-20
kbfsgit: don't send MD updates for pushes that errored

strib:
Issue: KBFS-2717

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -1694,10 +1694,28 @@ func (r *runner) handlePushBatch(ctx context.Context, args [][]string) (
 		_, err = r.output.Write([]byte(result + "\n"))
 	}
 
-	err = libgit.UpdateRepoMD(ctx, r.config, r.h, fs,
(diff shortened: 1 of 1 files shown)


kbfsgit/runner.go
```diff
@@ -1694,10 +1694,28 @@ func (r *runner) handlePushBatch(ctx context.Context, args [][]string) (
...
+		refStr := refspec.String()
+		start := strings.Index(refStr, ":") + 1
+		dst := refStr[start:]
+
+		if results[dst] != nil {
+			r.log.CDebugf(
+				ctx, "Removing commit result for errored push on refspec %s",
+				refspec)
+			refName := plumbing.ReferenceName(refspec.Src())
+			delete(commits, refName)
```
songgao: `commits` is returned by the function too. Is it OK to not include the errorred ones in returned `commits` either?

jzila: Yeah that's pretty much the idea.

songgao: Oh I see 😅

PR #1528, opened 2018-03-21
tlf_handle_resolve: use the team suffix when checking canonicals

strib:
When someone passes a TLF name like "a,b,c (conflicted copy 2018-03-22 #1)" to FUSE, we need to figure out if it's canonical, but it's hard, because the "#1" is needed for implicit teams, but during the initial parsing we don't know yet whether the folder is team-backed or not without making expensive service calls.

So this PR does makes the right suffix when it knows for sure about the team-backed-status of the TLF; otherwise, if the user passed in a "#1", it assumes the user (likely a KBFS server test) already knows that it should be an implicit team TLF.

Also it wrap the TLFNameNotCanonical error with a stack, because this was really hard to debug without that.

Issue: KBFS-2652

### DIFF
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -7,7 +7,6 @@ package libkbfs
 // This file has the online resolving functionality for TlfHandles.
 
 import (
-	"errors"
 	"fmt"
 	"reflect"
 	"sort"
@@ -20,6 +19,7 @@ import (
 	"github.com/keybase/kbfs/kbfscodec"
 	"github.com/keybase/kbfs/kbfsmd"
 	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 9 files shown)


libkbfs/tlf_handle_resolve.go
```diff
@@ -769,12 +776,33 @@ func parseTlfHandleLoose(
...
+		var canonExtensionString string
+		// If this resolve is being done in a "quick" way that doesn't
+		// check for implicit teams, we might not know if this handle
+		// is backed by a team or not.  But (mostly for tests) we need
+		// to make sure the suffix reflects the right format,
+		// otherwise we'll get too many levels of non-canonical
+		// redirects.  So if the user explicitly specified a "#1" in
+		// their suffix, let's keep it there and assume the user knows
+		// what they're doing.
+		if h.IsBackedByTeam() || strings.Contains(extensionSuffix, "#1") {
```
jzila: Why is this whitelisting `#1`? What about `#2`? `TestSBSConflicts` generates more than one conflict for each TLF.

strib: The whole issue is that, historically, we have not shown `#1` for non-team conflict TLFs, only showing the number starting at `#2`.  But max wanted it to be simpler for teams, so we always show the number now, even `#1`.  Ideally we would just switch non-team TLFs to do the same thing, but we can't because that old string is saved in places like chats and needs to remain the same going forward.  So this whole PR is just about dealing with the `#1` discrepancy, and making it show up only when it is supposed to.

jzila: OK I think I follow now. What is supposed to happen if the user passed a `#1` but it's not an implicit team?

strib: Ideally, it would give back a non-canonical name error leading it to the one without the "#1".  But this PR breaks that behavior.  I don't really think it's worth worrying about though, as I'm not sure why anyone would ever make a folder name like that -- they'd have to type it in manually.

PR #1535, opened 2018-03-29
Be consistent with official docs in systemd service file

kgizdov:
Service file `kbfs.service` uses the keyword `mountdir` to get configured mount point from user configuration and mount keybase FS. However, the official documentation say that users should set the entry for keyword `mountpoint` instead. Thus users are left with bad configuration until they manually dig this up and fix it themselves.

### DIFF
--- a/packaging/linux/systemd/kbfs.service
+++ b/packaging/linux/systemd/kbfs.service
@@ -9,7 +9,7 @@ Type=notify
 # means that error codes from this command are ignored. Without this line,
 # `systemctl --user restart kbfs.service` will hit mount failures if there
 # are any running shells cd'd into a Keybase folder.
-ExecStartPre=-/bin/sh -c 'fusermount -uz "$(keybase config get mountdir|sed s/\x5c"//g)"'
+ExecStartPre=-/bin/sh -c 'fusermount -uz "$(keybase config get mountpoint|sed s/\x5c"//g)"'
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Pretty sure it is the docs which are wrong, e.g.

https://github.com/keybase/client/blob/v1.0.46/packaging/linux/run_keybase#L13

eli-schwartz:
Like I thought: https://github.com/keybase/client/blob/607fc79192221b8eb1a9e770f609f5dacd2d5b9d/go/libkb/config.go#L798

strib:
> However, the official documentation say that users should set the entry for keyword mountpoint instead. 

@kgizdov which "official documentation" are you referring to?  https://keybase.io/docs/kbfs/understanding_kbfs correctly says `mountdir`.

eli-schwartz:
No idea but FWIW this came together with a bugreport to Arch which claims the service file is also somehow wrong because fusermount is reporting no mountpoint in /etc/mtab (works for me, pretty sure fusermount is not broken, guessing this is just the case where ExecStartPre is permitted to fail because nothing needed to be unmounted).

https://bugs.archlinux.org/task/58032

kgizdov:
@strib I swear that page said `mountpoint` when I looked at it before Easter. I guess this doesn't apply anymore.

PR #1536, opened 2018-04-03
Pages: don't error on broken proofs

songgao:
Please see #kbfs on kbchat; this might not be needed at all. But I already have it so am submitting it for review for now in case we still want it in.

Depends on: https://github.com/keybase/client/pull/11188


libkbfs/keybase_daemon_local.go
```diff
@@ -256,7 +256,17 @@ func (k *KeybaseDaemonLocal) Resolve(ctx context.Context, assertion string) (
...
+	name libkb.NormalizedUsername, utid keybase1.UserOrTeamID, err error) {
+	defer func() {
+		var asUser keybase1.UID
+		asUser, err = utid.AsUser()
+		if err != nil {
+			return
+		}
+		// This is required for every identify call. The userBreak function
+		// will take care of checking if third arg (breaks) is nil or not.
+		getExtendedIdentify(ctx).userBreak(name, asUser, nil)
```
songgao: @strib @jzila This was needed to make tests not hang. `userBreak` has to be called on every identify. We were already doing that in the real thing, but not in `KeybaseDaemonLocal` yet, which is used in test. Please let me know if you guys see a problem here.

songgao: (It's only a problem when there's an `extendedIdentify` object attached to `ctx`, which is why it didn't cause any problems for tests before this PR.)

songgao:
Closing following Patrick's comment from client: https://github.com/keybase/client/pull/11188#pullrequestreview-110089587

PR #1540, opened 2018-04-06
libfuse: assign constant inodes to each node

strib:
Otherwise, bazil.org/fuse assigns inodes based on hash(parentInode, name), and if a file gets moved it can end up having the same inode as a new node that took its old name.

This instead has the *FS object keep a monotonically-increasing uint64 and assigns a new inode to every new node, excepting those special files and aliases that can't actually be renamed and cause problems.

Issue: KBFS-2853

### DIFF
--- a/libfuse/fs.go
+++ b/libfuse/fs.go
@@ -146,11 +146,11 @@ func NewFS(config libkbfs.Config, conn *fuse.Conn, debug bool, platformParams Pl
 	return fs
 }
 
-func (fs *FS) assignInode() uint64 {
-	fs.inodeLock.Lock()
-	defer fs.inodeLock.Unlock()
-	next := fs.nextInode
-	fs.nextInode++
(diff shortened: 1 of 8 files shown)


libfuse/fs.go
```diff
@@ -117,28 +120,40 @@ func NewFS(config libkbfs.Config, conn *fuse.Conn, debug bool, platformParams Pl
...
 		time.AfterFunc(d, f)
 	}
 	return fs
 }
 
+func (f *FS) assignInode() uint64 {
+	f.inodeLock.Lock()
+	defer f.inodeLock.Unlock()
+	next := f.nextInode
+	f.nextInode++
```
songgao: It seems this is simple enough to use `atomic.AddUint64` without introducing mental overhead. But I guess assigning new `inode` isn't that frequent either.

PR #1544, opened 2018-04-10
Pages: add a cleartext cache for password and a rate limiter for bcrypt

songgao:
As suggested by @maxtaco a month ago, this adds a cleartext cache for passwords for the HTTP Basic Auth. Also added a per-site rate limiter for bcrypt.

### DIFF
--- a/libpages/config/config_v1.go
+++ b/libpages/config/config_v1.go
@@ -5,12 +5,15 @@
 package config
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"strings"
 	"sync"
+	"time"
 
 	"golang.org/x/crypto/bcrypt"
+	"golang.org/x/time/rate"
 )
 
 const (
@@ -47,6 +50,13 @@ type V1 struct {
 	// users should be authenticated.
 	Users map[string]string `json:"users"`
 
+	cleartextPasswordCacheLock sync.RWMutex
+	// [username -> password] map. Only available when a user has been
+	// authenticated.
+	cleartextPasswordCache map[string]string
+
+	bcryptLimiter *rate.Limiter
+
 	// ACLs is a path -> AccessControlV1 map that defines ACLs for different
 	// paths.
 	ACLs map[string]AccessControlV1 `json:"acls"`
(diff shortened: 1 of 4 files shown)


libpages/config/config_v1.go
```diff
@@ -92,13 +106,43 @@ func (c *V1) Version() Version {
...
+
 // Authenticate implements the Config interface.
-func (c *V1) Authenticate(username, password string) bool {
+func (c *V1) Authenticate(ctx context.Context, username, cleartextPassword string) bool {
+	if c.EnsureInit() != nil {
+		return false
+	}
+
+	if cleartext, ok := c.getCachedCleartextPassword(username); ok {
+		return cleartext == cleartextPassword
```
jzila: Shouldn't the `false` case here actually be treated as a cache miss, and so we should bcrypt it?

libpages/config/config_v1.go
```diff
@@ -92,13 +106,43 @@ func (c *V1) Version() Version {
 	return Version1
 }
 
+func (c *V1) getCachedCleartextPassword(username string) (cleartext string, ok bool) {
+	c.cleartextPasswordCacheLock.RLock()
+	defer c.cleartextPasswordCacheLock.RUnlock()
+	cleartext, ok = c.cleartextPasswordCache[username]
```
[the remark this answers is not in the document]
songgao: Yeah seems like so: https://play.golang.org/p/O-fTcR6oM_p

libpages/config/config_v1.go
```diff
@@ -75,15 +85,19 @@ func DefaultV1() *V1 {
 	return v1
 }
 
-func (c *V1) initACLChecker() {
+const bcryptRateLimitInterval = time.Second / 2
+
+func (c *V1) init() {
+	c.bcryptLimiter = rate.NewLimiter(rate.Every(bcryptRateLimitInterval), 1)
+	c.cleartextPasswordCache = make(map[string]string)
```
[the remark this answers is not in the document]
songgao: Yup; the entire config gets blown away at upper layer: https://github.com/keybase/kbfs/blob/a83f2423ebe9e518254ffabe131f9adfcc9b6c0a/libpages/site.go#L28

libpages/config/config_v1.go
jzila: What if the password changes?

songgao: The entire config object gets blown away when it expires, so bcrypt will run again. It's currently 16s but I'm thinking increase it to 2m.

songgao:
@jzila Are you OK with merging this?

songgao:
> The more I think about it, we probably shouldn't have this auth stuff at all.

@jzila Agreed. Perhaps we should remove the auth part from the docs as well.

PR #1546, opened 2018-04-12
redirector: fix volume and favorite names in macOS

strib:
This shows up when browsing the root disk of the local computer in Finder, and it's confusing if it says "keybase-redirector". "keybase-redirector" will still show up in `mount` though.

Also, the favorite was showing up as "Keybase (strib)" instead of just "Keybase".  This fixes that as well.

cc: @mlsteele 

Issue: KBFS-2866

### DIFF
--- a/libfuse/mounter.go
+++ b/libfuse/mounter.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path"
 	"runtime"
+	"strings"
 
 	"bazil.org/fuse"
 	"github.com/keybase/client/go/libkb"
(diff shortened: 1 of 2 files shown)


libfuse/mounter.go
```diff
@@ -97,12 +98,16 @@ func (m *mounter) Unmount() (err error) {
...
-// volumeName returns the directory (base) name
+// volumeName returns the first word of the directory (base) name
 func volumeName(dir string) (string, error) {
 	volName := path.Base(dir)
 	if volName == "." || volName == "/" {
 		err := fmt.Errorf("Bad volume name: %v", volName)
 		return "", err
 	}
-	return volName, nil
+	s := strings.Split(volName, " ")
```
songgao: There's also `strings.Fields()`

PR #1547, opened 2018-04-12
libkbfs: implement migrating TLFs to implicit team TLFs

strib:
When the service is ready to migrate a TLF to an implicit team, it can call `FinalizeMigration`.  This re-resolves the handle of a TLF using the implicit team, and then generates a new MD for the TLF using team-keying.

When setting the MD successor, we check that the migration is valid: that is, the new implicit team has the same writers, readers, and canonical name as the original MD.

Issue: KBFS-2606

### DIFF
--- a/libkbfs/root_metadata.go
+++ b/libkbfs/root_metadata.go
@@ -285,6 +285,25 @@ func (md *RootMetadata) MakeSuccessor(
 	return newMd, nil
 }
 
+// MakeSuccessorWithNewHandle does the same thing as MakeSuccessor,
+// plus it changes the handle.  (The caller is responsible for
+// ensuring that the handle change is valid.)
+func (md *RootMetadata) MakeSuccessorWithNewHandle(
+	ctx context.Context, newHandle *TlfHandle, latestMDVer kbfsmd.MetadataVer,
+	codec kbfscodec.Codec, keyManager KeyManager, merkleGetter merkleRootGetter,
+	teamKeyer teamKeysGetter, mdID kbfsmd.ID, isWriter bool) (
+	*RootMetadata, error) {
+	mdCopy, err := md.deepCopy(codec)
+	if err != nil {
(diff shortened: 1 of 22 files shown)


libkbfs/root_metadata.go
```diff
@@ -285,6 +285,25 @@ func (md *RootMetadata) MakeSuccessor(
...
+	ctx context.Context, newHandle *TlfHandle, latestMDVer kbfsmd.MetadataVer,
+	codec kbfscodec.Codec, keyManager KeyManager, merkleGetter merkleRootGetter,
+	teamKeyer teamKeysGetter, mdID kbfsmd.ID, isWriter bool) (
+	*RootMetadata, error) {
+	mdCopy, err := md.deepCopy(codec)
+	if err != nil {
+		return nil, err
+	}
+
+	mdCopy.tlfHandle = newHandle.deepCopy()
```
strib: On the way home I realized this isn't enough, I also need to fix up `md.bareMd` based on this handle.  The new code should be pretty isolated and probably reviewable on its own, so feel free to review this as is first and then the new commits when I finish them tomorrow, or feel free to wait til it's all ready.

strib: Ok I fixed it, ready to review in full again.

tlf/handle.go
```diff
@@ -414,3 +414,27 @@ func (h *Handle) DeepEqual(other Handle) bool {
...
+	// care if `b` matches the exact number of duplicates).
+	aMap := make(map[keybase1.UserOrTeamID]bool)
+	for _, u := range a {
+		aMap[u] = true
+	}
+	for _, u := range b {
+		if !aMap[u] {
+			return false
+		}
+		delete(aMap, u)
```
songgao: It looks like we are assuming `b` doesn't contain any duplicate (contradictory to the comment above)?

PR #1548, opened 2018-04-13
kbfsgit: update go-git to avoid failing on conflict files

strib:
If conflict files ended up in `objects` or `objects/pack`, go-git would still try to treat their full names as hashes, and end up trying to access a file named by the 0 hash, which would fail.  Instead, just skip files like that.

Issue: keybase/client#11366


strib:
> What if the full name happens to be a valid hex, or is that not possible?

If the full name of some random file put in the directory happens to exactly match the format for packfile names, go-git (and command line git too, I assume) would try to treat it as a normal packfile, and error out when reading it.  That seems pretty touch to protect against.  But this PR is mainly about protecting against conflict files, which should never match that format.

PR #1551, opened 2018-04-18
libkbfs: cache merkle roots for revoked and reset keys

strib:
In the future this will let us order MD writes against key revokes/resets with less server trust.

Issue: KBFS-2890


strib:
> Is this a heuristic/start-point for when client needs to search for merkle roots to verify writes?

Yeah, a starting point.  We'll need to send those merkle roots to the mdserver.

PR #1557, opened 2018-04-27
HTTP server in KBFS for serving content to the front end

songgao:
This needs to go in together with https://github.com/keybase/client/pull/11568 . Will merge the client one first, update vendor here, then merge this.

After they are in, we'll also need a third PR in client to add public server in `InitParams` in `go/bind`.


libhttpserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+	s.tokens.Add(token, nil)
+	return token, nil
+}
+
+func (s *Server) handleInvalidToken(w http.ResponseWriter) {
+	w.WriteHeader(http.StatusForbidden)
+	io.WriteString(w, `
+    <html>
+        <head>
+            <title>.kbfs_http_token_invalid</title>
```
jzila: `<title>KBFS HTTP Token Invalid</title>`

libhttpserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (s *Server) getHTTPFileSystem(
+	req *http.Request) (toStrip string, fs http.FileSystem, err error) {
+	fields := strings.Split(req.URL.Path, "/")
+	if len(fields) < 3 {
+		return "", nil, errors.New("bad path")
+	}
+	var tlfType tlf.Type
+	switch strings.ToLower(fields[0]) {
+	case "private":
```
jzila: Do we still not have a way to do this in the `tlf` library? We should make a `tlf.ParseTlfType()`.

libhttpserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+    </html>
+    `)
+}
+
+func (s *Server) handleBadRequest(w http.ResponseWriter) {
+	w.WriteHeader(http.StatusBadRequest)
+}
+
+func (s *Server) getHTTPFileSystem(
+	req *http.Request) (toStrip string, fs http.FileSystem, err error) {
```
jzila: IMO this method should accept `ctx context.Context, path string`. It doesn't need to know about requests.

libhttpserver/server.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	toStrip = path.Join(fields[0], fields[1])
+
+	fav := tlfHandle.ToFavorite()
+	if fsCached, ok := s.fs.Get(fav); ok {
+		if tlfFS, ok := fsCached.(*libfs.FS); ok {
+			return toStrip, tlfFS.ToHTTPFileSystem(req.Context()), nil
```
jzila: Can we cache these `ToHTTPFileSystem` calls?

songgao: This is merely a wrapper of the underlying `*libfs.FS`. I don't think we can cache its result since it attaches the request's context to it.

libhttpserver/server.go
songgao: Sure. On mobile we'll need to rely on this "field" to detect this failure, so I was kind of intentionally making it opaque to avoid something else hitting it. But what you suggest is good too I think, so will change.

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+			return toStrip, tlfFS.ToHTTPFileSystem(ctx), nil
+		}
+	}
+
+	tlfFS, err := libfs.NewFS(ctx,
+		s.config, tlfHandle, "", "", keybase1.MDPriorityNormal)
+	if err != nil {
+		return "", nil, err
+	}
+	s.fs.Add(fav, tlfFS)
```
strib: We might need to add hooks to delete or rename FSes from the favorites list when the handle changes.  For example, if an old TLF `a,b@rooter` becomes a conflict folder, or a subteam gets renamed (and a new subteam gets made with the old name).  You can subscribe with KBFSOps to be an observer of handle change callbacks.

strib: (I guess that change could be in a separate PR if you want, but we should probably do it sooner rather than later.)

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+	tlfHandle, err := libkbfs.GetHandleFromFolderNameAndType(ctx,
+		s.config.KBPKI(), s.config.MDOps(), fields[1], tlfType)
+	if err != nil {
+		return "", nil, err
+	}
+
+	toStrip = path.Join(fields[0], fields[1])
+
+	fav := tlfHandle.ToFavorite()
+	if fsCached, ok := s.fs.Get(fav); ok {
```
strib: It might be better to do the caching for `toStrip` , rather than the handle, because parsing and resolving the handle is kind of expensive.  Since all the calls will be coming in from the front end, where the names are already normalized, there shouldn't be any downside to it.

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+		s.config, tlfHandle, "", "", keybase1.MDPriorityNormal)
+	if err != nil {
+		return "", nil, err
+	}
+	s.fs.Add(fav, tlfFS)
+
+	return toStrip, tlfFS.ToHTTPFileSystem(ctx), nil
+}
+
+// serve accetps "/<fs path>?token=<token>"
```
strib: "accetps" -> "accepts"

libhttpserver/server_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func makeTestKBFSConfig(t *testing.T) (
+	kbfsConfig libkbfs.Config, shutdown func()) {
+	ctx := libkbfs.BackgroundContextWithCancellationDelayer()
+	cfg := libkbfs.MakeTestConfigOrBustLoggedInWithMode(
+		t, 0, libkbfs.InitSingleOp, "alice", "bob")
+
+	tempdir, err := ioutil.TempDir(os.TempDir(), "journal_server")
```
strib: Nit: if this function hits an error before it returns, `tempdir` will never be cleaned up.  Usually I'd have this function return an error, and then use a `defer` that does clean up if `err != nil`.

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+// For example:
+//     /team/keybase/file.txt?token=1234567890abcdef1234567890abcdef
+func (s *Server) serve(w http.ResponseWriter, req *http.Request) {
+	token := req.URL.Query().Get("token")
+	if len(token) == 0 || !s.tokens.Contains(token) {
+		s.handleInvalidToken(w)
+		return
+	}
+	toStrip, fs, err := s.getHTTPFileSystem(req.Context(), req.URL.Path)
```
strib: Are tokens going to be unique to a session or unique to a file?  If the latter, should we be deleting the token from the cache now that it's used?    If the former, do we have to worry about some other program eavesdropping on the "http" localhost connection and reusing the same tokens?

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return toStrip, tlfFS.ToHTTPFileSystem(ctx), nil
+}
+
+// serve accetps "/<fs path>?token=<token>"
+// For example:
+//     /team/keybase/file.txt?token=1234567890abcdef1234567890abcdef
+func (s *Server) serve(w http.ResponseWriter, req *http.Request) {
+	token := req.URL.Query().Get("token")
+	if len(token) == 0 || !s.tokens.Contains(token) {
```
strib: Will the frontend have some kind of retry mechanism to get a new token if the old one is kicked out of the cache?

libhttpserver/server.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return ext, mimeType
+}
+
+const portStart = 7000
+const portEnd = 8000
+const requestPathRoot = "/files/"
+
+// Init implements the libkbfs.LocalHTTPServer interface.
+func (s *Server) Init(
```
strib: Maybe we should wrap the init code in a `sync.Once`, or somehow return an error if it's already initialized?

libmime/mime_types.go
```diff
@@ -0,0 +1,117 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+// This file is auto-generated and should not be edited by hand.
+
+package libmime
+
+var mimeTypes = map[string]string{
```
strib: Would it make sense to alphabetize this, if it's generated by a script we wrote?  (Also, if we wrote the script to generate it, should that be checked in too?)

libmime/patch.go
```diff
@@ -0,0 +1,26 @@
...
+type Overrider func(ext string, mimeType string) (newExt string, newMimeType string)
+
+func dontOverride(ext string, mimeType string) (newExt string, newMimeType string) {
+	return ext, mimeType
+}
+
+// Patch patches the mime types Go uses by calling mime.AddExtensionType on
+// each from a private list in this package. Optionally provide a non-nil
+// Overrider to override any mime type defined in the list.
+func Patch(override Overrider) {
```
strib: Is this basically a required init function for anything using mime types?  If so, this seems like a weird function name, and something like `Init()` would be better.

Or does "mime" already have most of these defaults, and most of these `AddExtensionType` calls are no-ops?

libpages/server.go
```diff
@@ -21,14 +21,18 @@ import (
...
 	"github.com/keybase/kbfs/libpages/config"
-	_ "github.com/keybase/kbfs/libpages/mime" // side-effects: add mime types
 	"github.com/keybase/kbfs/tlf"
 	"go.uber.org/zap"
 	"golang.org/x/crypto/acme"
 	"golang.org/x/crypto/acme/autocert"
 )
 
+func init() {
+	libmime.Patch(nil)
```
strib: Is there a reason not to put this in `ListenAndServe()`?  The port number is already hardcoded in that function, so it's not like it can be called twice.

Jenkinsfile
```diff
@@ -366,6 +366,12 @@ def runNixTest(prefix) {
             sh './test.fuse -test.timeout 12m'
         }
     }
+    tests[prefix+'libhttpserver'] = {
```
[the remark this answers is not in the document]
songgao: Good catch. I didn't look closely to realize this didn't include appveyor

libhttpserver/server.go
songgao: Good point; Thanks! I'll do it in a follow-up PR soon.

libmime/patch.go
songgao: `mime` loads from the a couple places on the filesystem. So if the machine has nginx or apache installed, `mime` can already use them. But in many cases these are not no-ops. Also for `html` ones, we are overriding them into plain text to avoid being executed by frontend web view components.

`Init()` sounds good!

libhttpserver/server.go
songgao: This limits the HTTP server's access to those that can access the service socket, meaning the user running the app and `root`. If we didn't have the token mechanism, other users on the same machine can request content easily.

(I was wrong about enabling HSTS. That's irrelevant actually.)

songgao: assuming the OS doesn't allow non-root user to snoop traffic, which I think is the default

strib: Ah ok.  So other non-root users on the same machine definitely can't listen to regular HTTP traffic on the `lo` interface?

songgao: I think you need certain permissions to do that. It's possible to grant non-root user that permissions though: https://askubuntu.com/a/632189

libmime/patch.go
songgao: Actually I think patch is still better, based on your comment. Let me know if you still want `Init`

songgao:
I think we can move this out of config. Will try to do that in the PR.

songgao:
@strib Actually, although we can start their own servers in libfuse/libdokan/gobind, SimpleFS needs to access a method of the `LocalHTTPServer` in order to get the token, and the only easy way seems to be through `libkbfs.Config`, unless we are willing to let `SimpleFS` spawn up the HTTP server. So I'm gonna keep this for now. Let me know you want me to do that, and I'll do it in a follow-up PR.

strib:
Yeah my strawman proposal above was that either simplefs starts the http server, or that's it's done in the same wrapper function that starts SimpleFS, and it gets passed into it.  Is there any reason not to do it that way?  That is, do you expect something completely unrelated to simplefs or libpages to need access to it?  Maybe some day we'd want the address listed in `.kbfs_status`, for example?  Though maybe there should be a more general mechanism for that, not sure.

songgao:
>  Is there any reason not to do it that way?

Not really. I don't think we are using it anywhere other than front end right now. `libpages` uses `libfs` directly.

I think I just had a wrong impression from earlier discussion that we didn't want it to be in SimpleFS. I'll move it in a follow-up PR then. Thanks!

PR #1559, opened 2018-04-29
Fix file truncation on Windows with bash

### LINKED ISSUE #1494: File not overwritten by >file under Cygwin
I use Cygwin on windows, tjek this:

```
$ printf "A\nB\n">t

$ cat t
A
B

$ printf "Z\n">t

$ cat t
Z
B
```

taruti:
When a file is opened with FILE_OPEN_IF with Dokan and the file
exists it should be returned, but additionally ErrObjectNameCollision
should be returned.

Fixes KBFS-2812 and Fixes #1494

This ended up requiring quite a lot of restructuring code since
we need to pass up whether a file was created or an existing
one is returned. Doing this on all the return sites would have
been even more messy.

### DIFF
--- a/dokan/filesystem.go
+++ b/dokan/filesystem.go
@@ -37,7 +37,7 @@ type FileSystem interface {
 	WithContext(context.Context) (context.Context, context.CancelFunc)
 
 	// CreateFile is called to open and create files.
-	CreateFile(ctx context.Context, fi *FileInfo, data *CreateData) (file File, isDirectory bool, err error)
+	CreateFile(ctx context.Context, fi *FileInfo, data *CreateData) (file File, status CreateStatus, err error)
 
 	// GetDiskFreeSpace returns information about disk free space.
 	// Called quite often by Explorer.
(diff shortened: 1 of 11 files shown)


libdokan/tlf.go
```diff
@@ -157,13 +157,14 @@ func (tlf *TLF) GetFileInformation(ctx context.Context, fi *dokan.FileInfo) (st
 }
 
 // open tries to open a file.
-func (tlf *TLF) open(ctx context.Context, oc *openContext, path []string) (dokan.File, bool, error) {
+func (tlf *TLF) open(ctx context.Context, oc *openContext, path []string) (
+	dokan.File, dokan.CreateStatus, error) {
 	if len(path) == 0 {
 		if err := oc.ReturningDirAllowed(); err != nil {
-			return nil, true, err
+			return nil, 0, err
```
jzila: Is this behavior correct? It used to return `true`, which should map to `dokan.ExistingDir`. Was that previously a bug?

dokan/filesystem.go
```diff
@@ -54,6 +54,34 @@ type FileSystem interface {
 	ErrorPrint(error)
 }
 
+// CreateStatus marks status of successfull create/open operations.
+type CreateStatus uint32
+
+const (
+	// 1 Bit 0 == always set for valid values,
```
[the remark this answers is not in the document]
taruti: Will use constants.

libdokan/tlf.go
taruti: Here CreateFile fails so the CreateStatus is ignored.

taruti:
Addressed review, rebased to master for CI-passing and handle also FileSupersede and FileOverwriteIf similarly.

libdokan/tlf.go
jzila: SGTM

PR #1561, opened 2018-05-01
additional review feedbacks for #1557

songgao:
Circle CI will probably fail, but I tested locally that `LOCAL_CLIENT=1 LOCAL_KBFS=1 yarn run rn-gobuild-android` builds. So should be fine as long as this goes in together with https://github.com/keybase/client/pull/11612 .

Related: #1557

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -855,15 +855,17 @@ func (o folderHandleChangeObserver) TlfHandleChange(
 	o()
 }
 
-// SubscribeToEndOfLife returns a context that will be canceled when this *FS
+// SubscribeToEndOfLife returns a channel that will be closed when this *FS
 // reaches end-of-life, meaning if user of this object caches it for long term
 // use, it should invalide this entry and create a new one using NewFS.
-func (fs *FS) SubscribeToEndOfLife() (ctx context.Context, err error) {
-	ctx, cancel := context.WithCancel(context.Background())
-	if err = fs.config.Notifier().RegisterForChanges(
+func (fs *FS) SubscribeToEndOfLife() (<-chan struct{}, error) {
+	c := make(chan struct{})
+	once := &sync.Once{}
+	cancel := folderHandleChangeObserver(func() { once.Do(func() { close(c) }) })
+	if err := fs.config.Notifier().RegisterForChanges(
 		[]libkbfs.FolderBranch{fs.root.GetFolderBranch()},
 		folderHandleChangeObserver(cancel)); err != nil {
 		return nil, err
 	}
-	return ctx, nil
+	return c, nil
 }
--- a/libhttpserver/server.go
+++ b/libhttpserver/server.go
@@ -70,13 +70,13 @@ func (s *Server) handleBadRequest(w http.ResponseWriter) {
 }
 
 type endOfLifeTrackingFS struct {
-	fs  *libfs.FS
-	ctx context.Context
+	fs *libfs.FS
+	ch <-chan struct{}
 }
 
 func (e endOfLifeTrackingFS) isEndOfLife() bool {
 	select {
-	case <-e.ctx.Done():
+	case <-e.ch:
 		return true
 	default:
 		return false
@@ -117,12 +117,12 @@ func (s *Server) getHTTPFileSystem(ctx context.Context, requestPath string) (
 		return "", nil, err
 	}
 
-	fsLifeCtx, err := tlfFS.SubscribeToEndOfLife()
+	fsLifeCh, err := tlfFS.SubscribeToEndOfLife()
 	if err != nil {
 		return "", nil, err
 	}
 
-	s.fs.Add(toStrip, endOfLifeTrackingFS{fs: tlfFS, ctx: fsLifeCtx})
+	s.fs.Add(toStrip, endOfLifeTrackingFS{fs: tlfFS, ch: fsLifeCh})
 
 	return toStrip, tlfFS.ToHTTPFileSystem(ctx), nil
 }
@@ -131,16 +131,16 @@ func (s *Server) getHTTPFileSystem(ctx context.Context, requestPath string) (
 // For example:
 //     /team/keybase/file.txt?token=1234567890abcdef1234567890abcdef
 func (s *Server) serve(w http.ResponseWriter, req *http.Request) {
-	s.logger.Info("incoming request from %q: %s", req.UserAgent(), req.URL)
+	s.logger.Debug("Incoming request from %q: %s", req.UserAgent(), req.URL)
 	token := req.URL.Query().Get("token")
 	if len(token) == 0 || !s.tokens.Contains(token) {
-		s.logger.Info("invalid token %q", token)
+		s.logger.Info("Invalid token %q", token)
 		s.handleInvalidToken(w)
 		return
 	}
 	toStrip, fs, err := s.getHTTPFileSystem(req.Context(), req.URL.Path)
 	if err != nil {
-		s.logger.Warning("bad request. error=%v", err)
+		s.logger.Warning("Bad request; error=%v", err)
 		s.handleBadRequest(w)
 		return
 	}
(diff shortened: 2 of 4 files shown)


libfs/fs.go
```diff
@@ -841,3 +841,29 @@ func (fs *FS) ToHTTPFileSystem(ctx context.Context) http.FileSystem {
...
+}
+func (o folderHandleChangeObserver) TlfHandleChange(
+	context.Context, *libkbfs.TlfHandle) {
+	o()
+}
+
+// SubscribeToEndOfLife returns a context that will be canceled when this *FS
+// reaches end-of-life, meaning if user of this object caches it for long term
+// use, it should invalide this entry and create a new one using NewFS.
+func (fs *FS) SubscribeToEndOfLife() (ctx context.Context, err error) {
```
strib: I think this should return a channel that gets closed, rather than a context.  You're not using any other part of the context, right?

libhttpserver/server.go
```diff
@@ -112,13 +131,16 @@ func (s *Server) getHTTPFileSystem(ctx context.Context, requestPath string) (
 // For example:
 //     /team/keybase/file.txt?token=1234567890abcdef1234567890abcdef
 func (s *Server) serve(w http.ResponseWriter, req *http.Request) {
+	s.logger.Info("incoming request from %q: %s", req.UserAgent(), req.URL)
```
strib: Can this be `Debug` rather than `Info`?

strib: Also usually we try to capitalize the first letter of log messages, here and below.

libhttpserver/server.go
```diff
@@ -136,40 +158,66 @@ func overrideMimeType(ext, mimeType string) (newExt, newMimeType string) {
...
+	".py":    "text/plain",
+	".zsh":   "text/plain",
+	".fish":  "text/plain",
+	".cs":    "text/plain",
+	".rb":    "text/plain",
+	".m":     "text/plain",
+	".mm":    "text/plain",
+	".swift": "text/plain",
+	".flow":  "text/plain",
+
```
strib: Why this blank line?

libmime/patch.go
```diff
@@ -14,13 +14,20 @@ func dontOverride(ext string, mimeType string) (newExt string, newMimeType strin
 }
 
 // Patch patches the mime types Go uses by calling mime.AddExtensionType on
-// each from a private list in this package. Optionally provide a non-nil
-// Overrider to override any mime type defined in the list.
-func Patch(override Overrider) {
+// each from a private list in this package.  Both parameters are optional.
+// Provide a non-nil Overrider to override any mime type defined in the list.
+// Provide a non-nil additional map (ext->mimeType) to add additional mime
+// types.
```
strib: Maybe mention that the `overrider` can also override items in the `additional` list?  Might make sense to swap the order of the params too, since that's the case.

libmime/patch.go
```diff
@@ -14,13 +14,20 @@ func dontOverride(ext string, mimeType string) (newExt string, newMimeType strin
...
+func Patch(override Overrider, additional map[string]string) {
 	if override == nil {
 		override = dontOverride
 	}
 	for ext, mimeType := range mimeTypes {
 		mime.AddExtensionType(override(ext, mimeType))
 	}
+	if additional != nil {
```
strib: No need for this check/indentation.

libfs/fs.go
songgao: No I'm not using any other part, but this allows me not to worry about closing the channel more than once without building an extra layer. I guess we could use `WaitGroup` here too, but I figured `Context` was more appropriate. WDYT?

libmime/patch.go
songgao: Hmm that's actually not the case now but I think it makes sense. So will make it so.

songgao: Oh sorry that's actually the case. Will update comment/parameter order. Thanks!

libfs/fs.go
strib: I don't think it makes sense as a context or a waitgroup, sorry.  Those make it more confusing I think.  What about something like:

```go
c := make(chan struct{})
once := &sync.Once{}
fhco := folderHandleChangeObserver(func() { once.Do(close(c)) })
```

songgao: Sure; will do!

libhttpserver/server.go
```diff
@@ -103,7 +117,12 @@ func (s *Server) getHTTPFileSystem(ctx context.Context, requestPath string) (
 		return "", nil, err
 	}
 
-	s.fs.Add(toStrip, tlfFS)
+	fsLifeCh, err := tlfFS.SubscribeToEndOfLife()
```
jzila: I'd prefer something like `SubscribeToShutdown()` here

libhttpserver/server.go
```diff
@@ -71,6 +69,20 @@ func (s *Server) handleBadRequest(w http.ResponseWriter) {
 	w.WriteHeader(http.StatusBadRequest)
 }
 
+type endOfLifeTrackingFS struct {
```
jzila: I'd prefer `shutdownTrackingFS`

libhttpserver/server.go
```diff
@@ -136,40 +158,65 @@ func overrideMimeType(ext, mimeType string) (newExt, newMimeType string) {
 	return ext, mimeType
```
jzila: Why do we have a custom mechanism here for `javascript` and `html`, but below we also have `additionalMimeTypes`? Why not add those extensions to the `additionalMimeTypes` list?

simplefs/simplefs.go
```diff
@@ -111,23 +115,28 @@ type handle struct {
 // make sure the interface is implemented
 var _ keybase1.SimpleFSInterface = (*SimpleFS)(nil)
 
-// NewSimpleFS creates a new SimpleFS instance.
-func NewSimpleFS(config libkbfs.Config) keybase1.SimpleFSInterface {
-	return newSimpleFS(config)
-}
-
-func newSimpleFS(config libkbfs.Config) *SimpleFS {
+func newSimpleFS(g *libkb.GlobalContext, config libkbfs.Config) *SimpleFS {
```
jzila: Why do we need the `GlobalContext`?

libhttpserver/server.go
songgao: It's not a shutdown though; as it can still be used.

libhttpserver/server.go
songgao: `additionalMimeTypes` is based on exact ext/mimeType paris, while the overrider allows other ways of matching. For example, there could be some other ext that matches to `text/html`. The overrider will be able to catch them but `additionalMimeTypes` can't. The reason why `additionalMimeTypes` exists is that on some platform (where cat /etc/apache2/mime.types is populated), stuff like `.pub` are application types, and cause the webview to prompt a download dialog. So we explicitly patch them.

simplefs/simplefs.go
songgao: `libkb.HTTPSrv` needs it.

libhttpserver/server.go
songgao: Also since this isn't the same as shutting down a folder branch ops, and we already have `Shutdown` methods on various types, I think it'd be better to name this one differently to avoid confusion.

libhttpserver/server.go
jzila: OK. Do we have that documented in the code somewhere?

simplefs/simplefs.go
jzila: oh I see. Too bad there isn't an easy way of working around that :(

libhttpserver/server.go
songgao: I had some comments noting that we should update both list together. But will add more about this in `libmime`.

libhttpserver/server.go
songgao: It means user of this type shouldn't use this anymore (as mentioned on the method comment). I was trying to avoid mentioning "handle change" as that's a concept that the `libfs` layer hasn't had to expose yet. If end-of-life doesn't make sense here, I'd gladly consider other names. Would `SubscribeToObsolete` be better?

simplefs/simplefs.go
songgao: It looks like it's only used for logging. If we care about it enough, perhaps we can change `libkb.HTTPSrv` to accept a `Logger` interface instead.

libhttpserver/server.go
jzila: Oh yeah that would make more sense to me.

simplefs/simplefs.go
jzila: maybe later :)

PR #1564, opened 2018-05-03
dokan: Move DLL loading logic from C to Go and add debugging

taruti:
Will test later today.

### DIFF
--- a/dokan/loaddll.go
+++ b/dokan/loaddll.go
@@ -17,8 +17,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"runtime"
-	"strings"
 	"unsafe"
 
 	"golang.org/x/sys/windows"
@@ -51,7 +51,7 @@ func doLoadDLL(epc *errorPrinter, path string) (windows.Handle, error) {
 		path = shortPath
 		guessPath = true
 	} else {
-		path = strings.Replace(path, `/`, `\`, -1)
+		path = filepath.FromSlash(path)
 	}
 	const loadLibrarySearchSystem32 = 0x800
 	const flags = loadLibrarySearchSystem32
--- a/dokan/loaddll.go
+++ b/dokan/loaddll.go
@@ -39,14 +39,14 @@ func (ep *errorPrinter) Printf(s string, os ...interface{}) {
 // loadLibrary calls win32 LoadLibrary and logs the result.
 func loadLibrary(epc *errorPrinter, path string) (windows.Handle, error) {
 	hdl, err := windows.LoadLibrary(path)
-	epc.Printf("LoadLibrary(%q) -> %v,%v", path, hdl, err)
+	epc.Printf("LoadLibrary(%q) -> %v,%v\n", path, hdl, err)
 	return hdl, err
 }
 
 // doLoadDLL tries to load the dokan DLL from various locations.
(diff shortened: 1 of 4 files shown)


taruti:
Github bugs out with assigning reviewers, ping @jzila and @zanderz please look at this. Will do more testing later today.

dokan/loaddll.go
```diff
@@ -0,0 +1,144 @@
...
+	hdl, err := windows.LoadLibraryEx(path, 0, flags)
+	epc.Printf("loadDokanDLL LoadLibraryEx(%q,0,%x) -> %v,%v", path, flags, hdl, err)
+	if err == nil || !guessPath {
+		return hdl, err
+	}
+	// User probably has not installed KB2533623 which is a security update
+	// from 2011. Without this Windows security update loading libraries
+	// is unsafe on Windows.
+
+	if runtime.GOARCH == `386` {
```
zanderz: runtime.GOARCH is set at compile time and is unrelated to the architecture the program is running on. I assume you wanted that? The updater gets it this way: https://github.com/keybase/go-updater/blob/master/keybase/platform_windows.go#L127

jzila:
My C is a bit rusty, but it might be good to get strib to take a look as well.

dokan/loaddll.go
taruti: Actually doing it by the compile time version is fine in this case. If it was compiled as 32-bit then we want:
+ 32-bit windows: syswow64 does not exist, system32 exists and gets used.
+ 64-bit windows: syswow64 exists and gets used instead of system32.
The old C version also worked like this (but it just checked sizeof(void*)),

zanderz: ah ok then, NVM!

dokan/loaddll.go
```diff
@@ -0,0 +1,144 @@
...
+
+// doLoadDLL tries to load the dokan DLL from various locations.
+func doLoadDLL(epc *errorPrinter, path string) (windows.Handle, error) {
+	var guessPath bool
+	epc.Printf("loadDokanDLL %q\n", path)
+	if path == "" {
+		path = shortPath
+		guessPath = true
+	} else {
+		path = strings.Replace(path, `/`, `\`, -1)
```
strib: Can you use `filepath.FromSlash()` here instead?

dokan/loaddll.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	hdl, err = loadLibrary(epc, system32+shortPath)
+	if err == nil {
+		return hdl, nil
+	}
+	hdl, err = loadLibrary(epc, shortPath)
+	if err == nil {
+		return hdl, nil
+	}
+	err = fmt.Errorf("loadDokanDLL: cannot load Dokan DLL: %v", err)
```
strib: Would it be worthwhile to log each of the above non-nil errors along with which path caused them, in case each is different?

dokan/loaddll.go
taruti: Will do.

dokan/loaddll.go
taruti: The loadLibrary with the epc argument logs every call to the epc with the path included, and that gets attached to the error value in loadDokanDLL.

strib: Ah right.  So I guess it's not strictly needed to log `err` again here.

PR #1567, opened 2018-05-05
handle backgroud properly in libhttpserver

songgao:
see #kbfs for details

### DIFF
--- a/libhttpserver/server.go
+++ b/libhttpserver/server.go
@@ -32,6 +32,7 @@ type Server struct {
 	config libkbfs.Config
 	server *libkb.HTTPSrv
 	logger logger.Logger
+	g      *libkb.GlobalContext
 
 	tokens *lru.Cache
 	fs     *lru.Cache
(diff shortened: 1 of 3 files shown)


libhttpserver/server.go
```diff
@@ -187,26 +188,46 @@ const portStart = 7000
...
-	s.server = libkb.NewHTTPSrv(
-		g, libkb.NewPortRangeListenerSource(portStart, portEnd))
-	// Have to start this first to populate the ServeMux object.
-	if err = s.server.Start(); err != nil {
+	if err = s.startServer(); err != nil {
 		return nil, err
 	}
-	s.server.Handle(requestPathRoot,
-		http.StripPrefix(requestPathRoot, http.HandlerFunc(s.serve)))
+	go s.monitorAppState()
```
strib: This needs to be shutdown somewhere.

libhttpserver/server.go
```diff
@@ -187,26 +188,46 @@ const portStart = 7000
...
+	}
+	s.server.Handle(requestPathRoot,
+		http.StripPrefix(requestPathRoot, http.HandlerFunc(s.serve)))
+	return nil
+}
+
+func (s *Server) monitorAppState() {
+	state := keybase1.AppState_FOREGROUND
+	for {
+		state = <-s.g.AppState.NextUpdate(&state)
```
strib: On mobile will you be sharing the same `AppState` as the service?  Does that mean only one of you will get this update?  It looks like `AppState.NextUpdate()` takes a lock, so there can only be one caller at once.  Maybe this doesn't share a `g` with the service?

strib: Oh I see, it actually doesn't block, and saves all the channels.  Ok nevermind.

songgao: That’s actually a great point. I was rushing this in without thinking about that. Thanks for making sure!

libhttpserver/server.go
```diff
@@ -187,26 +189,52 @@ const portStart = 7000
...
-	// Have to start this first to populate the ServeMux object.
-	if err = s.server.Start(); err != nil {
+	if err = s.startServer(); err != nil {
 		return nil, err
 	}
-	s.server.Handle(requestPathRoot,
-		http.StripPrefix(requestPathRoot, http.HandlerFunc(s.serve)))
+	ctx, cancel := context.WithCancel(context.Background())
+	go s.monitorAppState(ctx)
+	s.cancel = cancel
```
jzila: these 3 lines could happen in `s.startServer`, which would allow you not to have to re-assign `s.cancel`.

PR #1569, opened 2018-05-07
md_ops: check revoked keys using the merkle tree

strib:
[This shouldn't be merged until we deploy the FindNextMD RPC.]

The algorithm is:

* Make a preliminary check against the untrusted server timestamps, as before.  If that fails, fail the verification.
* Fetch the merkle info from the mdserver, using the key revocation merkle root.
* If the server doesn't return anything, verify that the revision has a valid chain of updates up to the current head of the TLF.  If everything checks out, consider it valid. TODO in a future PR: make sure there's not a big merkle time gap between the revision and the head, or else the server might be lying.
* Otherwise, for private and team folders, we need to decrypt the merkle leaf, but we don't know what key generation was used for it. So, starting at the revision we care about, iterate forwards and try each new key generation we find.
* If the leaf revision corresponding to the revocation is smaller than the MD we're verifying, fail the verification because the MD was written too late.
* Otherwise, verify the chain of MDs between the revision we're verifying and the leaf revision.  If everything checks out, consider it valid.

Still TODO in a future PR: cross-check the global merkle root info with the service, to make sure it all fits together with our view of the global merkle tree.

Issue: KBFS-2904

### DIFF
--- a/libkbfs/md_ops.go
+++ b/libkbfs/md_ops.go
@@ -53,36 +53,270 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
 	return UnverifiableTlfUpdateError{tlf, writer, err}
 }
 
+type ctxMDOpsSkipKeyVerificationType int
+
+// This context key indicates that we should skip verification of
+// revoked keys, to avoid recursion issues.  Any resulting MD
+// that skips verification shouldn't be trusted or cached.
+const ctxMDOpsSkipKeyVerification ctxMDOpsSkipKeyVerificationType = 1
+
+func (md *MDOpsStandard) decryptMerkleLeaf(
+	ctx context.Context, rmd ReadOnlyRootMetadata,
+	kbfsRoot *kbfsmd.MerkleRoot, leafBytes []byte) (
+	leaf *kbfsmd.MerkleLeaf, err error) {
+	var eLeaf kbfsmd.EncryptedMerkleLeaf
+	err = md.config.Codec().Decode(leafBytes, &eLeaf)
+	if err != nil {
+		return nil, err
+	}
+
+	// The private key we need to decrypt the leaf does not live in
+	// key bundles; it lives only in the MDs that were part of a
+	// specific keygen.  But we don't yet know what the keygen was, or
+	// what MDs were part of it, except that they have to have a
+	// larger revision number than the given `rmd`.  So all we can do
+	// is iterate up from `rmd`, looking for a key that will unlock
+	// the leaf.
+	//
+	// Luckily, in the common case we'll be trying to verify the head
+	// of the folder, so we should be able to use
+	// rmd.data.TLFPrivateKey.
+
+	// Fetch the latest MD so we have all possible TLF crypt keys
+	// available to this device.
+	head, err := md.getForTLF(
+		ctx, rmd.TlfID(), rmd.BID(), rmd.MergedStatus(), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var uid keybase1.UID
+	if rmd.TlfID().Type() != tlf.Public {
+		session, err := md.config.KBPKI().GetCurrentSession(ctx)
+		if err != nil {
+			return nil, err
+		}
+		uid = session.UID
+	}
+
+	currRmd := rmd
+	//var fetched []*RootMetadataSigned
+	for {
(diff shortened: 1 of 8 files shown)


libkbfs/md_ops.go
```diff
@@ -53,14 +53,270 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
...
+	if rmd.TlfID().Type() != tlf.Public {
+		session, err := md.config.KBPKI().GetCurrentSession(ctx)
+		if err != nil {
+			return nil, err
+		}
+		uid = session.UID
+	}
+
+	currRmd := rmd
+	//var fetched []*RootMetadataSigned
```
jzila: commented code?

libkbfs/md_ops.go
```diff
@@ -53,14 +53,270 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
...
+	}
+
+	// Otherwise it's valid, as long as there's a valid chain of MD
+	// revisions between the two, and the global root info checks out
+	// as well.  By using `ctx`, we'll avoid infinite loops in the
+	// writer-key-checking code by skipping revoked key verification.
+	// This is ok, because we only care about the hash chain for the
+	// purposes of verifying `irmd`.
+	chain, err := getMergedMDUpdatesWithEnd(
+		ctx, md.config, irmd.TlfID(), irmd.Revision()+1, leaf.Revision, nil)
```
jzila: In the `decryptMerkleLeaf` case, we would have already retrieved this chain (via repeated iterations of `fetchLoop`). I wonder if we should return that to minimize repeated work.

libkbfs/md_ops.go
```diff
@@ -252,13 +496,34 @@ func (md *MDOpsStandard) processMetadata(ctx context.Context,
 	}
 
 	key := rmds.GetWriterMetadataSigInfo().VerifyingKey
-	*rmds = RootMetadataSigned{}
 	irmd := MakeImmutableRootMetadata(rmd, key, mdID, localTimestamp, true)
 
-	err = md.config.MDCache().Put(irmd)
-	if err != nil {
+	// Then, verify the verifying keys.  We do this after decrypting
+	// the MD and making the ImmutableRootMetadata, since we way need
```
jzila: s/we way/we may

libkbfs/md_ops.go
strib: Maybe we would have already fetched it, but maybe not.  They could be fetching different ranges.  It's possible `decryptMerkleLeaf` could have already verified part of the chain, but it could be only a small part of it, depending on what `leaf.Revision` is.  In any case, I'm specifically using `md_util` functions here, rather than calling `mdOps.getRange()` directly, because I want it to get MDs from the cache whenever possible.  So it should already avoid re-fetching and re-verifying any MDs more than once, as long as they were readable and cacheable.

jzila: Cool, thanks for the explanation.

libkbfs/md_ops.go
```diff
@@ -53,14 +53,274 @@ func (md *MDOpsStandard) convertVerifyingKeyError(ctx context.Context,
...
+	irmd ImmutableRootMetadata) (cacheable bool, err error) {
+	err = md.config.KBPKI().HasVerifyingKey(ctx, uid, verifyingKey,
+		rmds.untrustedServerTimestamp)
+	var info revokedKeyInfo
+	switch e := errors.Cause(err).(type) {
+	case nil:
+		return true, nil
+	case RevokedDeviceVerificationError:
+		// TODO(KBFS-2963): Support team-keyed TLFs.
+		if irmd.TypeForKeying() == tlf.TeamKeying {
```
strib: FYI I added this check because I realized that the decryption totally doesn't work for team-backed TLFs, since the private key isn't stored in the private MD in that case but needs to be fetched from the service.  I made a new ticket for that and will have a separate PR for it.

strib:
> LGTM. Do you think there's a way (e.g. have a special file) for us to trigger checking verifyKey check on every metadata of all TLFs in an admin build so that we can confidently ship this to the public?

Yeah, you can `cat .kbfs_update_history` to verify every MD in a TLF.  I'll definitely be doing that on a folder that I know has writes from a revoked device before we ship this.

PR #1575, opened 2018-05-10
Pages: sha256 password hash (config v2); auto generate random password

songgao:
After this gets merged, I'll have follow-up PRs to move `kbpagesconfig` into keybase/client as `keybase pages` command.

cc @maxtaco


libpages/config/config_v2.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
```
songgao: This is pretty much the same as `config_v1.go`, except that the authentication logic changes to use SHA256.

libpages/config/config_v2_test.go
```diff
@@ -0,0 +1,331 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
```
songgao: Tests are basically the same as `config_v1_test.go`

libpages/config/config_v2.go
strib: Is there no way to just extend the `V1` class (say by embedding it) and override just the `init` and `Authenticate` functions?

kbpagesconfig/common.go
```diff
@@ -0,0 +1,138 @@
...
+	originalConfigStr string,
+	newConfig config.Config,
+	configPath string,
+	p prompter) (err error) {
+	buf := &bytes.Buffer{}
+	if err := newConfig.Encode(buf, true); err != nil {
+		return fmt.Errorf("encoding config error: %v", err)
+	}
+	newConfigStr := buf.String()
+	if newConfigStr == originalConfigStr {
```
strib: Is there some guarantee that the json encoder will output struct fields in the same order every time?  I don't see that in the documentation.  Are you sure this is safe?

libpages/config/config_v2_test.go
strib: Again, would it be possible to re-factor the tests so you can just pass a config object in for most of them?

libpages/config/config_v2.go
```diff
@@ -0,0 +1,162 @@
...
+// InvalidPasswordHash is the error that happens when there's an invalid
+// password hash in the config.
+type InvalidPasswordHash struct{}
+
+// Error implements the error interface.
+func (InvalidPasswordHash) Error() string {
+	return "invalid passwordhash"
+}
+
+const sha256PasswordHashPrefix = "sha256:"
```
strib: Seems like this should be defined (and used) in `config/utils.go`.

libpages/config/utils.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+// GenerateSHA256PasswordHash generates a sha256 hashed password hash from
+// plaintextPassword in the form of "sha256:<hash>". This is what should go
+// into the config file.
+func GenerateSHA256PasswordHash(plaintextPassword string) string {
+	hash := sha256.Sum256([]byte(plaintextPassword))
```
strib: Would it be good practice to add a nonce in here (that's also stored in the config), so it's harder to tell when passwords are being reused?

songgao: Wait actually, what do we get from making it harder to tell when passwords are reused?

strib:
> I thought about just doing it in V1, but then it wouldn't have a clear cutoff for when we want to get rid of bcrypt. I was thinking after the migration is done and we disable the v1, we can just remove the V1 code altogether, so I didn't bother making the two types share code or tests. WDYT?

How would you ever know that everyone had upraded v1 to v2?  Even if you do know that at some point, we'll have to maintain both sets of code in the meantime.

I wasn't saying you shouldn't have a v2, just that it could share a bunch of code pretty easily, I think.  I might be missing something though.

libpages/config/utils.go
strib: It just seems like the right thing to do from a security standpoint.  Obviously Keybase still has to handle the plaintext passwords because that's how http auth works.  But you could imagine someone putting a config file in a public or team folder, but not wanting to expose the fact that two users have the same password to other readers of that directory.

songgao:
> How would you ever know that everyone had upraded v1 to v2? Even if you do know that at some point, we'll have to maintain both sets of code in the meantime.

By that I meant at the point we turn off V1 support on the kbpagesd instance we are hosting. We can message people who have V1. But now that I think of this, it'd be a lot of them. So that's actually an argument for making this into V1 so that people who don't have passwords don't have to upgrade ...

songgao:
or still bump to v2, but when we cut of bcrypt, keep v1 supported but cut bcrypt. So if user has v2, they know clearly nothing needs to be done. But if they have v1, they may or may not need to upgrade depending on if they have authed users.

libpages/config/utils.go
songgao: That's a good case. Will do!

kbpagesconfig/common.go
songgao: Good question. It does mention that map keys are sorted, but didn't mention anything about object keys that are from structs. It does seem like it's sorted by the order they are defined in the struct, but not being in the doc means it might go away.

Although I think it's safe in this case though, since even if the order change, we'd be showing an annoying diff to user and write the config when it's not necessary. But we'd never skip a write when we should. I can add a test for the struct ordering so that we catch it when this behavior change in the stdlib.

PR #1584, opened 2018-05-15
On Windows, use the restart exit code if the service is restarting

zanderz:


zanderz:
so, clearly this fails because the android build is pulling in client master, which does not have https://github.com/keybase/client/pull/11641. Is there a way to specify that branch for CI? or should it just wait until client merges that?

songgao:
Yeah in this case I would just merge both together (after review), and watch the build on master from both repos.

PR #1588, opened 2018-05-17
md_ops: check merkle time gaps when verifying a revoked key

strib:
(Note: this depends on #1569, which is blocking on a server deploy.)

This adds interface methods fetch timestamps for both global and KBFS Merkle roots, and uses that to make sure that a) if the mdserver doesn't return any Merkle info after a revoke, that the gap since the last merkle root hasn't been too big; and b) if it does return a Merkle root, there wasn't too big a gap since the revoke happened.

Issue: KBFS-2956

### DIFF
--- a/libkbfs/md_ops.go
+++ b/libkbfs/md_ops.go
@@ -29,8 +29,23 @@ const (
 	// we shouldn't be trusting the server.
 	maxAllowedMerkleGap = 3*time.Hour + 15*time.Minute
 
-	merkleGapEnforcementStartString = "2018-May-17"
+	// merkleGapEnforcementStartString indicates when the mdserver
+	// started rejecting new writes based on the lack of recent merkle
+	// updates (according to `maxAllowedMerkleGap` above).
+	merkleGapEnforcementStartString = "2018-05-17T00:00:00+07:00"
 )
+
+var merkleGapEnforcementStart time.Time
+
+func init() {
+	var err error
+	merkleGapEnforcementStart, err = time.Parse(
+		"2006-01-02T15:04:05+07:00", merkleGapEnforcementStartString)
+	if err != nil {
+		// Can never happen without a bad global const string.
+		panic(err)
+	}
+}
 
 // MDOpsStandard provides plaintext RootMetadata objects to upper
 // layers, and processes RootMetadataSigned objects (encrypted and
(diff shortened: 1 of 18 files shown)


libkbfs/md_ops.go
```diff
@@ -20,6 +21,17 @@ import (
...
 
+const (
+	// Our contract with the server states that it won't accept KBFS
+	// writes if more than 3 hours have passed since the last Merkle
+	// roots (both global and KBFS) were published.  Add some padding
+	// to that, and if we see any gaps larger than this, we will know
+	// we shouldn't be trusting the server.
+	maxAllowedMerkleGap = 3*time.Hour + 15*time.Minute
+
+	merkleGapEnforcementStartString = "2018-May-17"
```
songgao: Perhaps use the RFC3339 format here so that it includes timezone?

Also, if we screw up mdmerkle at some point (especially while the write-ordering work is still on-going), is this what we dial forward to "unbreak" clients?

libkbfs/md_ops.go
```diff
@@ -183,6 +195,94 @@ func (md *MDOpsStandard) makeMerkleLeaf(
...
+		return keybase1.MerkleTreeID_KBFS_PUBLIC
+	case tlf.SingleTeam:
+		return keybase1.MerkleTreeID_KBFS_PRIVATETEAM
+	default:
+		panic(fmt.Sprintf("Unexpected TLF keying type: %d",
+			irmd.TypeForKeying()))
+	}
+}
+
+func (md *MDOpsStandard) checkMerkleTimesAreRecent(ctx context.Context,
```
songgao: This function isn't used anywhere. Did you forget to remove it?

libkbfs/md_ops.go
```diff
@@ -245,15 +369,37 @@ func (md *MDOpsStandard) verifyKey(
...
 		return true, nil
 	}
 
+	// Check the gap between the revocation and the global/KBFS roots
+	// that include the revocation, to make sure they fall within the
+	// expected error window.  The server didn't begin enforcing this
+	// until some time into KBFS's existence though.
+	revokeTime := keybase1.FromTime(info.Time)
+	merkleGapEnforcementStart, err := time.Parse(
+		"2006-Jan-02", merkleGapEnforcementStartString)
```
songgao: Could just parse this once at init. That means it can't be `const` though.

libkbfs/md_ops.go
```diff
@@ -227,6 +327,30 @@ func (md *MDOpsStandard) verifyKey(
...
+		// TODO: cache the latest KBFS merkle root somewhere for a while?
+		latestKbfsRoot, err := md.config.MDServer().GetMerkleRootLatest(
+			ctx, treeID)
+		if err != nil {
+			return false, err
+		}
+		serverOffset, _ := md.config.MDServer().OffsetFromServerTime()
+		currServerTime := md.config.Clock().Now().Add(-serverOffset)
+		err = md.checkMerkleTimes(
+			ctx, latestRootTime, latestKbfsRoot, currServerTime,
```
songgao: `currServerTime` here is basically server-trust right? I wonder if there's a way to use a NTP time, or perhaps we should just use local time. It's probably reasonable to assume the local time is roughly in sync with a NTP time. If the local time is terribly out-of-sync (e.g. dead motherboard battery and no NTP service running), HTTPS would be broken too.

libkbfs/md_ops.go
strib: > Perhaps use the RFC3339 format here so that it includes timezone?

Sure, will do.

> Also, if we screw up mdmerkle at some point (especially while the write-ordering work is still on-going), is this what we dial forward to "unbreak" clients?

Depends what you mean by "screw up".  I'll add a comment to clarify, but basically this should reflect when we deploy the mdserver version that rejects writes after it's been too long since merkle published.  It doesn't help any other kinds of screw-ups.

(I'll have to update this value to whatever the next mdserver prod deploy is.)

libkbfs/md_ops.go
strib: As you know I'm very reluctant to use local time for anything.  I have seen clocks be out of sync too many times, even server clocks run by companies.

You're right though, that this particular instance of using server trust is a bit problematic.  Luckily it should be pretty rare, since it should only happen when a client is reading an MD update within ~3 hours of the device being revoked.

How about I just add a check here to make sure that `abs(serverOffset)` is less than 1 hour?  If that breaks, than the worst that can happen is the client will have to wait for the next merkle run.

libkbfs/md_ops.go
songgao: Right; I meant deploy/ops screwup. Thanks for clearing up!

libkbfs/md_ops.go
songgao: > How about I just add a check here to make sure that abs(serverOffset) is less than 1 hour?

That sounds good with me. Thanks!

strib:
@songgao: FYI I added 0d95a4d to this PR, which is temporary until CORE-7924 is done.

libkbfs/md_ops.go
```diff
@@ -20,6 +21,33 @@ import (
...
+	// roots (both global and KBFS) were published.  Add some padding
+	// to that, and if we see any gaps larger than this, we will know
+	// we shouldn't be trusting the server.  TODO: reduce this once
+	// merkle computation is faster.
+	maxAllowedMerkleGap = 8*time.Hour + 15*time.Minute
+
+	// merkleGapEnforcementStartString indicates when the mdserver
+	// started rejecting new writes based on the lack of recent merkle
+	// updates (according to `maxAllowedMerkleGap` above).
+	merkleGapEnforcementStartString = "2018-05-22T13:19:00-07:00"
```
jzila: Will this need to change?

PR #1589, opened 2018-05-17
md_ops: decrypt team Merkle leafs

strib:
(Note: this depends on #1569, which is blocking on a server deploy.)

If a team MD was written by a revoked device and we need to decrypt the Merkle leaf, this PR calls out to the service to get it done (since only the service has access to the team's per-key-generation private key).

If refactors the `kbfsmd.EncryptedMerkleLeaf` class a little to move the actual decryption into `kbfscrypto`, to match some of our other encrypted data types.

Issue: KBFS-2963

### DIFF
--- a/kbfscrypto/encrypted_data.go
+++ b/kbfscrypto/encrypted_data.go
@@ -228,3 +228,59 @@ func DecryptTLFCryptKeys(
 
 	return oldKeys, nil
 }
+
+// EncryptedMerkleLeaf is an encrypted MerkleLeaf object.
+type EncryptedMerkleLeaf struct {
+	encryptedData
+}
+
+// MakeEncryptedMerkleLeaf constructs an EncryptedMerkleLeaf.
+func MakeEncryptedMerkleLeaf(
+	version EncryptionVer, data []byte, nonce *[24]byte) EncryptedMerkleLeaf {
+	return EncryptedMerkleLeaf{
+		encryptedData{
+			Version:       version,
+			EncryptedData: data,
+			Nonce:         nonce[:],
+		},
+	}
+}
+
+// PrepareMerkleLeaf verifies the correctness of the given leaf, and
+// returns its nonce.
+func PrepareMerkleLeaf(encryptedMerkleLeaf EncryptedMerkleLeaf) (
+	nonce [24]byte, err error) {
+	if encryptedMerkleLeaf.Version != EncryptionSecretbox {
+		return [24]byte{},
+			errors.WithStack(UnknownEncryptionVer{
+				Ver: encryptedMerkleLeaf.Version})
+	}
+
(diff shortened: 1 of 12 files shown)


kbfscrypto/encrypted_data.go
```diff
@@ -228,3 +228,59 @@ func DecryptTLFCryptKeys(
...
+			Nonce:         nonce[:],
+		},
+	}
+}
+
+// PrepareMerkleLeaf verifies the correctness of the given leaf, and
+// returns its nonce.
+func PrepareMerkleLeaf(encryptedMerkleLeaf EncryptedMerkleLeaf) (
+	nonce [24]byte, err error) {
+	if encryptedMerkleLeaf.Version != EncryptionSecretbox {
```
jzila: Should this be `<`, so it'll handle future versions?

kbfscrypto/encrypted_data.go
```diff
@@ -228,3 +228,59 @@ func DecryptTLFCryptKeys(
...
+		return [24]byte{},
```
jzila: Not a big deal since the optimizer will probably take care of it, but this could just be `nonce`.

kbfscrypto/encrypted_data.go
```diff
@@ -228,3 +228,59 @@ func DecryptTLFCryptKeys(
...
+			errors.WithStack(InvalidNonceError{
+				Nonce: encryptedMerkleLeaf.Nonce})
+	}
+	copy(nonce[:], encryptedMerkleLeaf.Nonce)
+	return nonce, nil
+}
+
+// DecryptMerkleLeaf decrypts an EncryptedMerkleLeaf
+// using the given device private key and ephemeral public key.
```
jzila: Not a device key but a TLF key, right?

kbfscrypto/encrypted_data.go
strib: I don't follow how that would help handle future versions -- wouldn't `encryptedMerkleLeaf` be bigger in the case of a future version?  Since this version of the code supports exactly one version, it makes sense to me to use `!=`.  Later, if we introduce a new version, we'd have to change this code anyway to add a `switch` or something, right?

jzila: Yeah that was the idea: a future version would have a higher `encryptedMerkleLeaf.Version`, and presumably the code was going to be compatible. If there is no such presumption, then you can leave it as `!=`.

strib: Ah, yeah I don't know how we can make this code compatible with a new version it doesn't know about yet.  Would be awesome if we could, but I don't see it happening.  Ideally we'd make this into a popup error so the user would know to upgrade, but that seems out of scope for now.

PR #1591, opened 2018-05-23
folder_branch_ops: use fresh ctx when prefetching TLF root block

strib:
Otherwise the caller could cancel `ctx` after we return, and the block retriever will cancel the block fetch, and the bserver will log a warning about a context being canceled.

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -1628,8 +1628,14 @@ func (fbo *folderBranchOps) SetInitialHeadFromServer(
 	}()
 
 	if md.IsReadable() && fbo.config.Mode().PrefetchWorkers() > 0 {
-		// We `Get` the root block to ensure downstream prefetches occur.
-		_ = fbo.config.BlockOps().BlockRetriever().Request(ctx,
+		// We `Get` the root block to ensure downstream prefetches
+		// occur.  Use a fresh context, in case `ctx` is canceled by
+		// the caller before we complete.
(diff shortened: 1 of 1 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -1628,8 +1628,14 @@ func (fbo *folderBranchOps) SetInitialHeadFromServer(
 	}()
 
 	if md.IsReadable() && fbo.config.Mode().PrefetchWorkers() > 0 {
-		// We `Get` the root block to ensure downstream prefetches occur.
-		_ = fbo.config.BlockOps().BlockRetriever().Request(ctx,
+		// We `Get` the root block to ensure downstream prefetches
+		// occur.  Use a fresh context, in case `ctx` is canceled by
+		// the caller before we complete.
+		prefetchCtx := fbo.ctxWithFBOID(context.Background())
```
songgao: Should we still have a timeout here?

songgao: > Are you asking if we should add a timeout to the new context?

Yes that's what I meant. Sorry I didn't make it clear! I was a bit worried if not having a timeout here would cause goroutines hanging around when there are spotty network connections. I didn't look into the downstream code that uses this context though, so maybe we don't necessarily need it.

strib: Ah ok.  I think this will just end up using the normal block fetcher goroutines and not launch any new ones.

songgao: Great. Thanks!

PR #1593, opened 2018-05-29
libkbfs: stop QR early when an unverifiable TLF

strib:
Apparently some users have sigchains with revocations that are missing the root sequence number of when the revocation happened.  In this case, the service tells us they were revoked at seqno 0.  The API server rejects this as an invalid seqno, and gives back an opaque error type, which the mdserver sends back to the client.  A background process like QR treats this as a transient error and tries the TLF again, over and over.

This issue will go away once the service lands CORE-7806, but for now we can prevent this from happening forever by catching the problem early and returning a specific error type that will stop QR early.

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -1179,19 +1179,16 @@ func (fbm *folderBlockManager) reclaimQuotaInBackground() {
 		}
 
 		err := fbm.doReclamation(timer)
-		if e := errors.Cause(err); e != nil {
-			_, isWriteError := e.(WriteAccessError)
-			_, isFinalError := e.(kbfsmd.MetadataIsFinalError)
-			_, isRevokeError := e.(RevokedDeviceVerificationError)
-			if isWriteError || isFinalError || isRevokeError {
-				// If we can't write the MD, don't bother with the timer
-				// anymore. Don't completely shut down, since we don't
-				// want forced reclamations to hang.
-				timer.Stop()
-				timerChan = make(chan time.Time)
-				fbm.log.CDebugf(context.Background(),
-					"Permanently stopping QR due to error: %+v", err)
-			}
+		switch errors.Cause(err).(type) {
+		case WriteAccessError, kbfsmd.MetadataIsFinalError,
+			RevokedDeviceVerificationError:
+			// If we can't write the MD, don't bother with the timer
+			// anymore. Don't completely shut down, since we don't
+			// want forced reclamations to hang.
+			timer.Stop()
+			timerChan = make(chan time.Time)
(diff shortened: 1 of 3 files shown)


libkbfs/folder_block_manager.go
```diff
@@ -1179,16 +1179,19 @@ func (fbm *folderBlockManager) reclaimQuotaInBackground() {
...
-		_, isFinalError := err.(kbfsmd.MetadataIsFinalError)
-		if isWriteError || isFinalError {
-			// If we can't write the MD, don't bother with the timer
-			// anymore. Don't completely shut down, since we don't
-			// want forced reclamations to hang.
-			timer.Stop()
-			timerChan = make(chan time.Time)
-			fbm.log.CDebugf(context.Background(),
-				"Permanently stopping QR due to error: %+v", err)
+		if e := errors.Cause(err); e != nil {
```
jzila: Might as well do:
```go
switch errors.Cause(err).(type) {
    case WriteAccessError, kbfsmd.MetadataIsFinalError, RevokedDeviceVerificationError:
        // If we can't write the MD...
}
```

libkbfs/md_ops.go
```diff
@@ -221,6 +221,13 @@ func (md *MDOpsStandard) verifyKey(
 				"Skipping revoked key verification due to recursion")
 			return false, nil
 		}
+		if e.info.MerkleRoot.Seqno <= 0 {
+			md.log.CDebugf(ctx, "Can't verify an MD written by a revoked "+
+				"device if there's no valid root seqno to check: %d",
+				e.info.MerkleRoot.Seqno)
+			return false, err
```
songgao: It seems returning `err` here cause downstream failures because we can't verify keys. Do we not have legitimate revokes from earlier that didn't include seqno?

strib: @songgao: I don't follow what "downstream faillures" you're talking about.  But yes, as I said in the description, the API server returns 0 for some revokes, and that's a known issue that we have to work around.  I'll be posting another PR soon to decrease reliance on the field that can be 0.

songgao: Sorry downstream is probably not the right term. What I meant was when we return an error in this case, whatever that called `verifyKey` would get an error. Doesn't that cause read failures?

strib: Yes, but we were already returning an error in that case, because `FindNextMD` fails with an error instead (see keybase/client#12101).  This just gives a better error message and fixes the QR problem.

songgao: Ah I see; I misunderstood this PR as the fix making it work. Thanks.

PR #1594, opened 2018-05-29
keybase_service_base: use `FindNextMerkleRootAfterRevoke`

strib:
Instead of relying on the `PrevMerkleRootSigned` field in the revocation (which is the seqno that the client performing the revocation saw before pushing it), use the service to look up the exact merkle root that first contains the revocation.

This is more accurate and gets around the issue where some old revocations have a 0 root seqno.

Issue: keybase/client#12101

### DIFF
--- a/libkbfs/keybase_service_base.go
+++ b/libkbfs/keybase_service_base.go
@@ -182,7 +182,9 @@ func filterKeys(keys map[keybase1.KID]keybase1.PublicKeyV2NaCl) (
 	return verifyingKeys, cryptPublicKeys, kidNames, nil
 }
 
-func filterRevokedKeys(
+func (k *KeybaseServiceBase) filterRevokedKeys(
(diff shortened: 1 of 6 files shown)


libkbfs/keybase_service_base.go
```diff
@@ -197,7 +199,21 @@ func filterRevokedKeys(
 		var info revokedKeyInfo
 		if key.Base.Revocation != nil {
 			info.Time = key.Base.Revocation.Time
-			info.MerkleRoot = key.Base.Revocation.PrevMerkleRootSigned
+			res, err := k.userClient.FindNextMerkleRootAfterRevoke(ctx,
```
songgao: Is this server trust or does it verify the revoke signature is included in the merkle root? If former, do you think we can prefer `PrevMerkleRootSigned` when it's greater than zero, since it's from a client device?

PR #1595, opened 2018-05-30
keybase_service_base: don't refine revoked merkle root when prev=0

strib:
Also don't try to FindNextMD in that case either.  Old revoke sigchain entries just don't have enough information for us to check the merkle info when verifying writes by revoked devices, so we have to rely on server trust.


strib:
cc: @maxtaco @mmaxim @mlsteele: anyone want to try this out to make sure it fixes KBFS for you?

PR #1596, opened 2018-05-31
libkbfs: verify the merkle root and KBFS merkle nodes returned by FindNextMD

strib:
So far, we've been trusting the merkle data returned from the mdserver in `FindNextMD`.  In this PR, we start verifying it.  We call into the service to verify that the global root we have is legitimate and contains the KBFS root we expect it to.  And then we check the nodes of the KBFS merkle tree to make sure they contain the merkle leaf it is supposed to have.

Issue: KBFS-2954

### DIFF
--- a/libkbfs/merkle_checker.go
+++ b/libkbfs/merkle_checker.go
@@ -64,6 +64,8 @@ func verifyMerkleNodes(
 	config := merkle.NewConfig(
 		merkle.SHA512Hasher{}, 256, 512, kbfsmd.MerkleLeaf{})
 	tree := merkle.NewTree(mc, config)
+	// If any of the nodes returned by the server fail to match their
+	// expected hashes, `Find` will return an error.
 	foundLeaf, rootHash, err := tree.Find(ctx, merkle.Hash(tlfID.Bytes()))
 	if err != nil {
 		return err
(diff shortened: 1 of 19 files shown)


libkbfs/merkle_checker.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	if !rootHash.Eq(kbfsRoot.Hash) {
+		return errors.Errorf("Root hashes don't match: found=%v, rootHash=%v",
+			rootHash, kbfsRoot.Hash)
+	}
+	// If the checker returned all the nodes except the last one
+	// (which is the encoded leaf), we know they all verified and we
+	// reached the expected leaf.
+	if mc.nextNode != len(nodes)-1 {
```
songgao: Hmm apparently this passed the tests, but it seems if `LookupNode` hit the leaf, `nextNode` should've been updated  to `len(nodes)` assuming the last one in `nodes` is the leaf. Is it not?

libkbfs/merkle_checker.go
```diff
@@ -0,0 +1,94 @@
...
+	ctx context.Context, kbfsRoot *kbfsmd.MerkleRoot, nodes [][]byte,
+	tlfID tlf.ID) error {
+	// Verify the merkle nodes by pretending to look up the nodes
+	// using a merkle.Tree, which verifies all the nodes along the
+	// path of the lookup.
+	mc := &merkleChecker{kbfsRoot, nodes, 0}
+	config := merkle.NewConfig(
+		merkle.SHA512Hasher{}, 256, 512, kbfsmd.MerkleLeaf{})
+	tree := merkle.NewTree(mc, config)
+	foundLeaf, rootHash, err := tree.Find(ctx, merkle.Hash(tlfID.Bytes()))
```
songgao: Maybe add a comment here to mention this is where it's gonna error if hash of node returned by `LookupNode` doesn't match what we expect?

libkbfs/merkle_checker.go
strib: Yes, but `LookupNode` doesn't hit the leaf.  The leaf itself isn't really a separate Merkle node, it's actually just one of many values inside the lowest-most Merkle node.  The server returns it as the final node, though I guess that's not technically necessary since it can be extracted from the lowest-most Merkle node.  Oh well.

PR #1597, opened 2018-05-31
keybase_service_base: fill in revoked merkle info on demand

strib:
The `FindNextMerkleRootAfterRevoke` is expensive, so this commit only calls it for a key that's being explicitly asked for by the user (say, to read an MD update written by that key).

cc: @maxtaco 

Issue: KBFS-3009

### DIFF
--- a/libkbfs/keybase_service_base.go
+++ b/libkbfs/keybase_service_base.go
@@ -563,9 +563,6 @@ func (k *KeybaseServiceBase) ResolveImplicitTeamByID(
 func (k *KeybaseServiceBase) checkForRevokedVerifyingKey(
 	ctx context.Context, currUserInfo UserInfo, kid keybase1.KID) (
 	newUserInfo UserInfo, exists bool, err error) {
-	k.log.CDebugf(ctx, "Checking merkle info for user %s, revoked key %s",
-		currUserInfo.UID, kid)
-
 	for key, info := range currUserInfo.RevokedVerifyingKeys {
 		if !key.KID().Equal(kid) {
 			continue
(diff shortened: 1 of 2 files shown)


libkbfs/keybase_service_base.go
```diff
@@ -200,30 +200,16 @@ func (k *KeybaseServiceBase) filterRevokedKeys(
...
-					return nil, nil, nil, err
-				}
-				if res.Res != nil {
-					info.MerkleRoot = *res.Res
-				}
-			}
+			// If we don't have a prev seqno, then we already have the
+			// best merkle data we're going to get.
+			info.filledInMerkle = info.MerkleRoot.Seqno <= 0
+			info.sigChainLocation = key.Base.Revocation.SigChainLocation
```
jzila: lol `key.Base`

strib:
@jzila @songgao this is the PR that nojima said he wants to fast-track.

libkbfs/keybase_service_base.go
```diff
@@ -574,6 +560,53 @@ func (k *KeybaseServiceBase) ResolveImplicitTeamByID(
...
 }
 
+func (k *KeybaseServiceBase) checkForRevokedVerifyingKey(
+	ctx context.Context, currUserInfo UserInfo, kid keybase1.KID) (
+	newUserInfo UserInfo, exists bool, err error) {
+	k.log.CDebugf(ctx, "Checking merkle info for user %s, revoked key %s",
+		currUserInfo.UID, kid)
+
+	for key, info := range currUserInfo.RevokedVerifyingKeys {
+		if !key.KID().Equal(kid) {
```
jzila: Would be nice if `RevokedVerifyingKeys` was indexed by `KID` so we could just look it up rather than linear searching. But not a big deal.

strib: The problem is `KBPKIClient` does the opposite thing, where it has the `VerifyingKey` and needs to do a lookup in `RevokedVerifyingKeys`.  So there's not a great solution without reworking the interface APIs.  Probably a project better left for some other time.

jzila: Yeah agreed.

PR #1598, opened 2018-05-31
kbfsedits: introduce `NotificationMessage`

strib:
In the future, we'll write a json-encoded version of this message for each new operation, in a kbfs-edits chat channel.

Issue: KBFS-2992

### DIFF
--- a/kbfsedits/data_types.go
+++ b/kbfsedits/data_types.go
@@ -1,0 +1,81 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package kbfsedits
+
+import (
+	"time"
+
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/kbfs/kbfscrypto"
+	"github.com/keybase/kbfs/kbfsmd"
+	"github.com/keybase/kbfs/tlf"
+)
+
+// NotificationVersion is the numeric version of the edit notification
+// data structure.
+type NotificationVersion int
+
+const (
+	// NotificationV1 is unused, but represents the original,
+	// MD-ops-based edit notification list.
+	NotificationV1 NotificationVersion = 1
+	// NotificationV2 is the first version that stores JSON-encoded
+	// notifications in chat messages.
+	NotificationV2 NotificationVersion = 2
+)
+
+// NotificationOpType indicates the type of the operation that caused
+// the notification.
+type NotificationOpType string
+
+const (
+	// NotificationCreate is the type of an edit notification
+	// representing a new file or directory creation.
+	NotificationCreate NotificationOpType = "create"
+	// NotificationModify is the type of an edit notification
+	// representing a file modification.
+	NotificationModify NotificationOpType = "modify"
+	// NotificationRename is the type of an edit notification
+	// representing a file or directory getting renamed.
+	NotificationRename NotificationOpType = "rename"
+	// NotificationDelete is the type of an edit notification
+	// representing a file or directory getting deleted.
+	NotificationDelete NotificationOpType = "delete"
+)
+
+// EntryType indicates the type of the file that was edited.
+type EntryType string
+
+const (
+	// EntryTypeFile is for files that have been edited.  Note that
+	// covers both regular files and executables.
+	EntryTypeFile EntryType = "file"
+	// EntryTypeDir is for directories that have been edited.
+	EntryTypeDir EntryType = "dir"
+	// EntryTypeSym is for symlinks that have been edited.
+	EntryTypeSym EntryType = "sym"
+)
+
+// NotificationParams is used for op-type-specific data.
+type NotificationParams struct {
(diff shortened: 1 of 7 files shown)


strib:
(Note: when keybase/client#12146 is merged I'm going to vendor it into this PR, so that UID gets properly marshaled.)

kbfsedits/data_types.go
```diff
@@ -0,0 +1,81 @@
...
+	// EntryTypeSym is for symlinks that have been edited.
+	EntryTypeSym EntryType = "sym"
+)
+
+// NotificationParams is used for op-type-specific data.
+type NotificationParams struct {
+	OldFileName string `json:",omitempty"` // for renames
+	Offset      uint64 `json:",omitempty"` // for modifies
+	Length      uint64 `json:",omitempty"` // for modifies
+}
```
songgao: If I copy a file into KBFS and override an existing file, would that be an array of modifies, first one being truncate and following being writes?

kbfsedits/data_types.go
```diff
@@ -0,0 +1,81 @@
...
+	Filename string
+	Type     NotificationOpType
+	Time     time.Time // server-reported time
+	FileType EntryType
+	Revision kbfsmd.Revision
+	Device   kbfscrypto.VerifyingKey
+	UID      keybase1.UID
+	FolderID tlf.ID
+	Params   *NotificationParams `json:",omitempty"`
+}
```
songgao: It doesn't seem like we have a way to notify execution bit. Do we want to notify user about that?

tlf/id.go
```diff
@@ -227,6 +231,21 @@ func (id *ID) UnmarshalText(buf []byte) error {
 	return id.UnmarshalBinary(bytes)
 }
 
+// MarshalJSON implements the json.Marshaler interface for ID.
+func (id ID) MarshalJSON() ([]byte, error) {
+	return keybase1.Quote(id.String()), nil
+}
```
songgao: As an alternative where we don't use Quote, I think implementing encoding.TextMarshaler (which would just return `id.String()`) would also work.

kbfsedits/data_types.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// NotificationVersion is the numeric version of the edit notification
+// data structure.
+type NotificationVersion int
+
+const (
+	// NotificationV1 is unused, but represents the original,
+	// MD-ops-based edit notification list.
+	NotificationV1 NotificationVersion = 1
```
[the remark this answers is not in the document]
strib: Just for humans, it's not used anywhere.

kbfsedits/data_types.go
strib: Good question.  Internally, in the MD objects, that's described with a `createOp` operation that includes unreferenced blocks, indicating that a file was overwritten.  In this new protocol, we could change that potentially to show a delete followed by a create, to make it more explicit.

kbfsedits/data_types.go
strib: I mentioned that in the comment above on `EntryTypeFile` that I'm choosing to treat executables and files the same, since there's nothing in the design that requires MD notifications.  If you think that's a mistake and we should be future-proofing in case design changes their mind, let me know, but I couldn't think of a great reason to have it in now.

tlf/id.go
strib: I was being consistent with the KID mashaling. But I don't really follow your point.  Does the golang json encoder call the `TextMarshaler` interface if its available?  I don't see that mentioned in the docs.

kbfsedits/data_types.go
songgao: I see. Would there still be a bunch of "modifies" in this case? It seems it'd be nice if we could buffer them and collapse notifications. (Not in scope of this PR though.)

kbfsedits/data_types.go
songgao: OK sounds good; just making sure we've thought about that.

tlf/id.go
songgao: Yeah it's mentioned in the `encoding/json` doc: 

> If no MarshalJSON method is present but the value implements encoding.TextMarshaler instead, Marshal calls its MarshalText method and encodes the result as a JSON string. 

It might have been introduced later than the KID stuff was written. I think this seems cleaner but wouldn't cry if you choose to keep Quote here!

kbfsedits/data_types.go
strib: The write batching system already collapses them into ~1 second windows, so at first I was just planning to mirror that.  Doing anything else will require editing the chat history, or delaying sends for a while to batch stuff together.

The list we'll be returning to the frontend could be have more-collapsed ranges, but actually I think the frontend list will be way more collapsed, and will probably only contain filenames and timestamps. I could be wrong though.

tlf/id.go
strib: Oh I see, I don't know how I missed that when I searched the documentation, thanks.  Actually `tlf.ID` already has `MarshalText`, so I guess I'll just remove this part of the change entirely.

PR #1600, opened 2018-06-04
libkbfs: add a chat interface and an initial RPC implementation

strib:
This is untested, but it's also unused.  Plumbing it through will require a bit more work, plus a client<->kbfs commit dance, so I thought we should get this piece reviewed first so the PR doesn't get too unwieldy.

Issue: KBFS-2993

### DIFF
--- a/libkbfs/chat_rpc.go
+++ b/libkbfs/chat_rpc.go
@@ -162,7 +162,7 @@ func (c *ChatRPC) GetConversationID(
 		TopicType:        chatType,
 		TlfVisibility:    vis,
 		MembersType:      membersTypeFromTlfType(tlfType),
-		IdentifyBehavior: keybase1.TLFIdentifyBehavior_CHAT_CLI,
+		IdentifyBehavior: keybase1.TLFIdentifyBehavior_KBFS_CHAT,
 	}
 
 	// Try creating the conversation to get back the ID -- if the
@@ -185,7 +185,7 @@ func (c *ChatRPC) SendTextMessage(
 		TlfName:          string(tlfName),
 		TlfPublic:        tlfType == tlf.Public,
 		Body:             body,
-		IdentifyBehavior: keybase1.TLFIdentifyBehavior_CHAT_CLI,
+		IdentifyBehavior: keybase1.TLFIdentifyBehavior_KBFS_CHAT,
 	}
 	_, err := c.client.PostTextNonblock(ctx, arg)
 	return err
@@ -260,7 +260,7 @@ func (c *ChatRPC) ReadChannel(
 	arg := chat1.GetThreadLocalArg{
 		ConversationID:   convID,
 		Pagination:       pagination,
-		IdentifyBehavior: keybase1.TLFIdentifyBehavior_CHAT_CLI,
+		IdentifyBehavior: keybase1.TLFIdentifyBehavior_KBFS_CHAT,
 	}
 	res, err := c.client.GetThreadLocal(ctx, arg)
 	if err != nil {
--- a/libkbfs/chat_rpc.go
+++ b/libkbfs/chat_rpc.go
@@ -1,0 +1,296 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import (
+	"context"
+	"time"
+
+	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/client/go/logger"
+	"github.com/keybase/client/go/protocol/chat1"
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-framed-msgpack-rpc/rpc"
+	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
+)
+
+// ChatRPC is an RPC based implementation for chat.
+type ChatRPC struct {
+	config   Config
+	log      logger.Logger
+	deferLog logger.Logger
+	client   chat1.LocalInterface
+}
+
+var _ rpc.ConnectionHandler = (*ChatRPC)(nil)
+
+// NewChatRPC constructs a new RPC based chat implementation.
+func NewChatRPC(config Config, kbCtx Context) *ChatRPC {
+	log := config.MakeLogger("")
+	deferLog := log.CloneWithAddedDepth(1)
+	c := &ChatRPC{
+		log:      log,
+		deferLog: deferLog,
+		config:   config,
+	}
+	conn := NewSharedKeybaseConnection(kbCtx, config, c)
+	c.client = chat1.LocalClient{Cli: conn.GetClient()}
+	return c
+}
+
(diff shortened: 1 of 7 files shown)


libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,296 @@
...
+
+// ShouldRetryOnConnect implements the ConnectionHandler interface.
+func (c *ChatRPC) ShouldRetryOnConnect(err error) bool {
+	_, inputCanceled := err.(libkb.InputCanceledError)
+	return !inputCanceled
+}
+
+var _ Chat = (*ChatRPC)(nil)
+
+// Chat notes (will remove/rework these once the implementation is complete):
```
strib: No need to review these notes for correctness/grammar, I'll fix them in a later PR.  For now they're basically just notes to myself.

libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,296 @@
...
+	if tlfType == tlf.Public {
+		vis = keybase1.TLFVisibility_PUBLIC
+	}
+
+	arg := chat1.NewConversationLocalArg{
+		TlfName:          string(tlfName),
+		TopicType:        chatType,
+		TlfVisibility:    vis,
+		MembersType:      membersTypeFromTlfType(tlfType),
+		IdentifyBehavior: keybase1.TLFIdentifyBehavior_CHAT_CLI,
```
songgao: Would it make sense to have a new identify behavior, even though for now it's identical to CHAT_CLI?

libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	res, err := c.client.NewConversationLocal(ctx, arg)
```
songgao: Does this return existing conversation if we already created it before?

libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,296 @@
...
+	ctx context.Context, tlfName tlf.CanonicalName, tlfType tlf.Type,
+	convID chat1.ConversationID, body string) error {
+	arg := chat1.PostTextNonblockArg{
+		ConversationID:   convID,
+		TlfName:          string(tlfName),
+		TlfPublic:        tlfType == tlf.Public,
+		Body:             body,
+		IdentifyBehavior: keybase1.TLFIdentifyBehavior_CHAT_CLI,
+	}
+	_, err := c.client.PostTextNonblock(ctx, arg)
```
songgao: The comment on the interface says "synchronously". Is either this method call or the comment not accurate?

libkbfs/keybase_daemon.go
```diff
@@ -119,3 +119,15 @@ func (k keybaseDaemon) NewCrypto(config Config, params InitParams, ctx Context,
 	}
 	return crypto, nil
 }
+
+func (k keybaseDaemon) NewChat(
```
songgao: [My PR](https://github.com/keybase/kbfs/pull/1604) changes the init code a little bit. If we end up deciding merging that, this could probably go into "AdditionalProtocolCreaters".

songgao: Ah nvm I confused myself. This is a client not handler.

libkbfs/chat_rpc.go
strib: Yes, according to mike.  I'll clarify the comment above and move it here.

libkbfs/chat_rpc.go
strib: Good catch, the comment is wrong.  Mike recommended this RPC instead of the one I was originally planning and I forgot to update the comment.

libkbfs/chat_rpc.go
strib: Ok done: keybase/client#12277

I'll vendor it in here, once it's merged.

libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,298 @@
...
+// Finally an inbox layer that can read the server inbox, and also
+// checks the journal status, to return the top set of conversations
+// at any given time.  Maybe it also subscribes to inbox notifications
+// of some kind.
+
+func membersTypeFromTlfType(tlfType tlf.Type) chat1.ConversationMembersType {
+	if tlfType == tlf.SingleTeam {
+		return chat1.ConversationMembersType_TEAM
+	}
+	return chat1.ConversationMembersType_IMPTEAMNATIVE
```
jzila: what about `chat1.ConversationMembersType_IMPTEAMUPGRADE`?

libkbfs/chat_rpc.go
```diff
@@ -0,0 +1,298 @@
...
+//   * Read each channel until getting N updates for each writer
+//     * chat1.GetThreadLocal with pagination
+//     * No prev filled in on next pagination to go backward
+//   * How to reconcile renames, etc across channels?
+//     * It's hard to know if a long-ago writer updated a file, and
+//       later it was renamed by a prolific writer who made N more updates
+//       afterward.
+//     * For performance reasons, we probably just need to show the old
+//       update under the old file name.  Should be rare enough that
+//       it doesn't matter.
```
jzila: This sounds reasonable.

libkbfs/chat_rpc.go
strib: This behavior is what Mike told me to do.  I think `IMPTEAMUPGRADE` is a temporary core-internal thing, if I understand it correctly, just used during migration.

jzila: OK cool, thanks.

songgao: I think there's a difference between the teams that are upgraded from an old TLF (`impteamupgrade`) and teams that are born so (`impteamnative`). IIRC they'll stay like that. For example, my chat has 103 `impteamupgrade` conversations.

strib: Ah ok thanks.  But in this case, since it will always be new notifications, we never need to mark them as upgraded.

songgao: You mean notifications use conversations that are newly created from now, and they'll always be `impteamnative` since it's enabled globally now? Yeah that sounds good.

jzila: OK now I'm confused. I thought we're using existing conversations, just different `topic-type`s.

strib: I'm pretty sure a new topic type implies a new underlying conversation (though it could of course have the same name as some existing conversation).

strib: Analogy: private TLFs vs public TLFs.

jzila: I thought new conversation implied new key set. Topic types are just channels that are separate from the crypto aspect. Maybe @mmaxim could elucidate.

strib: Either way, the thing that matters is that it's a new conversation ID (every channel has a separate ID as well), and the members type is specific to the ID I believe.

mmaxim: Also, a "conversation ID" is equal to `SHA256((TeamID, Topic Type, Topic ID))`

PR #1604, opened 2018-06-07
fix protocol registration race and cleanup init code

songgao:
This would need to go in along with https://github.com/keybase/client/pull/12268

Circle CI will fail, but I tested an iOS build locally and it works.

### DIFF
--- a/libdokan/start.go
+++ b/libdokan/start.go
@@ -47,9 +47,9 @@ func startMounting(options StartOptions,
 func Start(options StartOptions, kbCtx libkbfs.Context) *libfs.Error {
 	// Hook simplefs implementation in.
 	createSimpleFS := func(
-		libkbfsCtx libkbfs.Context, cfg libkbfs.Config) (rpc.Protocol, error) {
+		libkbfsCtx libkbfs.Context, config libkbfs.Config) (rpc.Protocol, error) {
 		return keybase1.SimpleFSProtocol(
-			simplefs.NewSimpleFS(libkbfsCtx.GetGlobalContext(), cfg)), nil
+			simplefs.NewSimpleFS(libkbfsCtx.GetGlobalContext(), config)), nil
 	}
 	// Hook git implementation in.
 	shutdownGit := func() {}
@@ -65,7 +65,7 @@ func Start(options StartOptions, kbCtx libkbfs.Context) *libfs.Error {
 	}()
 
 	// Patch the kbfsParams to inject two additional protocols.
-	options.KbfsParams.AdditionalProtocolCreaters = []libkbfs.AdditionalProtocolCreater{
+	options.KbfsParams.AdditionalProtocolCreators = []libkbfs.AdditionalProtocolCreator{
 		createSimpleFS, createGitHandler,
 	}
 
--- a/libdokan/start.go
+++ b/libdokan/start.go
@@ -12,6 +12,7 @@ import (
 	"github.com/keybase/client/go/libkb"
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/go-framed-msgpack-rpc/rpc"
 	"github.com/keybase/kbfs/dokan"
 	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libgit"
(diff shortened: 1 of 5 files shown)


libkbfs/keybase_daemon_rpc.go
```diff
@@ -333,6 +336,9 @@ func (k *KeybaseDaemonRPC) OnConnect(ctx context.Context,
 		return err
 	}
 
+	// Set k.server only if err == nil.
+	k.server = server
```
strib: Should we set it to `nil` in `OnDisconnected`?  I don't think it would really affect things, but probably good practice in case we add protocols later that don't get added until after startup.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -255,47 +277,28 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
 	} else {
 		k.protocols = protocols
 	}
+
+	// If we are already connected, register these protocols.
+	k.lock.RLock()
```
strib: Is `k.lock` supposed to be protecting `k.protocols` as well?  If so, the above code modifies it outside the lock.  (Of course that problem was pre-existing, but since you added `k.lock` above `k.protocols` in the struct, it's now a little more confusing.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -255,47 +277,28 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
...
+			}
+		}
+	}
 }
 
 // OnConnect implements the ConnectionHandler interface.
 func (k *KeybaseDaemonRPC) OnConnect(ctx context.Context,
 	conn *rpc.Connection, rawClient rpc.GenericClient,
 	server *rpc.Server) error {
+	k.lock.Lock()
```
strib: I'm slightly nervous about holding a lock while we do the Hello RPC below (in case of deadlocks, etc) but I think it's pretty safe in this case and probably for the best.  So maybe just add a comment explaining why that's safe?

libkbfs/init.go
```diff
@@ -86,14 +89,9 @@ type InitParams struct {
...
 
-	// CreateSimpleFSInstance creates a SimpleFSInterface from config.
-	// If this is nil then simplefs will be omitted in the rpc api.
-	CreateSimpleFSInstance func(
-		*libkb.GlobalContext, Config) keybase1.SimpleFSInterface
-
-	// CreateGitHandlerInstance creates a KBFSGitInterface from config.
-	// If this is nil then git will be omitted in the rpc api.
-	CreateGitHandlerInstance func(Config) keybase1.KBFSGitInterface
+	// AdditionalProtocolCreaters are for addin additional protocols that we
```
strib: "addin" -> "adding"

libkbfs/keybase_daemon_rpc.go
songgao: good catch!

libkbfs/keybase_daemon_rpc.go
songgao: Good point. I ended up just holding a R lock at beginning and take the lock again at the end.

libdokan/start.go
```diff
@@ -45,19 +46,29 @@ func startMounting(options StartOptions,
...
+		handler, shutdownGit = libgit.NewRPCHandlerWithCtx(
+			libkbfsCtx, config, &options.KbfsParams)
+		return keybase1.KBFSGitProtocol(handler), nil
+	}
 	defer func() {
 		shutdownGit()
 	}()
 
+	// Patch the kbfsParams to inject two additional protocols.
+	options.KbfsParams.AdditionalProtocolCreaters = []libkbfs.AdditionalProtocolCreater{
```
jzila: s/Creater/Creator, here and everywhere else.

libkbfs/keybase_daemon.go
```diff
@@ -25,10 +26,18 @@ func (k keybaseDaemon) NewKeybaseService(config Config, params InitParams, ctx C
 		if err != nil {
 			return nil, err
 		}
+
+		var additionalProtocols []rpc.Protocol
+		for _, creater := range params.AdditionalProtocolCreaters {
```
jzila: s/creater/creator, here and everywhere else.

libfuse/start.go
```diff
@@ -85,19 +86,29 @@ func startMounting(ctx context.Context,
...
 	// Hook git implementation in.
 	shutdownGit := func() {}
-	options.KbfsParams.CreateGitHandlerInstance =
-		func(config libkbfs.Config) (i keybase1.KBFSGitInterface) {
-			i, shutdownGit = libgit.NewRPCHandlerWithCtx(
-				kbCtx, config, &options.KbfsParams)
-			return i
-		}
+	createGitHandler := func(
+		libkbfsCtx libkbfs.Context, config libkbfs.Config) (rpc.Protocol, error) {
```
jzila: above you use `cfg` and here you use `config`. I don't have a preference as to which one, but they should be consistent.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -250,64 +272,54 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
...
+	err = func() error {
+		k.lock.RLock()
+		defer k.lock.RUnlock()
+
+		for _, p := range k.protocols {
+			err := server.Register(p)
+			if err != nil {
+				if _, ok := err.(rpc.AlreadyRegisteredError); !ok {
+					return err
+				}
```
jzila: Instead of nested `if` statements here, I'd prefer a type switch.
```go
switch err.(type) {
case nil:
case rpc.AlreadyRegisteredError:
default:
    return err
}
return nil
```

libkbfs/keybase_daemon_rpc.go
```diff
@@ -250,64 +272,54 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
 	if protocols == nil {
 		return
 	}
+
+	k.lock.RLock()
```
jzila: this should be a `k.lock.Lock()`, right? Since you append or write to `k.protocols`.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -250,64 +272,54 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
...
-			if _, ok := err.(rpc.AlreadyRegisteredError); !ok {
+	server *rpc.Server) (err error) {
+	err = func() error {
+		k.lock.RLock()
+		defer k.lock.RUnlock()
+
+		for _, p := range k.protocols {
+			err := server.Register(p)
+			switch err.(type) {
+			case nil, rpc.AlreadyRegisteredError:
```
jzila: nice, even better than my suggestion!

PR #1605, opened 2018-06-07
keybase_service_base: fetch the merkle root following a reset

strib:
Just like we did for revokes, we also need to get the next Merkle root following an account reset (which effectively revokes all the old devices).

Tested it manually with a TLF containing a write by a pre-reset device.

Issue: KBFS-3029

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -36,6 +36,8 @@ type revokedKeyInfo struct {
 
 	// These fields never need copying.
 	sigChainLocation keybase1.SigChainLocation
+	resetSeqno       keybase1.Seqno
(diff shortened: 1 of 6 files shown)


libkbfs/data_types.go
```diff
@@ -36,6 +36,8 @@ type revokedKeyInfo struct {
 
 	// These fields never need copying.
 	sigChainLocation keybase1.SigChainLocation
+	resetSeqno       keybase1.Seqno
+	isReset          bool
```
songgao: It seems when `isReset` being `false` implies it's a revoke. Not a big deal but if/when you need to change this again, it'd be nice to have a comment here.

PR #1606, opened 2018-06-08
kbpagesd: use new way to specify protocols in params

strib:


strib:
Merging without CI, since this fixes CI.  I ran lint and vet locally.

PR #1611, opened 2018-06-12
chat: register for new channels and messages

strib:
(Based on #1610, which needs to be merged first.)

This registers KBFS for kbfs-edits notifications, and adds the interface methods needed to invalid local edit notification state (including gregor reconnects).

Issue: KBFS-2995

### DIFF
--- a/libkbfs/chat_rpc.go
+++ b/libkbfs/chat_rpc.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"github.com/keybase/client/go/libkb"
@@ -23,6 +24,9 @@ type ChatRPC struct {
 	log      logger.Logger
 	deferLog logger.Logger
 	client   chat1.LocalInterface
+
+	convLock sync.RWMutex
+	convCBs  map[string][]ChatChannelNewMessageCB
 }
 
 var _ rpc.ConnectionHandler = (*ChatRPC)(nil)
@@ -35,6 +39,7 @@ func NewChatRPC(config Config, kbCtx Context) *ChatRPC {
 		log:      log,
 		deferLog: deferLog,
 		config:   config,
+		convCBs:  make(map[string][]ChatChannelNewMessageCB),
 	}
 	conn := NewSharedKeybaseConnection(kbCtx, config, c)
 	c.client = chat1.LocalClient{Cli: conn.GetClient()}
@@ -42,7 +47,7 @@ func NewChatRPC(config Config, kbCtx Context) *ChatRPC {
 }
 
 // HandlerName implements the ConnectionHandler interface.
-func (ChatRPC) HandlerName() string {
+func (c *ChatRPC) HandlerName() string {
 	return "Chat"
 }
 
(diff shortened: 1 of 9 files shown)


libkbfs/chat_rpc.go
```diff
@@ -296,3 +309,180 @@ func (c *ChatRPC) ReadChannel(
...
+		return err
+	}
+	switch activityType {
+	case chat1.ChatActivityType_NEW_CONVERSATION:
+		// If we learn about a new conversation for a given TLF,
+		// attempt to route it to the TLF.
+		info := arg.Activity.NewConversation()
+		tlfType := tlf.Private
+		if info.Conv.Visibility == keybase1.TLFVisibility_PUBLIC {
+			tlfType = tlf.Public
```
jzila: Does this mean that notifications for public TLFs will be public?

I guess I only subscribe to them if I have a public TLF as a favorite, correct?

libkbfs/chat_rpc.go
```diff
@@ -296,3 +309,180 @@ func (c *ChatRPC) ReadChannel(
...
+		} else if info.Conv.MembersType == chat1.ConversationMembersType_TEAM {
+			tlfType = tlf.SingleTeam
+		}
+		tlfHandle, err := GetHandleFromFolderNameAndType(
+			ctx, c.config.KBPKI(), c.config.MDOps(), info.Conv.Name, tlfType)
+		if err != nil {
+			return err
+		}
+		c.config.KBFSOps().NewNotificationChannel(
+			ctx, tlfHandle, info.ConvID, info.Conv.Channel)
```
jzila: Why does the registration happen inside `NewNotificationChannel` instead of here? Seems like we're introducing a back-and-forth dependency.

libkbfs/chat_rpc.go
strib: Yeah, notifications are public, but as chris said today in the call, we'll only include your own notifications for your own public folder in the overall list.  The RPC inbox call doesn't even return public chats, we're going to have to add that in manually at a higher layer.

libkbfs/chat_rpc.go
strib: Because there might not yet be an FBO/edit history object for the corresponding TLF yet, and so we won't have anything to register.  And we don't want to create one if one isn't being used yet.  (We'll only create edit history objects for the top N chats in our inbox.)

This call is supposed to be a fire-and-forget type thing; any registration resulting from it should probably happen in a new goroutine.

jzila: Understood, thanks.

libkbfs/interfaces.go
```diff
@@ -2227,4 +2236,8 @@ type Chat interface {
 	ReadChannel(
 		ctx context.Context, convID chat1.ConversationID, startPage []byte) (
 		messages []string, nextPage []byte, err error)
+
+	// RegisterForMessages registers a callback that will be called
+	// for each new messages that reaches convID.
+	RegisterForMessages(convID chat1.ConversationID, cb ChatChannelNewMessageCB)
```
songgao: Is this gonna be used by SimpleFS methods, and KBFS in the GUI will call SimpleFS methods to get a list of recent activities?

Somehow I was under the impression that FE code will use the chat messages directly. But I guess calling SimpleFS to consolidate all information makes more sense.

strib: > Is this gonna be used by SimpleFS methods, and KBFS in the GUI will call SimpleFS methods to get a list of recent activities?

@songgao There will be another layer or two in between SimpleFS and this chat interface, I'm working on that now.  I think we want the same info to go to both SimpleFS, the GUI, and a special `.kbfs_` file, so it should be abstracted away.  Plus my sense from GUI folks is that we want to minimize the work the GUI has to do, so I'm putting as much of the logic into the Go code that I can.

PR #1612, opened 2018-06-12
tlf_journal: return an estimated end time for flushing bytes

strib:
This adds an exponentially-weighted moving average for estimated upload speed in bytes per second to each TLF journal, and adds an end time prediction to the journal status based on that estimate.  For the overall journal status, it uses the latest end-time estimate among all the TLFs.

If the status is obtained before any uploads for that TLF have completed, it predicts an arbitrary end estimate of 10 seconds from now.

If multiple TLFs are uploading at the same time, the hope is that their bandwidth estimates will decrease because of the increased competition for the TCP channel among the TLF journals.  If this doesn't work well, we change this to have a system-wide bandwidth estimate, but since all the TLFs upload data in parallel, it's not clear exactly how this should work.

Note that a traditional exponentially-weighted moving average would probably input sample measurements at constant time intervals, but since uploading is sporadic we can't really do that.  Instead, our decay is exponential in the number of samples we've taken, so it could have a long memory in terms of time, if there haven't been many uploads recently.

Issue: KBFS-3044

### DIFF
--- a/libkbfs/journal_server_util.go
+++ b/libkbfs/journal_server_util.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"errors"
+	"time"
 
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/kbfs/tlf"
@@ -60,6 +61,7 @@ func fillInJournalStatusUnflushedPaths(ctx context.Context, config Config,
 	unflushedPaths := make(chan []string, len(tlfIDs))
 	storedBytes := make(chan int64, len(tlfIDs))
 	unflushedBytes := make(chan int64, len(tlfIDs))
+	endEstimates := make(chan *time.Time, len(tlfIDs))
 	errIncomplete := errors.New("Incomplete status")
 	statusFn := func() error {
 		for tlfID := range statusesToFetch {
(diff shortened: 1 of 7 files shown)


libkbfs/journal_server_util.go
```diff
@@ -139,6 +143,15 @@ func fillInJournalStatusUnflushedPaths(ctx context.Context, config Config,
 		for ub := range unflushedBytes {
 			jStatus.UnflushedBytes += ub
 		}
+		// Pick the latest end estimate.
```
songgao: > latest

Just to make sure I understand correctly, there would be exactly 1 item in `endEstimates` when we reach here, right?

strib: That's correct.

libkbfs/journal_server_util.go
```diff
@@ -139,6 +143,15 @@ func fillInJournalStatusUnflushedPaths(ctx context.Context, config Config,
...
+		for e := range endEstimates {
+			if e != nil {
+				if jStatus.EndEstimate == nil ||
```
jzila: These two `if` statements can be combined:
```go
if e != nil && (jStatus.EndEstimate == nil || jStatus.EndEstimate.Before(*e)) {
    jStatus.EndEstimate = e
}
```

PR #1616, opened 2018-06-13
turn on a bunch of things for mobile

songgao:


libkbfs/modes.go
```diff
@@ -283,32 +283,27 @@ func (mc modeConstrained) RekeyWorkers() int {
...
 func (mc modeConstrained) ConflictResolutionEnabled() bool {
-	// TODO: turn this on once we allow mobile writes.
-	return false
+	return true
 }
 
 func (mc modeConstrained) QuotaReclamationEnabled() bool {
-	// TODO: turn this on once we allow mobile writes.
-	return false
+	return true
```
jzila: Do we need QR on mobile?

songgao: I wonder how expensive it is. The phone could be the only device that user uses.

strib: Yeah as discussed today, let's keep this off for now until we add a new mode method to indicate that this device should only self-QR.  Also need to verify that the QR timer doesn't cause app wakeups.

PR #1618, opened 2018-06-13
tlf_handle_resolve: don't use finalized TLF IDs from iteam info

strib:
When chats got migrated to implicit teams, the current TLF ID got written into the iteam sigchain.  KBFS uses that ID in the sigchain, if it's available, in preparation for when we turn on KBFS iteams. However, if the TLF gets finalized (because a user resets), the old ID remains in the sigchain and KBFS will use it even when looking up the non-finalized TLF ID.

We can fix this more permanently when we migrate existing KBFS TLFs to iteams; at that point we can override the TLF ID using a new sigchain link.

Issue: KBFS-3045

### DIFF
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -238,6 +238,23 @@ func makeTlfHandleHelper(
 		panic(fmt.Sprintf("Unknown TLF type: %s", t))
 	}
 
+	if !isImplicit && tlfID != tlf.NullID && finalizedInfo == nil &&
+		idGetter != nil {
+		// Implicit team chat migration might have set the ID to an
+		// old folder that has since been reset.  Check with the
+		// server to see if that has happened, and if so, don't use
+		// that ID.  When we migrate existing TLFs to iteams, we can
+		// probably override the sigchain link with a new one
+		// containing the correct TLF ID.
+		latestHandle, err := idGetter.GetLatestHandleForTLF(ctx, tlfID)
+		if err != nil {
+			return nil, err
+		}
+		if latestHandle.IsFinal() {
+			tlfID = tlf.NullID
+		}
+	}
+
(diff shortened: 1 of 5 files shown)


strib:
@jzila: any chance you can test this out and see if it gives you a consistent view of a TLF on all devices?

jzila:
Yup, give me a bit.

libkbfs/tlf_handle_resolve.go
```diff
@@ -756,7 +764,31 @@ func parseTlfHandleLoose(
...
+			_, finalizedInfo := extensionList.Splat()
+
+			if finalizedInfo == nil && idGetter != nil {
+				// Implicit team chat migration might have set the ID to an
+				// old folder that has since been reset.  Check with the
+				// server to see if that has happened, and if so, don't use
+				// that ID.  When we migrate existing TLFs to iteams, we can
+				// probably override the sigchain link with a new one
+				// containing the correct TLF ID.
+				latestHandle, err := idGetter.GetLatestHandleForTLF(
```
songgao: I forgot how often `parseTlfHandleLoose` is called; is this gonna slow down the happy path?

strib: This is on the path every time a TLF name is resolved into a handle, and it will slow things down with a round-trip to the mdserver in the case where there's an implicit team corresponding to the TLF, and that team has a TLF ID in its sigchain.  I'm not sure how often that will be the case -- until a few days ago I thought it wouldn't happen until we turned on iteam migration for KBFS, but I was wrong.

I will spend a few minutes thinking about whether I can cache this effectively somehow.

strib:
@songgao @jzila: I worked this PR on @songgao's prompting to hit the cache whenever possible, on the premise that any IDs from the handle->ID cache came from the server.  Please take another look, and ideally @jzila you could test this version of it too.

PR #1625, opened 2018-06-15
chat: implement local chat server, and tweak interface a bit

strib:
This is in preparation for having notification tests in libkbfs, though it is currently untested.

Issue: KBFS-2996

### DIFF
--- a/libkbfs/chat_local.go
+++ b/libkbfs/chat_local.go
@@ -6,74 +6,247 @@ package libkbfs
 
 import (
 	"context"
+	"fmt"
+	"sort"
+	"sync"
+	"time"
 
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/chat1"
+	"github.com/keybase/kbfs/kbfscrypto"
 	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
 )
 
-// ChatLocal is a local implementation for chat.
-//
-// TODO: fill in the logic to allow for testing.
-type ChatLocal struct {
+type convLocal struct {
+	convID   chat1.ConversationID
+	chanName string
+	messages []string
+	cbs      []ChatChannelNewMessageCB
+	mtime    time.Time
+}
+
+type convLocalByIDMap map[string]*convLocal
+
+type convLocalByTypeMap map[tlf.Type]convLocalByIDMap
+
+type convLocalByNameMap map[tlf.CanonicalName]convLocalByTypeMap
+
+type newConvCB func(context.Context, *TlfHandle, chat1.ConversationID, string)
+
+type chatLocalSharedData struct {
+	lock          sync.Mutex
+	newChannelCBs []newConvCB
+	convs         convLocalByNameMap
+	convsByID     convLocalByIDMap
+}
+
+// chatLocal is a local implementation for chat.
+type chatLocal struct {
 	config   Config
 	log      logger.Logger
 	deferLog logger.Logger
+	data     *chatLocalSharedData
 }
 
-// NewChatLocal constructs a new local chat implementation.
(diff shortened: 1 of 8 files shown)


libkbfs/chat_local.go
```diff
@@ -6,74 +6,247 @@ package libkbfs
...
+	messages []string
+	cbs      []ChatChannelNewMessageCB
+	mtime    time.Time
+}
+
+type convLocalByIDMap map[string]*convLocal
+
+type convLocalByTypeMap map[tlf.Type]convLocalByIDMap
+
+type convLocalByNameMap map[tlf.CanonicalName]convLocalByTypeMap
```
jzila: Why did you choose to do name -> type -> ID instead of type -> name -> ID? Seems better for locality to go in order of increasing cardinality.

libkbfs/chat_local.go
```diff
@@ -6,74 +6,247 @@ package libkbfs
...
+			for _, conv := range byID {
+				if conv.mtime.After(hAndT.mtime) {
+					hAndT.mtime = conv.mtime
+				}
+			}
+			handlesAndTimes = append(handlesAndTimes, hAndT)
+		}
+	}
+
+	sort.Slice(handlesAndTimes, func(i, j int) bool {
```
jzila: Why are you using `sort.Slice` here, but in other places you implement the full `sort.Interface`? `sort.Slice` uses reflection so it'll be slower.

libkbfs/chat_local.go
```diff
@@ -6,74 +6,247 @@ package libkbfs
...
+		// Reverse sort so newest conversation is at index 0.
+		return handlesAndTimes[i].mtime.After(handlesAndTimes[j].mtime)
+	})
+	for i := 0; i < maxChats && i < len(handlesAndTimes); i++ {
```
jzila: I'd prefer `i < len(handlesAndTimes)` to be first, since it's the primary range we care about. `maxChats` is just used to clip.

libkbfs/chat_local.go
```diff
@@ -6,74 +6,247 @@ package libkbfs
...
 	ctx context.Context, tlfName tlf.CanonicalName, tlfType tlf.Type,
 	chatType chat1.TopicType) (
 	convIDs []chat1.ConversationID, channelNames []string, err error) {
-	return nil, nil, nil
+	if chatType != chat1.TopicType_KBFSFILEEDIT {
+		panic(fmt.Sprintf("Bad topic type: %d", chatType))
+	}
+
+	c.data.lock.Lock()
+	defer c.data.lock.Unlock()
```
jzila: These can be `RLock`/`RUnlock`.

libkbfs/chat_local.go
```diff
@@ -6,74 +6,247 @@ package libkbfs
...
 }
 
 // ReadChannel implements the Chat interface.
-func (c *ChatLocal) ReadChannel(
+func (c *chatLocal) ReadChannel(
 	ctx context.Context, convID chat1.ConversationID, startPage []byte) (
 	messages []string, nextPage []byte, err error) {
-	return nil, nil, nil
+	c.data.lock.Lock()
+	defer c.data.lock.Unlock()
```
jzila: These can be `RLock`/`RUnlock`.

libkbfs/chat_local.go
strib: It was arbitrary.  I'll change it.

libkbfs/chat_local.go
strib: Because it was less code and I didn't have a need for a real type on the list.  But if you have a preference, I'll change it.

jzila: Thanks for changing it! Yeah I'd prefer to use `sort.Sort` pretty much always.

PR #1626, opened 2018-06-15
kbfsedits: add a class that can construct per-TLF edit histories

strib:
This introduces `kbfsedits.TlfHistory`, which receives raw notification strings from a caller for each writer in a TLF, and constructs the edit history from each writer (up to `maxEditsPerWriter` edits each), up to `maxWritersPerHistory` writers per TLF.

The collected sets of per-writer edits in the history are sorted in reverse order, by the most recent revision from each writer.  Within each set, a writer's edits are sorted in reverse order by revision.

The constructed history only includes creates and modifies of files (no directories or symlinks).  If a file was modified and later renamed, the notification for the modify shows the new file name.  If a file was modified but later deleted, it doesn't appear in the history.  Each writer only has the most recent modification or create for each file.

For performance reasons, it's possible that not all deletes or renames will be processed, because they might have been done by a writer prior to their `maxEditsPerWriter` edits, and so those won't be reflected in the modifications of any of the other writers.

The idea is that each `libkbfs.FolderBranchOps`, subscribe to the kbfs-edits chat channels for the TLF's notifications, and fill in these histories with notifications as needed.  Then a new, yet-to-be-written object in `kbfsedits` will collate together a complete history for the logged-in user, using the individual `TlfHistory` instances corresponding to the most-recently-written TLFs.

NOTE: This doesn't yet take journal revisions into account.  That support will come in a future commit so it's easier to review.

Issue: KBFS-2996

### DIFF
--- a/kbfsedits/tlf_history.go
+++ b/kbfsedits/tlf_history.go
@@ -22,7 +22,7 @@ type writerNotifications struct {
 	notifications notificationsByRevision
 }
 
-// writersByRevision sort sets of per-writer notifications in reverse
+// writersByRevision sorts sets of per-writer notifications in reverse
 // order by the revision of the latest notification for each writer.
 type writersByRevision []*writerNotifications
 
@@ -122,8 +122,6 @@ func (th *TlfHistory) AddNotifications(
 			newEdits = append(newEdits, revMsg)
 		}
 	}
-	sort.Sort(newEdits)
-	newEdits = newEdits.uniquify()
 
 	th.lock.Lock()
 	defer th.lock.Unlock()
@@ -160,8 +158,8 @@ type recomputer struct {
 	numProcessed  map[string]int             // writer name -> num
 }
 
-func newRecomputer() recomputer {
-	return recomputer{
+func newRecomputer() *recomputer {
+	return &recomputer{
 		byWriter:      make(map[string]*writerNotifications),
 		modifiedFiles: make(map[string]map[string]bool),
 		fileEvents:    make(map[string]fileEvent),
@@ -169,7 +167,7 @@ func newRecomputer() recomputer {
 	}
 }
 
-// processNotification add the notification to the recomputer's
+// processNotification adds the notification to the recomputer's
 // history if it is a create/modify for a file that hasn't yet been
 // deleted.  If the file is renamed in a future revision, the added
 // notification has the new name of the file.  processNotification
@@ -178,7 +176,7 @@ func newRecomputer() recomputer {
 //
 // It returns true if it has added enough notifications for the given
 // writer, and the caller should not send any more for that writer.
-func (r recomputer) processNotification(
+func (r *recomputer) processNotification(
 	writer string, notification NotificationMessage) (doTrim bool) {
 	filename := notification.Filename
 	r.numProcessed[writer]++
--- a/kbfsedits/tlf_history.go
+++ b/kbfsedits/tlf_history.go
@@ -1,0 +1,343 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package kbfsedits
+
+import (
+	"container/heap"
+	"encoding/json"
+	"sort"
+	"sync"
+)
+
+const (
+	// The max number of edits needed for each writer.
+	maxEditsPerWriter    = 10
+	maxWritersPerHistory = 10
+)
+
+type writerNotifications struct {
+	writerName    string
+	notifications notificationsByRevision
+}
+
+// writersByRevision sort sets of per-writer notifications in reverse
+// order by the revision of the latest notification for each writer.
+type writersByRevision []*writerNotifications
+
+func (wbr writersByRevision) Len() int {
+	return len(wbr)
+}
+
+func (wbr writersByRevision) Less(i, j int) bool {
+	// Some revisions come before no revisions.
+	iHasZero := len(wbr[i].notifications) == 0
+	jHasZero := len(wbr[j].notifications) == 0
+	if jHasZero {
+		return iHasZero
+	} else if iHasZero {
+		return false
+	}
+
+	// Reverse sort, so latest revisions come first.
+	return wbr[i].notifications[0].Revision > wbr[j].notifications[0].Revision
+}
+
+func (wbr writersByRevision) Swap(i, j int) {
+	wbr[i], wbr[j] = wbr[j], wbr[i]
+}
+
+func (wbr *writersByRevision) Push(x interface{}) {
+	wn := x.(*writerNotifications)
+	*wbr = append(*wbr, wn)
+}
+
+func (wbr *writersByRevision) Pop() interface{} {
+	// The item to remove is the last item; heap has already swapped
+	// it to the end.
+	old := *wbr
+	n := len(old)
+	item := old[n-1]
+	*wbr = old[0 : n-1]
+	return item
+}
+
+// TlfHistory maintains a history of the last N file edits from each
+// writer in the TLF.
+//
+// There will be two users of a TlfHistory instance:
+//
+//   * One user (likely something outside of the kbfsedits package,
+//     e.g. libkbfs.folderBranchOps) will read notifications from the
+//     corresponding TLF and add them to this history.  After adding a
+//     batch or several batches of messages, it should call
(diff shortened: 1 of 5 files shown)


kbfsedits/tlf_history_test.go
```diff
@@ -0,0 +1,261 @@
...
+	require.NoError(t, err)
+	checkTlfHistory(t, th, expected)
+
+	// Then input the rest, and we'll have a complete set.
+	err = th.AddNotifications(aliceName, aliceMessages[:maxEditsPerWriter/2])
+	require.NoError(t, err)
+	expected = writersByRevision{
+		{aliceName, allExpected},
+	}
+	checkTlfHistory(t, th, expected)
```
jzila: Nice tests!

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,356 @@
...
+
+		if len(wn.notifications) == maxEditsPerWriter {
+			// We have enough edits for this user.
+			return true
+		}
+	case NotificationRename:
+		// If the file already has a final event, move that to the old
+		// filename.  Otherwise, this is the final event.
+		if hasEvent {
+			r.fileEvents[notification.Params.OldFilename] = event
```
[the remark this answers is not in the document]
strib: Only renames have `notification.Params.OldFilename`, and it's the previous name the file was renamed from.  `eventFilename` is the original new name of the file (before being replaced by whatever's in `lastEvents`).

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,354 @@
...
+		filename = event.newName
+	}
+
+	// Keep only the creates and modifies for non-deleted files,
+	// but remember the renames and deletes.
+	switch notification.Type {
+	case NotificationCreate, NotificationModify:
+		// Disregard any file that's already been deleted.
+		if hasEvent && event.delete {
+			return false
```
songgao: If a file is deleted then recreated, does it come here or hit a different path?

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,354 @@
...
+		doTrim := r.processNotification(nextWriter, nextNotification)
+
+		// Remove that notification, and fix the heap because this
+		// writer has a different newest revision.
+		if doTrim {
+			// Trim all earlier revisions because they won't be needed
+			// for the cached history.
+			numProcessed := r.numProcessed[nextWriter]
+			th.byWriter[nextWriter].notifications =
+				th.byWriter[nextWriter].notifications[numProcessed:]
```
songgao: Is this essentially to make `th.byWriter[nextWriter].notifications` an empty slice? If so, why not just make it `nil`?

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,354 @@
...
+		if wn == nil || len(wn.notifications) < maxEditsPerWriter {
+			writersWhoNeedMore[writerName] = true
+		}
+	}
+	sort.Sort(history)
+	// Garbage-collect any writers that don't appear in the history.
+	for i := maxWritersPerHistory; i < len(history); i++ {
+		delete(th.byWriter, history[i].writerName)
+		delete(writersWhoNeedMore, history[i].writerName)
+	}
```
songgao: It might be better to move this `for` block into the `if` block below, not for performance reasons but for readability. It took me a while to associate them :)

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,354 @@
...
+	th.computed = true
+	th.cachedHistory = history
+	return history, writersWhoNeedMore
+}
+
+func (th *TlfHistory) getHistoryIfCached() (
+	cached bool, history writersByRevision) {
+	th.lock.RLock()
+	defer th.lock.RUnlock()
+	return th.computed, th.cachedHistory
```
songgao: Should this return `th.computed, nil` if `th.computed == false` because of the method name? Not a big deal though since it's only used in `getHistory()` below.

kbfsedits/tlf_history.go
strib: > If a file is deleted then recreated, does it come here or hit a different path?

Remember this is processing events backwards.  So we see the re-create first, which gets added to the set of notifications, and _then_ we see the delete.  Everything earlier than the delete matching that filename gets ignored when it hits this code.

kbfsedits/tlf_history.go
strib: No, `numProcessed` could be more than 0 (and will be, in the common case).  So this is removing items from the end of the slice that were not used in the making of the history, not necessarily making it `nil`.

strib: Actually this might be backwards, this seems to be removing items from the beginning of the slice, the ones that were actually processed.  I'm going to try to fix that and add a test.

kbfsedits/tlf_history.go
songgao: I see; that makes sense. Thanks!

kbfsedits/tlf_history.go
songgao: > No, numProcessed could be more than 0 (and will be, in the common case).

By reading the code, I was under the impression that `numProcessed` will always be equal to `len(th.byWriter[nextWriter].notifications)`, making this an empty slice. (Since the `numProcessed` marks the start of the slicing).

strib: It's not, because of this: https://github.com/keybase/kbfs/blob/15309b94d93e4d6d3b900e0b0d5374ca53ef7749/kbfsedits/tlf_history.go#L232-L234

When we get enough notifications for the user, we return `doTrim=true`, which causes the caller to truncate the rest of the notifications away from the slice.  Or that's what it will do, with my new commit.

kbfsedits/tlf_history.go
```diff
@@ -311,13 +311,13 @@ func (th *TlfHistory) recomputeLocked() (
...
-		delete(th.byWriter, history[i].writerName)
-		delete(writersWhoNeedMore, history[i].writerName)
-	}
 	if len(history) > maxWritersPerHistory {
 		history = history[:maxWritersPerHistory]
+		// Garbage-collect any writers that don't appear in the history.
+		for i := maxWritersPerHistory; i < len(history); i++ {
+			delete(th.byWriter, history[i].writerName)
+			delete(writersWhoNeedMore, history[i].writerName)
+		}
```
songgao: I think this still needs to be above the `history = history[:maxWritersPerHistory]` right?

kbfsedits/tlf_history.go
```diff
@@ -289,7 +289,7 @@ func (th *TlfHistory) recomputeLocked() (
 			// for the cached history.
 			numProcessed := r.numProcessed[nextWriter]
 			th.byWriter[nextWriter].notifications =
-				th.byWriter[nextWriter].notifications[numProcessed:]
+				th.byWriter[nextWriter].notifications[:numProcessed]
```
songgao: Ah; that makes more sense! I was confused by the earlier version.

kbfsedits/tlf_history.go
strib: Oops, excellent call.

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,362 @@
...
+	oldLen := len(wn.notifications)
+	newEdits = append(newEdits, wn.notifications...)
+	sort.Sort(newEdits)
+	wn.notifications = newEdits.uniquify()
+	if len(wn.notifications) == oldLen {
+		// No new messages.
+		return nil
+	}
+	if !existed {
+		wn.writerName = writerName
```
songgao: `writerName` is already populated above right?

kbfsedits/tlf_history.go
```diff
@@ -0,0 +1,362 @@
...
+		if r.modifiedFiles[writer][filename] {
+			return false
+		}
+
+		wn, ok := r.byWriter[writer]
+		if !ok {
+			wn = &writerNotifications{writer, nil}
+			r.byWriter[writer] = wn
+		}
+		wn.writerName = writer
```
songgao: Same here; is it necessary to assign this again?

kbfsedits/tlf_history.go
strib: I must have had a reason but now I can't remember it, and the tests pass without it, so I'll fix in a new PR.  Thanks for giving it another look!

PR #1628, opened 2018-06-18
libkbfs: power edit histories with v2 notifications

strib:
(Circle will fail until keybase/client#12429 is merged.)

This PR changes `folderBranchOps` to monitor the kbfs-edits chat channels corresponding to the TLF, and use that to power the TLF edit history for the folder.  Each `folderBranchOps` now has a `kfsbedits.TlfHistory` that it keeps up-to-date.

It also introduces a new `kbfsedits.UserHistory` object, which sorts all the TLFs in descending order by the timestamp of the last edit, and serves per-TLF edit histories using the structs defined in keybase/client#12429.

When a user logs out, it clears the edit history for all non-public TLFs.

When KBFS reconnects to either the Gregor or Keybase service, it may have missed chat updates, so it has to re-fetch all the messages for all the channels from scratch.

The per-TLF `.kbfs_edit_history` file is now powered by these chat channels, but since we haven't enabled sending notifications in non-test environments yesterday, it will always appear empty.  That's ok; I don't know of any users of that file.

Coming in a future PR: Querying the chat inbox to figure out which TLFs need to be monitored, to power the overall per-user edit history.

Issue: KBFS-2996


libkbfs/modes.go
```diff
@@ -319,6 +334,11 @@ func (mc modeConstrained) TLFEditHistoryEnabled() bool {
 	return false
 }
 
+func (mc modeConstrained) SendEditNotificationsEnabled() bool {
+	// TODO: turn this on once we allow mobile writes.
```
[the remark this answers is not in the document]
songgao: Oh good point. We'll need that for the journal stuff too. I'll include that in #1616 .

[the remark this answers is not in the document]
songgao: Oh I confused myself. This is not in libkbfs yet. So I'll merge my PR as-is. Thanks for the heads up!

strib:
FYI, the jenkins failures on this (and all dependent PRs) are due to https://keybase.atlassian.net/browse/CORE-8174, which will be fixed soon.

kbfsedits/user_history.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/keybase/kbfs/tlf"
+)
+
+const (
+	// The max number of TLF writer clusters to return in a user history.
+	maxClusters = 10
+)
+
+type tlfKey struct {
+	tlfName tlf.CanonicalName
```
songgao: In case of SBS, will we end up with two `tlfWithHistory` objects?

strib: Well, `folderBranchOps.monitorEditsChat` only gets the name once at the beginning of its loop, and uses that same name every time it calls `UpdateHistory`.  So no, I don't see how there would be two `TlfHistory` objects for the same TLF, but the existing one will continue to have the old name I think, even after the assertion is proven.  The next KBFS restart will fix it.

libkbfs/folder_branch_ops.go
```diff
@@ -6721,24 +6732,13 @@ func (fbo *folderBranchOps) GetUpdateHistory(ctx context.Context,
...
 	lState := makeFBOLockState()
-	head, err := fbo.getMDForReadNeedIdentify(ctx, lState)
-	if err != nil {
-		return nil, err
-	}
-
-	return fbo.editHistory.GetComplete(ctx, head)
+	md, _ := fbo.getHead(lState)
+	name := md.GetTlfHandle().GetCanonicalName()
+	return fbo.config.UserHistory().GetTlfHistory(name, fbo.id().Type()), nil
```
songgao: @strib what about here where we use the name from head handle to get TLF history? Would it fail if the name in there was before the SBS?

strib: Yes, you're right it would -- though that only affects `/keybase/type/name/.kbfs_edit_history`, not the full user history that will appear in the widget.

If you think it's important, it probably won't be too hard to fix up the names when the handles change, I just didn't feel like complicating things more when I was writing this the first time through.

songgao: Got it. That's OK; just wanted to make sure this doesn't affect the main path. I'm OK with leaving this as-is. Thanks!

libkbfs/folder_branch_ops.go
```diff
@@ -475,7 +488,6 @@ func (fbo *folderBranchOps) Shutdown(ctx context.Context) error {
 	fbo.merkleFetches.Wait(ctx)
 	fbo.cr.Shutdown()
 	fbo.fbm.shutdown()
-	fbo.editHistory.Shutdown()
```
jzila: Nice cleanup with all the shutdowns :)

libkbfs/folder_branch_ops.go
```diff
@@ -5711,7 +5727,7 @@ func (fbo *folderBranchOps) SyncFromServer(ctx context.Context,
 	if err := fbo.fbm.waitForDeletingBlocks(ctx); err != nil {
 		return err
 	}
-	if err := fbo.editHistory.Wait(ctx); err != nil {
+	if err := fbo.editActivity.Wait(ctx); err != nil {
```
jzila: I know this isn't new here, but it doesn't look like we have any starvation protection. So as long as edit activity keeps coming in faster than we can process it, another client on a TLF could starve all other clients.

libkbfs/folder_branch_ops.go
```diff
@@ -6852,12 +6861,178 @@ func (fbo *folderBranchOps) KickoffAllOutstandingRekeys() error {
...
+		fbo.log.CDebugf(nil, "Asking for an edit re-init after reconnection")
+		fbo.editActivity.Add(1)
+		fbo.editChannels <- editChannelActivity{nil, "", ""}
+	}
+}
+
+func (fbo *folderBranchOps) receiveNewEditChat(
+	convID chat1.ConversationID, message string) {
+	fbo.editActivity.Add(1)
+	fbo.editChannels <- editChannelActivity{convID, "", message}
```
jzila: What happens if we reach 100 `editChannels`? This will block...

libkbfs/folder_branch_ops.go
```diff
@@ -6852,12 +6861,178 @@ func (fbo *folderBranchOps) KickoffAllOutstandingRekeys() error {
...
+		if finishActivity {
+			fbo.editActivity.Done()
+		}
+
+		select {
+		case <-fbo.shutdownChan:
+			fbo.log.CDebugf(ctx, "Shutting down chat monitoring")
+			return
+		case a := <-fbo.editChannels:
+			finishActivity = true
```
jzila: We set this to true once and then never reset it to `false`.

IMO it'd be more readable to have this inside a `func()` with `defer fbo.editActivity.Done()`

libkbfs/modes.go
```diff
@@ -252,6 +261,12 @@ func (mso modeSingleOp) TLFEditHistoryEnabled() bool {
 	return false
 }
 
+func (mso modeSingleOp) SendEditNotificationsEnabled() bool {
+	// We don't want git, or other single op writes, showing up in the
+	// notification history.
```
jzila: That said, long term this might be nice to consider.

kbfsedits/user_history.go
```diff
@@ -0,0 +1,210 @@
...
+	uh.lock.Lock()
+	defer uh.lock.Unlock()
+	if currIndex, ok := uh.tlfs.indices[key]; ok {
+		uh.tlfs.histories[currIndex].history = history
+	} else {
+		uh.tlfs.indices[key] = len(uh.tlfs.indices)
+		uh.tlfs.histories = append(
+			uh.tlfs.histories, tlfWithHistory{key, history})
+	}
+	sort.Sort(uh.tlfs)
```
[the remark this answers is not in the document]
strib: > I wonder if this should be a heap too...

Actually on further thought, this doesn't need to be sorted at all -- the sorting only needs to happen in `Get()`.  I'm going to turn this into a simple map.

libkbfs/folder_branch_ops.go
strib: True, but a) a well-behaved client with batching will only push one edit per TLF per second, and b) the starvation would only happen in `SyncFromServer`, which outside of testing I think is only used in single-op mode, which wouldn't subscribe to notifications anyway.

libkbfs/folder_branch_ops.go
strib: > What happens if we reach 100 editChannels? This will block...

It will block when a single FBO has 100 separate unprocessed edit notifications in the queue, but only until it finishes processing one.  The blocking has to happen somewhere, but I chose 100 pretty arbitrarily.  Let me know if you have a number or blocking plan you think is more reasonable.

libkbfs/modes.go
strib: If you think we should keep them in the history, maybe with a special type in the notification so we can ignore them in the edit history.  I was worried this would flood the channel with irrelevant info and force us to scroll back through more messages than necessary to reconstruct the history.

It might be better in a completely different kbfs-edits channel, something like "<writer>:git" maybe.  In that case, we can probably do it at some later time if desired.

libkbfs/folder_branch_ops.go
strib: Refactored this whole function to make it cleaner.

libkbfs/folder_branch_ops.go
jzila: Naw that sounds reasonable.

libkbfs/modes.go
jzila: Agreed, I was thinking something like a separate channel for later.

PR #1629, opened 2018-06-19
kbfs_ops: use inbox to initialize FBOs needed to construct the edit history

strib:
(Depends on #1628.)

Also, causes a notification about a new channel to fire if we get a new message on a conversation we haven't yet subscribed to.  That way we can launch a new `folderBranchOps` for a TLF that was written to but didn't previously appear high enough in our inbox.

Re-initialize the FBOs after each re-connect and re-login as well.

Issue: KBFS-2996

### DIFF
--- a/libkbfs/init.go
+++ b/libkbfs/init.go
@@ -622,12 +622,6 @@ func doInit(
 	config.SetTLFValidDuration(params.TLFValidDuration)
 	config.SetBGFlushPeriod(params.BGFlushPeriod)
 
-	kbfsOps := NewKBFSOpsStandard(config)
-	config.SetKBFSOps(kbfsOps)
-	config.SetNotifier(kbfsOps)
-	config.SetKeyManager(NewKeyManagerStandard(config))
-	config.SetMDOps(NewMDOpsStandard(config))
-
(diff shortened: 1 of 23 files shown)


libkbfs/init.go
```diff
@@ -639,15 +633,29 @@ func doInit(
...
-	config.SetKeybaseService(service)
+	config.SetChat(chat)
 
-	// Initialize KBPKI client (needed for MD Server).
+	// Initialize KBPKI client (needed for KBFSOps and MD Server).
 	k := NewKBPKIClient(config, kbfsLog)
 	config.SetKBPKI(k)
 
+	kbfsOps := NewKBFSOpsStandard(config)
+	config.SetKBFSOps(kbfsOps)
```
songgao: If we set this before `keybaseServiceCn.NewChat` above, do we still need those `!= nil` checks in `*ChatRPC`?

strib: The changes in this file were to make sure chat is initialized _before_ KBFSOps, because now KBFSOps depends on it.  So unfortunately it's a cyclical dependency and I needed to add the `!= nil` checks.  If you see a better way to structure things (besides building an internal, async notification system), let me know.

PR #1630, opened 2018-06-20
More verbose logging about the system upon mount failures with Dokan

taruti:


libdokan/dokan_info.go
```diff
@@ -0,0 +1,82 @@
...
+	buf bytes.Buffer
+}
+
+func (ep *errorPrinter) Printf(s string, os ...interface{}) {
+	fmt.Fprintf(&ep.buf, s, os...)
+}
+
+func debugFileInfo(epc *errorPrinter, path string) {
+	f, err := os.Open(path)
+	epc.Printf("debugFileInfo: open(%q) -> %v, %v\n", path, f, err)
```
jzila: `%+v` for the error, here and everywhere else.

PR #1631, opened 2018-06-20
libfuse/libdokan: expose user history in root .kbfs_edit_history

strib:
(Depends on #1629.)

Also add DSL tests for the user edit history and fix a few small bugs:

* Use the full canonical path to the filenames in the edit notifications, like we do for the journal unflushed paths.
* Always set the topic name in `NewConversationLocal`.
* kbfs-edits conversations can legitimately have some weird messages and states, so don't error on those.
* A notification about a new channel should launch the corresponding FBO, if one doesn't exist yet.
* Send edit notifications in a goroutine when journaling is enabled, to avoid deadlocks.

Issue: KBFS-2999

### DIFF
--- a/libkbfs/ops.go
+++ b/libkbfs/ops.go
@@ -521,7 +521,8 @@ func (co *createOp) ToEditNotification(
 	rev kbfsmd.Revision, revTime time.Time, device kbfscrypto.VerifyingKey,
 	uid keybase1.UID, tlfID tlf.ID) *kbfsedits.NotificationMessage {
 	n := makeBaseEditNotification(rev, revTime, device, uid, tlfID, co.Type)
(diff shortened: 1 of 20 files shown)


libkbfs/ops.go
```diff
@@ -1537,3 +1541,17 @@ func RegisterOps(codec kbfscodec.Codec) {
 	codec.RegisterIfaceSliceType(reflect.TypeOf(opsList{}), opsListCode,
 		opPointerizer)
 }
+
+type pathSortedOps []op
```
songgao: Might be useful to have a comment on this type to explain that it sorts by length and when it's useful. So basically what you put where it's used:

> Make sure the ops are in increasing order by path length, so e.g. file creates come before file modifies.

PR #1632, opened 2018-06-20
folder_branch_ops: enable notification messages

strib:
(Depends on #1631.)

Issue: KBFS-3000

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -2290,14 +2290,6 @@ func (fbo *folderBranchOps) getConvID(
 
 func (fbo *folderBranchOps) sendEditNotifications(
 	ctx context.Context, rmd ImmutableRootMetadata, body string) error {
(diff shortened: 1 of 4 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -2287,14 +2287,6 @@ func (fbo *folderBranchOps) getConvID(
 
 func (fbo *folderBranchOps) sendEditNotifications(
 	ctx context.Context, rmd ImmutableRootMetadata, body string) error {
-	// For now only write out the notifications if we're in test mode,
-	// just in case we decide to change the notification format before
-	// we launch.  TODO: turn this on for admins once we can test it
-	// on staging.
```
songgao: Do we still want the "admins" gate?

strib: Now that I have an end-to-end kbfsdocker test working, I'm ok enabling it for non-admins too.

PR #1633, opened 2018-06-22
kbfsedits: includes at least 3 self-clusters in the overall user history

strib:
(Depends on #1631.)

This is the first attempt at making sure your own recent write history appears in the overall edit history, even if they were a long time ago.  The desired behavior, for now:

* Make sure there are always 3 TLF-writer clusters for yourself in the list.
* Sort them into the list according to the regular algorithm -- i.e., if they happened a long time ago, they will still appear at the bottom.
* Don't do anything special for the public folder -- if it happens to be one of the top-3 self-clusters, it will make the list; if not, it won't.

To make this happen, this PR:
* Creates a special "#self" kbfs-edits notification channel in the logged-in user's private conversation.
* On every "normal" notification, it writes the name and type of the TLF into this "#self" channel.
   * But as a best-effort optimization, it tries to make sure that whatever it writes is _different_ from the last write to the "#self" channel, so we don't build up a bunch of useless duplicate entries.
* When loading the inbox, it makes sure the last 3 unique TLFs from the "#self" channel are included.
* `TlfHistory` never trims the logged-in user's writes from the history.
* `UserHistory` always includes at least 3 clusters by the logged-in user in the overall history.

Issue: KBFS-3080

### DIFF
--- a/kbfsedits/tlf_history.go
+++ b/kbfsedits/tlf_history.go
@@ -20,7 +20,6 @@ const (
 type writerNotifications struct {
 	writerName    string
 	notifications notificationsByRevision
-	isLoggedIn    bool
 }
 
 // writersByRevision sorts sets of per-writer notifications in reverse
@@ -82,10 +81,11 @@ func (wbr *writersByRevision) Pop() interface{} {
 //     `getHistory()` from each one.  It may also construct pretty
 //     versions of individual edit histories for a particular TLF.
 type TlfHistory struct {
-	lock          sync.RWMutex
-	byWriter      map[string]*writerNotifications
-	computed      bool
-	cachedHistory writersByRevision
+	lock               sync.RWMutex
+	byWriter           map[string]*writerNotifications
+	computed           bool
+	cachedHistory      writersByRevision
+	cachedLoggedInUser string
 }
 
 // NewTlfHistory constructs a new TlfHistory instance.
@@ -100,7 +100,7 @@ func NewTlfHistory() *TlfHistory {
 // the caller should call `Recompute` to find out if more messages
 // should be added for any particular writer.
 func (th *TlfHistory) AddNotifications(
-	writerName string, messages []string, isLoggedIn bool) (err error) {
+	writerName string, messages []string) (err error) {
 	newEdits := make(notificationsByRevision, 0, len(messages))
 
 	// Unmarshal and sort the new messages.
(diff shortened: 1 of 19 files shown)


kbfsedits/tlf_history.go
```diff
@@ -313,11 +316,24 @@ func (th *TlfHistory) recomputeLocked() (
...
+			}
 			delete(th.byWriter, history[i].writerName)
 			delete(writersWhoNeedMore, history[i].writerName)
 		}
-		history = history[:maxWritersPerHistory]
+		if loggedInIndex > 0 {
+			// Keep the logged-in user as the last entry.
+			loggedIn := history[loggedInIndex]
+			history = history[:maxWritersPerHistory]
+			history[maxWritersPerHistory-1] = loggedIn
```
songgao: Did you mean to do a swap here instead of duplicating `loggedIn` like what it's doing now?

kbfsedits/tlf_history.go
```diff
@@ -20,6 +20,7 @@ const (
 type writerNotifications struct {
 	writerName    string
 	notifications notificationsByRevision
+	isLoggedIn    bool
```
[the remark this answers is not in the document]
strib: I could, if I passed the logged-in user name to both `UserHistory.UpdateHistory()` and `TlfHistory.Recompute()`.  I'm not sure which is cleaner, but I can try it that way and see if it seems better.

kbfsedits/tlf_history.go
strib: This doesn't duplicate anything, because the `for` loop above starts with `i := maxWritersPerHistory`; therefore, the loggedInIndex will always be beyond the cutoff that happens in the line above.  I'll clarify with a comment.

jzila: I wonder if you can do:
```go
history = append(history[:maxWritersPerHistory-1], history[loggedInIndex])
```

strib: I probably could, though I'm not sure if that's worse performance under the hood.  But if you think it's clearer, I'm happy to change it.

jzila: It's easier to read, and I doubt the performance will be a problem.

songgao: I see; thanks for explaining!

`append` reuses the underlying capacity if available, I think.

strib:
Appveyor is being fixed separately.

PR #1635, opened 2018-06-22
lock better and log better

songgao:
It's possible that `AddProtocol` is called after we iterate protocols in
`OnConnect`, but before it manages to finish everything and assign
the `(*KeybaseDaemonRPC).server`, so `AddProtocol` would find `server`
to be nil, and skips registering additional protocols. So just do a full
lock at the beginning of `OnConnect` to avoid this race.

Also add logging for each registered protocol to help with debugging.

### DIFF
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -315,8 +315,6 @@ func (k *KeybaseDaemonRPC) OnConnect(ctx context.Context,
 		}
 	}
 
-	return nil
-
 	// Using conn.GetClient() here would cause problematic
 	// recursion.
 	c := keybase1.NotifyCtlClient{Cli: rawClient}
--- a/libkbfs/keybase_daemon_rpc.go
+++ b/libkbfs/keybase_daemon_rpc.go
@@ -267,6 +267,18 @@ func (*KeybaseDaemonRPC) HandlerName() string {
 	return "KeybaseDaemonRPC"
 }
 
+func (k *KeybaseDaemonRPC) registerProtocolLocked(p rpc.Protocol) error {
+	k.log.Debug("registering protocol %q", p.Name)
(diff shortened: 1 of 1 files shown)


libkbfs/keybase_daemon_rpc.go
```diff
@@ -297,23 +306,13 @@ func (k *KeybaseDaemonRPC) AddProtocols(protocols []rpc.Protocol) {
...
-		}
+	k.lock.Lock()
+	defer k.lock.Unlock()
 
-		return nil
-	}()
-	if err != nil {
-		return err
+	for _, p := range k.protocols {
+		if err = k.registerProtocolLocked(p); err != nil {
```
jzila: this needs to `server.Register`, not `k.server.Register`.

libkbfs/keybase_daemon_rpc.go
```diff
@@ -267,6 +267,18 @@ func (*KeybaseDaemonRPC) HandlerName() string {
 	return "KeybaseDaemonRPC"
 }
 
+func (k *KeybaseDaemonRPC) registerProtocolLocked(p rpc.Protocol) error {
```
jzila: This should accept a server too.

songgao: Yup! I was thinking the same thing. The latest commit does exactly that.

songgao:
@jzila PTAL thanks!

songgao:
NP; I didn’t think of that race either. Hope it doesn’t introduce side effects!

strib:
I asked max and mike, and they say the HelloIAm code doesn't make any RPC calls, so _should_ be fine...

jzila:
Hmm this broke lint. Not sure why it was missed in the PR.

jzila:
Actually nvmind, it was something else.

PR #1640, opened 2018-06-25
add gomock and mockgen to jenkinsfile

jzila:


libkbfs/node_cache_test.go
```diff
@@ -464,11 +464,9 @@ func TestNodeCacheGCReal(t *testing.T) {
 
 	childNode1 = nil
 	runtime.GC()
-	_ = <-finalizerChan
+	<-finalizerChan
 
-	if len(ncs.nodes) != 2 {
-		t.Errorf("Expected %d nodes, got %d", 2, len(ncs.nodes))
-	}
+	require.Len(t, ncs.nodes, 1)
```
jzila: Fixed the test; @strib want to validate that this is okay? Seems to be consistent new GC behavior.

strib: Seems weird to me -- which extra node is it GCing? I guess it must be `childNode2`, but it's supposed to keep that live until the end of the function.  Maybe I'll try `runtime.Keepalive` and see if it helps.

jzila: Aren't there 3 nodes to start? There's `childNode1`, `childNode2`, and ...?

strib: And the parent node, which both other nodes have references to.

PR #1645, opened 2018-06-26
md_ops: check historic team writership

strib:
This PR checks, if the writer of an MD is not currently a writer of the team, that the user _was_ a writer at the "time" the revision was made.  Here "time" is defined by the merkle trees, just as it is with checking revoked devices.  We ask the service for the merkle root following when the user was last downgraded from being a writer, and then we use that to call `FindNextMD`.  This required some refactoring to share that code with the device-revocation-checking code.

Two notes:
1. Currently, the service can only tell us about writership downgrades, not readership downgrades.  Our MD validators call `IsTeamReader` for the last modifying user, since a reader is allowed to copy a writer's MD to make a rekey request.  So the best we can do now is just fake it, which should be fine since a) we do the full verification on the real writer of the MD, and b) readers can only make rekey requests, and it seems ok to be slack about the verification of those, as long as the writer stuff is fully verified.
2. The service has a small bug right now where it won't return an error if the user was NEVER a writer of the team.  Max is fixing this soon as CORE-8199.

Issue: KBFS-2955

### DIFF
--- a/libkbfs/keybase_service_base.go
+++ b/libkbfs/keybase_service_base.go
@@ -683,6 +683,37 @@ func (k *KeybaseServiceBase) LoadUserPlusKeys(ctx context.Context,
 	return userInfo, nil
 }
 
+func (k *KeybaseServiceBase) getLastWriterInfo(
+	ctx context.Context, teamInfo TeamInfo, tlfType tlf.Type, user keybase1.UID,
+	verifyingKey kbfscrypto.VerifyingKey) (TeamInfo, error) {
+	if _, ok := teamInfo.LastWriters[verifyingKey]; ok {
+		// Already cached, nothing to do.
+		return teamInfo, nil
+	}
+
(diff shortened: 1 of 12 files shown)


libkbfs/keybase_service_base.go
```diff
@@ -715,12 +747,26 @@ func (k *KeybaseServiceBase) LoadTeamPlusKeys(
 			// If the user is in the writer map, that satisfies none, reader
 			// or writer desires.
 			satisfiesDesires = cachedTeamInfo.Writers[desiredUser.Uid]
-			// If not, and the desire role is a reader, we need to
-			// check the reader map explicitly.
-			if !satisfiesDesires &&
-				(desiredRole == keybase1.TeamRole_NONE ||
-					desiredRole == keybase1.TeamRole_READER) {
-				satisfiesDesires = cachedTeamInfo.Readers[desiredUser.Uid]
+			if !satisfiesDesires {
```
jzila: If only life was this easy.

PR #1647, opened 2018-06-28
unflushed_path_cache: check for nil channel before closing

strib:
I guess this got triggered by simple FS starting to call for the status all of the sudden?  Still investigating how it could get into this state, but this should stop the panics.

Issue: KBFS-3101


strib:
I'll fix the appveyor test bug separately, it's unrelated.

PR #1648, opened 2018-06-28
tlf_journal: save unflushedPathCache during initialization

strib:
(This is the real fix for KBFS-3101.)

Otherwise, if the journal flushes and the cache is cleared, we don't try to complete the initialization on the new instance of the object.

Issue: KBFS-3101


strib:
Appveyor failure unrelated, fixed in another PR.

PR #1653, opened 2018-06-30
kbfsedits: show unflushed notifications in the edit history

strib:
This PR adds a set of unflushed notifications to the TLF edit history, which supersedes any other known notifications for conflicting revisions.  As revisions flush from the journal, the unflushed notifications are removed from the history (and should be replaced by normal notifications coming through the chat mechanism).

If the device restarts with unflushed journals, this PR makes FBO objects for them (which initialize their unflushed paths on startup).

Issue: KBFS-3081


libkbfs/keybase_service_util.go
```diff
@@ -36,6 +36,7 @@ func serviceLoggedIn(ctx context.Context, config Config, session SessionInfo,
 			log.CWarningf(ctx,
 				"Failed to enable existing journals: %v", err)
 		}
+		jServer.MakeFBOsForExistingJournals(ctx)
```
jzila: Should this happen even if the above failed?

libkbfs/journal_server.go
```diff
@@ -280,6 +280,67 @@ func (j *JournalServer) hasTLFJournal(tlfID tlf.ID) bool {
...
+	}
+
+	handle, err := MakeTlfHandle(
+		ctx, headBareHandle, tlfID.Type(), j.config.KBPKI(),
+		j.config.KBPKI(), constIDGetter{tlfID})
+	if err != nil {
+		return err
+	}
+
+	_, _, err = j.config.KBFSOps().GetRootNode(ctx, handle, MasterBranch)
```
jzila: This will also trigger the prefetcher for all existing journals. Should be fine, just something to be aware of.

libkbfs/keybase_service_util.go
strib: Probably not, good point.

PR #1654, opened 2018-07-02
kbfsedits: handle renames of parent directories

strib:
This PR adds support for renamed parent directories to `TlfHistory`, to make sure that the full paths of edited files reflect any renames on their parent directories.

Also, there was a bug with sending edit notifications for rename operations -- instead of a single rename, they were being sent as separated delete+create notifications.  This PR fixes that, and adds tests.

Issue: KBFS-3073

### DIFF
--- a/kbfsedits/tlf_history.go
+++ b/kbfsedits/tlf_history.go
@@ -293,6 +293,23 @@ func (r *recomputer) processNotification(
 			return false
 		}
 
+		// See if any of the parent directories were renamed, checking
+		// backwards until we get to the TLF name.
+		prefix := filename
+		suffix := ""
+		for strings.Count(prefix, "/") > 4 {
+			var finalElem string
+			prefix, finalElem = path.Split(prefix)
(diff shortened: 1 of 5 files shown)


kbfsedits/tlf_history.go
```diff
@@ -281,6 +283,23 @@ func (r *recomputer) processNotification(
 			return false
 		}
 
+		// See if any of the parent directories were renamed, checking
+		// backwards until we get to the TLF name.
+		prefix := filename
+		suffix := ""
+		for strings.Count(prefix, "/") > 4 {
```
jzila: What about Windows?

strib: > What about Windows?

The path in the file edit notifications come from `libkbfs.Path`, which explicitly uses `/` to construct them: https://github.com/keybase/kbfs/blob/ae58f965dbcd49300d922dd09d0fff829b916de7/libkbfs/path.go#L29-L48

Any translation to system-specific paths should happen in the GUI.

PR #1656, opened 2018-07-03
rekey_queue: put queue size into mode, and drop reqs when full

strib:
Launching a goroutine for rekey requests when the queue is full doesn't make a lot of sense if we're trying to conserve memory (like on a phone), since it probably uses a lot more memory than just expanding the queue.  So this PR just drops requests when the queue is full, and increases the queue size on desktop to handle more requests at a time.  Any dropped requests will get processed again at the next rekey interval.

This only changes behavior for people with more than 2K TLFs, which is probably only two users.

Issue: KBFS-3072

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -1837,6 +1837,8 @@ type InitMode interface {
 	PrefetchWorkers() int
 	// RekeyWorkers returns the number of rekey workers to run.
 	RekeyWorkers() int
+	// RekeyQueueSize returns the size of the rekey queue.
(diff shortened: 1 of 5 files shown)


strib:
> This is fine, but you could also consider using an InfiniteChannelWrapper if we can handle the memory usage:
> https://github.com/keybase/kbfs/blob/master/libkbfs/channels.go#L19

I did think about that, but since the problem I'm trying to solve here is the memory consumption (caused by goroutines in nojima's case, but could in theory happen with the queue as well), I thought it wouldn't be appropriate.

PR #1657, opened 2018-07-03
kbfsedits: ignore macOS "._" files and don't send empty updates

strib:


kbfsedits/tlf_history.go
```diff
@@ -168,6 +170,11 @@ func newRecomputer() *recomputer {
 	}
 }
 
+func ignoreFile(filename string) bool {
+	_, base := path.Split(filename)
+	return strings.HasPrefix(base, "._")
```
songgao: Do we wanna include `.Trashes`, `.fseventsd`, `.DS_Store` as well?

PR #1658, opened 2018-07-03
libkbfs: turn on QR for mobile, but limit it to self-QR

strib:
Once mobile devices can write, we don't want them wasting cycles cleaning up after other device writes (which would have happened after 24 hours if the other device didn't do it first).

Instead, move all the QR stuff into the mode, and make sure mobile devices (running in constrained mode) only run QR when the latest writer was the local device.

Issue: KBFS-3121

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -735,7 +735,7 @@ func (fbm *folderBlockManager) deleteBlocksInBackground() {
 func (fbm *folderBlockManager) isOldEnough(rmd ImmutableRootMetadata) bool {
 	// Trust the server's timestamp on this MD.
 	mtime := rmd.localTimestamp
-	unrefAge := fbm.config.QuotaReclamationMinUnrefAge()
+	unrefAge := fbm.config.Mode().QuotaReclamationMinUnrefAge()
 	return mtime.Add(unrefAge).Before(fbm.config.Clock().Now())
 }
 
@@ -779,7 +779,7 @@ func (fbm *folderBlockManager) getMostRecentOldEnoughAndGCRevisions(
 				fbm.isOldEnough(rmd) {
 				fbm.log.CDebugf(ctx, "Revision %d is older than the unref "+
 					"age %s", rmd.Revision(),
-					fbm.config.QuotaReclamationMinUnrefAge())
+					fbm.config.Mode().QuotaReclamationMinUnrefAge())
 				mostRecentOldEnoughRev = rmd.Revision()
 			}
 
@@ -997,8 +997,12 @@ func (fbm *folderBlockManager) isQRNecessary(
 	// written by this device.  We want to avoid fighting with other
 	// active writers whenever possible.
 	if !selfWroteHead {
+		minHeadAge := fbm.config.Mode().QuotaReclamationMinHeadAge()
+		if minHeadAge <= 0 {
+			return false
+		}
 		headAge := fbm.config.Clock().Now().Sub(head.localTimestamp)
-		if headAge < fbm.config.QuotaReclamationMinHeadAge() {
+		if headAge < minHeadAge {
 			return false
 		}
 	}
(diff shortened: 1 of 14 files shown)


strib:
@songgao @jzila it would be awesome if one of you can test this on ios/android and make sure the QR timer doesn't cause the app to be un-backgrounded.

jzila:
(haven't tested it yet)

songgao:
Hmm we don't have deletes in GUI yet, so it seems hard to trigger a large QR with normal operations. Any tips on how to have a timer and trigger it after e.g. 2min?

strib:
> Hmm we don't have deletes in GUI yet, so it seems hard to trigger a large QR with normal operations. Any tips on how to have a timer and trigger it after e.g. 2min?

Any writes causes QR, because the old versions of the parent directory blocks are deleted.

But you don't really even have to do any writes to test this, because QR wakes up on a 1 minute timer all the time, even when there hasn't been any writes.  So start it up running this branch, wait a little bit, then background it, and make sure it doesn't get un-backgrounded.

If there's no obvious way to tell, you could add a log message around here: https://github.com/keybase/kbfs/blob/strib/KBFS-3121-mobile-self-qr-only/libkbfs/folder_block_manager.go#L1181

songgao:
Oh I see; thanks. Will do it now.

strib:
@jzila @songgao @mmaxim: I added a couple commits that pause QR when the app isn't foregrounded.  Ready for a re-review, and I'd appreciate another test with a real app to make sure this works ok.  There are log messages for pause/resume.

libkbfs/folder_block_manager.go
```diff
@@ -1029,7 +1037,7 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
 	ctx, cancel := context.WithCancel(fbm.ctxWithFBMID(context.Background()))
 	fbm.setReclamationCancel(cancel)
 	defer fbm.cancelReclamation()
-	defer timer.Reset(fbm.config.QuotaReclamationPeriod())
+	defer timer.Reset(fbm.config.Mode().QuotaReclamationPeriod())
```
songgao: This fires the timer in modes where `QuotaReclamationPeriod()` is 0. Just making sure you've considered it!

libkbfs/folder_block_manager.go
```diff
@@ -1161,18 +1169,29 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
...
 			// Use a channel that will never fire instead.
 			timerChan = make(chan time.Time)
 		}
+
+		state := keybase1.AppState_FOREGROUND
 		select {
 		case <-fbm.shutdownChan:
 			return
+		case state = <-fbm.g.AppState.NextUpdate(&state):
+			for state != keybase1.AppState_FOREGROUND {
```
songgao: After app is backgrounded, it can be suspended by the OS. So if a QR has started at this point, would it make sense to try to stop it if it doesn't finish within a short period? Perhaps could pass in a context with deadline into `doReclaimation()`.

libkbfs/folder_block_manager.go
```diff
@@ -1161,18 +1169,29 @@ func (fbm *folderBlockManager) doReclamation(timer *time.Timer) (err error) {
...
-	timer := time.NewTimer(fbm.config.QuotaReclamationPeriod())
+	timer := time.NewTimer(fbm.config.Mode().QuotaReclamationPeriod())
 	timerChan := timer.C
 	for {
 		// Don't let the timer fire if auto-reclamation is turned off.
-		if fbm.config.QuotaReclamationPeriod().Seconds() == 0 {
+		if fbm.config.Mode().QuotaReclamationPeriod().Seconds() == 0 {
 			timer.Stop()
 			// Use a channel that will never fire instead.
 			timerChan = make(chan time.Time)
```
songgao: The doc suggests something like this:

```
if !t.Stop() {
	<-t.C
}
```

Not sure which way is better though.

libkbfs/folder_block_manager.go
songgao: Oh I meant to delete this comment but forgot. Ignore this!

libkbfs/folder_block_manager.go
strib: QR can't be going at this point, since the loop waits on each QR before doing the `select`.  Are you proposing having a second background goroutine that waits on the signals?  That seems a bit too complicated for this PR -- I vote that if we get backgrounded, we just let the QR finish and/or get suspended.  We can revisit later if it becomes a problem.

libkbfs/folder_block_manager.go
strib: Thanks, but I don't think I want to mess with code that is working, in this PR anyway.

libkbfs/folder_block_manager.go
songgao: Oh I didn't realize it waits on the QR. "Revisit later if it becomes a problem" sounds good!

PR #1659, opened 2018-07-04
[KBFS-3052] vendor: pull new SimpleFSRecursive API from client

jzila:
Will need to re-vendor from master when keybase/client#12682 is merged. CircleCI will fail until then.

This is ready for review. LMK if you'd like some tests to validate the refactor.

Reviewers: [w=1](https://github.com/keybase/kbfs/pull/1659/files?w=1) is your friend here.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -485,6 +485,106 @@ func (k *SimpleFS) SimpleFSList(ctx context.Context, arg keybase1.SimpleFSListAr
 		})
 }
 
+// listRecursiveToDepthAsync returns a function that recursively lists folders,
+// up to a given depth. A depth of -1 is treated as unlimited. The function
+// also updates progress for the passed-in opID as it progresses, and then sets
+// the result for the opID when it completes.
+//
+// TODO: refactor SimpleFSList to use this too (finalDepth = 0)
+//
+func (k *SimpleFS) listRecursiveToDepth(opID keybase1.OpID,
+	path keybase1.Path, filter keybase1.ListFilter,
+	finalDepth int) func(context.Context) error {
+	return func(ctx context.Context) (err error) {
+		// A stack of paths to process - ordering does not matter.
+		// Here we don't walk symlinks, so no loops possible.
+		type pathStackElem struct {
+			path  string
+			depth int
+		}
+		var paths []pathStackElem
+
+		fs, finalElem, err := k.getFS(ctx, path)
+		switch err.(type) {
+		case nil:
+		case libfs.TlfDoesNotExist:
+			// TLF doesn't exist yet; just return an empty result.
+			k.setResult(opID, keybase1.SimpleFSListResult{})
+			return nil
+		default:
+			return err
+		}
+
+		// With listing, we don't know the totals ahead of time,
+		// so just start with a 0 total.
+		k.setProgressTotals(opID, 0, 0)
+		fi, err := fs.Stat(finalElem)
+		if err != nil {
+			return err
+		}
+		var des []keybase1.Dirent
+		if !fi.IsDir() {
+			var d keybase1.Dirent
+			err := setStat(&d, fi)
+			if err != nil {
+				return err
+			}
+			des = append(des, d)
+			// Leave paths empty so we can skip the loop below.
+		} else {
+			// Start with a depth of 0
+			paths = append(paths, pathStackElem{finalElem, 0})
+		}
+
+		for len(paths) > 0 {
+			// Take last element and shorten.
+			pathElem := paths[len(paths)-1]
+			paths = paths[:len(paths)-1]
+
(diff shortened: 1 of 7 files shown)


simplefs/simplefs.go
```diff
@@ -485,82 +485,115 @@ func (k *SimpleFS) SimpleFSList(ctx context.Context, arg keybase1.SimpleFSListAr
...
+				// Only recurse if the caller requested infinite depth (-1), or
+				// if the current path has a depth less than the desired final
+				// depth of recursion.
+				if fi.IsDir() && (finalDepth == -1 || pathElem.depth < finalDepth) {
+					paths = append(paths, pathStackElem{stdpath.Join(pathElem.path, fi.Name()), pathElem.depth + 1})
 				}
-				k.updateReadProgress(arg.OpID, 0, int64(len(fis)))
 			}
-			k.setResult(arg.OpID, keybase1.SimpleFSListResult{Entries: des})
+			k.updateReadProgress(opID, 0, int64(len(fis)))
```
strib: It kind of seems like this should only add one to the FilesRead progress, but should update the FilesTotal progress by len(fis). (And you'd start the whole thing out with a total of 1.) That way the progress tracking will be a little more useful as it can kinda predict the latency of the ReadDirs.

Maybe that's overkill though.

simplefs/simplefs.go
```diff
@@ -485,82 +485,115 @@ func (k *SimpleFS) SimpleFSList(ctx context.Context, arg keybase1.SimpleFSListAr
...
 
-			return nil
-		})
+		return nil
+	}
+}
+
+// SimpleFSListRecursiveToDepth - Begin recursive list of items in directory at path up to a given depth.
+func (k *SimpleFS) SimpleFSListRecursiveToDepth(ctx context.Context, arg keybase1.SimpleFSListRecursiveToDepthArg) error {
+	return k.startAsync(ctx, arg.OpID, keybase1.AsyncOps_LIST_RECURSIVE,
```
strib: `LIST_RECURSIVE_TO_DEPTH`?

simplefs/simplefs_test.go
```diff
@@ -102,6 +102,9 @@ func checkPendingOp(ctx context.Context,
 	case keybase1.AsyncOps_LIST_RECURSIVE:
 		list := o.ListRecursive()
 		assert.Equal(t, list.Path, src, "Expected matching path in operation")
+	case keybase1.AsyncOps_LIST_RECURSIVE_TO_DEPTH:
```
strib: This case won't actually matter until there's a test that calls the `ToDepth` variant, right?  Might be nice if this PR added something like that.  (I know the regular recursive call tests the code with -1, but would be good to have something that stops early.

simplefs/simplefs.go
jzila: Hmm. I'm not really sure what the implications are on the UX with respect to listing directories. I follow your argument, but not sure what the right answer is. @songgao WDYT?

simplefs/simplefs_test.go
jzila: Will do. We don't currently have a test for `ListRecursive` either, so I'll add one.

jzila:
@strib: ready for another review. New commit diff is https://github.com/keybase/kbfs/pull/1659/files/9b9c8e500d9ad715d18b595fa50da90b28d21f3d..26fc9fff58abbb09bb819747521193be9f937f32

simplefs/simplefs.go
```diff
@@ -529,6 +529,7 @@ func (k *SimpleFS) listRecursiveToDepth(opID keybase1.OpID,
 			if err != nil {
 				return err
 			}
+			d.Name = finalElem
```
strib: This isn't technically needed, right?  `setStat` should already set it, I think.  (Am fine with having it here too for clarity, though.)

jzila: Correct, technically not needed. Originally I thought it'd give absolute-ish paths (`/private/jdoe/a`), but it doesn't, making this line extraneous.

jzila:
@strib: not so much a bug, rather that there's no way to reconstruct the path tree from the current `SimpleFSListRecursive` output. There's a bug with the flags though that I'm fixing in the client PR.

strib:
> @strib: not so much a bug, rather that there's no way to reconstruct the path tree from the current SimpleFSListRecursive output. There's a bug with the flags though that I'm fixing in the client PR.

Ah ok.  In the #kbfs channel your first and second command outputs looked identical, even though there was a non-empty directory that should have been recursed into, but maybe that was a bug with the `--recusive` flag on the client side.

jzila:
Yep, because the client is misconfigured to accept `--recurse`, where the help says `--rec` or `--recursive`.

jzila:
CI flakes are unrelated.

PR #1663, opened 2018-07-05
simplefs: allow subscriptions for TLF notifications

strib:
This lets simplefs users pass a `RefreshSubscription` flag into a `SimpleFSList[Recursive]` call.  If `true`, `SimpleFS` will begin sending `FSPathUpdate` notifications through the notify router for that TLF.  Works only for KBFS paths, for non-empty TLFs.

Note that this sends notifications for anything in the TLF, not just the listed subdir.  But the interfaces let us make that optimization later if we need to.  Also, once the favorite subscription stuff is worked out, we can subscribe to favorite notifications too.

This vendors in keybase/client#12699 (before merging, I'll fix the commit hash).

Issue: KBFS-3120


simplefs/simplefs.go
```diff
@@ -1356,26 +1451,40 @@ func (k *SimpleFS) SimpleFSFolderEditHistory(
...
+}
+
+// BatchChanges implements the libkbfs.Observer interface for SimpleFS.
+func (k *SimpleFS) BatchChanges(
+	ctx context.Context, changes []libkbfs.NodeChange, _ []libkbfs.NodeID) {
+	k.subscribeLock.RLock()
+	defer k.subscribeLock.RUnlock()
+	for _, nc := range changes {
+		fb := nc.Node.GetFolderBranch()
+		if fb == k.subscribeCurrFB {
```
jzila: So if we get unmerged then we don't notify updates?

strib: You should still get updates for unmerged updates on this device, the same way FUSE does.  What made you think otherwise?

jzila: It looks like you're only notifying that the path is updated if the folder-branch hasn't changed.

strib: Oh, when the TLF goes onto an unmerged branch, it (confusingly) doesn't change the actual `FolderBranch` it's referenced with.  That concept of a branch is unused today, but it was meant to indicate other historical views, or forks, of a given TLF.  So it might be useful for the time travel stuff, but for now all TLFs have `libkbfs.MasterBranch` in their `FolderBranch`, even when unmerged.

jzila: Understood, thanks for clarifying.

libkbfs/reporter_kbpki.go
```diff
@@ -267,6 +283,15 @@ func (r *ReporterKBPKI) send(ctx context.Context) {
 				r.log.CDebugf(ctx, "ReporterDaemon: error sending "+
 					"notification: %s", err)
 			}
+		case path, ok := <-r.notifyPathBuffer:
+			if !ok {
+				return
+			}
+			if err := r.config.KeybaseService().NotifyPathUpdated(
```
songgao: It seems it makes sense to make this affected by suppression too. WDYT?

songgao: Also, it seems this notification would include changes from the local device too. So perhaps we can remove the suppression from `Notify`.

strib: I added suppression logic for the path updates, but I didn't want to mess with the `Notify` logic since I have no idea if the frontend is still using them or if you've switched completely over to path notifications yet (I'm guessing not, since I haven't even turned them on yet!).

PR #1664, opened 2018-07-06
keybase_service_util: don't block on initializing FBOs on login

strib:
`serviceLoggedIn` can be called as a side-effect of the mdserver `OnConnect` call, if that is the first time the current session is discovered.  Initializing FBOs depends on the mdserver, and so that can lead to a deadlock.

Instead, initialize them in the background in that case, while the other cases can still wait for the initialization to complete.

Issue: KBFS-3155

### DIFF
--- a/libkbfs/keybase_service_util.go
+++ b/libkbfs/keybase_service_util.go
@@ -5,6 +5,8 @@
 package libkbfs
 
 import (
+	"sync"
+
 	"github.com/keybase/client/go/libkb"
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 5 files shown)


libkbfs/keybase_service_util.go
```diff
@@ -27,7 +29,7 @@ func EnableAdminFeature(ctx context.Context, runMode libkb.RunMode, config Confi
 // serviceLoggedIn should be called when a new user logs in. It
 // shouldn't be called again until after serviceLoggedOut is called.
 func serviceLoggedIn(ctx context.Context, config Config, session SessionInfo,
-	bws TLFJournalBackgroundWorkStatus) {
+	bws TLFJournalBackgroundWorkStatus) (wg *sync.WaitGroup) {
```
songgao: It seems only tests rely on the returned `wg` for now. But in case in the future we start using this too, perhaps it makes sense to initialize `wg` at beginning to a `sync.WaitGroup{}` so we never return a `nil` pointer here.

PR #1665, opened 2018-07-07
kbfs_ops: on TLF history init, don't run TLF identifies

strib:
Just start up the chat monitoring, so we don't trigger any unintended popups.

If there are unflushed journals for some TLFs, we could theoretically trigger popups for them since they are initialized differently, but in that case the user would probably have recently written to the TLF and so would have already gone through the identify flow.

Also, remove a `folderBranchOps` function that's been unused for a long time, and confused me.

Issue: CORE-8290

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -1206,12 +1206,18 @@ func (fs *KBFSOpsStandard) initTlfsForEditHistories() {
 	// that doesn't have one yet.  Also make sure there's one for the
 	// logged-in user's public folder.
 	for _, h := range handles {
-		fs.log.CDebugf(ctx, "Initializing TLF %s for the edit history",
-			h.GetCanonicalPath())
-		_, _, err := fs.GetRootNode(ctx, h, MasterBranch)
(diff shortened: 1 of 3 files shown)


libkbfs/kbfs_ops.go
```diff
@@ -1206,12 +1206,18 @@ func (fs *KBFSOpsStandard) initTlfsForEditHistories() {
...
-		if err != nil {
-			fs.log.CWarningf(ctx, "Couldn't get root node for %s: %+v",
-				h.GetCanonicalName(), err)
+		if h.tlfID != tlf.NullID {
+			fs.log.CDebugf(ctx, "Initializing TLF %s (%s) for the edit history",
+				h.GetCanonicalPath(), h.tlfID)
+			ops := fs.getOpsNoAdd(ctx, FolderBranch{h.tlfID, MasterBranch})
+			// Don't initialize the entire TLF, because we don't want
+			// to run identifies on it.  Instead, just start the
+			// chat-monitoring part.
```
jzila: Cool. This also avoids prefetching that TLF.

PR #1667, opened 2018-07-07
implement SimpleFSSuppressNotifications

songgao:
KBFS side of https://github.com/keybase/client/pull/12737

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -875,7 +875,7 @@ type Reporter interface {
 	NotifySyncStatus(ctx context.Context, status *keybase1.FSPathSyncStatus)
 	// SuppressNotifications suppresses notifications. See
 	// protocol/avdl/keybase1/simple_fs.avdl for more details.
-	SuppressNotifications(ctx context.Context, nextSuppressIn time.Duration)
+	SuppressNotifications(ctx context.Context, suppressDuration time.Duration)
 	// Shutdown frees any resources allocated by a Reporter.
 	Shutdown()
 }
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -873,6 +873,9 @@ type Reporter interface {
 	Notify(ctx context.Context, notification *keybase1.FSNotification)
 	// NotifySyncStatus sends the given path sync status to any sink.
 	NotifySyncStatus(ctx context.Context, status *keybase1.FSPathSyncStatus)
+	// SuppressNotifications suppresses notifications. See
+	// protocol/avdl/keybase1/simple_fs.avdl for more details.
+	SuppressNotifications(ctx context.Context, nextSuppressIn time.Duration)
 	// Shutdown frees any resources allocated by a Reporter.
 	Shutdown()
 }
--- a/libkbfs/reporter_kbpki.go
+++ b/libkbfs/reporter_kbpki.go
@@ -81,6 +81,7 @@ type ReporterKBPKI struct {
 	log              logger.Logger
 	notifyBuffer     chan *keybase1.FSNotification
 	notifySyncBuffer chan *keybase1.FSPathSyncStatus
+	suppressCh       chan time.Duration
 	canceler         func()
 }
 
@@ -92,6 +93,7 @@ func NewReporterKBPKI(config Config, maxErrors, bufSize int) *ReporterKBPKI {
 		log:              config.MakeLogger(""),
 		notifyBuffer:     make(chan *keybase1.FSNotification, bufSize),
 		notifySyncBuffer: make(chan *keybase1.FSPathSyncStatus, bufSize),
+		suppressCh:       make(chan time.Duration, 1),
 	}
 	var ctx context.Context
 	ctx, r.canceler = context.WithCancel(context.Background())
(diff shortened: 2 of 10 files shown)


libkbfs/interfaces.go
```diff
@@ -873,6 +873,9 @@ type Reporter interface {
 	Notify(ctx context.Context, notification *keybase1.FSNotification)
 	// NotifySyncStatus sends the given path sync status to any sink.
 	NotifySyncStatus(ctx context.Context, status *keybase1.FSPathSyncStatus)
+	// SuppressNotifications suppresses notifications. See
+	// protocol/avdl/keybase1/simple_fs.avdl for more details.
```
songgao: Teleport gate: https://github.com/keybase/client/pull/12737/files#diff-b4c73240d3e8f6f7b5d70397eade2ea6R310

libkbfs/reporter_simple.go
```diff
@@ -98,6 +99,12 @@ func (r *ReporterSimple) NotifySyncStatus(_ context.Context,
 	// ignore notifications
 }
 
+// SuppressNotifications implements the Reporter interface for ReporterSimple.
+func (k *ReporterSimple) SuppressNotifications(
+	ctx context.Context, nextSuppressIn time.Duration) {
```
strib: Use `_` for param names please.

libkbfs/reporter_kbpki.go
```diff
@@ -230,23 +232,37 @@ func (r *ReporterKBPKI) NotifySyncStatus(ctx context.Context,
 	}
 }
 
+// SuppressNotifications implements the Reporter interface for ReporterKBPKI.
+func (r *ReporterKBPKI) SuppressNotifications(
+	ctx context.Context, nextSuppressIn time.Duration) {
+	r.suppressCh <- nextSuppressIn * 2
```
strib: I don't understand why we `*2` here, why not just let the caller set the real value?  Also, why is it called `nextSuppressIn` everywhere?  Isn't this supposed to be the `suppressDuration` or something?

simplefs/simplefs.go
```diff
@@ -1379,3 +1379,8 @@ func (k *SimpleFS) SimpleFSFolderEditHistory(
 	// Now get the edit history.
 	return k.config.KBFSOps().GetEditHistory(ctx, node.GetFolderBranch())
 }
+
+func (k *SimpleFS) SimpleFSSuppressNotifications(ctx context.Context, nextSuppressIn int) error {
+	k.config.Reporter().SuppressNotifications(ctx, time.Second*time.Duration(nextSuppressIn))
```
strib: I don't understand this formula.  If `nextSuppressIn` is supposed to be in seconds, shouldn't this be just `time.Second*nextSuppressIn`?

libkbfs/reporter_simple.go
```diff
@@ -98,6 +99,12 @@ func (r *ReporterSimple) NotifySyncStatus(_ context.Context,
 	// ignore notifications
 }
 
+// SuppressNotifications implements the Reporter interface for ReporterSimple.
+func (k *ReporterSimple) SuppressNotifications(
```
strib: Should be `r *ReporterSimple`.

simplefs/simplefs.go
songgao: Yes, but you can’t multiply int with time.Duration, this the type conversation. Maybe I should convert time.Second to int64 first?

libkbfs/reporter_kbpki.go
songgao: The idea is this is like a heartbeat. If caller hasn’t called this RPC for 2*nextSuppressIn, the suppression stops. It needs to be larger than the actual calling interval to count for delays. I guess calling it suppressDuration and pass in the 2*’ed value is clearer. So will do that. Let me know if that’s not what you mean!

strib: I just think `nextSuppressIn` sounds like the next suppression isn't supposed to happen for that amount of time, not that it will _be_ suppressed for that amount of time.

libkbfs/interfaces.go
```diff
@@ -873,6 +873,9 @@ type Reporter interface {
...
+	SuppressNotifications(ctx context.Context, nextSuppressIn time.Duration)
```
strib: param name is still misleading.

libkbfs/reporter_kbpki.go
```diff
@@ -230,23 +232,37 @@ func (r *ReporterKBPKI) NotifySyncStatus(ctx context.Context,
...
+	ctx context.Context, suppressDuration time.Duration) {
+	r.suppressCh <- suppressDuration
+}
+
 // Shutdown implements the Reporter interface for ReporterKBPKI.
 func (r *ReporterKBPKI) Shutdown() {
 	r.canceler()
 	close(r.notifyBuffer)
 	close(r.notifySyncBuffer)
+	close(r.suppressCh)
```
jzila: This is a problem in all the other channels as well, but after `Shutdown`, any calls to `SuppressNotifications` will panic. It's probably fine, just something to consider.

songgao: It's already shut down, so yeah not really a big issue. I guess we can just leave them all open if care?

jzila: No they need to be closed so the loops finish (or we'd need a shutdown channel otherwise). I'm just wary of races where something calls into SimpleFS's `SuppressNotifications` after shutdown has been called.

songgao: The canceler should release the loop, so ctx.Done() is sort of a shutdown channel here. Although it seems we are not checking it now (despite of passing it in). I'll put up a PR to fix it.

PR #1668, opened 2018-07-09
fix kbpki reporter shutdown as suggested by jzila

songgao:


libkbfs/reporter_kbpki.go
```diff
@@ -302,6 +299,8 @@ func (r *ReporterKBPKI) send(ctx context.Context) {
 				stagedStatus = nil
 			}
 			suppressed = false
+		case <-ctx.Done():
```
jzila: You can now remove all the `if !ok {  return }` lines above, unless you put the above channel `close`s back in.

libkbfs/reporter_kbpki.go
```diff
@@ -241,9 +241,6 @@ func (r *ReporterKBPKI) SuppressNotifications(
 // Shutdown implements the Reporter interface for ReporterKBPKI.
 func (r *ReporterKBPKI) Shutdown() {
 	r.canceler()
-	close(r.notifyBuffer)
-	close(r.notifySyncBuffer)
-	close(r.suppressCh)
```
jzila: We still probably want these to panic so that we are made aware of races (instead of infinitely blocking).

libkbfs/reporter_kbpki.go
songgao: ah now I'm realizing the "before" is better, as channel selection is random.

libkbfs/reporter_kbpki.go
```diff
@@ -235,15 +235,17 @@ func (r *ReporterKBPKI) NotifySyncStatus(ctx context.Context,
 // SuppressNotifications implements the Reporter interface for ReporterKBPKI.
 func (r *ReporterKBPKI) SuppressNotifications(
 	ctx context.Context, suppressDuration time.Duration) {
-	r.suppressCh <- suppressDuration
+	select {
+	case r.suppressCh <- suppressDuration:
+	case <-ctx.Done():
```
songgao: This is the incoming call's ctx, but it should still eventually timeout if something is waiting on it during shutdown.

jzila: You could save the canceler context in the reporter, and select on both contexts.

songgao: I think it'd still be racy. The channels can be closed at the same time sending to `r.suppressCh` happens, and it's a random selection here.

jzila: if `suppressCh` successfully receives the `suppressDuration`, then it'll get processed by the loop.

songgao: Oh you are saying we just select on that context, but still keep suppressCh open in shutdown?

jzila: yep

libkbfs/reporter_kbpki.go
```diff
@@ -83,6 +83,7 @@ type ReporterKBPKI struct {
 	notifySyncBuffer chan *keybase1.FSPathSyncStatus
 	suppressCh       chan time.Duration
 	canceler         func()
+	ctx              context.Context
```
strib: I know @jzila suggested this, but I disagree: you're really not supposed to save a context in a struct.  Can you get by with just saving the `Done()` channel instead?  The `ctx` should continue to be passed around explicitly.

songgao: I think so:

> Successive calls to Done return the same value.

I don't have a strong opinion on any of these. So if that's a solution both of you agree on, I'll make the change. Please let me know. @jzila

jzila: Strib is correct about the general case, but this isn't a real context here. It's a cancelation proxy.

strib: So why not just have the cancellation piece of it, instead of the whole context?

Also, it's being used for more than that, it's passed into all the `CDebugf` statements.

jzila: But sure, we can save the `Done()` channel instead.

PR #1669, opened 2018-07-10
kbfsfuse: create mountdir if it does not exist yet

### LINKED ISSUE #1655: kbfsfuse does not know how to handle missing directories during the mount stage.
The default Linux mountdir is currently `/run/user/$UID/keybase/kbfs`, which does not exist by default and causes kbfs to fail to mount if it doesn't exist. kbfs should try to create it, rather than logging the error

```
2018-07-03T09:31:50.875363-04:00 ▶ [ERRO kbfs mount_interrupter.go:52] 050 Mounting the filesystem failed: stat /run/user/1001/keybase/kbfs: no such file or directory
````

Interestingly, it doesn't actually consider this a failure state -- kbfsfuse continues to run, rather that exiting with a non-zero return code.

### PR DESCRIPTION
**Pull request by eli-schwartz**
Fixes #1655

### DIFF
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -80,6 +80,10 @@ func start() *libfs.Error {
 	} else {
 		mountDir = flag.Arg(0)
 	}
+	err := os.MkdirAll(mountDir, os.ModePerm)
+	if err != nil {
+		fmt.Println(err)
+	}
 
 	if len(flag.Args()) > 1 {
 		fmt.Print(getUsageString(ctx))
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -10,6 +10,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"runtime"
 
 	"bazil.org/fuse"
 
@@ -83,6 +84,16 @@ func start() *libfs.Error {
 	err := os.MkdirAll(mountDir, os.ModePerm)
 	if err != nil {
 		fmt.Println(err)
+		switch os := runtime.GOOS; os {
+		case "linux":
+			fmt.Println("failed to create mount directory")
+		case "darwin":
+			// If the user did not override the defaults this isn't expected to work.
+			// The default is in a directory only writable by root.
+			fmt.Println("mount directory does not exist and should have been created by system service")
+		case "windows":
+			fmt.Println("I honestly have no idea what to say here, maybe someone can tell me what is supposed to create this on Windows")
+		}
 	}
 
 	if len(flag.Args()) > 1 {

### CODE REVIEW

File: kbfsfuse/main.go [hunk 1]
```diff
@@ -80,6 +80,13 @@ func start() *libfs.Error {
 	} else {
 		mountDir = flag.Arg(0)
 	}
+	dir, err := os.Stat(mountDir)
+	if err != nil || !dir.IsDir() {
```
strib: You can check for a more specific `err` here by using `os.IsNotExist(err)`, instead of `err != nil`.  Presumably if you get some other error (say the disk is completely borked), you'd want to return a `libfs.InitError` as above, rather than trying to `mkdir`.

kbfsfuse/main.go
```diff
@@ -80,6 +80,13 @@ func start() *libfs.Error {
...
+		err = os.MkdirAll(mountDir, os.ModePerm)
```
strib: `os.ModePerm` seems kind of permissive.  Why not `0700`?

jzila:
I agree with @strib. If we're going to do something like that, it should go in the systemd unit, not in the Go code because that needs to stay cross-platform.

kbfsfuse/main.go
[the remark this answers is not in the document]
eli-schwartz: I was looking for something like mkdir -p, which should return its own error in that case, but on second look it seems like MkdirAll should handle existing directories and "is already a file" itself.

kbfsfuse/main.go
eli-schwartz: Shouldn't this be restricted by umask, and additionally by the permissions of the parent directory (which in the case of keybase's XDG_RUNTIME_DIR subdirectory are already 700, on top of 700 for the XDG_RUNTIME_DIR itself).

eli-schwartz:
> On macOS, the default mount directory is under /Volumes/, which isn't writeable by a non-root user. So trying to make the mount in kbfsfuse is ineffective, but more philosophically I'm worried that having the code here at all might cause problems down the road if someone sees it, thinks it should work in the general case, begins to rely on it, and then macOS breaks.

Maybe comments would help? :D What does Windows do here?

> It kind of seems like a program that creates a temporary mountpoint should also clean up after itself on exit whenever possible. I think that's why I'm more comfortable having this in run_keybase or in the systemd definitions, since those manage the entire, repeatable life-cycle of the process in the context of the full system, and not just one instance of a program. Are there other examples of arch programs that create mountpoints on the fly, and never clean them up?

run_keybase doesn't clean up either, so I don't see why this should.
For XDG_RUNTIME_DIR it is cleaned up on reboot.
For the general case why keep deleting and remaking new directories every time you restart kbfsfuse?

> ** Side note about cleaning up the mountdir: on macOS we're pretty strict about removing it after Keybase is quit, to prevent users from writing stuff in there when it's not mounted and causing confusion and future mount problems. We don't worry about that as much with Linux users, maybe because we don't provide an easy way to quit Keybase.

You don't seem to worry about it at all AFAICT, though honestly I'd expect people to wonder where their missing subdirectories that they tried to create files in, went to.

> I guess overall, if we are going to make a change here, I'd prefer it to be in the systemd scripts, rather than in kbfsfuse itself, but I could be convinced otherwise. @jzila @songgao any thoughts?

My thought on the matter is, the systemd service is a wrapper on top of starting a binary, and the binary should act as expected regardless of how it is started. If you're going to move this outside of kbfsfuse and into some wrapper, how about `/usr/bin/kbfsfuse` is a shellscript which runs mkdir then runs `/usr/bin/kbfsfuse_real`? (Yes, this is a joke.)

songgao:
Thanks for the contribution @eli-schwartz! I agree with Strib and John that if we do this it should be in the thing that starts `kbfsfuse`, i.e. systemd unit or the `run_keybase` script.

eli-schwartz:
> not in the Go code because that needs to stay cross-platform.

 The go code cannot use whatever language-specific version of #ifdef __linux__ or checking sys.platform is available?

eli-schwartz:
What about the current version?

strib:
> What about the current version?

Would it be a good compromise to add a new flag to explicitly create the directory? Then your arch distribution could set that flag, and we wouldn't risk issues with our official distributions.  I'd be ok with that.

eli-schwartz:
Perhaps it would make sense to go back to `stat`ing for the directory and setting a flag iff kbfsfuse tries to create the directory, and if it does try, then delete it in the SIGTERM handler?

Alternatively only do anything on Linux. But it appears from my cursory rethat keybase respects the "mountdir" config key on macOS too, so there would be situations where it's entirely valid to try to create it because it might be in user-controlled areas...

jzila:
Closing for archival.

PR #1672, opened 2018-07-10
kbfs_ops: avoid tracker popups on new writes post-startup

strib:
Previously, I fixed the case where kbfs edit channels were causing tracker popups on startup for no reason.  Unfortunately, this can also happen after startup, if some other device writes to the folder and causes your local device to make an FBO for it.  Fix it by just monitoring the chats, not creating the FBO.


strib:
(If we end up doing a hotfix release, we should try to get this in, but it's probably not worth delaying the release for.)

PR #1673, opened 2018-07-11
libhttpserver: Rework mime type logic

taruti:


libhttpserver/content_type_utils.go
```diff
@@ -5,40 +5,35 @@
...
 }
 
 var _ http.ResponseWriter = (*contentTypeOverridingResponseWriter)(nil)
 
 func newContentTypeOverridingResponseWriter(
-	original http.ResponseWriter) *contentTypeOverridingResponseWriter {
+	uri *url.URL, original http.ResponseWriter) *contentTypeOverridingResponseWriter {
 	return &contentTypeOverridingResponseWriter{
 		original: original,
+		mimeType: mimeTypeFromExtension(path.Ext(uri.Path)),
```
songgao: I'm not sure checking the extension here is needed. Go's `http.FileSystem` implementation already  favors extension-based mime-type detection over content-based: https://github.com/golang/go/blob/becd2a83c7a22c28bb24ac95364171b23a3d27d2/src/net/http/fs.go#L132-L135

So whatever we get here must be from the extension as long as the `mime` package knows about it, either through a system-wide mime-type file (such as `/etc/apache2/mime.types`), or from our amended mime-type file here: https://github.com/keybase/kbfs/blob/master/libmime/mime_types.go and xml is one of them.

Looks like you just need to add `xml` into the old `overrideMimeType`.

PR #1676, opened 2018-07-13
data_types: add a PrevRevisions skiplist to EntryInfo

strib:
This tracks the MD revision numbers for revisions of a file (or directory) in a skiplist-like fashion, basically giving pointers into places in the file's history, which will allow for easier access to past versions of the file.

To account for the extra metadata in the fileinfo, this also increases the max directory size by 50%.

For testing, this exposes the PrevRevision list in the `.kbfs_fileinfo` file (along with some refactoring to clean that up a bit and avoid kernel caching issues), and adds a few very simple checks to some DSL tests to make sure it shows the right number of revisions in the basic cases.

(This isn't strictly needed for time travel M1, but the earlier we get it in, the more useful it will be for people later on.)

Issue: KBFS-3194

### DIFF
--- a/libkbfs/prev_revisions.go
+++ b/libkbfs/prev_revisions.go
@@ -43,6 +43,7 @@ func (pr PrevRevisions) addRevision(
 	}
 	ret = make(PrevRevisions, newLength)
 	copy(ret, pr)
+	numDropped := 0
 	for i, prc := range ret {
 		if prc.Count == 0 {
 			// A count of 0 indicates an empty slot.
@@ -58,9 +59,13 @@ func (pr PrevRevisions) addRevision(
 				Revision: kbfsmd.RevisionUninitialized,
 				Count:    0,
 			}
+			numDropped++
 			continue
 		}
 		ret[i].Count++
+	}
+	if numDropped > 1 {
+		ret = ret[:len(ret)-(numDropped-1)]
 	}
 	for i := len(ret) - 1; i >= 1; i-- {
 		toMove := ret[i-1]
--- a/libkbfs/prev_revisions.go
+++ b/libkbfs/prev_revisions.go
@@ -5,8 +5,6 @@
 package libkbfs
 
 import (
-	"fmt"
-
 	"github.com/keybase/go-codec/codec"
 	"github.com/keybase/kbfs/kbfsmd"
 )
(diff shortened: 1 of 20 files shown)


libkbfs/prev_revisions.go
```diff
@@ -0,0 +1,105 @@
...
+			ret[i] = PrevRevisionAndCount{
+				Revision: kbfsmd.RevisionUninitialized,
+				Count:    0,
+			}
+			numDropped++
+			continue
+		} else if numDropped > 0 {
+			panic("Once we've dropped one, we should drop all the rest")
+		}
+		ret[i].Count++
```
jzila: Comment saying that here `prc` represents a revision within the range requested (`prc.Revision < r && prc.Revision > minRev`), so we count `r` in this slot.

libkbfs/prev_revisions.go
```diff
@@ -0,0 +1,105 @@
...
+		newLength++
+	}
+	ret = make(PrevRevisions, newLength)
+	copy(ret, pr)
+	earliestGoodSlot := 0
+	numDropped := 0
+
+	for i, prc := range ret {
+		if prc.Count == 255 {
+			panic("Previous revision count is about to overflow")
```
jzila: It seems like this is a distinct possibility. Shouldn't we do something other than panic?

libkbfs/prev_revisions.go
```diff
@@ -0,0 +1,105 @@
...
+	if earliestGoodSlot > 0 {
+		ret = ret[earliestGoodSlot:]
+	}
+
+	// Drop revisions off the end that are too old.
+	if numDropped == len(ret) {
+		// Leave the first slot available for overwriting.
+		ret = ret[:1]
+	} else if numDropped > 1 {
+		ret = ret[:len(ret)-(numDropped-1)]
```
jzila: so if `numDropped` is 1, then we leave the slice the same?

libkbfs/prev_revisions.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	for i := len(ret) - 1; i >= 1; i-- {
+		toMove := ret[i-1]
+		if ret[i].Count == 0 || toMove.Count >= minPrevRevisionSlotCounts[i] {
+			ret[i] = toMove
+			ret[i-1] = PrevRevisionAndCount{
+				Revision: kbfsmd.RevisionUninitialized,
+				Count:    0,
+			}
```
jzila: You can skip `toMove`, and then this can be:
```go
if ret[i].Count == 0 || ret[i-1].Count >= minPrevRevisionSlotCounts[i] {
    ret[i], ret[i-1] = ret[i-1], PrevRevisionAndCount{...}
}
```

libkbfs/prev_revisions.go
strib: I'll add a comment, but I don't think it's possible for the count to go to 255.  Outline of proof:

* The count in a slot only gets incremented when there's a new revision of the file.
* As soon as a slot has the count equal to the min-count of the slot to its right, it gets moved over, replacing whatever was there before.
* So the question is, once something gets to 100, how many file-revisions will it be before something else overwrites it?
* Someone better at math than me can probably figure out an exact answer, but intuitively the answer is less than 150, given the other min slot counts of 1,5,20,50.

Also, I modified the playground code from before to panic if it reaches 255, and set it to run for 100,000 file revisions, and it doesn't panic: https://play.golang.org/p/ltVzr1PH-MG.  The max count it sees is 178.  (It might be a partial coincidence, but that's the sum of all the slots, + 2.

libkbfs/prev_revisions.go
strib: Yes, except that the final element has been emptied out in the code above, so it will be overwritten by the shifting code below.  This will be clarified in my new comments.

strib:
Thanks @jzila!  Pushed two more commits that hopefully clear things up.

PR #1678, opened 2018-07-16
kbfs_status: show archived bytes

strib:
Hopefully right now this is zero for most users, though old bugs could have left some archived bytes around.  This will be more interesting once #1675 is live.

Issue: KBFS-3187

### DIFF
--- a/libkbfs/bserver_remote.go
+++ b/libkbfs/bserver_remote.go
@@ -533,7 +533,15 @@ func (b *BlockServerRemote) GetUserQuotaInfo(ctx context.Context) (info *kbfsblo
 		b.log.LazyTrace(ctx, "BServer: GetUserQuotaInfo done (err=%v)", err)
 	}()
 	res, err := b.getConn.getClient().GetUserQuotaInfo(ctx)
-	return kbfsblock.ParseGetQuotaInfoRes(b.config.Codec(), res, err)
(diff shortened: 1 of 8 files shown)


libkbfs/bserver_remote.go
```diff
@@ -533,7 +533,15 @@ func (b *BlockServerRemote) GetUserQuotaInfo(ctx context.Context) (info *kbfsblo
...
 	}()
 	res, err := b.getConn.getClient().GetUserQuotaInfo(ctx)
-	return kbfsblock.ParseGetQuotaInfoRes(b.config.Codec(), res, err)
+	info, err = kbfsblock.ParseGetQuotaInfoRes(b.config.Codec(), res, err)
+	if err != nil {
+		return nil, err
+	}
+	for f, us := range info.Folders {
+		b.log.CDebugf(ctx, "Folder %s, archivedBytes=%d, gitArchiveBytes=%d",
+			f, us.Bytes[kbfsblock.UsageArchive], us.Bytes[kbfsblock.UsageGitArchive])
```
jzila: Do we really want to be logging this on every `GetUserQuotaInfo`?

strib: Oops, no we don't.  This was for my personal debugging.  Will remove, thanks for catching!

PR #1685, opened 2018-07-17
folder_branch_ops: support "archive" FBOs

strib:
Archive FBOs turn off all background work, and make the whole TLF read-only. 

In a future PR, `KBFSOpsStandard` will create these according to a revision number encoded in the branch name, and set the "head" of the FBO to that exact revision.

Issue: KBFS-3203


libkbfs/readonly_node.go
```diff
@@ -35,3 +35,7 @@ func (rn ReadonlyNode) Readonly(ctx context.Context) bool {
 func (rn ReadonlyNode) WrapChild(child Node) Node {
 	return &ReadonlyNode{rn.Node.WrapChild(child)}
 }
+
+func readonlyWrapper(node Node) Node {
```
jzila: wow I'm somewhat surprised we didn't have this already.

PR #1686, opened 2018-07-18
kbfs_ops: support revision branch names

strib:
This PR introduces branch names that can refer to a specific revision, and `KBFSOpsStandard` respects that when creating FBOs.  It includes the first successful test of a time-travelling TLF!

It depends on #1685, which should be merged first.

Issue: KBFS-3204

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -373,7 +373,7 @@ const (
 // MakeRevBranchName returns a branch name specifying an archive
 // branch pinned to the given revision number.
 func MakeRevBranchName(rev kbfsmd.Revision) BranchName {
-	return BranchName(fmt.Sprintf("%s%d", branchRevPrefix, rev))
+	return BranchName(branchRevPrefix + strconv.FormatInt(int64(rev), 10))
 }
 
 // RevisionIfSpecified returns a valid revision number and true if
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -7,6 +7,7 @@ package libkbfs
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -365,7 +366,34 @@ const (
 	// folder.  Set to the empty string so that the default will be
 	// the master branch.
 	MasterBranch BranchName = ""
+
+	branchRevPrefix = "rev="
 )
+
+// MakeRevBranchName returns a branch name specifying an archive
+// branch pinned to the given revision number.
+func MakeRevBranchName(rev kbfsmd.Revision) BranchName {
+	return BranchName(fmt.Sprintf("%s%d", branchRevPrefix, rev))
+}
+
+// HasRevision returns a valid revision number and true if `bn` is a
+// revision branch.
+func (bn BranchName) HasRevision() (kbfsmd.Revision, bool) {
+	if len(bn) <= len(branchRevPrefix) {
+		return kbfsmd.RevisionUninitialized, false
+	}
+
+	if bn[:len(branchRevPrefix)] != branchRevPrefix {
+		return kbfsmd.RevisionUninitialized, false
+	}
+
+	i, err := strconv.ParseInt(string(bn[len(branchRevPrefix):]), 10, 64)
+	if err != nil {
+		return kbfsmd.RevisionUninitialized, false
+	}
+
+	return kbfsmd.Revision(i), true
+}
 
 // FolderBranch represents a unique pair of top-level folder and a
 // branch of that folder.
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -379,7 +379,7 @@ func MakeRevBranchName(rev kbfsmd.Revision) BranchName {
 // HasRevision returns a valid revision number and true if `bn` is a
 // revision branch.
 func (bn BranchName) HasRevision() (kbfsmd.Revision, bool) {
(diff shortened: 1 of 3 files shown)


libkbfs/data_types.go
```diff
@@ -364,8 +365,35 @@ const (
...
 
+// MakeRevBranchName returns a branch name specifying an archive
+// branch pinned to the given revision number.
+func MakeRevBranchName(rev kbfsmd.Revision) BranchName {
+	return BranchName(fmt.Sprintf("%s%d", branchRevPrefix, rev))
+}
+
+// HasRevision returns a valid revision number and true if `bn` is a
+// revision branch.
+func (bn BranchName) HasRevision() (kbfsmd.Revision, bool) {
```
jzila: `HasRevision` would usually imply only a bool value being returned. `MaybeGetRevision` or `RevisionIfSpecified` would be clearer.

libkbfs/kbfs_ops.go
```diff
@@ -442,9 +455,21 @@ func (fs *KBFSOpsStandard) getOrInitializeNewMDMaster(ctx context.Context,
...
+	if rev, isRevBranch := fb.Branch.HasRevision(); isRevBranch {
+		fs.log.CDebugf(ctx, "Getting archived revision %d for branch %s",
+			rev, fb.Branch)
+		md, err = getSingleMD(
+			ctx, fs.config, h.tlfID, kbfsmd.NullBranchID, rev,
+			kbfsmd.Merged, nil)
+		// This will error if there's no corresponding MD.
+		if err != nil {
+			return false, ImmutableRootMetadata{}, tlf.NullID, err
+		}
```
jzila: I think I'd rather have an explicit `return false, md, h.tlfID, nil` here, rather than falling through. Then you don't need the comment above.

libkbfs/kbfs_ops.go
```diff
@@ -1115,11 +1148,13 @@ func (fs *KBFSOpsStandard) NewNotificationChannel(
 		fs.editActivity.Add(1)
 		go func() {
 			defer fs.editActivity.Done()
-			fs.log.CDebugf(ctx, "Initializing TLF %s for the edit history",
-				handle.GetCanonicalPath())
+			fs.log.CDebugf(ctx, "%p: Initializing TLF %s for the edit history",
```
jzila: Why output the pointer address?

libkbfs/data_types.go
```diff
@@ -364,8 +365,35 @@ const (
...
 	// the master branch.
 	MasterBranch BranchName = ""
+
+	branchRevPrefix = "rev="
 )
 
+// MakeRevBranchName returns a branch name specifying an archive
+// branch pinned to the given revision number.
+func MakeRevBranchName(rev kbfsmd.Revision) BranchName {
+	return BranchName(fmt.Sprintf("%s%d", branchRevPrefix, rev))
```
songgao: This is pretty straightforward; perhaps just use `strconv` here as it's cheaper?

libkbfs/data_types.go
songgao: Or just `Revision`?

libkbfs/data_types.go
```diff
@@ -364,8 +365,35 @@ const (
...
+// HasRevision returns a valid revision number and true if `bn` is a
+// revision branch.
+func (bn BranchName) HasRevision() (kbfsmd.Revision, bool) {
+	if !strings.HasPrefix(string(bn), branchRevPrefix) {
+		return kbfsmd.RevisionUninitialized, false
+	}
+
+	if bn[:len(branchRevPrefix)] != branchRevPrefix {
+		return kbfsmd.RevisionUninitialized, false
+	}
```
songgao: Is this check necessary? Shouldn't this already be covered by the `HasPrefix` test above?

libkbfs/kbfs_ops.go
strib: I think returning an error is what we want here -- otherwise someone who looks up `.kbfs_archived_rev=10000`, when the current head is only 9000, will get an error instead of it just looking empty.  I can clarify that in the comment.

libkbfs/kbfs_ops.go
strib: Just extra debugging I left in.  Will remove, thanks!

libkbfs/data_types.go
strib: Ok, going with `RevisionIfSpecified`, since `Revision()` doesn't convey that there's a bool involved.

libkbfs/kbfs_ops.go
jzila: I meant after the error check and return.

libkbfs/kbfs_ops.go
```diff
@@ -442,9 +455,23 @@ func (fs *KBFSOpsStandard) getOrInitializeNewMDMaster(ctx context.Context,
...
+			rev, fb.Branch)
+		md, err = getSingleMD(
+			ctx, fs.config, h.tlfID, kbfsmd.NullBranchID, rev,
+			kbfsmd.Merged, nil)
+		// This will error if there's no corresponding MD, which is
+		// what we want since that means the user input an incorrect
+		// MD revision.
+		if err != nil {
+			return false, ImmutableRootMetadata{}, tlf.NullID, err
+		}
```
jzila: so:
```go
if err != nil {
    return false, ImmutableRootMetadata{}, tlf.NullID, err
}
return false, md, h.tlfID, nil
```

PR #1689, opened 2018-07-18
[KBFS-3226] Fix up disk block cache start state checks

jzila:


libkbfs/disk_block_cache.go
```diff
@@ -102,9 +102,17 @@ func rateMeterToStatus(m metrics.Meter) MeterStatus {
 	}
 }
 
+type DiskBlockCacheStartState int
```
strib: Going to need comments on these exported types and consts to pass lint.

libkbfs/disk_block_cache.go
```diff
@@ -102,9 +102,17 @@ func rateMeterToStatus(m metrics.Meter) MeterStatus {
...
+
+const (
+	DiskBlockCacheStartStateStarting DiskBlockCacheStartState = 0
```
strib: Not down with the `iota`?

jzila:
@strib yes, it'll get rid of the stack traces. I believe it'll also set the state of the disk cache to "failed", since that's what I think happened on Chris's machine; we just weren't correctly handling that scenario.

PR #1694, opened 2018-07-20
vendor chat1 updates

joshblum:
Update `ChatRPC` to implement the `chat1.NotifyChatInterface` with the proposed changes in https://github.com/keybase/client/pull/12896

cc @mmaxim once the client since is approved i'll merge this so ci can pass on the client


joshblum:
i think there is a dependency cycle in getting this fix in for circle ci, once https://github.com/keybase/client/pull/12896 is approved and these both pass jenkins i'll merge

PR #1701, opened 2018-07-24
Vendor chat1 updates for ConvSettings

joshblum:
Once https://github.com/keybase/client/pull/12963 is merged will revendor to get the proper commit hashes

### DIFF
--- a/libkbfs/chat_rpc.go
+++ b/libkbfs/chat_rpc.go
@@ -757,10 +757,10 @@ func (c *ChatRPC) ChatSetTeamRetention(
 	return nil
 }
 
-// ChatSetConvMinWriterRole implements the chat1.NotifyChatInterface for
+// ChatSetConvSettings implements the chat1.NotifyChatInterface for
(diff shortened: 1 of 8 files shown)


strib:
@joshblum I know this is just a vendor PR, but it would still be great to have a chance to review it.  Also, please follow this up with a re-vendor to get the merged commit hash from the client repo (you can push that directly to master if it's only updating the hash).

joshblum:
@strib sorry about that! There is a circular dependency between client repo circle ci passing and ci passing here, when i had initially added this (just renamed here) @jzila gave the thumbs over chat to merge w/o ci so i thought it would be best to do so again to minimize ci breakage. in the future i can get thumbs there first before merging the client side pr.

the commit hash is from client master, which hash did you prefer?

strib:
Ah ok, I didn't notice that you had updated it from client master.  Thanks!

Yeah, it's best to get thumbs on both client and kbfs first in that situation, then merge both.  Thanks!

PR #1702, opened 2018-07-24
simplefs: support by-revision archived paths

strib:
This uses keybase/client#12957 -- it should be reviewed and merged first.  CircleCI will break until then.

This PR also adds support for custom branch names to `libfs.NewFS`.

Issue: KBFS-3206

### DIFF
--- a/simplefs/simplefs_test.go
+++ b/simplefs/simplefs_test.go
@@ -616,9 +616,10 @@ type fsBlockerMaker struct {
 
 func (maker fsBlockerMaker) makeNewBlocker(
 	ctx context.Context, config libkbfs.Config,
-	tlfHandle *libkbfs.TlfHandle, subdir string) (billy.Filesystem, error) {
+	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string) (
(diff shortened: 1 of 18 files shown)


simplefs/simplefs_test.go
```diff
@@ -261,6 +261,29 @@ func TestList(t *testing.T) {
...
+
+	checkPendingOp(
+		ctx, t, sfs, opid, keybase1.AsyncOps_LIST, pathArchived,
+		keybase1.Path{}, true)
+	err = sfs.SimpleFSWait(ctx, opid)
+	require.NoError(t, err)
+	listResult, err = sfs.SimpleFSReadList(ctx, opid)
+	require.NoError(t, err)
+	require.Len(
+		t, listResult.Entries, 0, "Expected 0 directory entries in listing")
```
jzila: Could you also test one where we have an archived revision with content in it? E.g. take the revision above (with 1 directory entry), add a file to the directory, then retrieve the archived revision that should have 1 DE.

jzila: Hmm, I guess you're doing that below, so up to you whether you want to take this advice.

PR #1705, opened 2018-07-25
md_util: exported helper function for lookup up MD by server time

strib:
This depends on keybase/client#12850.  Once it's merged, I'll update the vendor.

(We can merge this PR before the server RPC is deployed, since this doesn't have any way yet of letting a user call the RPC.)

Issue: KBFS-3207


libkbfs/mdserver_disk.go
```diff
@@ -383,6 +383,50 @@ func (md *MDServerDisk) GetForTLF(ctx context.Context, id tlf.ID,
...
+		return nil, err
+	}
+
+	rmd, err := tlfStorage.getForTLF(ctx, session.UID, kbfsmd.NullBranchID)
+	if err != nil {
+		return nil, err
+	}
+
+	for rmd.untrustedServerTimestamp.After(serverTime) {
+		nextRev := rmd.MD.RevisionNumber() - 1
```
jzila: Please comment that for the disk MDServer we're doing a linear search.

strib: Sure -- where do you think the comment should be?  Right here?

jzila: I'd say just above the loop.

libkbfs/mdserver_memory.go
```diff
@@ -610,7 +658,10 @@ func (md *MDServerMemory) Put(ctx context.Context, rmds *RootMetadataSigned,
 		return kbfsmd.ServerError{Err: err}
 	}
 
-	block := mdBlockMem{encodedMd, md.config.Clock().Now(), rmds.MD.Version()}
+	// Pretend the timestamp went over RPC, so we get the same
+	// resolution level as a real server.
+	t := keybase1.FromTime(keybase1.ToTime(md.config.Clock().Now()))
```
songgao: nice

test/dsl_test.go
```diff
@@ -172,7 +172,7 @@ func (o *opt) close() {
 func (o *opt) runInitOnce() {
 	o.initOnce.Do(func() {
 		o.clock = &libkbfs.TestClock{}
-		o.clock.Set(time.Unix(0, 0))
+		o.clock.Set(time.Unix(1, 0))
```
songgao: I'm curious why this is needed?

libkbfs/mdserver_memory.go
songgao: We should probably do it [here](https://github.com/keybase/kbfs/blob/ddf361693e24e7a1c011011ee58d8a331f76febd/libkbfs/mdserver_tlf_storage.go#L187) too?

test/dsl_test.go
strib: It's needed because otherwise the ToTime+FromTime trick strip turns the timestamp into a zero timestamp, and we have checks for that in the library (to make sure an `ImmutableRootMetadata` never has an empty timestamp, for example).

PR #1708, opened 2018-07-26
[KBFS-3240] Fix windows uploads

jzila:
Local paths need to be manipulated using the `path/filepath` package, not the `path` package.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -869,6 +869,7 @@ func (k *SimpleFS) doCopyFromSource(
 
 	src, err := srcFS.Open(srcFI.Name())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


strib:
@jzila won't this change break `keybase fs` commands?  Those use `.ToSlash` on all paths before sending them down to simplefs: https://github.com/keybase/client/blob/master/go/client/cmd_simplefs.go#L89.

I think the intent that all callers were supposed to do that, hence the use of `stdpath` instead of `filepath`.

jzila:
Sigh we really need some integration tests on `keybase fs` commands. Let me look into whether filepath will work properly.

jzila:
Or we can convert them in JavaScript too. Feel free to revert this change to unbreak `keybase fs` commands, and I'll put up a client PR on Sunday evening.

strib:
Ok, will revert.

PR #1709, opened 2018-07-26
libfuse/libdokan: file for by-relative-time archives

strib:
This depends on #1707, and won't pass circle until keybase/client#12850 is merged.

This adds support for a file like `.kbfs_archived_reltime=5m` at the root of a TLF, that will contain the name of the by-revision archive directory name, corresponding to a revision of the TLF that was live at the given relative time (e.g., 5 minutes ago, in the example).

It's a file, rather than a symlink, because if it were a symlink it would form a non-universal path that you wouldn't be able to send to other users or save for later use, without risking changing the revision you are pointing to.

The relative time is any string that can be parsed by `time.ParseDuration`.

Issue: KBFS-3211

### DIFF
--- a/libdokan/archive_reltime_file.go
+++ b/libdokan/archive_reltime_file.go
@@ -1,0 +1,34 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libdokan
+
+import (
+	"time"
+
+	"github.com/keybase/kbfs/libfs"
+	"github.com/keybase/kbfs/libkbfs"
+	"golang.org/x/net/context"
+)
+
+// NewArchiveRelTimeFile returns a special read file that contains a
+// text representation of the global KBFS status.
+func NewArchiveRelTimeFile(
+	fs *FS, handle *libkbfs.TlfHandle, filename string) *SpecialReadFile {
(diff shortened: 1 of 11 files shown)


libdokan/archive_reltime_file.go
```diff
@@ -0,0 +1,34 @@
...
+			data, isRel, err := libfs.FileDataFromRelativeTimeString(
+				ctx, fs.config, handle, filename)
+			if err != nil {
+				return nil, time.Time{}, err
+			}
+			if !isRel {
+				panic("ArchiveRelTimeFile should only be used with " +
+					"reltime file names")
+			}
+			return data, time.Time{}, nil
```
songgao: I didn't find an equivalent of `entryValid` for dokan. The closest thing seems to be the [timestamps](https://github.com/keybase/kbfs/blob/537e297f17abe0e7a6d8173574ac008fd05564a9/libdokan/special_read_file.go#L39-L41) in `GetFileInformation`. So perhaps we should return `time.Now()` here?

strib: I'm just following the lead of `libfs/status_file.go` and the like, all of which return `time.Time{}`.  I don't think that returned time has anything to do with the cache time of the dir entry (which is what `entryValid` controls for fuse), it's just the atime/ctime/mtime in the fileinfo.

PR #1714, opened 2018-08-01
simplefs: support SimpleFSGetRevisions

strib:
(CircleCI will be broken until keybase/client#13082 is merged.)

This fetches either a scattered set of revisions from the past (default) or the five most recent revisions (slightly more expensive). This could involve multiple round trips to the server, though they are parallelized when possible.  Progress is available via `SimpleFSCheck`.

Issue: KBFS-3215


simplefs/simplefs.go
```diff
@@ -1388,6 +1389,231 @@ func (k *SimpleFS) SimpleFSStat(ctx context.Context, path keybase1.Path) (de key
...
+				_, prevPRs, err := k.getRevisionsFromPath(ctx, pathToPrev)
+				if _, isGC := err.(libkbfs.RevGarbageCollectedError); isGC {
+					k.log.CDebugf(ctx, "Hit a GC'd revision: %d",
+						lastRevision-1)
+					break
+				} else if err != nil {
+					return nil, err
+				}
+				if len(prevPRs) == 0 {
+					return nil, simpleFSError{"No previous revisions"}
```
jzila: Why is this an error? Shouldn't we still return if there are only 3 revisions?

strib: It's because if the file can be stat'd, the previous revisions should always include at least it's own revision (since we're looking at a revision from a previous list, so we know the file should be there). So I guess this error text is a bit confusing, I'll change the text.

strib: Let me amend that slightly: we looking up `lastRevision - 1` here, but we know there's another slot in the previous revisions list at this point (because `nextSlot < len(prs)`, so we know that there is _some_ other revision between the last one and the next one, so there should definitely be a stat'able file and a non-empty previous revisions list here.

jzila: Got it, that makes much more sense than my interpretation. Maybe a comment explaining that here would help.

strib: Sure, done in https://github.com/keybase/kbfs/commit/e0c60848371b4e0169b2cd81b8aff79c06bdc68b.

PR #1716, opened 2018-08-02
prev_revisions: avoid panic when squashing all revisions in list

strib:
If the prev revisions list is full of revision numbers that are less than the new number `r`, we were clearing out the entire list without leaving an extra slot.  I think this can happen when squashing many journal revisions together into one, where the squashed one has a equal or lower revision number than the others.

Issue: keybase/client#12174


strib:
I'm going to merge this as soon as it passes CI, to get it into the morning's linux build.  But will address any later comments in a later PR.

PR #1721, opened 2018-08-06
add filter to SimpleFSSyncStatus

songgao:
Will update vendor to master when merging. Client PR: https://github.com/keybase/client/pull/13165

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -1755,7 +1755,7 @@ func (k *SimpleFS) SimpleFSDumpDebuggingInfo(ctx context.Context) error {
 }
 
 // SimpleFSSyncStatus - Get sync status.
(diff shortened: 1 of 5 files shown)


simplefs/simplefs.go
```diff
@@ -1771,6 +1771,18 @@ func (k *SimpleFS) SimpleFSSyncStatus(ctx context.Context) (keybase1.FSSyncStatu
...
+			if isFiltered(filter, p) {
+				continue
+			}
+			syncingPaths = append(syncingPaths, p)
+		}
+	}
+
 	k.log.CDebugf(ctx, "Sending sync status response with %d syncing bytes",
 		status.UnflushedBytes)
 	return keybase1.FSSyncStatus{
```
jzila: Looks like you forgot to return `syncingPaths` here.

songgao: whoops good catch

PR #1723, opened 2018-08-06
don't automatically add favorites in SimpleFS operations

songgao:
This would affect `keybase fs` too, but as John suggested we can just add `keybase fs favorites add/ignore`.

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -705,6 +705,27 @@ const (
 	FavoritesOpNoChange
 )
 
+func (o FavoritesOp) String() string {
+	switch o {
+	case FavoritesOpAdd:
+		return "Add"
+	case FavoritesOpAddNewlyCreated:
+		return "AddNewlyCreated"
+	case FavoritesOpRemove:
+		return "Remove"
+	case FavoritesOpNoChange:
+		return "NoChange"
+	default:
+		return "Unknown"
+	}
+}
+
+// CtxKeyFavoritesOp is the context key which when present, controls whether an
+// operation should result in adding a TLF into favorites. Explicit favorites
+// related calls (e.g. AddFavorite and DeleteFavorite) are not affected by
+// this.
+type CtxKeyFavoritesOp struct{}
+
 // RekeyResult represents the result of an rekey operation.
 type RekeyResult struct {
 	DidRekey      bool
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -353,10 +353,21 @@ func (fs *KBFSOpsStandard) getOpsIfExists(
 	return fs.ops[fb]
 }
 
+func (fs *KBFSOpsStandard) getFavoritesOpFromContextWithDefault(ctx context.Context, defaultFop FavoritesOp) FavoritesOp {
+	if ctxFop, ok := ctx.Value(CtxKeyFavoritesOp{}).(FavoritesOp); ok {
+		fs.deferLog.CDebugf(ctx, "getOps: Using FavoritesOp from ctx: %s", ctxFop)
+		return ctxFop
+	}
+	fs.deferLog.CDebugf(ctx, "getOps: Using default FavoritesOp: %s", defaultFop)
+	return defaultFop
+}
+
 func (fs *KBFSOpsStandard) getOps(ctx context.Context,
-	fb FolderBranch, fop FavoritesOp) *folderBranchOps {
+	fb FolderBranch, defaultFop FavoritesOp) *folderBranchOps {
 	ops := fs.getOpsNoAdd(ctx, fb)
-	if err := ops.doFavoritesOp(ctx, fs.favs, fop, nil); err != nil {
+	if err := ops.doFavoritesOp(ctx, fs.favs,
+		fs.getFavoritesOpFromContextWithDefault(ctx, defaultFop),
+		nil); err != nil {
 		// Failure to favorite shouldn't cause a failure.  Just log
 		// and move on.
 		fs.log.CDebugf(ctx, "Couldn't add favorite: %v", err)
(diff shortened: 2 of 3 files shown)


songgao:
> Code looks fine, but should we wait to merge it until we have the explicit add/remove calls as well and are calling them in the right places in the GUI?

Sure; I opened the PR partly as a continuous discussion. Now that it sounds good to you folks, I'm gonna add the two commands to `keybase fs`.

> when people get a new folder notification and click into it from the GUI, it won't show up in their favorites list and they won't see it in Finder.

Good point; I didn't realize new folders don't actually get added into the list until user goes into it. It sounds like this is actually one of those situations where we should actually automatically add to favorites. WDYT?

strib:
> It sounds like this is actually one of those situations where we should actually automatically add to favorites. WDYT?

What situation are you talking about?  The whole reason favorites are added on first access to begin with is because we don't want them auto-added on creation, because then people could just spam your favorites list by making folders with you.  I think we should preserve that behavior in the GUI -- if we can't do it the way we've been doing that for technical reasons related to async operations, we should do it with explicit calls.

I'm not exactly convinced that there's not a better way to handle this, by say serializing requests coming in through the GUI to make sure that folder accesses and ignores happen strictly one after another.  But I have no clue how that code works, and that might not be possible, so I defer to your judgement on the best way to do that.

jzila:
> Good point; I didn't realize new folders don't actually get added into the list until user goes into it. It sounds like this is actually one of those situations where we should actually automatically add to favorites. WDYT?

Have we considered just handling this in the UI? I.e. if you've navigated into a TLF that you haven't favorited, have a banner at the top that asks you to favorite it.

strib:
> Have we considered just handling this in the UI? I.e. if you've navigated into a TLF that you haven't favorited, have a banner at the top that asks you to favorite it.

I don't remember discussions about that, but I wasn't in on the design meetings.  But in general it seems like a bad policy to make the UI more complicated, in order to compensate for buggy/non-deterministic code.

jzila:
I don't agree that not favoriting is buggy/non-deterministic.

We currently favorite automatically upon navigation into a TLF because the filesystem doesn't have UX capabilities to allow the user to choose.

Now that we have a UI where we have full freedom in making things opt-in or opt-out, we should at least step back to reevaluate what is the right default behavior for the user.

jzila:
Oh I follow your logic now, strib. You were saying that we shouldn't be looking at a UI-based solution to address the current buggy/non-deterministic ignore/favorite race.

Fair enough, but I think these two issues are in the grand scheme of things orthogonal. Yes, we could fix the buggy behavior to behave just like the FS does. But maybe this is a good opportunity to evaluate the right experience for the user.

strib:
Yep, we're on the same page now.  I'm just channeling my inner-malgorithms and venturing that the design team will favor fewer options for the user in almost all cases.  I think that's why we never had a `keybase fs add` command in the first place.

songgao:
> What situation are you talking about?

I'm talking about the flow where user sees the badged folder, and goes into the folder from GUI, where we should probably automatically add it (or better, prompt user to add it by clicking a button, but need design).

> The whole reason favorites are added on first access to begin with is because we don't want them auto-added on creation, because then people could just spam your favorites list by making folders with you.

If spamming is a concern, badging is kind of the same right? I guess user can click "Ignore", but with the above proposal, I think it works the same way.

> I'm not exactly convinced that there's not a better way to handle this, by say serializing requests coming in through the GUI to make sure that folder accesses and ignores happen strictly one after another.

I think there are things we can do, either in the KBFS process or in the frontend. This is not because there isn't another way to solve it, but that with GUI work, there are now ways to add/remove favorites other than through the FUSE interface, so I took it as a chance to make a change to hopefully improve the UX.

For now, it probably won't make much of a difference, but once we support opening arbitrary path in GUI, I feel like a more predictable behavior about adding stuff to favorites would be desirable.

strib:
> If spamming is a concern, badging is kind of the same right? I guess user can click "Ignore", but with the above proposal, I think it works the same way.

Yeah, but as you say it's at least in a place where ignores can happen, and it feels better to have the spammy stuff separated from the regular stuff with a label/badge, which isn't possible in the favorites list in the CLI/finder.

> I took it as a chance to make a change to hopefully improve the UX.

I disagree that it would improve the UI.  It seems better to just do the expected thing under the hood, without bothering the user to make a decision about it. Just personal opinion though.

> For now, it probably won't make much of a difference, but once we support opening arbitrary path in GUI, I feel like a more predictable behavior about adding stuff to favorites would be desirable.

I don't understand this part about opening arbitrary paths.

songgao:
Just to be clear, the proposal is to:

* not add to favorites in any of simpleFS operations;
* if a folder is badged (newly created), add it to favorites automatically when user goes into that folder in GUI, or access it from `keybase fs`. This is what I meant by the situation where we automatically add favorites.

With above two points, it's actually completely the same for GUI now, since at this time the only way for user to enter or interact with a TLF is through a TLF list, which for now we include the favorited TLFs, and badged new TLFs.  So By combining above two points, we are not changing any behavior. The difference comes when we support opening arbitrary KBFS path in GUI.

For example, we can start to make stuff like `/keybase/private/strib,songgao/wifi-password` into a link, and when user clicks it, we show that path inside GUI's File's tab. This would make it possible to navigate into a TLF without going through a TLF list, so it's possible to (in GUI) interact with a TLF that's neither favorited nor badged. With the above 2 points, this wouldn't result in favoriting the TLF. But with what we have in master now (i.e. before this PR), it would favorite the TLF.

songgao:
Ah and I just saw the discussion above you had @strib @jzila . Somehow GitHub didn't show me or bring them to my intention earlier. Sorry for missing them.

Re whether to have a banner/button to ask user to add a non-favorited TLF to favorites: I don't think either way it'd make a big difference now because the only way user goes into a TLF for now is through the TLF list that we show, which has favorited TLFs and badged new ones. So perhaps we can probably bring it up with the design team and see how they think?

songgao:
Yeah basically, but not entirely the same. We'd only favorite new badged TLFs, at the time when user goes into it. But if user does stuff with SimpleFS (or an old Keybase window somehow) in an ignored (or `rmdir`ed) TLF, this wouldn't favorite them again. For normal flows, yes the behavior would be the same.

strib:
> For example, we can start to make stuff like /keybase/private/strib,songgao/wifi-password into a link, and when user clicks it, we show that path inside GUI's File's tab. This would make it possible to navigate into a TLF without going through a TLF list, so it's possible to (in GUI) interact with a TLF that's neither favorited nor badged. With the above 2 points, this wouldn't result in favoriting the TLF. But with what we have in master now (i.e. before this PR), it would favorite the TLF.

Seems to me like this _should_ favorite the TLF.  If the user interacted in a useful way with the folder, we should always favorite it so they can find it later.  That's just my personal opinion though, it should be a question for design.

jzila:
Closing for archival. Not sure if this PR is relevant any more.

PR #1724, opened 2018-08-07
libfs: return ErrNotExist if a non-final path element exists but is not a dir

songgao:


libfs/fs.go
```diff
@@ -270,6 +280,7 @@ func (fs *FS) lookupParentWithDepth(
 			return nil, "", "", err
 		}
 
+		fmt.Printf("SONGGAO parts=%#+v\n", parts)
```
jzila: 🔥

PR #1725, opened 2018-08-07
it would be useful to know right when os.Exit(0) is called

maxtaco:


kbfsfuse/main.go
```diff
@@ -114,5 +114,6 @@ func main() {
 
 		os.Exit(err.Code)
 	}
+	fmt.Fprintf(os.Stderr, "kbfsfuse exiting with code=0")
```
strib: Should end in `\n`.

jzila:
Closing for archival.

PR #1732, opened 2018-08-10
file_data: generalize functions for getting blocks

strib:
When we introduce indirect directory blocks, we'll want to reuse the logic for navigating a tree of blocks.  Do this by adding generic offset types and some new helper functions to the Block interface, and refactoring functions from `fileData` into a new generic `blockTree` struct that can operate on any type of block.

This commit shouldn't change any logicc, it's just a refactor

Issue: KBFS-3299

### DIFF
--- a/libkbfs/block_tree.go
+++ b/libkbfs/block_tree.go
@@ -1,0 +1,247 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import "context"
+
+// blockGetter is a function that gets a block suitable for reading or
+// writing, and also returns whether the block was already dirty.  It
+// may be called from new goroutines, and must handle any required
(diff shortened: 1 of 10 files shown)


strib:
To review, I recommend:

1. `interfaces.go`
2. `block_types.go`
3. `block_tree.go` (This is just some functions ported over from `file_data.go` and made generic.)
4. `file_data.go` (This is just converting types and some logic to use the new `blockTree` and types.)
5. Tests.  (Again, just converted over to use new types.)

libkbfs/block_tree.go
```diff
@@ -0,0 +1,247 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import "context"
+
+// blockGetter is a function that gets a block suitable for reading or
```
jzila: `blockGetterFn`

PR #1735, opened 2018-08-11
dir_data: implement the read-only methods

strib:
This PR introduces a `dirData` class that can understand multi-block directories, and it implements read-only operations on it and tests for them.  (But it's not in use in the rest of the code yet.)

A future PR will implement the more complicated modifying methods.

Depends on #1732 and #1734.

Issue: KBFS-3301

### DIFF
--- a/libkbfs/block_types.go
+++ b/libkbfs/block_types.go
@@ -38,24 +38,34 @@ func (i Int64Offset) Less(other Offset) bool {
 // StringOffset represents the offset of a block within a directory.
 type StringOffset string
 
-var _ Offset = StringOffset("")
+var _ Offset = (*StringOffset)(nil)
 
 // Equals implements the Offset interface for StringOffset.
-func (s StringOffset) Equals(other Offset) bool {
-	otherS, ok := other.(StringOffset)
+func (s *StringOffset) Equals(other Offset) bool {
+	if s == nil {
+		return other == nil
+	} else if other == nil {
+		return false
+	}
+	otherS, ok := other.(*StringOffset)
 	if !ok {
 		panic(fmt.Sprintf("Can't compare against non-string offset: %T", other))
 	}
-	return string(s) == string(otherS)
+	return string(*s) == string(*otherS)
 }
 
 // Less implements the Offset interface for StringOffset.
-func (s StringOffset) Less(other Offset) bool {
-	otherS, ok := other.(StringOffset)
+func (s *StringOffset) Less(other Offset) bool {
+	if s == nil {
+		return other != nil
+	} else if other == nil {
+		return false
+	}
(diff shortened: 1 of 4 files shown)


libkbfs/dir_data.go
```diff
@@ -0,0 +1,110 @@
...
+	_, blocks, _, err := dd.tree.getBlocksForOffsetRange(
+		ctx, dd.rootBlockPointer(), topBlock, topBlock.FirstOffset(), nil,
+		false, true)
+	if err != nil {
+		return nil, err
+	}
+
+	numEntries := 0
+	for _, b := range blocks {
+		numEntries += len(b.(*DirBlock).Children)
```
jzila: Is it worth two iterations to get the benefit of a single allocation of the slice?

strib: Actually yes!  I wrote a small program: https://play.golang.org/p/pZZG5V372W1

Executed on my machine, two loops is always faster than one loop:

```sh
$ ./x 10
One loop: 1.059µs
Two loops: 300ns
$ ./x 100
One loop: 13.799µs
Two loops: 426ns
$ ./x 1000
One loop: 20.654µs
Two loops: 7.353µs
$ ./x 10000
One loop: 384.259µs
Two loops: 71.895µs
$ ./x 100000
One loop: 2.366071ms
Two loops: 235.291µs
$ ./x 1000000
One loop: 14.520121ms
Two loops: 3.193464ms
$ ./x 10000000
One loop: 139.688947ms
Two loops: 27.569866ms
```

songgao: THere's a third version :) https://play.golang.org/p/5l_xphR2NqZ

```
%  go run main.go 10000              ✘ 130
One loop: 346.775µs
One loop2: 86.786µs
Two loops: 115.841µs
```

strib: @songgao yeah, but that one doesn't apply to this code since we need to compute the new size from the data in the slice, we can't simply use the slice length.

songgao: Ah that's right! I mis-read. Thanks.

libkbfs/block_types.go
```diff
@@ -38,24 +38,34 @@ func (i Int64Offset) Less(other Offset) bool {
 // StringOffset represents the offset of a block within a directory.
 type StringOffset string
 
-var _ Offset = StringOffset("")
+var _ Offset = (*StringOffset)(nil)
```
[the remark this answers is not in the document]
jzila: That sounds reasonable, thanks.

libkbfs/dir_data.go
jzila: Great test, thanks!

PR #1736, opened 2018-08-13
block_tree: refactor tree update methods out of fileData

strib:
In a future PR, `dirData` will start using these methods.

Also, a few other cleanups:

* Separate the indirect pointer methods of blocks out into their own interface, so we don't have to implement them in `CommonBlock` and test block types.
* Have `dirData` and `fileData` use the fields of `blockTree` for most of their internal fields, to simplify things and avoid unnecessary overhead.

Issue: KBFS-3301

### DIFF
--- a/libkbfs/block_types.go
+++ b/libkbfs/block_types.go
@@ -113,6 +113,8 @@ type CommonBlock struct {
 	cachedEncodedSize uint32
 }
 
+var _ Block = (*CommonBlock)(nil)
+
 // GetEncodedSize implements the Block interface for CommonBlock
 func (cb *CommonBlock) GetEncodedSize() uint32 {
 	cb.cacheMtx.RLock()
@@ -147,21 +149,6 @@ func (cb *CommonBlock) IsIndirect() bool {
 	return cb.IsInd
 }
 
-// FirstOffset implements the Block interface for CommonBlock.
-func (cb *CommonBlock) FirstOffset() Offset {
-	panic("CommonBlock doesn't implement indirect pointer methods")
-}
-
-// NumIndirectPtrs implements the Block interface for CommonBlock.
-func (cb *CommonBlock) NumIndirectPtrs() int {
-	panic("CommonBlock doesn't implement indirect pointer methods")
-}
-
-// IndirectPtr implements the Block interface for CommonBlock.
-func (cb *CommonBlock) IndirectPtr(_ int) (BlockInfo, Offset) {
-	panic("CommonBlock doesn't implement indirect pointer methods")
-}
-
 // OffsetExceedsData implements the Block interface for CommonBlock.
 func (cb *CommonBlock) OffsetExceedsData(_, _ Offset) bool {
 	panic("CommonBlock doesn't implement data methods")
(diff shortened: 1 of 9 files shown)


strib:
Once again:

1. `interfaces.go`
2. `block_types.go`
3. `dir_data.go` and test (just a refactor)
4. `file_data.go` and test (just a refactor)

libkbfs/block_types.go
```diff
@@ -175,6 +162,11 @@ func (cb *CommonBlock) Set(other Block) {
 	cb.SetEncodedSize(otherCopy.GetEncodedSize())
 }
```
songgao: Super nit: I can't comment there, but the `otherCopy` above is not a copy, so perhaps `otherCommon`?

libkbfs/interfaces.go
```diff
@@ -115,6 +115,22 @@ type Block interface {
...
+	// is greater than what currently could be stored in this block.
+	OffsetExceedsData(startOff, off Offset) bool
+	// BytesCanBeDirtied returns the number of bytes that should be
+	// marked as dirtied if this block is dirtied.
+	BytesCanBeDirtied() int64
+}
+
+// BlockWithPtrs defines methods needed for interacting with indirect
+// pointers.
+type BlockWithPtrs interface {
```
jzila: Wouldn't it be better for this to be called `IndirectBlock` or `BlockIndirect`?

libkbfs/block_types.go
```diff
@@ -440,26 +542,63 @@ func (fb *FileBlock) IndirectPtr(i int) (BlockInfo, Offset) {
 	return iptr.BlockInfo, iptr.Off
 }
 
-// OffsetExceedsData implements the Block interface for FileBlock.
-func (fb *FileBlock) OffsetExceedsData(startOff, off Offset) bool {
-	if fb.IsInd {
-		panic("OffsetExceedsData called on an indirect file block")
+// AppendNewIndirectPtr implements the Block interface for FileBlock.
```
jzila: Some of these say they implement the `Block` interface but they actually implement the `BlockWithPtrs` interface.

libkbfs/block_types.go
```diff
@@ -294,13 +360,22 @@ type FileBlock struct {
...
 func NewFileBlock() Block {
 	return &FileBlock{
 		Contents: make([]byte, 0, 0),
 	}
 }
 
+// NewFileBlockWithPtrs creates a new, empty FileBlock.
+func NewFileBlockWithPtrs(isInd bool) BlockWithPtrs {
+	fb := NewFileBlock().(*FileBlock)
+	fb.IsInd = isInd
```
jzila: Is it possible to make a `BlockWithPtrs` that isn't indirect? I don't see this parameter used.

libkbfs/interfaces.go
strib: It's doesn't necessarily represent an indirect block though.  It represents a block type that _can_ have indirect pointers (as opposed to `CommonBlock` and other test blocks).  But it can just as easily be a direct `DirBlock` or `FileBlock`, where `NumIndirectPtrs() == 0`.  I'm definitely open to a different name, but I don't think these names are right.

libkbfs/block_types.go
strib: Yes, it's used here: https://github.com/keybase/kbfs/pull/1736/files#diff-27199ce3c4667e207829d68cf8646446R543

Like I said above, any dir/file block can be a `BlockWithPtrs`, and here it's used to create a new right-most leaf block for the block tree.

jzila: OK so we're talking about any block that can be converted to an indirect block.

strib: Yes, right.

jzila: All right let's keep it. Not a big deal anyway.

PR #1738, opened 2018-08-15
dir_data: implement dirblock-splitting and add/remove/update methods

strib:
This adds a `BlockSplitter` function for splitting up directory blocks that get too big, and implements `addEntry`, `removeEntry` and `updateEntry` methods for `dirData`.

It also adds a bunch of tests for these new methods in `dir_data_test.go` (which github has collapsed).

The basic algorithm is the same as for files.  If a leaf directory block has more child entries than it can handle, traverse up the path of the block tree until we find a parent with an empty indirect pointer slot, and add a block to the right.  (If no parent has room, add another layer of indirection at the top.)  Then shift the new block over at its level in the tree, until reaching the place where its offset fits in.

Note that this builds on #1736, and also modifies the existing block-shifting code a bit because, unlike for file trees, directory trees are **not** necessarily balanced when they are built.  This means that any indirect-pointer list in the tree, and not just those along the right edge, can be less than full, which required a little extra code to handle.

If we decide it's important later, we can write code to rebalance the trees.  It won't change anything about the data structures or the way directories are read, so it's not imperative that we do it now, and I think it would be tricky and take a long time to get right.

Also note that this code doesn't yet try to handle the updated pointers and block refs that are dirtied when the leaf blocks are modified.  The next PR will likely need to fix a few issues with block accounting that may occur when integrating this into `folderBlockOps`.  So none of that stuff is tested yet, or made available in the `dirData` interface.

Issue: KBFS-3301

### DIFF
--- a/libkbfs/bsplitter_simple.go
+++ b/libkbfs/bsplitter_simple.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/keybase/kbfs/kbfscodec"
 )
@@ -16,6 +17,7 @@ type BlockSplitterSimple struct {
 	maxSize                 int64
 	maxPtrsPerBlock         int
 	blockChangeEmbedMaxSize uint64
+	maxDirEntriesPerBlock   int
 }
 
 // NewBlockSplitterSimple creates a new BlockSplittleSimple and
@@ -87,6 +89,7 @@ func NewBlockSplitterSimple(desiredBlockSize int64,
 		maxSize:                 maxSize,
 		maxPtrsPerBlock:         maxPtrs,
 		blockChangeEmbedMaxSize: blockChangeEmbedMaxSize,
+		maxDirEntriesPerBlock:   0, // disabled for now
 	}, nil
 }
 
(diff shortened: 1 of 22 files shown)


libkbfs/bsplitter_simple.go
```diff
@@ -147,3 +150,36 @@ func (b *BlockSplitterSimple) ShouldEmbedBlockChanges(
...
+	// Sort the entries and split them down the middle.
+	names := make([]string, 0, len(block.Children))
+	for name := range block.Children {
+		names = append(names, name)
+	}
+
+	sort.Strings(names)
+	// Delete the second half of the names from the original block,
+	// and add to the new block.
+	newBlock := NewDirBlock().(*DirBlock)
```
jzila: I'm thinking that `NewDirBlock()` should just return a `*DirBlock` instead of a `Block`; or do you depend on the nil interface behavior somewhere?

libkbfs/bsplitter_simple.go
```diff
@@ -147,3 +150,36 @@ func (b *BlockSplitterSimple) ShouldEmbedBlockChanges(
...
+	startOff := int(len(names) / 2)
+	for i := startOff; i < len(names); i++ {
+		name := names[i]
```
jzila: Since you don't use `i` anywhere, this could be:
```go
for name := range names[int(len(names) / 2):] {
```

libkbfs/bsplitter_simple.go
strib: Not sure what you mean by "nil interface", but `getBlockHelperLocked()` takes a function that returns a new `Block`, and `NewDirBlock` is one thing that's passed to it.  (It's worked that way forever.)

jzila: nil interface = `(Block)(nil)`, which is different from `(*DirBlock)(nil)`.

i.e. `((Block)(nil) == (*DirBlock)(nil)) == false`

jzila: https://play.golang.org/p/EZ9-nxvjjGN

strib: Oh i think I see.  I depend on the function type of `NewDirBlock` to return exactly a `Block`, so that I can pass it into `getBlockHelperLock`.  To extend your example, this wouldn't work if I changed it to return a `Block`: https://play.golang.org/p/vhmzqPo1z34

jzila: Understood. That answers my question, thanks!

PR #1739, opened 2018-08-16
root_metadata: add an interface for KeyMetadata + root dir entry

strib:
And use this in `crChains`.

It seems kind of random, but this will be needed in an upcoming PR when we need to replace some of directory entry cache stuff in `folderBlockOps` with `dirData`.

Issue: KBFS-3302

### DIFF
--- a/libkbfs/conflict_resolver.go
+++ b/libkbfs/conflict_resolver.go
@@ -2709,11 +2709,10 @@ func (cr *ConflictResolver) resolveOnePath(ctx context.Context,
 				newPtrs[ptr] = true
 			}
 
+			mdInfo := unmergedChains.mostRecentChainMDInfo
 			nodeMap, cache, err := cr.fbo.blocks.SearchForNodes(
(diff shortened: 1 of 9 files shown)


libkbfs/conflict_resolver.go
```diff
@@ -2714,8 +2711,9 @@ func (cr *ConflictResolver) resolveOnePath(ctx context.Context,
 
 			nodeMap, cache, err := cr.fbo.blocks.SearchForNodes(
 				ctx, cr.fbo.nodeCache, ptrs, newPtrs,
-				unmergedChains.mostRecentChainMDInfo.kmd,
-				unmergedChains.mostRecentChainMDInfo.rootInfo.BlockPointer)
+				unmergedChains.mostRecentChainMDInfo,
+				unmergedChains.mostRecentChainMDInfo.GetRootDirEntry().
+					BlockPointer)
```
jzila: I'd rather not break up a single parameter here.

PR #1744, opened 2018-08-20
folder_update_prepper: use `dirData` when prepping paths

strib:
This updates `folderUpdatePrepper` to use `dirData` when updating dir entries, and to ready any dirty blocks within the directory when prepping the paths for commit.

The ready path() isn't yet exercised in any tests though; a future PR may end up having to fix bugs here.  It's split out in this way, pre-testing, for ease of review.

In addition, this PR:
* refactors ready functions out of `fileData` into `blockTree` (no behavior change)
* makes `blockTree` use a restricted interface for checking dirtiness, to make it easier to pass in other kinds of dirty-block-holders, like `localBcache`.

Issue: KBFS-3302

### DIFF
--- a/libkbfs/block_tree.go
+++ b/libkbfs/block_tree.go
@@ -10,6 +10,7 @@ import (
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/kbfs/kbfsblock"
+	"github.com/keybase/kbfs/tlf"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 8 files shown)


libkbfs/block_tree.go
```diff
@@ -794,3 +799,114 @@ func (bt *blockTree) markParentsDirty(parentBlocks []parentBlockAndChildIndex) (
...
+
+	// Starting from the leaf level, ready each block at each level,
+	// and put the new BlockInfo into the parent block at the level
+	// above.  At each level, only ready each block once. Don't ready
+	// the root block though; the folderUpdatePrepper code will do
+	// that.
+	for level := len(pathsFromRoot[0]) - 1; level > 0; level-- {
+		for i := 0; i < len(pathsFromRoot); i++ {
+			// Ready the dirty block.
+			pb := pathsFromRoot[i][level]
```
jzila: I know this is just a refactor, but are we guaranteed that each `pathsFromRoot[i]` will have `len(pathsFromRoot[0])` levels?

strib: That's a great question.  I think the answer is yes, because of the way making new blocks work.  If a block is split, and all parent blocks are full, we add a new block level above the top of the whole tree, and then create a new branch on the right side extending all the way down.  So the tree should have a balanced number of levels.

So actually I think comments I've written in the past, about the directory tree not necessarily being balanced, weren't correct.

PR #1749, opened 2018-08-21
folder_update_prepper: account properly for indirect dir pointers

strib:
Manipulating a multi-level directory block tree ends up adding and unreferencing pointers, and we need to track these in the MD in the same way that we do for files.  So track the unreferenced pointers in `folderBlockOps` as they are dirtied, and reference new pointers made as part of readying the directory.

Also, `StateChecker` now properly checks indirect directory blocks.

To test this, I locally enabled directory block-splitting for a complicated concurrency test, but I don't want to check that in quite yet because that particular test gets very slow; I want to build a real way to test via the DSL tests first.

Issue: KBFS-3302


libkbfs/folder_block_ops.go
```diff
@@ -941,6 +971,20 @@ func (fbo *folderBlockOps) newDirDataWithLBC(lState *lockState,
...
 }
 
+func (fbo *folderBlockOps) makeDirDirtyLocked(
+	lState *lockState, ptr BlockPointer, unrefs []BlockInfo) func() {
+	fbo.blockLock.AssertLocked(lState)
+	oldUnrefs, wasDirty := fbo.dirtyDirs[ptr]
+	fbo.dirtyDirs[ptr] = append(oldUnrefs, unrefs...)
+	return func() {
+		if wasDirty {
+			fbo.dirtyDirs[ptr] = oldUnrefs[:len(oldUnrefs)]
```
songgao: Did you mean to do `fbo.dirtyDirs[ptr] = oldUnrefs[:len(oldUnrefs):len(oldUnrefs)]` here? It seems the length of `oldUnrefs` never changed.

libkbfs/state_checker.go
```diff
@@ -48,19 +48,40 @@ func (sc *StateChecker) findAllFileBlocks(ctx context.Context,
 	return nil
 }
 
+// findAllDirBlocks adds all dir blocks found under this block to
+// the blocksFound map, if the given path represents an indirect
```
songgao: s/blocksFound/blocksSizes

libkbfs/folder_block_ops.go
jzila: This is an undo function, so it's restoring the state before `makeDirDirtyLocked` was applied.

strib: Yeah I was just worried that the above `append` would have changed the length of `oldUnrefs`, and I was trying to get exactly the original slice.  Does the double-colon syntax help do that, I've never seen it before.  I can try it.

jzila: Looks like the third index sets the capacity of the slice. Seems redundant in this situation.
I wonder if you can just do `oldUnrefs[:]` here though.

songgao: The double colon limits the underlying capacity, to make sure an `append` at a later time can't override underlying data. So for example, if you have

```
a := []int{1, 2, 3}
b := a[:2]
c := a[:2:2]
```

Then `d := append(b, 1)` can override the third element in `a`, but `d := append(c, 1)` can't: https://play.golang.org/p/xACLZaiHrMk

If all you care is the length, then just returning `oldUnrefs` would be the same right?

jzila: Wow, TIL! That definitely breaks an assumption I had about how slices work.

songgao: I used to get confused about how slice works, but this showed up and explained it very well: https://blog.golang.org/go-slices-usage-and-internals

strib: Cool thanks, good explanation!  You're right it can probably just be `unrefs`, but it doesn't seem worth changing at this point and this way is still a little clearer to me.

songgao: Yeah; not suggesting a change; Just a discussion. What you have now looks great to me!

PR #1750, opened 2018-08-21
throttle NotifySync and NotifyPathUpdated to one per second

songgao:
The NotifySync notifications tend to come in clustered (up to more than 10 within a second), so this seems pretty helpful.

This will need to go in at the same time as https://github.com/keybase/client/pull/13408

### DIFF
--- a/libkbfs/reporter_kbpki.go
+++ b/libkbfs/reporter_kbpki.go
@@ -82,9 +82,7 @@ type ReporterKBPKI struct {
 	notifyBuffer     chan *keybase1.FSNotification
 	notifyPathBuffer chan string
 	notifySyncBuffer chan *keybase1.FSPathSyncStatus
-	suppressCh       chan time.Duration
 	canceler         func()
-	done             <-chan struct{}
 }
 
 // NewReporterKBPKI creates a new ReporterKBPKI.
@@ -96,11 +94,9 @@ func NewReporterKBPKI(config Config, maxErrors, bufSize int) *ReporterKBPKI {
 		notifyBuffer:     make(chan *keybase1.FSNotification, bufSize),
 		notifyPathBuffer: make(chan string, bufSize),
 		notifySyncBuffer: make(chan *keybase1.FSPathSyncStatus, bufSize),
-		suppressCh:       make(chan time.Duration, 1),
 	}
 	var ctx context.Context
 	ctx, r.canceler = context.WithCancel(context.Background())
-	r.done = ctx.Done()
 	go r.send(ctx)
 	return r
 }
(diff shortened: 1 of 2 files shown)


simplefs/simplefs.go
```diff
@@ -1885,8 +1886,7 @@ func (k *SimpleFS) TlfHandleChange(_ context.Context, _ *libkbfs.TlfHandle) {
 // SimpleFSSuppressNotifications suppresses FSEvent and FSSyncEvent
 // notifications.
 func (k *SimpleFS) SimpleFSSuppressNotifications(ctx context.Context, suppressDurationSec int) error {
-	k.config.Reporter().SuppressNotifications(ctx, time.Duration(suppressDurationSec)*time.Second)
-	return nil
+	return errors.New("this rpc is deprecated")
```
songgao: I'll have a separate set of PRs to actually remove this from the protocol after these are merged.

libkbfs/reporter_kbpki.go
```diff
@@ -282,9 +269,17 @@ func (r *ReporterKBPKI) send(ctx context.Context) {
 			if !ok {
 				return
 			}
-			if suppressed {
-				stagedNotification = notification
-			} else if err := r.config.KeybaseService().Notify(ctx,
+			nt := notification.NotificationType
+			// Only these notifications are used in frontend:
+			// https://github.com/keybase/client/blob/0d63795105f64289ba4ef20fbefe56aad91bc7e9/shared/util/kbfs-notifications.js#L142-L154
+			if nt != keybase1.FSNotificationType_REKEYING &&
```
jzila: This can still get pretty spammy :(

songgao:
> It seems like this might be a good opportunity to do a bit of a refactor:

This would change behavior though. With the current approach, it always keep track of the newest notification. But with the proposed refactor, it'd always keep the oldest.

libkbfs/reporter_kbpki.go
songgao: Fortunately they don't happen often, and are going away in the future. So I'm inclined to just leave it here rather than coming up some deduplication logic. (I assume just sending the latest one wouldn't be the same.) WDYT?

jzila:
Is that an issue?

songgao:
I can't thing of any for now, but it seems to me that delivering the latest is better than oldest. Also, I'm curious why the refactor would be cleaner. It just moves dropping messages from receiver to the sender right?

jzila:
It's cleaner because there is no caching needed, so you don't need to track state across a loop iteration--everything that needs to happen happens in one iteration.

songgao:
This still feels pretty subjective to me, but sure, I can make change.

PR #1764, opened 2018-08-29
folder_block_ops: option for excluding non-live blocks from sums

strib:
(I won't merge it until the corresponding kbfs-server work is merged and deployed.)

We can check with the server to figure out whether a block is "live" or not, and if not, we can avoid counting its size in the sum.  This is then used by the `folderUpdatePrepper` to know exactly which blocks need to have their sizes included in the "unref" sum, so that the MD byte counts can be managed correctly.  This will matter when other PRs start including dirData refs and unrefs in `resolutionOp`s.

Issue: KBFS-3303

### DIFF
--- a/libkbfs/block_ops.go
+++ b/libkbfs/block_ops.go
@@ -93,7 +93,7 @@ func (b *BlockOpsStandard) GetEncodedSize(ctx context.Context, kmd KeyMetadata,
 		if err != nil {
 			return 0, 0, err
 		}
-		if found {
+		if found && size > 0 {
 			return size, keybase1.BlockStatus_LIVE, nil
 		}
 	}
--- a/libkbfs/block_ops.go
+++ b/libkbfs/block_ops.go
@@ -5,6 +5,7 @@
 package libkbfs
 
 import (
+	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/kbfscrypto"
 	"github.com/keybase/kbfs/tlf"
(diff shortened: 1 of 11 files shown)


libkbfs/block_ops.go
```diff
@@ -83,31 +84,22 @@ func (b *BlockOpsStandard) Get(ctx context.Context, kmd KeyMetadata,
...
-	block := NewCommonBlock()
-	errCh := b.queue.Request(ctx, defaultOnDemandRequestPriority, kmd,
-		blockPtr, block, NoCacheEntry)
-	err := <-errCh
-	if err != nil {
-		return 0, err
-	}
-	return block.GetEncodedSize(), nil
+	return b.config.BlockServer().GetEncodedSize(
+		ctx, kmd.TlfID(), blockPtr.ID, blockPtr.Context)
```
jzila: Why are we skipping the caches for this? `NoCacheEntry` is about what to do with the block that's returned, not about whether to retrieve from caches.

Additionally, I don't think this is the right approach here--`GetEncodedSize` happens quite often and needs to go through the request de-duplication with other `GetEncodedSize` and `GetBlock` requests. By using the `blockRetrievalQueue`, this de-duping comes for free.

strib: The whole point here _is_ to hit the servers, because we need to know the server's current view of the status of the block.  That's not something we can productively cache, as far as I know.

`GetEncodedSize` shouldn't actually happen quite often.  It only happens during real conflict resolution.  Only one of those happens at a time, and it will only be called for each block once during that CR.

I don't understand your point about `NoCacheEntry` or using the retrieval queue -- the queue doesn't support `GetEncodedSize` yet as far as I know, and I don't see much benefit in making it do so.

strib: (This no longer fetches the whole block, in case that wasn't clear.  Just the size and the status.)

jzila: I mean split out the functions, such that we only retrieve the status via RPC when we needed, so that when we only need the size we use the cached value.

jzila: We also `GetEncodedSize` in `syncAllLocked`: https://github.com/keybase/kbfs/blob/ed63bec510192b0739c519293d0133a6205bd281/libkbfs/folder_branch_ops.go#L4593-L4596

strib: Oh, but on the `syncAllLocked` path, [this line](https://github.com/keybase/kbfs/pull/1764/files#diff-63b0a199cafd8f676224371308119991R611) is always going to come out to `false`, and so we _will_ use sizes from the cache because of [this](https://github.com/keybase/kbfs/pull/1764/files#diff-ca9a77b130940ad6ae13c6cb2d59dcd8R289).

We only have to hit the server during actual CR, when the status of the blocks we are unreferencing is ambiguous.

> I mean split out the functions, such that we only retrieve the status via RPC when needed, so that when we only need the size we use the cached value.

That's essentially what the code is doing as explained above -- it uses the cached size whenever the status isn't actually needed. If we split it into two functions, then to avoid two RPCs we need to add a separate cache for the block sizes somewhere (since we won't be downloading the actual block, so it can't go in the `BlockCache`).  That seems like overkill.

jzila: > Oh, but on the syncAllLocked path, this line is always going to come out to false, and so we will use sizes from the cache because of this.
Cool, I'm happy with this. Thanks for the explanation!

jzila: Oh actually, if the block isn't cached in memory, you're still skipping the disk cache.

libkbfs/folder_update_prepper.go
```diff
@@ -604,10 +604,14 @@ func (fup *folderUpdatePrepper) updateResolutionUsageLockedCache(
 	// we don't know whether these are files or directories, just look
 	// them up generically.  Ignore any recoverable errors for unrefs.
 	// Note that we can't combine these with the above ref fetches
-	// since they require a different MD.
+	// since they require a different MD.  If the merged changes
+	// didn't changes any blocks (in particular, the root block), we
+	// can assume all the blocks we are unreferencing were live;
```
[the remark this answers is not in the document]
strib: > I was gonna ask what if another client has changed the blocks resulting in block status change, but it seems we'd have been or will be rebased to the latest MD so that should never be a problem. Does it sound right?

Yeah exactly.  If there was a change in status that we don't picj up, it will result in an MD conflict, and we'll have to run CR again later.

PR #1765, opened 2018-08-29
bsplitter: add an env var for setting dir block split size

strib:
To make testing easier without enabling it by default (and slowing everything down).

Issue: KBFS-3303

### DIFF
--- a/libkbfs/bsplitter_simple.go
+++ b/libkbfs/bsplitter_simple.go
@@ -6,7 +6,9 @@ package libkbfs
 
 import (
 	"fmt"
+	"os"
 	"sort"
+	"strconv"
 
 	"github.com/keybase/kbfs/kbfscodec"
 )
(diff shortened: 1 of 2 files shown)


libkbfs/bsplitter_simple.go
```diff
@@ -20,6 +22,18 @@ type BlockSplitterSimple struct {
...
+func getMaxDirEntriesPerBlock() (int, error) {
+	dirEnv := os.Getenv("KEYBASE_BSPLIT_MAX_DIR_ENTRIES")
+	if len(dirEnv) > 0 {
+		maxDirEntriesPerBlock, err := strconv.Atoi(dirEnv)
+		if err != nil {
+			return 0, err
+		}
+		return maxDirEntriesPerBlock, nil
+	}
+	return 0, nil // disabled by default
```
jzila: Can you introduce a constant that is currently 0 by default? I'm assuming eventually we plan on changing this constant to some nonzero value.

strib: It won't be a constant in the future.  It will be derived from the max block size (just like the number of i-pointers per block).  See https://github.com/keybase/kbfs/pull/1772/commits/e7b58b67c9d98db79c26ef74f5d222c602f1dc58 in #1772.  Does that sound ok?

PR #1766, opened 2018-08-29
folder_block_ops: improve locking situation with dirData

strib:
When fetching child directory blocks in parallel in `dirData` (via `blockTree`), we need to hold `blockLock` the whole time.  That means we have to lock `blockLock` when we create the `dirData`, and release it when we're done with that instance.

So if it's not already locked, have the `dirData` constructor method return an undo function that must be called when the `dirData` goes out of scope, which unlocks `blockLock`.

And fix up all other `folderBlockOps` calls that previously locked it during the course of conflict resolution.

Issue: KBFS-3303

### DIFF
--- a/libkbfs/conflict_resolver.go
+++ b/libkbfs/conflict_resolver.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/keybase/client/go/protocol/keybase1"
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/kbfscrypto"
 	"github.com/keybase/kbfs/kbfsmd"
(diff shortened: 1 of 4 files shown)


libkbfs/conflict_resolver.go
```diff
@@ -2331,8 +2337,11 @@ func (cr *ConflictResolver) doActions(ctx context.Context,
 		// Now get the directory blocks.  For unmerged directories, we
 		// can use a nil local block cache, because unmerged blocks
 		// should never be changed during the CR process (since
-		// they're just going away).
-		unmergedDir := cr.fbo.blocks.newDirDataWithLBC(
+		// they're just going away).  This call will lock `blockLock`,
+		// and the subsequent `newDirData` calls can assume it's
+		// locked already.
+		var unmergedDir *dirData
+		unmergedDir, undoFn = cr.fbo.blocks.newDirDataWithLBC(
```
songgao: The "undoFn" here doesn't actually undo what'd done by `newDirDataWithLBC`. It just releases the lock. So perhaps it should be called something else? (here and other occurrences of `undoFn` in this PR)

libkbfs/folder_block_ops.go
```diff
@@ -941,17 +948,17 @@ func (fbo *folderBlockOps) newDirDataWithLBC(lState *lockState,
 				return block, true, nil
 			}
 
-			if rtype == blockReadParallel {
-				return nil, false, errors.New(
-					"LBC must be used with parallel reads")
+			lState := lState
```
jzila: I know this is canonical, but because we potentially reassign it a few lines later, I'd prefer if the shadowed variable had a different name.

libkbfs/conflict_resolver.go
jzila: @strib normally in a similar situation, you'd create a `func()` inside the loop, with a `defer undoFn` once the function is assigned. This time you've chosen to go with the nillable method variable with several defers. Any particular reason why?

strib: @jzila: the span of the subsequent code seemed too complicated to put into an anonymous `func()`, but I can try to split it into a new, real function on `cr` instead.  It'll cause conflicts with other PRs I think, but I can manage that.

@songgao: I was just trying to encapsulate the logic a bit -- I didn't want the caller to really be aware of the locking so I picked something generic.  But I'll try to come up with a better name.

jzila: @strib you can always implement that as a later PR to avoid conflicts.

PR #1769, opened 2018-08-29
cr: collect dir block unrefs while applying actions

strib:
(Depends on #1768.)

And during the resolution process, make sure we account both for these unrefs, and dir block refs/unrefs that occurred in previous resolutionOps that haven't made it to the server yet.

Issue: KBFS-3303

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -2239,7 +2239,7 @@ type crAction interface {
 	// part of this conflict resolution.
 	do(
 		ctx context.Context, unmergedCopier, mergedCopier fileBlockDeepCopier,
-		unmergedDir, mergedDir *dirData) ([]BlockInfo, error)
+		unmergedDir, mergedDir *dirData) (unrefs []BlockInfo, err error)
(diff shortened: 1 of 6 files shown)


libkbfs/interfaces.go
```diff
@@ -2232,12 +2232,14 @@ type crAction interface {
...
+	// conflict, and potentially uses the provided
 	// `fileBlockDeepCopier`s to obtain copies of other blocks (along
-	// with new BlockPointers) when requiring a file copy.
+	// with new BlockPointers) when requiring a block copy.  It
+	// returns a set of block infos that need to be unreferenced as
+	// part of this conflict resolution.
 	do(
 		ctx context.Context, unmergedCopier, mergedCopier fileBlockDeepCopier,
-		unmergedDir, mergedDir *dirData) error
+		unmergedDir, mergedDir *dirData) ([]BlockInfo, error)
```
songgao: Change this into named return form to make it clear from signature that returned `BlockInfo`s are unrefs?

PR #1771, opened 2018-08-30
block_tree: fix shifting logic, and add a useful String method

strib:
While debugging new multi-block directory tests, I noticed some `blockTree` errors.  I added a `String` method to help with debugging, and fixed two bugs:

* If a block got shifted all the way to the left side of an existing block, and that was its final location, we were failing to update the new parent pointers to that block.  Instead, update those pointers at the top of the loop, before breaking when we figure out that it's in the right location.  This PR adds a new test that fails without the fix.
* `shiftBlocksToFillHole` was dropping some of the unref pointers when setting parent offsets.  This will be tested in a future PR that adds more multi-block dir DSL tests.

Issue: KBFS-3304

### DIFF
--- a/libkbfs/block_tree.go
+++ b/libkbfs/block_tree.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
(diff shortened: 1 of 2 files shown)


libkbfs/block_tree.go
```diff
@@ -609,6 +610,45 @@ func (bt *blockTree) setParentOffsets(
...
 
+func (bt *blockTree) String() string {
+	block, _, err := bt.getter(
+		nil, bt.kmd, bt.rootBlockPointer(), bt.file, blockRead)
+	if err != nil {
+		return "ERROR: " + err.Error()
+	}
+
+	level := []BlockWithPtrs{block}
+	res := "\n---------------\n"
```
songgao: Perhaps use a `bytes.Buffer` here to reduce allocations? It doesn't seem like this will be used much though, so not a big deal.

libkbfs/block_tree.go
```diff
@@ -644,13 +684,28 @@ func (bt *blockTree) shiftBlocksToFillHole(
...
 	// Swap left as needed.
+	loopedOnce := false
 	for {
 		var leftOff Offset
 		var newParents []parentBlockAndChildIndex
 		immedPblock := immedParent.pblock
 		if currIndex > 0 {
 			_, leftOff = immedPblock.IndirectPtr(currIndex - 1)
 		} else {
+			if loopedOnce {
```
jzila: This can be put in the loop index: https://play.golang.org/p/A8TtNwDZQFv
```go
for loopedOnce := false; ; loopedOnce = true {
	...
}
```

libkbfs/block_tree.go
strib: Yeah, it's completely unused for now, I just wanted it here for whenever I'm debugging something tricky.  I'll put in a TODO to make it better if ever used in production.

PR #1772, opened 2018-08-30
test: set dir entries by block size, and add multi-block dir DSL tests

strib:
This adds a way to compute how many dir entries will fit in a block, and sets the max dir entry size to that for all tests.  That means if a test sets a low max block size, it will also allow a small number of entries per block.  In production, this function sets the number to 1456.

It also adds a number of new tests with small block sizes, that focus on directory operations.

Notably, it also bumps the min data version that can be read by the client.  This is needed for the tests, and we might as well enable reading for production clients now too.  (Actually dir splitting is still off until we start calling `SetMaxDirEntriesByBlockSize` during the regular init code.)

Issue: KBFS-3304


strib:
@jzila, @songgao: This is ready for review now, and is the last PR we need to get in before Monday's release.

libkbfs/folder_branch_ops.go
```diff
@@ -2864,8 +2864,10 @@ func (fbo *folderBranchOps) checkNewDirSize(ctx context.Context,
 	// Just an approximation since it doesn't include the size of the
 	// directory entry itself, but that's ok -- at worst it'll be an
 	// off-by-one-entry error, and since there's a maximum name length
-	// we can't get in too much trouble.
-	if currSize+uint64(len(newName)) > fbo.config.MaxDirBytes() {
+	// we can't get in too much trouble. TODO(KBFS-3306): remove this
+	// check.
+	if dirPath.tailPointer().DirectType == DirectBlock &&
```
songgao: Could you explain why this check is needed now and will be be removed when we enable block splitting (KBFS-3306)?

libkbfs/bsplitter_simple.go
```diff
@@ -112,6 +114,52 @@ func NewBlockSplitterSimple(desiredBlockSize int64,
 	}, nil
 }
 
+// SetMaxDirEntriesByBlockSize sets the maximum number of directory
+// entries per directory block, based on the maximum block size.  If
+// the `KEYBASE_BSPLIT_MAX_DIR_ENTRIES` is set, this function does
+// nothing.
+func (b *BlockSplitterSimple) SetMaxDirEntriesByBlockSize(
+	codec kbfscodec.Codec) error {
+	dirEnv := os.Getenv("KEYBASE_BSPLIT_MAX_DIR_ENTRIES")
```
songgao: Hmm would it be better to just call `getMaxDirEntriesPerBlock()` here so we don't have to worry about env var in multiple functions. Although if you agree, we can also change this when we do KBFS-3306.

songgao: Also I guess I'm not sure how we use this env var. Do we rely on it in tests?

libkbfs/folder_branch_ops.go
strib: Yes: https://github.com/keybase/kbfs/blob/master/libkbfs/prev_revisions.go#L22-L26

(It's 5.)

libkbfs/bsplitter_simple.go
strib: > Hmm would it be better to just call getMaxDirEntriesPerBlock() here so we don't have to worry about env var in multiple functions. Although if you agree, we can also change this when we do KBFS-3306.

No, because there's a difference between an unset envvar and the envvar set to "0", and the return value of that function doesn't show the difference.

> Also I guess I'm not sure how we use this env var. Do we rely on it in tests?

No, I just found it useful for local testing, to run all the tests in dir-splitting mode without having to recompile.

libkbfs/folder_branch_ops.go
songgao: Thanks; this seem to answer my question in the first comment instead of this though?

strib: Oops sorry, weird.  This check is needed temporarily now that we estimate the block size for indirect directories using `MaxBlockSizeBytesDefault`.  For tests that set the block size to something small, like 20 bytes, they will make indirect directories, but the size of the directories will appear to be very large because of that size heuristic.  So this check causes unneeded errors for those tests.

In KBFS-3306 we will remove the directory size check completely, so this whole function will go away.

songgao: I see; thanks for explaining!

PR #1773, opened 2018-08-31
folder_update_prepper: estimate directory size of indirect blocks

strib:
Now that dirs can have multiple levels of indirection, it doesn't make sense to always use the plain (encoded, but not encrypted) size of the top block as the `DirEntry.Size`.  But it's probably too expensive to download all leaf blocks, and encode them, to find out the "correct" value during every folder update.

Instead, use a simple heuristic based on the number of indirect pointers that isn't great, but probably good enough.

Issue: KBFS-3305

### DIFF
--- a/libkbfs/block_types.go
+++ b/libkbfs/block_types.go
@@ -367,7 +367,8 @@ func (db *DirBlock) SetIndirectPtrInfo(i int, info BlockInfo) {
 	db.IPtrs[i].BlockInfo = info
 }
 
-func (db *DirBlock) totalPlainSizeEstimate(plainSize int) int {
+func (db *DirBlock) totalPlainSizeEstimate(
+	plainSize int, bsplit BlockSplitter) int {
 	if !db.IsIndirect() || len(db.IPtrs) == 0 {
 		return plainSize
 	}
@@ -380,7 +381,10 @@ func (db *DirBlock) totalPlainSizeEstimate(plainSize int) int {
 	//   pointer, assume N-1 of them are full.
 	//
 	// * If there are N child pointers, and the first one is an
-	//   indirect pointer, just give up and max out at N full blocks.
+	//   indirect pointer, just give up and max out at the maximum
+	//   number of indirect pointers in a block, assuming that at
+	//   least one indirect block is full of pointers when there are
+	//   at least 2 indirect levels in the tree.
 	//
 	// This isn't great since it overestimates in many cases
 	// (especially when removing entries), and can vastly unerestimate
@@ -390,7 +394,7 @@ func (db *DirBlock) totalPlainSizeEstimate(plainSize int) int {
 	if db.IPtrs[0].DirectType == DirectBlock {
 		return MaxBlockSizeBytesDefault * (len(db.IPtrs) - 1)
 	}
-	return MaxBlockSizeBytesDefault * len(db.IPtrs)
+	return MaxBlockSizeBytesDefault * bsplit.MaxPtrsPerBlock()
 }
 
 // FileBlock is the contents of a file
--- a/libkbfs/block_types.go
+++ b/libkbfs/block_types.go
@@ -388,7 +388,7 @@ func (db *DirBlock) totalPlainSizeEstimate(plainSize int) int {
 	// unlikely that directory byte size matters for anything in real
 	// life.  Famous last words, of course...
 	if db.IPtrs[0].DirectType == DirectBlock {
-		return MaxBlockSizeBytesDefault*len(db.IPtrs) - 1
+		return MaxBlockSizeBytesDefault * (len(db.IPtrs) - 1)
 	}
 	return MaxBlockSizeBytesDefault * len(db.IPtrs)
 }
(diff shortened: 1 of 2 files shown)


libkbfs/block_types.go
```diff
@@ -367,6 +367,32 @@ func (db *DirBlock) SetIndirectPtrInfo(i int, info BlockInfo) {
...
+	// * If there are N child pointers, and the first one is an
+	//   indirect pointer, just give up and max out at N full blocks.
+	//
+	// This isn't great since it overestimates in many cases
+	// (especially when removing entries), and can vastly unerestimate
+	// if there are more than 2 levels of indirection.  But it seems
+	// unlikely that directory byte size matters for anything in real
+	// life.  Famous last words, of course...
+	if db.IPtrs[0].DirectType == DirectBlock {
+		return MaxBlockSizeBytesDefault*len(db.IPtrs) - 1
```
songgao: Based on the description in comment it seems this should've been

```
		return MaxBlockSizeBytesDefault * (len(db.IPtrs) - 1)
```

strib: D'oh, good catch, thanks!

strib:
> Could you explain the heuristic? It seems it makes little difference between if the first pointer is direct vs indirect.

Hopefully with the bug fix you pointed out, and the existing explanation in the comment, the heuristic should be clear now.

> Also, would it be better to go through all pointers in the root block and estimate based one each of them being direct vs indirect (still without fetching) instead of only looking at the first one?

I'm pretty sure that the tree should be balanced, because we add new levels at the top of the tree (and then down along the right side), rather than the bottom.  So all the pointers on the first level should have the same type.

Although, now that I write this out, the tree-making algorithm I describe is actually much worse for this heuristic, since when a new level is added, the number of indirect pointers coming out of the root will go from N to 2, and the reported size of the directory will shrink.  Hrmm.

I guess if the first pointer is indirect, I should not max it out at `N*len(db.IPtrs)`, but at `N*BlockSplitter.MaxPtrsPerBlock()`.  Basically, if there are two levels of indirect pointers, we can assume at lease one of those levels was completely full.  I'll change it and update the comment.

songgao:
Thanks for the explanation. Although I'm still a bit confused. Even with the fix, it still seems the fact the first point is direct vs indirect makes little difference; they both give a size close to when we have first layer full. If we are making estimates based on if the first pointer is indirect, shouldn't we do e.g. `MaxBlockSizeBytesDefault * len(db.IPtrs) * BlockSplitter.MaxPtrsPerBlock()` when first pointer is indirect (with the assumption that second layer is full)?

strib:
> Even with the fix, it still seems the fact the first point is direct vs indirect makes little difference; they both give a size close to when we have first layer full.

I don't think that's true.  If the first pointer is direct, and there are only two pointers total, then the total size is just `MaxBlockSizeBytesDefault`.  If the first pointer is indirect, then it's `MaxBlockSizeBytesDefault*BlockSplitter.MaxPtrsPerBlock()`, which is a lot more.

> If we are making estimates based on if the first pointer is indirect, shouldn't we do e.g. MaxBlockSizeBytesDefault * len(db.IPtrs) * BlockSplitter.MaxPtrsPerBlock() when first pointer is indirect (with the assumption that second layer is full)?

We can't really say whether there are multiple full levels below this, we only know for sure that there is one full level.  For example, when the third level is added to a tree, the root block will have two indirect pointers. The first one points to an indirect block with a full set of pointers, but the second one points to a block with only one pointer.

songgao:
I see; that makes sense. Thanks for the explanation!

libkbfs/block_types.go
```diff
@@ -367,6 +367,36 @@ func (db *DirBlock) SetIndirectPtrInfo(i int, info BlockInfo) {
...
+	//
+	// * If there are N child pointers, and the first one is an
+	//   indirect pointer, just give up and max out at the maximum
+	//   number of indirect pointers in a block, assuming that at
+	//   least one indirect block is full of pointers when there are
+	//   at least 2 indirect levels in the tree.
+	//
+	// This isn't great since it overestimates in many cases
+	// (especially when removing entries), and can vastly unerestimate
+	// if there are more than 2 levels of indirection.  But it seems
```
jzila: Might be overcomplicated, but we could cache the depth of the tree at each level.

strib: I did think about putting the depth into the pointer itself, but that means that when we add a new level at the top of the tree, we'll have to fetch and update all the pointers for the entire tree, which seems like it would be too complicated and costly for something (a dir with more than two levels) that will probably be used almost never, and even then almost no one will ever care about the reported size of the directory in bytes.

jzila: If you track the descending depth then you wouldn't need to do that.

strib: Oh, duh, of course.  Not sure what I was thinking.  I'll file a ticket, though I still don't think it's crucial enough to worry about right now.

jzila: Agreed, thanks!

PR #1779, opened 2018-09-04
vendor go-keychain

mmaxim:
cc @malgorithms


mmaxim:
Possibly, I also pushed it to keybase/keybase. It also affects 1.10.4 which is how I noticed.

PR #1782, opened 2018-09-05
vendor: Update go-git and go-billy

taruti:
This is not elegant. The subdirectories have been manually added as separate instances before. This continues to do that.

Alternatively govendor supports the "-tree" flag, should we just remove the existing per subdirectory vendors and use -tree to get the whole thing?


strib:
> Alternatively govendor supports the "-tree" flag, should we just remove the existing per subdirectory vendors and use -tree to get the whole thing?

There's a lot of subdirectories under `go-git.v4` that we _don't_ want to vendor, such as all the example directories, so I think it makes sense to continue this way of doing things.

strib:
Tests pass and basic functionality seems to work on the CLI, so I'm merging.  Thanks @taruti!

PR #1789, opened 2018-09-11
move NormalizeSocialAssertion and AssertionParseAndOnly to RPC call

joshblum:
depends on changes in https://github.com/keybase/client/pull/13624. Once the client side is merged,  I will update the vendor hashes to match client master. In the client PR we are starting to add support for a dynamic set of proof services for generic social proofs. This changes how a bunch of previously static calls in the `externals.` package will work so instead of calling these methods directly we wrap them in RPC calls the service. 

- [x] vendor in master hashes


libkbfs/key_manager_test.go
```diff
@@ -49,9 +51,21 @@ func keyManagerInit(t *testing.T, ver kbfsmd.MetadataVer) (mockCtrl *gomock.Cont
 		gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes().
 		Return(ImplicitTeamInfo{}, errors.New("No such team"))
 
+	mockNormalizeSocialAssertion(config)
 	return mockCtrl, config, ctx
 }
 
+func mockNormalizeSocialAssertion(config *ConfigMock) {
```
strib: Please move this func definition above its first use.

libkbfs/tlf_handle_resolve.go
```diff
@@ -688,8 +687,8 @@ func (ra resolvableAssertion) resolve(ctx context.Context) (
 			id:   id,
 		}, keybase1.SocialAssertion{}, tlfID, nil
 	case NoSuchUserError:
-		socialAssertion, ok := externals.NormalizeSocialAssertion(ra.assertion)
-		if !ok {
+		socialAssertion, serr := ra.resolver.NormalizeSocialAssertion(ctx, ra.assertion)
```
strib: I think we should just call the error `err` and return it below.

joshblum: i left it as the original implementation since there are some error handlers that depend on this error type and i didn't want to subtly change the functionality. should we just log `serr` and return `err`?

strib: Hmm, ok, that makes sense.  Unfortunately we don't have easy access to a proper logger here.  Just add a comment about why we're dropping it?

joshblum: will do!

joshblum:
ran go tests locally, merging w/o CI so circle can not fail for others

PR #1790, opened 2018-09-11
use different port if the pinned one is taken

songgao:
KBFS side of https://github.com/keybase/client/pull/13666

### DIFF
--- a/libhttpserver/server.go
+++ b/libhttpserver/server.go
@@ -162,11 +162,22 @@ const requestPathRoot = "/files/"
 func (s *Server) restart() (err error) {
 	s.serverLock.Lock()
 	defer s.serverLock.Unlock()
-	s.server.Stop()
-	// Have to start this first to populate the ServeMux object.
-	if err = s.server.Start(); err != nil {
+	if s.server != nil {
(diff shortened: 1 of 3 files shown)


libhttpserver/server.go
```diff
@@ -162,11 +162,22 @@ const requestPathRoot = "/files/"
...
+	if s.server != nil {
+		s.server.Stop()
+		err = s.server.Start()
+	}
+	if s.server == nil ||
+		// If pinned port is in use, just pick a new one like we never had a
+		// server before.
+		err == kbhttp.ErrPinnedPortInUse {
+		s.server = kbhttp.NewSrv(s.logger,
+			kbhttp.NewPortRangeListenerSource(portStart, portEnd))
```
jzila: Do we need to stop the previous server first?

songgao: Got you; yeah in this case the error is returned from here: https://github.com/keybase/client/blob/df15a7336dec72486ad6a45a2ad9ad162ef92cec/go/kbhttp/srv.go#L122 where we don't even have a `http.Server` yet. So nothing's leaky.

PR #1792, opened 2018-09-11
libkbfs: start splitting directories and remove DirTooBig checks

strib:
[I won't merge this until the mobile releases go out.]

Also remove the unused FileTooBig error.

Issue: KBFS-3306


strib:
Good call, I don't think so.  I'll get rid of that comment.

PR #1794, opened 2018-09-12
Default to application/octet-stream for unknown types

taruti:


libhttpserver/content_type_utils.go
```diff
@@ -37,11 +37,12 @@ func (w *contentTypeOverridingResponseWriter) overrideMimeType(
 	case strings.HasPrefix(ty, "audio/") ||
 		strings.HasPrefix(ty, "image/") ||
 		strings.HasPrefix(ty, "video/") ||
-		ty == "application/pdf":
+		ty == "application/pdf" ||
+		ty == "text/plain":
```
jzila: Shouldn't this be `strings.HasPrefix(ty, "text/")`?

And IIRC we need it to always return `text/plain` when it's any type of text.

taruti: Sure we can do that, Note however that we already do html etc interception before. Will have a revised patch up in 2min

libhttpserver/content_type_utils.go
```diff
@@ -28,20 +28,22 @@ func (w *contentTypeOverridingResponseWriter) overrideMimeType(
...
 	case strings.HasPrefix(ty, "audio/") ||
 		strings.HasPrefix(ty, "image/") ||
 		strings.HasPrefix(ty, "video/") ||
 		ty == "application/pdf":
 		return ty
-	// Otherwise text/plain.
+	// Otherwise default to binary data.
 	default:
-		return "text/plain"
+		return "application/octet-stream"
```
songgao: Could you explain why we default to binary now? It seems it's safer to default to text to avoid stuff getting executed or parsed by browsers accidentally. If we need to add support for particular types, we can add them explicitly right?

jzila:
@songgao: https://keybase.atlassian.net/browse/KBFS-3396

songgao:
@jzila @taruti thanks for the background. I'm still a bit concerned this would cause unexpected behavior in some situations. What do you all think if we return something like `text/keybase-unknown` as default? Or `application/keybase-unknown`. Then in frontend we can explicitly route user to `default-view.js` on that content type.

PR #1802, opened 2018-09-13
libgit: don't hold server locks for GC for too long

strib:
Instead, refactor the autogit strategy and only create a "working" file while holding the lock.  If another worker has already created the file and is within an hour time limit according to the timestamp on the file, just return `nil` to avoid showing an error to the user. If the other worker fails, they'll get another message about doing GC on a later repo operation.

Issue: KBFS-3407

### DIFF
--- a/libgit/util.go
+++ b/libgit/util.go
@@ -1,0 +1,75 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libgit
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/keybase/client/go/logger"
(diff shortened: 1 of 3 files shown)


libgit/util.go
```diff
@@ -0,0 +1,75 @@
...
+				"modCommonTime=%s, currCommonTime=%s, workTimeLimit=%s",
+				modCommonTime, currCommonTime, workTimeLimit)
+			// The other GC is still running within the time
+			// limit.
+			return false, nil
+		}
+		log.CDebugf(ctx, "Other GC expired; "+
+			"modCommonTime=%s, currCommonTime=%s, workTimeLimit=%s",
+			modCommonTime, currCommonTime, workTimeLimit)
+	}
```
songgao: I know this is from the other file, but it'd be nice if this was in a switch/case block.

PR #1803, opened 2018-09-14
Add missing libkb file

joshblum:
this file was missing in the vendor package

### DIFF
--- a/vendor/github.com/keybase/client/go/libkb/msgpack.go
+++ b/vendor/github.com/keybase/client/go/libkb/msgpack.go
@@ -1,0 +1,36 @@
+// Copyright 2018 Keybase, Inc. All rights reserved. Use of
+// this source code is governed by the included BSD license.
+
+package libkb
+
+import (
+	"fmt"
+
+	"github.com/keybase/go-codec/codec"
+)
+
+func MsgpackDecode(dst interface{}, src []byte) (err error) {
+	ch := codecHandle()
+	return codec.NewDecoderBytes(src, ch).Decode(dst)
+}
+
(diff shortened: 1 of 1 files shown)


strib:
Why isn't vendor.json updated with a new checksum?

joshblum:
these files are git ignored, the correct checksum was put into https://github.com/keybase/kbfs/pull/1789 but this file was mistakenly missing from the merge

strib:
I'm confused, `msgpack.go` is being gitignored?

joshblum:
```
$ git add vendor/github.com/keybase/client/go/libkb
The following paths are ignored by one of your .gitignore files:
vendor/github.com/keybase/client/go/libkb
Use -f if you really want to add them.
```

although i can't find any `.gitignore`s in this repo or my global config that are causing this, i'm not sure the root cause

joshblum:
but we should get this in since it breaks CI for all

strib:
I'd guess the root cause is that this is a brand new file that got added to master while your client PR was in review, and when you squashed yours in and re-merged, you didn't `git add --all vendor` in kbfs afterwards.

No idea about the gitignore thing.

Let me try it locally.

joshblum:
yeah i didn't realize that the gitignore was in play and that i should be force adding those vendor dirs

strib:
Hmm you're right.  I don't get the gitignore error, but `govendor` doesn't update the checksum when I re-vendor it, so something's weird with your local gitignore config.

joshblum:
sorry about that! i'll try to figure out what is setting it locally so it doesn't happen again

PR #1804, opened 2018-09-14
vendor: Update go-git for object walker git gc blob slow path

taruti:


taruti:
Rebased from master for CI failure.

PR #1805, opened 2018-09-14
cr_actions: add test and fix on setmtime + merged rename conflict

strib:
When the merged user sets the mtime and moves a file across directories, and the unmerged user just sets the mtime, we weren't issuing the correct local op notifications to the unmerged user after conflict resolution.  Previously, the notifications made it look like the unmerged file was moved somewhere else, and the file would appear to be missing on a subsequent lookup.

To fix, detect the situation explicitly and update the notifications to make it look like a new file (the renamed merged file) is being created instead.

Note that this doesn't solve KBFS-3408, it's just a test I tried adding while debugging that issue.  The test fails to repro that issue though, but it's worth fixing in its own right.

Issue: KBFS-3408

### DIFF
--- a/libkbfs/cr_actions.go
+++ b/libkbfs/cr_actions.go
@@ -588,6 +588,17 @@ func (rua *renameUnmergedAction) do(
 	return unrefs, nil
 }
 
+func removeRmOpFromChain(
+	original BlockPointer, chains *crChains, oldName string) {
+	chain := chains.byOriginal[original]
+	for i, op := range chain.ops {
+		if ro, ok := op.(*rmOp); !ok || ro.OldName != oldName {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


libkbfs/cr_actions.go
```diff
@@ -670,32 +681,62 @@ func (rua *renameUnmergedAction) updateOps(
...
+			so.AddUpdate(
+				unmergedEntry.BlockPointer, newMergedEntry.BlockPointer)
+			err = prependOpsToChain(
+				unmergedEntry.BlockPointer, mergedChains, so)
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		// The unmerged copy is changing it's name, so make a local
```
songgao: `s/it's/its/`?

PR #1817, opened 2018-09-25
libgit: add a git browser that doesn't require a checkout

strib:
`Browser` uses the gogit `Tree` instance to navigate directories and read files directly from a dotgit directory, without requiring an explicit checkout.

This will be used in future commits as the basis for the new `.kbfs_autogit` directory.

The test provided is pretty simple but covers the basics.  I'm happy to add more stuff to the test if there are specific things you think would be good to add.

Issue: KBFS-3421

### DIFF
--- a/libgit/browser.go
+++ b/libgit/browser.go
@@ -60,7 +60,7 @@ func NewBrowser(
 	}
 
 	if ref.Type() != plumbing.HashReference {
-		return nil, errors.Errorf("Can't browse reference type %s", ref.Type())
+		return nil, errors.Errorf("can't browse reference type %s", ref.Type())
 	}
 
 	c, err := repo.CommitObject(ref.Hash())
@@ -109,7 +109,7 @@ func (b *Browser) followSymlink(filename string) (string, error) {
 		return "", err
 	}
 	if path.IsAbs(link) {
-		return "", errors.Errorf("Can't follow absolute link: %s", link)
+		return "", errors.Errorf("can't follow absolute link: %s", link)
 	}
 
 	parts := strings.Split(filename, "/")
@@ -120,7 +120,7 @@ func (b *Browser) followSymlink(filename string) (string, error) {
 	newPath := path.Clean(path.Join(parentPath, link))
 	if strings.HasPrefix(newPath, "..") {
 		return "", errors.Errorf(
-			"Cannot follow symlink out of chroot: %s", newPath)
+			"cannot follow symlink out of chroot: %s", newPath)
 	}
 	return newPath, nil
 }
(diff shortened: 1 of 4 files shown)


libgit/browser.go
```diff
@@ -0,0 +1,301 @@
...
+			return nil, err
+		}
+		for _, e := range dirTree.Entries {
+			size += int64(len(e.Name))
+		}
+	}
+
+	// Git doesn't keep track of the mtime anywhere, so just use the
+	// time this browser was created.
+	return &browserFileInfo{entry, size, b.mtime}, nil
```
songgao: Would it be better to use the commit's timestamp instead?

libgit/browser_file.go
```diff
@@ -0,0 +1,89 @@
...
+	// Reader object and just scan starting from the beginning.
+	fullData := make([]byte, int64(len(p))+off)
+	r, err := bf.f.Reader()
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		_ = r.Close()
+	}()
+	n, err = r.Read(fullData)
```
songgao: I'm a bit worried about reading large objects here. Instead of read everything into ram, perhaps we can have an upper count for the buffer here, and call `r.Read` multiple times until we reach the specified offset? Although on the other hand, files in git are usually small enough to fit in RAM so maybe it's fine this way for now.

Same for `Seek()` below, but I guess for that we should just aim for the real implementation.

libgit/browser.go
strib: Good idea, will do.

libgit/browser_file.go
strib: Good idea, done.

strib:
Thanks @songgao, please review the new commits, especially the `ReadAt` one since it contains non-trivial new logic.

libgit/browser_file.go
```diff
@@ -43,27 +49,35 @@ func (bf *browserFile) Read(p []byte) (n int, err error) {
...
+	for dataToSkip > 0 {
+		bufSize := dataToSkip
+		if bufSize > bf.maxBufSize {
+			bufSize = bf.maxBufSize
 		}
-		return 0, err
+
+		buf := make([]byte, dataToSkip)
+		// Throwaway data.
+		n, err := r.Read(buf)
```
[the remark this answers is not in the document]
strib: I did this intentionally to make sure the `n` wouldn't accidentally leak to the return param somehow.

libgit/browser_file.go
```diff
@@ -43,27 +49,35 @@ func (bf *browserFile) Read(p []byte) (n int, err error) {
...
+	// Skip past the data we don't care about, one chunk at a time.
+	dataToSkip := off
+	for dataToSkip > 0 {
+		bufSize := dataToSkip
+		if bufSize > bf.maxBufSize {
+			bufSize = bf.maxBufSize
 		}
-		return 0, err
+
+		buf := make([]byte, dataToSkip)
```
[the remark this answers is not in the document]
strib: Good point, done in the latest commit.

PR #1819, opened 2018-09-26
simplefs: add FILTER_SYSTEM_HIDDEN_2 to only filter system hidden files

songgao:
As discussed in kbchat, we should only hide system hidden files, thus the change.

This needs to go in with https://github.com/keybase/client/pull/13927 . After that's in, I'll update all commit hashes to master.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -476,12 +476,29 @@ func (k *SimpleFS) updateWriteProgress(
 	}
 }
 
+var filesToIgnore = map[string]bool{
+	".Trashes":   true,
+	".fseventsd": true,
+	".DS_Store":  true,
+}
+var prefixesToIgnore = []string{"._"}
+
 func isFiltered(filter keybase1.ListFilter, name string) bool {
 	switch filter {
 	case keybase1.ListFilter_NO_FILTER:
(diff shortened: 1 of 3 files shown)


songgao:
Will wait for https://github.com/keybase/kbfs/pull/1811

simplefs/simplefs.go
```diff
@@ -476,12 +476,29 @@ func (k *SimpleFS) updateWriteProgress(
 	}
 }
 
+var filesToIgnore = map[string]bool{
+	".Trashes":   true,
+	".fseventsd": true,
+	".DS_Store":  true,
+}
+var prefixesToIgnore = []string{"._"}
```
jzila: On the one hand, for performance ideally this would be a regex instead of a loop. On the other, regexes are less easily debugged, and this set of prefixes will likely always remain small.

songgao: Hmm I guess if it grows large we could make it a trie? It feels to me that regex is gonna be slower to do a simple prefix match. Not sure!

jzila: Naw, regexes compile down to a state machine that would be functionally equivalent to a trie for cases where a trie would help.

PR #1822, opened 2018-09-27
libgit: autogit browses git repos directly, without a checkout

strib:
This also changes the directory structure, so there's no more need for the `tlf_type/tlf_name` prefix in the autogit path.  You can browse repos under TLFs for which you don't have write access.

However, this does NOT yet support invalidating kernel caches when the commits happen to the underlying repo.  So we shouldn't yet update kbpbot after this is merged, until subsequent PRs are also merged.

Due to some nasty potential deadlocks, I had to rework the way that NodeCache wraps its children, to make sure it happens without taking a lock.

Issue: KBFS-3425

### DIFF
--- a/libgit/autogit_node_wrappers.go
+++ b/libgit/autogit_node_wrappers.go
@@ -8,11 +8,12 @@ import (
 	"context"
 	"path"
 	"sync"
-	"time"
 
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libkbfs"
-	"github.com/keybase/kbfs/tlf"
-	"github.com/pkg/errors"
+	billy "gopkg.in/src-d/go-billy.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
 )
 
 // This file contains libkbfs.Node wrappers for implementing the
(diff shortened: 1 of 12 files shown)


libkbfs/node_cache.go
```diff
@@ -146,8 +161,16 @@ func (ncs *nodeCacheStandard) Get(ref BlockRef) Node {
 		panic(InvalidBlockRefError{ref})
 	}
 
+	var rootWrappers []func(Node) Node
+	defer func() {
+		if n != nil {
+			n = ncs.wrapNodeStandard(n, rootWrappers)
+		}
+	}()
```
songgao: `rootWrappers` is always assigned with `ncs.rootWrappers`, so it doesn't seem necessary. Is there a reason not to just use `ncs.rootWrappers` in the deferred function?

If the concern is that `ncs.rootWrappers` may change, I think the deferred function should still be called under the lock because even though it's assigned to a different variable, it still shares the same underlying buffer with all other references. But it doesn't seem like `ncs.rootWrappers` can change.

(same below)

libgit/autogit_node_wrappers.go
```diff
@@ -430,32 +176,68 @@ func (arn autogitRootNode) WrapChild(child libkbfs.Node) libkbfs.Node {
...
+	rn.lock.Lock()
+	defer rn.lock.Unlock()
+	if rn.fs == nil {
+		// Make the FS once, in a place where we know the NodeCache
+		// won't be locked (to avoid deadlock).
+
+		// Wrap this child so that it will show all the
+		// repos. TODO(KBFS-3429): fill in context.
+		ctx := context.TODO()
+		h, err := rn.am.config.KBFSOps().GetTLFHandle(ctx, rn)
```
songgao: Perhaps this context should be different from what's passed into `NewFS` below, so it can have a timeout.

libkbfs/node_cache.go
strib: There's no guarantee `ncs.rootWrappers` won't change, so I was trying to save it while under lock.  But you're right, those places should use a copy.  Will fix.

libgit/autogit_node_wrappers.go
strib: You're right, I should be using the passed-in `ctx` here. Will fix.

libkbfs/node_cache.go
songgao: Ah I missed the `AddRootWrapper ` method. As long as the deferred function is called under the lock (i.e. defer it after the deferred unlock) and it'd be fine right?

strib: Yes, but the whole point here is to apply the wrappers while *not* under lock, because that leads to deadlock.

songgao: Sounds good. Sorry for going back and forth and causing the confusion!

PR #1823, opened 2018-09-27
libgit: support for browsing branches

strib:
With the root of an autogit repo, you can now navigate to the branch of a version by using `.kbfs_autogit_branch_<name>`.  Branch names can contain slashes, so you can stack these.  For example, to navigate to a branch a/b/c in repo "test", use this path:

```
.kbfs_autogit/test/.kbfs_autogit_branch_a/.kbfs_autogit_branch_b/.kbfs_autogit_branch_c/
```

It's a bit unwieldy, but it will be fine for GUI-based browsing, and I don't see a good alternative for CLI-based browsing, besides shortening the prefix somehow.

If the branch names doesn't start with "refs/", then "refs/heads/" is prepended to it.  But e.g. tags can be shown with a full path like "refs/tags/v2.6.0".

Issue: KBFS-3426

### DIFF
--- a/libgit/autogit_node_wrappers.go
+++ b/libgit/autogit_node_wrappers.go
@@ -53,6 +53,9 @@ const (
 	// AutogitBranchPrefix is a prefix of a subdirectory name
 	// containing one element of a git reference name.
 	AutogitBranchPrefix = ".kbfs_autogit_branch_"
+	// branchSlash can substitute for slashes in branch names,
+	// following `AutogitBranchPrefix`.
+	branchSlash = ":"
 )
 
 type repoFileNode struct {
@@ -144,8 +147,9 @@ func (rdn repoDirNode) WrapChild(child libkbfs.Node) libkbfs.Node {
 			am:     rdn.am,
 			repoFS: rdn.repoFS,
 			subdir: "",
-			branch: plumbing.ReferenceName(
-				path.Join(string(rdn.branch), branchName)),
+			branch: plumbing.ReferenceName(path.Join(
+				string(rdn.branch),
+				strings.Replace(branchName, branchSlash, "/", -1))),
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


strib:
> I wonder if something like this would be better?

What's the advantage of your suggestion? The "simple" case is what I have now, and the slashy case seems to just add complexity.  Maybe you're intending to help with the fact that it's hard to tell whether one of the path prefixes is a prefix of a real branch name, but I don't think this helps, because the FS still has to return something if you, e.g., `ls .kbfs_autogit/test/.kbfs_autogit_branch_start/branch/name/`.

songgao:
I was trying to make it so that the branch name remains complete in the path. It seems `.kbfs_autogit/test/.kbfs_autogit_branch_start/a/b/c/.kbfs_autogit_branch_end/` is more readable and easy to type than `.kbfs_autogit/test/.kbfs_autogit_branch_a/.kbfs_autogit_branch_b/.kbfs_autogit_branch_c/` when going from CLI. (Agree that it makes no difference with GUI.

strib:
Oh right, I somehow forgot about the readability issue since I opened this.  Fair point.

Hmm.  I don't actually know which is better for the user.  Yours is definitely more readable, but it feels more brittle for some reason too.  If the user forgets to add the end tag, things will get pretty confusing.

I guess another option is to substitute some other character for `/`, if there is one that's legal in the FS but not in git branch names.

Sigh.  Maybe I'll sleep on it.

songgao:
> I guess another option is to substitute some other character for /, if there is one that's legal in the FS but not in git branch names.

Yeah that sounds good to me as well, and is likely more usable than my cumbersome macro suggestion. It seems we've got quite a few options here: https://stackoverflow.com/a/3651867

jzila:
> I guess another option is to substitute some other character for /, if there is one that's legal in the FS but not in git branch names.
+1, this seems like the right answer to me.

strib:
Cool thanks all.  I just pushed a new commit that (arbitrarily) uses ":" instead of "/" (while still maintaining the old option of use multiple tags in succession.  Happy to adjust it to other suggestions.

PR #1824, opened 2018-09-28
node_cache: add and implement AllNodeChildren function

strib:
This function returns all the children of the current node.  It will be useful to figure out which autogit nodes need to be invalidated, after a repo is updated.

This implementation iterates through all nodes, and back through all parents, to figure out which nodes to return.  A better implementation might track the a node's children directly, but that will be a bit tricky to get right, so let'd try this for now and see if it's good enough.

Issue: KBFS-3427

### DIFF
--- a/libkbfs/node_cache.go
+++ b/libkbfs/node_cache.go
@@ -402,6 +402,50 @@ func (ncs *nodeCacheStandard) AllNodes() (nodes []Node) {
 	return nodes
 }
 
+// AllNodeChildren implements the NodeCache interface for nodeCacheStandard.
+func (ncs *nodeCacheStandard) AllNodeChildren(n Node) (nodes []Node) {
+	var rootWrappers []func(Node) Node
+	defer func() {
+		for i, n := range nodes {
+			nodes[i] = ncs.wrapNodeStandard(n, rootWrappers)
+		}
+	}()
+
+	ncs.lock.Lock()
+	defer ncs.lock.Unlock()
+	rootWrappers = ncs.rootWrappers
+	nodes = make([]Node, 0, len(ncs.nodes))
+	entryIDs := make(map[NodeID]bool)
+	for _, entry := range ncs.nodes {
+		var pathIDs []NodeID
+		parent := entry.core.parent
+		for parent != nil {
+			// If the node's parent is what we're looking for (or on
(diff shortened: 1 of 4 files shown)


libkbfs/node_cache.go
```diff
@@ -402,6 +402,50 @@ func (ncs *nodeCacheStandard) AllNodes() (nodes []Node) {
...
+func (ncs *nodeCacheStandard) AllNodeChildren(n Node) (nodes []Node) {
+	var rootWrappers []func(Node) Node
+	defer func() {
+		for i, n := range nodes {
+			nodes[i] = ncs.wrapNodeStandard(n, rootWrappers)
+		}
+	}()
+
+	ncs.lock.Lock()
+	defer ncs.lock.Unlock()
```
songgao: Would `RLock` be sufficient here?

libkbfs/node_cache.go
```diff
@@ -402,6 +402,50 @@ func (ncs *nodeCacheStandard) AllNodes() (nodes []Node) {
...
+	rootWrappers = ncs.rootWrappers
+	nodes = make([]Node, 0, len(ncs.nodes))
+	entryIDs := make(map[NodeID]bool)
+	for _, entry := range ncs.nodes {
```
songgao: Is `ncs` per repo? Otherwise it seems this'd be slow for kbfs processes that have been heavily used.

libkbfs/node_cache.go
strib: Yes!  Good catch, thanks.

libkbfs/node_cache.go
strib: `ncs` is per-TLF, not per-repo.  But, a node's lifetime in the cache is controlled by the kernel; once the kernel forgets a node, it's removed from the cache.  So the size of each cache is limited by the number of nodes the OS is keeping about the TLF, which is presumably limited to a somewhat reasonable number.

But yes, I agree it would perform better if we keep more state connected a parent to its children.  But that's going to be a little complex, so I wanted to wait to see if performance was a problem first.  The next PR adds some debug log messages that should help figure that out.

songgao: Sounds good. Thanks for the background!

PR #1825, opened 2018-09-28
autogit: invalidate all AG nodes associated with an updated repo

strib:
When the root of a .kbfs_git repo is updated, invalidate the associated autogit root node, and any cached subnodes below it.  Achieve this by:

1. In `AutogitManager`, tracking the association between a `.kbfs_git` repo root node, and the associated `.kbfs_autogit` repo root node.
2. When the former is updated, ask `folderBranchOps` to invalidate all nodes associated with the latter.
3. `folderBranchOps` gets all of the autogit nodes for that repo from the node cache.  (This isn't optimally implemented, but it's probably good enough for now.)
4. `folderBranchOps` sends invalidation notifications for all of those nodes, which get delivered to the kernel in the usual way.

I've tested the invalidation by hand, and I plan to write a kbfsdocker autogit test in the future, which will exercise it more.

Issue: KBFS-3428

### DIFF
--- a/libkbfs/interfaces.go
+++ b/libkbfs/interfaces.go
@@ -484,6 +484,13 @@ type KBFSOps interface {
 	// newest version.  It works asynchronously, so no error is
 	// returned.
 	ForceFastForward(ctx context.Context)
+	// InvalidateNodeAndChildren sends invalidation messages for the
+	// given node and all of its children that are currently in the
+	// NodeCache.  It's useful if the caller has outside knowledge of
+	// data changes to thae node or its children that didn't come
+	// through the usual MD update channels (e.g., autogit nodes need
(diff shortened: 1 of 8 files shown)


libkbfs/interfaces.go
```diff
@@ -484,6 +484,13 @@ type KBFSOps interface {
 	// newest version.  It works asynchronously, so no error is
 	// returned.
 	ForceFastForward(ctx context.Context)
+	// InvalidateNodeAndChildren sends invalidation messages for the
+	// given node and all of its children that are currently in the
+	// NodeCache.  It's useful if the caller has outside knowledge of
+	// data changes to thae node or its children that didn't come
```
songgao: s/thae/that

PR #1831, opened 2018-10-01
autogit: remove old autogit checkouts when a repo is accessed

strib:
Whenever a repo is accessed within a particular TLF, delete all old checkouts in that TLF.

Also, remove self-checkouts on first autogit access to ANY repo, because some users (like kbpbot) only checkout repos from other TLFs, and use their own private folder as the destination.

To enable this, we allow for "unwrapped" FSes, to avoid getting entangled in recursive node wrapping.

Issue: KBFS-3431

### DIFF
--- a/libgit/autogit_manager.go
+++ b/libgit/autogit_manager.go
@@ -6,10 +6,14 @@ package libgit
 
 import (
 	"context"
+	"os"
 	"sync"
 
 	"github.com/keybase/client/go/logger"
+	"github.com/keybase/client/go/protocol/keybase1"
+	"github.com/keybase/kbfs/libfs"
 	"github.com/keybase/kbfs/libkbfs"
+	"github.com/pkg/errors"
 )
 
 type getNewConfigFn func(context.Context) (
@@ -40,6 +44,8 @@ type AutogitManager struct {
 	registeredFBs          map[libkbfs.FolderBranch]bool
 	repoNodesForWatchedIDs map[libkbfs.NodeID]*repoDirNode
 	watchedNodes           []libkbfs.Node // preventing GC on the watched nodes
+	deleteCancels          []context.CancelFunc
+	shutdown               bool
 }
 
 // NewAutogitManager constructs a new AutogitManager instance.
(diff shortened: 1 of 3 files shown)


libgit/autogit_manager.go
```diff
@@ -51,12 +60,108 @@ func NewAutogitManager(config libkbfs.Config) *AutogitManager {
...
+		ctx, am.config.KBPKI(), am.config.MDOps(),
+		string(session.Name), tlf.Private)
+	if err != nil {
+		am.log.CDebugf(ctx,
+			"Unable to get private handle; ignoring self-autogit delete: +%v",
+			err)
+		return
+	}
+
+	am.removeOldCheckoutsForHandle(ctx, h, libkbfs.MasterBranch)
```
songgao: Do we need to call `takeConfigLock` here like in https://github.com/keybase/kbfs/blob/5cac0cd933f685215f2eca6d9b7d0e0961e2137f/libgit/repo.go#L814 ? Same above.

libgit/autogit_manager.go
```diff
@@ -51,12 +60,108 @@ func NewAutogitManager(config libkbfs.Config) *AutogitManager {
...
+	if err != nil {
+		am.log.CDebugf(ctx, "Error making unwrapped FS for TLF %s: %+v",
+			h.GetCanonicalPath(), err)
+		return
+	}
+
+	fi, err := fs.Stat(AutogitRoot)
+	if os.IsNotExist(errors.Cause(err)) {
+		// No autogit repos to remove.
+		return
```
songgao: It'd be nice to have a log here to make it easy to tell if the cleanup has been finished.

libgit/autogit_manager.go
strib: That lock is for the actual `.kbfs_git` repos, and not the autogit checkouts.

You could argue that we need to take some sort of autogit lock, but if there are two new clients both trying to remove all files at the same time, that seems pretty safe to me.  And if there are an old client and a new client racing to access autogit, that also seems fine since we've agreed there are basically zero users of autogit, besides kbpbot.

Does that make sense?

songgao: That totally makes sense. Thanks!

libgit/autogit_manager.go
strib: I decided not to have a log here, because there was no previous log saying that cleanup had even started, and I didn't want to write to have these log messages that appear on the first autogit access every time.  I could be convinced otherwise though, if you think it's important.

songgao: I'm fine without a log; don't worry about it!

strib:
Thanks for all the reviews!  Is this one ready for a :heavy_check_mark: or are you still looking at the code?

PR #1839, opened 2018-10-04
always set the number on pagination param when fetching chat threads CORE-9096

mmaxim:
Otherwise it would send 0 over on the subsequent pages and get 0 results.

Also, do we really need the entire thread? Can we check in this loop to see if we have enough data to reasonably render the widget? It seems like we can abort out of this loop and save a bunch of traffic.


mmaxim:
Oh wait nvm, I see how it works, it does break out early.

PR #1840, opened 2018-10-04
Configurable disk cache limit

jakob223:
Modifiy the disk cache limiter to accept a variable size disk cache,
passed as a command line option at startup.

Also, a few documentation fixes and an error handling fix.


jakob223:
Moved to https://github.com/keybase/kbfs/pull/1843

PR #1841, opened 2018-10-04
tlf_handle_resolve: never skip implicit team resolution

strib:
For non-single-team folders, we should always check whether something is an implicit team, now that we've started migrating TLFs.  This was originally intended to avoid extra implicit team calls for chat, but now that chat is all upgraded, it almost never calls into KBFS, so this shouldn't change the behavior at all.

Issue: KBFS-3480


strib:
cc: @mmaxim.

PR #1842, opened 2018-10-04
keybase_service_base: on an identify error, notify user breaks

strib:
Otherwise, the original identifier will hang because `makeTlfBreaksIfNeeded` won't return.

Issue: KBFS-3480


mmaxim:
Is this needed before upgrading more TLFs?

strib:
> Is this needed before upgrading more TLFs?

No.  This is fixing the hang, but it shouldn't get into the situation that had the hang now that the other PR has been merged.

strib:
Thanks @mmaxim!

PR #1845, opened 2018-10-06
disk_block_cache: put blocks in working set if sync cache is full

strib:
Previously, if the sync cache was full, the overall cache put would fail (without logging an error, due to the unnamed return param). However, this means that when your synced directory is bigger than your disk cache, performance crawls to a halt in that directory because all new blocks need to be fetched again, even if there's room in the working set cache.

So, if the sync cache put fails, just put it in the working set and let eviction take care of it later.


strib:
@jzila: I can't remember if there was a good reason we did it the previous way.  If so, I can rethink this, but it was making performance pretty bad on my private repo (which is synced).

PR #1846, opened 2018-10-08
identify: call userBreak callback on all paths

strib:
And check context for cancellation on all paths as well.

The new widget behavior of use CHAT_GUI identify behavior on SimpleFSStat() calls, as the initializing call into a KBFS folder, was exercising new paths in the identify code that lead to deadlock. Specifically:

* When calling `IdentifyLite` on single-team folder names, we weren't passing any track-break messages back to the goroutine waiting for them.
* When trying to resolve implicit teams, if there were no track breaks, we weren't passing any track-break messages back to the goroutine either.  And now that we're trying to resolve implicit teams for all private and public folders, this was always happening.

Put these two together, and every type of folder was getting deadlocked if the widget stat'd it before its first access via FUSE.

Also fix another potential deadlock noticed via the CHAT_SKIP path.

Issue: KBFS-3487

### DIFF
--- a/libkbfs/identify_util.go
+++ b/libkbfs/identify_util.go
@@ -26,21 +26,47 @@ type extendedIdentify struct {
 	tlfBreaks  *keybase1.TLFBreak
 }
 
-func (ei *extendedIdentify) userBreak(username kbname.NormalizedUsername, uid keybase1.UID, breaks *keybase1.IdentifyTrackBreaks) {
+func (ei *extendedIdentify) userBreak(
+	ctx context.Context, username kbname.NormalizedUsername, uid keybase1.UID,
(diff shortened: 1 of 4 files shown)


maxtaco:
Does this deadlock only affected those with migrated private TLFs?

strib:
> Does this deadlock only affected those with migrated private TLFs?

@maxtaco: No, unrelated.  But it only applies to those running commit d8b3dc64e7c3b47f206c7dedeee7dbd37d180739, which is not in the non-Linux releases.  And for Linux users, it only applies to those who navigate using the GUI, which is probably a small subset.

maxtaco:
Ok, good news the impact is limited and we don't need a hotfix!

PR #1847, opened 2018-10-08
SimpleFS RPC for symlinking

jakob223:
Fixes KBFS-3443. Needs to be merged in conjunction with the corresponding client PR.

### DIFF
--- a/simplefs/simplefs.go
+++ b/simplefs/simplefs.go
@@ -165,24 +165,6 @@ func rawPathFromKbfsPath(path keybase1.Path) (string, error) {
 	}
 }
 
-func rawPathFromPath(path keybase1.Path) (string, error) {
-	pt, err := path.PathType()
-	if err != nil {
-		return "", err
-	}
-
-	switch pt {
-	case keybase1.PathType_KBFS:
-		return stdpath.Clean(path.Kbfs()), nil
-	case keybase1.PathType_KBFS_ARCHIVED:
-		return stdpath.Clean(path.KbfsArchived().Path), nil
-	case keybase1.PathType_LOCAL:
-		return stdpath.Clean(path.Local()), nil
-	default:
-		return "", errInvalidPathType
-	}
-}
-
 // remoteTlfAndPath decodes a remote path for us.
 func remoteTlfAndPath(path keybase1.Path) (
 	t tlf.Type, tlfName, middlePath, finalElem string, err error) {
@@ -989,30 +971,12 @@ func (k *SimpleFS) SimpleFSSymlink(ctx context.Context, arg keybase1.SimpleFSSym
 	}
 	defer func() { k.doneSyncOp(ctx, err) }()
 
-	// Get root FS of dst
-	t, tlfName, restOfDstPath, finalDstElem, err := remoteTlfAndPath(arg.Dest)
-	if err != nil {
-		return err
-	}
-	tlfHandle, err := libkbfs.GetHandleFromFolderNameAndType(
-		ctx, k.config.KBPKI(), k.config.MDOps(), tlfName, t)
-	if err != nil {
-		return err
-	}
-	fs, err := libfs.NewFS(
-		ctx, k.config, tlfHandle, libkbfs.MasterBranch, "", "",
-		keybase1.MDPriorityNormal)
-	if err != nil {
-		return err
-	}
-	srcRawPath, err := rawPathFromPath(arg.Src)
+	dstFS, finalDstElem, err := k.getFS(ctx, arg.Link)
 	if err != nil {
 		return err
 	}
 
-	err = fs.Symlink(
-		stdpath.Clean(srcRawPath),
-		stdpath.Join(restOfDstPath, finalDstElem))
+	err = dstFS.Symlink(arg.Target, finalDstElem)
 	return err
 }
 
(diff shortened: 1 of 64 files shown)


simplefs/simplefs.go
```diff
@@ -961,6 +980,42 @@ func (k *SimpleFS) SimpleFSCopy(ctx context.Context, arg keybase1.SimpleFSCopyAr
...
+// SimpleFSSymlink starts making a symlink of a file or directory
+func (k *SimpleFS) SimpleFSSymlink(ctx context.Context, arg keybase1.SimpleFSSymlinkArg) (err error) {
+	// This is not async.
+	ctx, err = k.startSyncOp(ctx, "Symlink", arg)
+	if err != nil {
+		return err
+	}
+	defer func() { k.doneSyncOp(ctx, err) }()
+
+	// Get root FS of dst
```
strib: Please end comments in punctuation.

simplefs/simplefs.go
```diff
@@ -164,6 +165,24 @@ func rawPathFromKbfsPath(path keybase1.Path) (string, error) {
 	}
 }
 
+func rawPathFromPath(path keybase1.Path) (string, error) {
```
strib: If you take my advice on the other PR, you won't need this function, since the target will just be a string.

simplefs/simplefs.go
```diff
@@ -961,6 +980,42 @@ func (k *SimpleFS) SimpleFSCopy(ctx context.Context, arg keybase1.SimpleFSCopyAr
...
+func (k *SimpleFS) SimpleFSSymlink(ctx context.Context, arg keybase1.SimpleFSSymlinkArg) (err error) {
+	// This is not async.
+	ctx, err = k.startSyncOp(ctx, "Symlink", arg)
+	if err != nil {
+		return err
+	}
+	defer func() { k.doneSyncOp(ctx, err) }()
+
+	// Get root FS of dst
+	t, tlfName, restOfDstPath, finalDstElem, err := remoteTlfAndPath(arg.Dest)
```
strib: Instead of this block, from here down to `NewFS()`, you should be able to use `k.getFS()`, like in `SetStat()`.  Though you might want to add a check that the dest path type is exactly `KBFS` first.

simplefs/simplefs.go
```diff
@@ -961,6 +980,42 @@ func (k *SimpleFS) SimpleFSCopy(ctx context.Context, arg keybase1.SimpleFSCopyAr
...
+		return err
+	}
+	srcRawPath, err := rawPathFromPath(arg.Src)
+	if err != nil {
+		return err
+	}
+
+	err = fs.Symlink(
+		stdpath.Clean(srcRawPath),
+		stdpath.Join(restOfDstPath, finalDstElem))
```
strib: If you use `getFS()` above, you won't need to do this join -- `finalDstElem` would be enough.

simplefs/simplefs.go
```diff
@@ -961,6 +980,42 @@ func (k *SimpleFS) SimpleFSCopy(ctx context.Context, arg keybase1.SimpleFSCopyAr
...
+		keybase1.MDPriorityNormal)
+	if err != nil {
+		return err
+	}
+	srcRawPath, err := rawPathFromPath(arg.Src)
+	if err != nil {
+		return err
+	}
+
+	err = fs.Symlink(
```
strib: Can just `return` here instead of assigning to `err` first.

simplefs/simplefs.go
```diff
@@ -64,6 +64,7 @@ var errOnlyRemotePathSupported = simpleFSError{"Only remote paths are supported
 var errInvalidRemotePath = simpleFSError{"Invalid remote path"}
 var errNoSuchHandle = simpleFSError{"No such handle"}
 var errNoResult = simpleFSError{"Async result not found"}
+var errInvalidPathType = simpleFSError{"Invalid path type"}
```
strib: This is useful, but please use it in `getFS()` as well.

strib:
By the way, it seems like the jenkins tests failed due to an overloaded test runner I think.  Hopefully they'll pass the next time...

simplefs/simplefs.go
```diff
@@ -989,30 +971,12 @@ func (k *SimpleFS) SimpleFSSymlink(ctx context.Context, arg keybase1.SimpleFSSym
...
-		return err
-	}
-	fs, err := libfs.NewFS(
-		ctx, k.config, tlfHandle, libkbfs.MasterBranch, "", "",
-		keybase1.MDPriorityNormal)
-	if err != nil {
-		return err
-	}
-	srcRawPath, err := rawPathFromPath(arg.Src)
+	dstFS, finalDstElem, err := k.getFS(ctx, arg.Dest)
```
strib: Please check here too that the path type of Dest is only KBFS.  (I know you check on in the command implementation, but there could be other callers of this RPC, like the GUI eventually.)

simplefs/simplefs.go
strib: I guess you don't need this anymore?

jakob223: Nope. And somehow I missed most of the comments here when I fixed stuff up  so I never saw this comment in the first place.. whoops

PR #1848, opened 2018-10-09
autogit: cache and reuse `*Browser` objects

strib:
Instead of creating a new `*Browser` for every wrapped Node, have `AutogitManager` create them on-demand and cache them for future use.

Previously, when the browsers were stored in the node objects themselves, they'd get cleaned upon kernel node garbage collection. With this new structure, we use an LRU cache of a limited size to
control the memory.

The browsers themselves are stuck at a particular commit, so they need to be cleared out everytime the underlying repo changes.

The improved the performance of reading an entirely-disk-cached ~30-entry directory from 6.35s to 0.66s on my machine.

The main downsides are that:

* Since the size of the LRU cache is independent of the kernel cache, we might end up throwing away browsers for nodes that are still in the cache, and they have to be re-made again on the next access. But that doesn't seem too likely in practice.

* Previously we were using the caller's context objects when constructing the browsers, which meant we would have the caller's log tags for debugging purposes.  This way uses the log tags of whoever first created the cached browser, for all subsequent calls. This is a bummer, but to fix it we'd need to do surgery in go-git to be able to update the storage layer of their `Tree` objects on the fly, which isn't possible right now.  We can revisit that later if it seems useful.

Issue: KBFS-3474


libgit/autogit_manager.go
```diff
@@ -191,23 +217,54 @@ func (am *AutogitManager) LocalChange(
...
+		if !ok {
+			continue
+		}
+		v, ok := tmp.(browserCacheValue)
+		if !ok {
+			continue
+		}
+		rootNodeID := v.repoFS.RootNode().GetID()
+		for _, nodeID := range repoNodeIDs {
+			if rootNodeID == nodeID {
```
songgao: Is the idea that rootNode will only be removed when all its children are forgotten already?

libgit/autogit_manager.go
```diff
@@ -191,23 +217,54 @@ func (am *AutogitManager) LocalChange(
...
 		if ok {
 			nodes = append(nodes, node)
+			repoNodeIDs = append(repoNodeIDs, nodeID)
+		}
+	}
+	return nodes, repoNodeIDs
+}
+
+func (am *AutogitManager) clearInvalidatedBrowsers(
+	repoNodeIDs []libkbfs.NodeID) {
```
songgao: Would it be better to make `repoNodeIDs` a `map` to make the `rootNodeID` check below faster?

libgit/autogit_manager.go
```diff
@@ -225,10 +282,84 @@ func (am *AutogitManager) TlfHandleChange(
...
+					return nil, nil, err
+				}
+				browser, ok := b.(*Browser)
+				if !ok {
+					return nil, nil, errors.Errorf("Bad browser type: %T", b)
+				}
+				am.browserCache.Add(
+					key, browserCacheValue{rootB.repoFS, browser})
+				return rootB.repoFS, browser, nil
+			}
```
songgao: If `!ok`, might as well return `errors.Errorf("Bad browser in cache: %T", tmp)` to keep consistent?

libgit/autogit_manager.go
```diff
@@ -225,10 +282,84 @@ func (am *AutogitManager) TlfHandleChange(
...
+	}
+
+	// It's kind of dumb to hold the browser lock through all of this,
+	// but it doesn't seem worthwhile to build the whole
+	// channel/notification system that would be needed to manage
+	// multiple concurrent requests for the same repo node.
+
+	am.log.CDebugf(ctx, "Making browser for repo=%s, branch=%s, subdir=%s",
+		repoName, branch, subdir)
+
```
songgao: I wonder if we can recurse into calling `am.GetBrowserForRepo(ctx, gitFS, repoName, branch, "")` here, Chroot, then `Add` the subdir one. This way the part below (during cache miss) can be simplified too.

libgit/autogit_manager.go
strib: Not sure which idea you're talking about, or which removal you mean.

`nodeID` represents a block ID that we were notified about in a metadata update.  It doesn't necessarily have to correspond to any `Node` currently in the `NodeCache`.

`rootNodeID` corresponds to a node that is definitely still in the node cache, because `repoFS` holds a reference to it, and also `AutogitManager.repoNodesForWatchedIDs` has a reference to it.

This code is just making sure we wipe all the browser associated with the repo from our browser cache, since the corresponding git repo has been updated.

libgit/autogit_manager.go
strib: Great idea!  Done in e730dd3.

libgit/autogit_manager.go
songgao: Ah I see. I was somehow under the impression that it may contain thousands of nodes. If it's very small then this is good. Thanks for explaining! (Though it might be helpful to add a comment when you touch this file again.)

libgit/autogit_manager.go
songgao: Ah; sorry for the vague question. I was referring to the part where we remove k from browserCache three lines below this, and that whether it was the best place to remove it here when `rootNodeID === nodeID`. I was (mistakenly) imagining that `repoNodeID` would be way more than normally just 1. So your comment above already answered my question. Thanks!

PR #1851, opened 2018-10-09
root_metadata: clear unresolved users from MD on migration

strib:
The unresolved users are now part of the implicit team.  Keeping them in the MD confuses the MD, and causes it to not return `TeamKeying` as the keying type.


strib:
cc: @mmaxim, this is what was causing SBS migrations to fail.

PR #1852, opened 2018-10-10
libhttpserver: Display less things inline (fix xlsx bug)

taruti:


libhttpserver/content_type_utils.go
```diff
@@ -29,12 +29,13 @@ func (w *contentTypeOverridingResponseWriter) calculateOverride(
 	ty := strings.ToLower(mimeType)
 	switch {
 	// First anything textual as text/plain.
-	// Includes javascript, html, xml. (note that the type may be e.g. application/xhtml+xml)
+	case strings.HasPrefix(ty, "text/"):
+		return "text/plain", "inline"
+	// Rest of javascript, html, xml. (note that the type may be e.g. application/xhtml+xml)
 	case strings.Contains(ty, "javascript") ||
 		strings.Contains(ty, "xml") ||
```
songgao: It seems the reason why it hit this path is `xml` is [in the mime type of xlsx string](https://github.com/keybase/kbfs/blob/e9256567b9a2990a93490849e4f44df08f4c70c6/libmime/mime_types.go#L112) unfortunately. Perhaps we could instead just change this to `strings.HasSuffix(ty, "/xml")` to cover both `application/xml` and `text/xml`? This way other mime types that has "xml" inside would just fall into `default:` and `attachment`. WDYT?

taruti: Yes, 

For more fun there is `image/svg+xml`, `application/xhtml+xml` etc

However currenltly `application/xml` is never produced, that suffix (`/xml`) would only match `text/xml`. So that would not be better.

taruti: I think this change is better than what we have at the moment.

For better results I could go through all the types in libmime and look at what would make sense and make the rules based on that.

songgao: Thanks for the context. I was just hoping we could still display the plain text of text/xml and text/html, and other text content. Do you think there'd be a way to detect non-binary xml types?

libhttpserver/content_type_utils.go
```diff
@@ -29,12 +29,13 @@ func (w *contentTypeOverridingResponseWriter) calculateOverride(
...
-		strings.Contains(ty, "html") ||
-		strings.HasPrefix(ty, "text/"):
-		return "text/plain", "inline"
+		strings.Contains(ty, "html"):
+		return "text/plain", "attachment"
```
songgao: If we are returning `attachment`, is it still necessary to override the type as `text/plain`?

PR #1853, opened 2018-10-10
cr: fill in final paths for ops on created chains

strib:
Even if the `getPaths()` caller doesn't explicitly request the paths of created chains, we still need to fill in the final paths on the ops of those chains, because they might be used later (for example, in the case where the same directory is created in both the unmerged and merged branches, and the child entries need to be compared for conflicts).

Also, to test this and help prevent all similar issues like this in the past, when in test mode the code now first clears all paths, and afterwards checks that paths were set on all ops.  Before this PR, this resulted in several test failures that are now fixed.

Issue: KBFS-3493

### DIFF
--- a/libkbfs/cr_chains.go
+++ b/libkbfs/cr_chains.go
@@ -1224,7 +1224,7 @@ func (ccs *crChains) getPaths(ctx context.Context, blocks *folderBlockOps,
 	}
 
 	if checkOpFinalPaths {
-		// If we plans to check all the paths, clear them out first.
+		// If we plan to check all the paths, clear them out first.
 		for _, chain := range ccs.byMostRecent {
 			for _, op := range chain.ops {
 				op.setFinalPath(path{})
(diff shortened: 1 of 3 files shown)


libkbfs/cr_chains.go
```diff
@@ -1133,6 +1134,50 @@ func (ccs *crChains) findPathForDeleted(mostRecent BlockPointer) path {
...
+	for ptr, chain := range ccs.byMostRecent {
+		for _, op := range chain.ops {
+			co, ok := op.(*createOp)
+			if !ok {
+				continue
+			}
+			for _, ref := range co.Refs() {
+				if ref == mostRecent {
+					// If the path isn't set yet, recurse.
+					p := co.getFinalPath()
```
songgao: is "final path" the parent path of the entry here?

libkbfs/cr_chains.go
```diff
@@ -1178,6 +1223,15 @@ func (ccs *crChains) getPaths(ctx context.Context, blocks *folderBlockOps,
 		}
 	}
 
+	if checkOpFinalPaths {
+		// If we plans to check all the paths, clear them out first.
```
songgao: s/plans/plan

libkbfs/cr_chains.go
strib: Yep, exactly!

PR #1854, opened 2018-10-10
tlf_handle_resolve: don't fetch team TLF ID if idGetter == nil

strib:
An `ls /keybase/team` results in a `Lookup()` call for each team in your favorites list.  This in turn requires parsing the TLF handle. The folderlist already uses `libfs.ParseTlfHandlePreferredQuick()`, which sets the `idGetter` to `nil` to skip looking up TLF IDs for private/public folders.  But team TLF IDs were treated differently, since they don't use `idGetter`.

Instead, if `idGetter` is `nil`, use that as a signal to skip looking up the team TLF ID, to make stats much faster in the team folder list.

### DIFF
--- a/libkbfs/tlf_handle_resolve.go
+++ b/libkbfs/tlf_handle_resolve.go
@@ -269,6 +269,7 @@ func makeTlfHandleHelper(
 
 type resolvableID struct {
 	resolver resolver
+	idGetter tlfIDGetter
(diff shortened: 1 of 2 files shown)


strib:
@maxtaco: this eliminates the `GetKBFSTeamSettings()` calls during `ls /keybase/team`.

PR #1856, opened 2018-10-12
Rewrite bt.getBlocksForOffsetRange to avoid spawning too many goroutines

jakob223:
rewrite BlockTree.getBlocksForOffsetRange to be iterative and
worker-pool based instead of recursive, so as to limit the
number of goroutines spawned.

### DIFF
--- a/libkbfs/block_tree.go
+++ b/libkbfs/block_tree.go
@@ -7,6 +7,8 @@ package libkbfs
 import (
 	"context"
 	"fmt"
+	"sort"
+	"sync"
 
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/client/go/protocol/keybase1"
@@ -269,6 +271,24 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
 	return ptr, parentBlocks, block, nextBlockStartOff, startOff, nil
 }
 
+// TODO: rename these structs
+// structs used for passing data to and from GetBlocksForOffsetRange tasks
+type task struct {
+	ptr        BlockPointer
+	pblock     *BlockWithPtrs
+	pathPrefix []parentBlockAndChildIndex
+	startOff   Offset
+	endOff     Offset
+	prefixOk   bool
+	getDirect  bool
+}
+type result struct {
+	pathFromRoot    []parentBlockAndChildIndex
+	ptr             *BlockPointer
+	block           Block
+	nextBlockOffset Offset
+}
+
 // getBlocksForOffsetRange fetches all the blocks making up paths down
 // the block tree to leaf ("direct") blocks that encompass the given
 // offset range (half-inclusive) in the data.  If `endOff` is nil, it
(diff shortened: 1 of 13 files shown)


libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-		// caller already checked the range for this block.
-		if getDirect {
-			// Return a child map with only this block in it.
-			return [][]parentBlockAndChildIndex{nil},
-				map[BlockPointer]Block{ptr: pblock}, nil, nil
+
+	// Make a queue for tasks and one for results. Queue for results is big to prevent workers from blocking on
+	// sending back their results.
+	jobs := make(chan task, 2)
+	results := make(chan result, 400)
```
strib: The results chan is being continuously drained by a separate goroutine which doesn't block on anything else, so I don't think this needs to be big.  In fact, it can probably be a 0-length channel and that would be fine.  Am I missing something though?

libkbfs/block_tree.go
```diff
@@ -269,6 +271,24 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+	pblock     *BlockWithPtrs
+	pathPrefix []parentBlockAndChildIndex
+	startOff   Offset
+	endOff     Offset
+	prefixOk   bool
+	getDirect  bool
+}
+type result struct {
+	pathFromRoot    []parentBlockAndChildIndex
+	ptr             *BlockPointer
```
strib: Why not use `result.Block != nil` as an indication that there's a block, instead of making this a pointer?

libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-			for _, p := range pfr {
-				newPath := append([]parentBlockAndChildIndex{{
-					pblock:     pblock,
-					childIndex: childIndex,
-				}}, p...)
-				r.pathsFromRoot = append(r.pathsFromRoot, newPath)
+			// process the subtask by passing it to another goroutine; if they're all occupied, recurse instead.
+			select {
+			case jobs <- subTask:
+				wg.Add(1)
```
strib: I'm pretty sure you need to `wg.Add(1)` _before_ sending to the channel (and call `wg.Done()` in the `default` case below).  Otherwise there's a race where the job can be processed and completed before Add is called.

libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-			return [][]parentBlockAndChildIndex{nil},
-				map[BlockPointer]Block{ptr: pblock}, nil, nil
+
+	// Make a queue for tasks and one for results. Queue for results is big to prevent workers from blocking on
+	// sending back their results.
+	jobs := make(chan task, 2)
+	results := make(chan result, 400)
+
+	// make a WaitGroup to keep track of whether there's still work to be done.
+	// TODO: simplify this so we don't need both the errgroup and the WaitGroup
```
strib: Yeah I think we can simplify all this a bit, and just use a `WaitGroup`:

1. Add an `error` field to `result`.
2. Make a new context using `context.WithCancel()`, and call `defer cancel()`.
3. Use `go` to launch your 16 workers, passing each of them the new context.
4. Use `wg` to track the length of the jobs queue, as you already do below.
5. If any of the results have a non-nil error (subject to the `DeadlineExceeded` check), just return it immediately, which will cancel the context via the `defer` and cause each of the workers will exit.

It's kind of like implementing your own `errgroup`, except now each error is tied precisely to the job that triggered it.  It wouldn't be a huge improvement, just a little one.  I'm open to other suggestions, if you already had some other simplification in mind.

libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-					pblock:     pblock,
-					childIndex: childIndex,
-				}}, p...)
-				r.pathsFromRoot = append(r.pathsFromRoot, newPath)
+			// process the subtask by passing it to another goroutine; if they're all occupied, recurse instead.
+			select {
+			case jobs <- subTask:
+				wg.Add(1)
+			default:
+				bt.processGetBlocksTask(ctx, wg, subTask, jobs, results)
```
strib: The error is dropped here, though if you implement my suggestion above and return it in `result` instead, you won't need to fix this.

strib: I don't really get the benefit of re-using this worker to process the next job immediately, rather than making the job queue the same size as the number of workers, and letting this worker just pick the next thing from the queue.  That would make things easier to reason about, I think, especially when trying to inspect stack traces.

libkbfs/block_tree.go
```diff
@@ -269,6 +271,24 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
 	return ptr, parentBlocks, block, nextBlockStartOff, startOff, nil
 }
 
+// TODO: rename these structs
+// structs used for passing data to and from GetBlocksForOffsetRange tasks
+type task struct {
+	ptr        BlockPointer
+	pblock     *BlockWithPtrs
```
strib: `BlockWithPtrs` is an interface, and can be `nil` on its own, so no need to have a pointer to it here.

libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
+
+		// We only need to fetch direct blocks if we've been asked
+		// to do so.  If the direct type of the pointer is
+		// unknown, we can assume all the children are direct
+		// blocks, since there weren't multiple levels of
+		// indirection before the introduction of the flag.
+		if job.getDirect || childPtr.DirectType == IndirectBlock {
+			// Create a task for the child, and either enqueue it or recurse to it.
+
+			subTask := task{
```
strib: Maybe slightly less error-prone to do something like:
```go
subTask := job
subtask.ptr = childPtr
subtask.pblock = nil
subtask.pathPrefix = childPath
```

But I don't feel strongly about it either way.

Even better might be making a `job.subTask()` function that takes in the new data and encapsulates this copy, so the logic is closer to where the fields are defined.

strib:
(By the way, you need to rebase on master to get CI to pass, I think.)

libkbfs/block_tree.go
jakob223: The jobs queue would need to be arbitrarily large under this architecture to allow that - each job can queue lots more jobs, and so the queue can become full, leading to a deadlock where all the workers are trying to enqueue jobs. If we switch to the workerpool you linked in the top comment, we avoid this.

strib: Ah fair point.  During my review I didn't remember that this was inside of a loop over all the pointers in the block, rather than a single call (after which the worker exited and pulled a new job off the queue anyway).

But yeah, if you like workerpool, then it'd be nice to get rid of this, to make the code cleaner.  I think letting the job queue get as big as it needs to be is fine in this case.  I don't expect that to ever approach memory limits.

libkbfs/block_tree.go
```diff
@@ -291,159 +311,214 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-	nextBlockOffset = pblock.FirstOffset()
-	if !pblock.IsIndirect() {
-		// Return a single empty path, under the assumption that the
-		// caller already checked the range for this block.
-		if getDirect {
-			// Return a child map with only this block in it.
-			return [][]parentBlockAndChildIndex{nil},
-				map[BlockPointer]Block{ptr: pblock}, nil, nil
+
+	// Make a queue for tasks and one for results. Queue for results is big to prevent workers from blocking on
```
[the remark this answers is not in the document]
jakob223: Whoops, my editor was set to wrap at 120 chars

libkbfs/block_tree.go
jakob223: I think this resulted from my misunderstanding where most of the time was going to be taken. I agree 0-length would be fine.

jakob223:
@strib I've addressed all your comments.

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
-		if getDirect {
-			// Return a child map with only this block in it.
-			return [][]parentBlockAndChildIndex{nil},
-				map[BlockPointer]Block{ptr: pblock}, nil, nil
-		}
-		// Return an empty child map with no blocks in it (since
-		// getDirect is false).
-		return [][]parentBlockAndChildIndex{nil}, nil, nil, nil
+// getBlocksForOffsetRangeTask is used for passing data to
+// getBlocksForOffsetRange tasks
```
strib: Please end comments with punctuation (here and below).

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+	ptr        BlockPointer
+	pblock     BlockWithPtrs
+	pathPrefix []parentBlockAndChildIndex
+	startOff   Offset
+	endOff     Offset
+	prefixOk   bool
+	getDirect  bool
+}
+
+func (task *getBlocksForOffsetRangeTask) subTask(childPtr BlockPointer,
```
strib: We aren't very consistent on this point and don't really have an official policy on it, but I personally prefer breaking function signatures after the `(`, so that if it's going to be too long, the parameters start on their own line.  (Here and below.) No worries if you prefer to keep it this way, though.

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+	nextBlockOffset Offset
+	err             error
+}
+
+// processGetBlocksTask examines the block it is passed, enqueueing any children
+// in range into wp, and passing data back through results.
+func (bt *blockTree) processGetBlocksTask(ctx context.Context,
+	wg *sync.WaitGroup, wp *workerpool.WorkerPool,
+	job getBlocksForOffsetRangeTask,
+	results chan getBlocksForOffsetRangeResult) {
```
strib: Can be `results chan<- getBlocksForOffsetRangeResult`.

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
-				}}, p...)
-				r.pathsFromRoot = append(r.pathsFromRoot, newPath)
+		// We only need to fetch direct blocks if we've been asked
+		// to do so.  If the direct type of the pointer is
+		// unknown, we can assume all the children are direct
+		// blocks, since there weren't multiple levels of
+		// indirection before the introduction of the flag.
+		if job.getDirect || childPtr.DirectType == IndirectBlock {
+			// Create a task for the child, and either enqueue it or recurse to it.
+
```
strib: Extra blank line.

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+}
+
+// processGetBlocksTask examines the block it is passed, enqueueing any children
+// in range into wp, and passing data back through results.
+func (bt *blockTree) processGetBlocksTask(ctx context.Context,
+	wg *sync.WaitGroup, wp *workerpool.WorkerPool,
+	job getBlocksForOffsetRangeTask,
+	results chan getBlocksForOffsetRangeResult) {
+	defer wg.Done()
+
```
strib: Maybe the first thing we should do here is check that `ctx` hasn't been cancelled yet?

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+	ptr BlockPointer, pblock BlockWithPtrs, startOff, endOff Offset,
+	prefixOk bool, getDirect bool) (pathsFromRoot [][]parentBlockAndChildIndex,
+	blocks map[BlockPointer]Block, nextBlockOffset Offset,
+	err error) {
+	// Make a WaitGroup to keep track of whether there's still work to be done.
+	var wg sync.WaitGroup
+
+	// Make a workerpool to limit the number of concurrent goroutines.
+	wp := workerpool.New(maxBlockFetchWorkers)
+	defer wp.Stop() // TODO: this may be unnecessary
```
strib: Unless you think a "TODO" requires further ticketed work, please address them and remove the comments before merging.  If they do require more work, make a ticket for them in JIRA and add the ticket number here, like `// TODO(KBFS-3503): ...`.

In this case, this is definitely not needed because you make the same call again in a later goroutine.  But in general, the workerpool documentation implies to me that the `Stop` call is needed to clean up the goroutines launched by it (though I haven't checked the code).

libkbfs/block_tree.go
```diff
@@ -269,179 +275,235 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
-				break outer
-			} else {
-				panic("No response ready when !prefixOk")
+	pathsFromRoot = [][]parentBlockAndChildIndex{}
+	for res := range results {
+		if res.err != nil {
+			// If we are ok with just getting the prefix, don't treat a
+			// deadline exceeded error as fatal.
+			if prefixOk && res.err == context.DeadlineExceeded {
+				//TODO: determine that this is correct
```
strib: I don't think that `break` is the right thing here, since results could come out of order and there might still be some valid results waiting in the queue.  Probably `continue` would be better.

Even with `continue`, though, I think you're introducing a new bug.  The idea with `prefixOk` is that a file system read is allowed to return less data than was requested.  So if the call asked for 100 bytes, but the read only returned 20 bytes, that's ok and doesn't result in an error.  Since we're no longer processing results in order here, the returned set of blocks might contain (for example) one block at the start of the range, and one at the end.  Then other code (in `file_data.go`) would incorrectly fill zeroes in between those two blocks, when the correct behavior would have been just to return the data in the first block, ignoring the last one.

I'm not quite sure about the right way to fix it.  I think that, after you sort the paths below, you can check that the `childIndex` of the final element of each path increases by 1 until it reaches the number of pointers in the parent block, at which point the `childIndex` of the parent element must increase by 1.  And if you find a hole, and `prefixOk == true`, then just remove all the paths after the hole from the slice being returned.

Maybe there's a less code-intensive way of doing it, but I'm drawing a blank right now.

jakob223:
@strib fixed all the small things and I think I fixed the remaining bug. I don't think we currently have any tests covering that case, which makes me slightly nervous.

libkbfs/block_tree.go
```diff
@@ -505,6 +519,63 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
+		var prevPath []parentBlockAndChildIndex
+		for pathIdx, path := range pathsFromRoot {
+			// First path should be at startOff. Each subsequent path should
+			// immediately follow the preceding path.
+			pathIsCorrect := true
+			if pathIdx == 0 {
+				lastBlockInPath := path[len(path)-1]
+				_, off := lastBlockInPath.pblock.IndirectPtr(
+					lastBlockInPath.childIndex)
+				if off != startOff {
```
strib: Hrm, this isn't true actually. startOff just has to be somewhere within the given block, not necessarily the first byte.  (For example, the block could start at byte 0 and be 1024 bytes long, but the read could start at byte 100.)

For that matter, we have this concept of "holes" in file data, since you can "truncate" a file to a very large size and it doesn't make sense to write all those zeroes.  So the first block could legitimately have an offset that is greater than `off`.  Bleh.

Maybe each job and result should track whether this is the left-most requested block for the request overall.  So the first job would have some bit set, and it would transfer that bit to the first job it requests, and so on.  And when a block or error is returned when the bit is set for that job also sets the bit in the result.  Then if the above code gets a `DeadlineExceeded` error and the result bit is set, you can just return the error since you know you didn't get any valid data.  (And this check here goes away.)

libkbfs/block_tree.go
```diff
@@ -269,178 +275,305 @@ func (bt *blockTree) getNextDirtyBlockAtOffset(ctx context.Context,
...
+	prefixOk bool, getDirect bool) (pathsFromRoot [][]parentBlockAndChildIndex,
+	blocks map[BlockPointer]Block, nextBlockOffset Offset,
+	err error) {
+	// Make a WaitGroup to keep track of whether there's still work to be done.
+	var wg sync.WaitGroup
+
+	// Make a workerpool to limit the number of concurrent goroutines.
+	wp := workerpool.New(maxBlockFetchWorkers)
+
+	// Make a context to cancel all the jobs if something goes wrong
```
strib: Please end in punctuation.

libkbfs/block_tree.go
```diff
@@ -466,15 +470,20 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
...
-				break
+			if prefixOk && res.err == context.DeadlineExceeded &&
+				len(errors) == 0 {
+				mustCheckForHoles = true
+			} else {
+				errors = append(errors, res.err)
 			}
-			return nil, nil, nil, res.err
+			wp.Stop()
+			cancel()
```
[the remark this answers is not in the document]
jakob223: `wp.Stop()` isn't blocking.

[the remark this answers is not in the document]
strib: Ah thanks, I forgot there was a separate `StopWait()`.  Ok thanks, this is fine then.

libkbfs/block_tree_test.go
```diff
@@ -0,0 +1,65 @@
...
+			t.Errorf("Paths incorrectly truncated from %v to %v",
+				paths, newPaths)
+		}
+	}
+
+	for idx, badEx := range badExamples {
+		paths := makePaths(badEx, mockBlock)
+		newPaths := checkForHolesAndTruncate(paths)
+		if len(newPaths) != correctLengths[idx] {
+			t.Errorf("Paths incorrectly truncated from %v to %v",
```
strib: Please use `require.Len(t, newPaths, correctLengths[idx])` instead of the `if` and the `t.Errorf`.

libkbfs/block_tree_test.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	// correctLengths holds the length that each bad example should be truncated
+	// to.
+	correctLengths := []int{1, 3, 3}
+
+	for _, goodEx := range goodExamples {
+		paths := makePaths(goodEx, mockBlock)
+		newPaths := checkForHolesAndTruncate(paths)
+		if !reflect.DeepEqual(paths, newPaths) {
+			t.Errorf("Paths incorrectly truncated from %v to %v",
```
strib: Please use `require.True(t, reflect.DeepEqual(paths, newPaths))` instead of the `if` and the `t.Errorf`. (We do this for new tests, but haven't fixed all the old tests yet.)  There's an optional third parameter for messages as well, if you really want to print `paths` and `newPaths` when the check fails.

libkbfs/block_tree_test.go
```diff
@@ -0,0 +1,65 @@
...
+	mockBlock := NewMockBlockWithPtrs(mockCtrl)
+	mockBlock.EXPECT().NumIndirectPtrs().Return(4).MinTimes(1)
+
+	goodExamples := [][][]int{
+		{{3, 2}, {3, 3}},
+		{{2, 3}, {3, 0}},
+		{{1, 0}, {1, 1}, {1, 2}, {1, 3}, {2, 0}, {2, 1}},
+		{{1, 2, 3}, {1, 3, 0}, {1, 3, 1}, {1, 3, 2}, {1, 3, 3}, {2, 0, 0}},
+	}
+	badExamples := [][][]int{
```
strib: Slight preference to moving `badExamples` and `correctLengths` definitions to right above the bad examples loop.

libkbfs/block_tree.go
```diff
@@ -344,6 +348,7 @@ func (bt *blockTree) processGetBlocksTask(ctx context.Context,
 				ptr:             job.ptr,
 				block:           pblock,
 				nextBlockOffset: nil,
+				firstBlock:      job.firstBlock,
```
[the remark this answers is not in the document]
jakob223: I claim we don't need to do it on the former line 382. That's not about a block that is in range for the job so it's sort of meaningless.

libkbfs/block_tree.go
```diff
@@ -496,6 +552,9 @@ func (bt *blockTree) getBlocksForOffsetRange(ctx context.Context,
 				res.nextBlockOffset.Less(minNextBlockOffset)) {
 			minNextBlockOffset = res.nextBlockOffset
 		}
+		if res.firstBlock {
+			gotFirstBlock = true
```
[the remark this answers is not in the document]
jakob223: Ooh, I didn't realize that indicated the end of a file. I added more logic to prevent us from moving on if the first block gets `context.DeadlineExceeded`.

libkbfs/block_tree.go
```diff
@@ -403,11 +409,59 @@ func (bt *blockTree) processGetBlocksTask(ctx context.Context,
...
 	}
 }
 
+func checkForHolesAndTruncate(
+	pathsFromRoot [][]parentBlockAndChildIndex) [][]parentBlockAndChildIndex {
+	var prevPath []parentBlockAndChildIndex
+	for pathIdx, path := range pathsFromRoot {
+		// Each path after the first must immediately follow the preceding path.
+		pathIsCorrect := true
+		if pathIdx != 0 {
```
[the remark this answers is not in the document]
jakob223: Heh, I did it this way to avoid having multiple `return` locations. But I can do it that way instead.

libkbfs/block_tree_test.go
jakob223: I do really want to print them, it was useful when debugging the originally incorrect implementation.

libkbfs/block_tree.go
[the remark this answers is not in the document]
strib: Multiple returns aren't a huge deal in golang, compared to other languages, because of the nice `defer` statements.  So I tend to prefer to optimize for indentation, but your mileage might vary.

PR #1860, opened 2018-10-15
leveldb: factor out some code that will be useful for other DBs

strib:
Suggested by jzila.

Issue: KBFS-3503


strib:
@jzila: I'm not sure if this is what you had in mind with refactoring the leveldb stuff.  Let me know if there are other pieces you think I should be able to pull out into common code.

PR #1861, opened 2018-10-16
kbfscrypto: set a Cause on libkb.DecryptionErrors

strib:
Otherwise there's a panic when sending the error back over RPC.

Issue: KBFS-3514


strib:
@jzila: I believe it panics here: https://github.com/keybase/kbfs/blob/2f3974677af77bfd4f89b23ad6afed1fd27af60a/vendor/github.com/keybase/client/go/libkb/rpc_exim.go#L1923

The stacktrace pointed to the `return` line, but I'm pretty sure this is the cause.

jzila:
OK we should fix that to not require the cause.

PR #1862, opened 2018-10-17
favorites: add env var for disabling favorites

strib:
Issue: KBFS-3530

### DIFF
--- a/libkbfs/favorites.go
+++ b/libkbfs/favorites.go
@@ -5,12 +5,17 @@
 package libkbfs
 
 import (
+	"os"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


libkbfs/favorites.go
```diff
@@ -70,6 +76,17 @@ type Favorites struct {
 }
 
 func newFavoritesWithChan(config Config, reqChan chan *favReq) *Favorites {
+	disableVal := os.Getenv(disableFavoritesEnvVar)
+	if len(disableVal) > 0 {
```
jzila: This should also check that the value isn't falsey.

PR #1864, opened 2018-10-18
disk_md_cache: introduce disk cache for the last known MD heads

strib:
The cache has a Stage+Commit flow, instead of the usual Put, because according to the KBFS Offline design, we might not want to cache the MD right after learning about it.  Instead, we might want to wait until syncing is done, or until the user actually accesses it for the first time.  In the meantime, the MD must be stored in encrypted form somewhere, so this cache stages them in memory until they are fully commit to disk or unstaged.

Issue: KBFS-3504

### DIFF
--- a/libkbfs/disk_md_cache.go
+++ b/libkbfs/disk_md_cache.go
@@ -1,0 +1,437 @@
+// Copyright 2018 Keybase Inc. All rights reserved.
+// Use of this source code is governed by a BSD
+// license that can be found in the LICENSE file.
+
+package libkbfs
+
+import (
+	"context"
+	"io"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/keybase/client/go/logger"
+	"github.com/keybase/kbfs/kbfsmd"
+	"github.com/keybase/kbfs/tlf"
+	"github.com/pkg/errors"
+	ldberrors "github.com/syndtr/goleveldb/leveldb/errors"
+	"github.com/syndtr/goleveldb/leveldb/opt"
+	"github.com/syndtr/goleveldb/leveldb/storage"
+)
+
+const (
+	headsDbFilename           string = "diskCacheMDHeads.leveldb"
+	initialDiskMDCacheVersion uint64 = 1
+	currentDiskMDCacheVersion uint64 = initialDiskMDCacheVersion
+	defaultMDCacheTableSize   int    = 50 * opt.MiB
+)
+
+// diskMDCacheConfig specifies the interfaces that a DiskMDCacheStandard
+// needs to perform its functions. This adheres to the standard libkbfs Config
+// API.
+type diskMDCacheConfig interface {
+	codecGetter
+	logMaker
+}
+
+type diskMDCacheStageKey struct {
+	tlfID tlf.ID
+	rev   kbfsmd.Revision
+}
+
+type diskMDBlock struct {
+	// Exported only for serialization.
+	Buf  []byte
+	Ver  kbfsmd.MetadataVer
+	Time time.Time
+}
+
+// DiskMDCacheLocal is the standard implementation for DiskMDCache.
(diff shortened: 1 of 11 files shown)


libkbfs/disk_md_cache.go
```diff
@@ -0,0 +1,483 @@
...
+	startErrCh := make(chan struct{})
+	cache = &DiskMDCacheLocal{
+		config:     config,
+		hitMeter:   NewCountMeter(),
+		missMeter:  NewCountMeter(),
+		putMeter:   NewCountMeter(),
+		log:        log,
+		headsDb:    headsDb,
+		tlfsCached: map[tlf.ID]kbfsmd.Revision{},
+		tlfsStaged: make(map[tlf.ID][]diskMDBlock),
```
songgao: Any reason to use two style of `map` initialization?

libkbfs/interfaces.go
```diff
@@ -1225,6 +1233,31 @@ type DiskBlockCache interface {
...
+	// Stage asks the disk cache to store the given MD in memory, but
+	// not yet write it to disk.  A later call to `Commit` or
+	// `Unstage` for `rev` or higher is required to avoid memory leaks.
+	Stage(ctx context.Context, tlfID tlf.ID, rev kbfsmd.Revision, buf []byte,
+		ver kbfsmd.MetadataVer, localTimestamp time.Time) error
+	// Commit writes a previously-staged MD to disk.  Trying to commit
+	// a revision that hasn't been staged is a no-op, to allow callers
+	// to call Commit without knowing whether Stage was called first
+	// (e.g., if the revision came from the cache in the first place).
+	Commit(ctx context.Context, tlfID tlf.ID, rev kbfsmd.Revision) error
```
songgao: Like `Unstage` below, perhaps mention the behavior for committing when multiple revisions are staged here too? It seems only first one with given revision is committed.

libkbfs/interfaces.go
```diff
@@ -1225,6 +1233,31 @@ type DiskBlockCache interface {
...
+	// Unstage unstages and forgets about a previously-staged MD.  (If
+	// multiple copies of the same revision have been staged, it only
+	// unstages one of them.)
+	Unstage(ctx context.Context, tlfID tlf.ID, rev kbfsmd.Revision) error
```
songgao: Make it clear that it only unstages *first of them* instead of "one of them"?

libkbfs/disk_md_cache.go
```diff
@@ -0,0 +1,483 @@
...
+	if cache.headsDb == nil {
+		return errors.WithStack(DiskCacheClosedError{method})
+	}
+	return nil
+}
+
+// Get implements the DiskMDCache interface for DiskMDCacheLocal.
+func (cache *DiskMDCacheLocal) Get(
+	ctx context.Context, tlfID tlf.ID) (
+	buf []byte, ver kbfsmd.MetadataVer, localTimestamp time.Time, err error) {
```
songgao: Could you explain how `localTimestamp` will be used?

libkbfs/disk_md_cache.go
```diff
@@ -0,0 +1,483 @@
...
+	foundMD := false
+	// The staged MDs list is unordered, so iterate through the whole
+	// thing to find what should remain after commiting `rev`.
+	for _, md := range stagedMDs {
+		if md.Revision > rev {
+			newStagedMDs = append(newStagedMDs, md)
+			continue
+		} else if md.Revision < rev {
+			continue
+		} else if foundMD && md.Revision == rev {
```
songgao: `&& md.Revision == rev` isn't necessary here I think.

libkbfs/interfaces.go
songgao: Also perhaps mention all revisions older than rev would be unstaged, like in your commit message.

PR #1867, opened 2018-10-22
simplefs: operation for resetting TLFs

strib:
This allows the user to reset their TLF by generating a new TLF ID for the implicit team, with the epoch incremented.  It clears the old TLF out of the running instance.  The new TLF ID is written to the implicit team sigchain, and overrides the old one.

It can only be done AFTER a Keybase admin does a server-side reset, so we can track which TLFs are being reset.

[This doesn't yet include the protocol changes in keybase/client#14358, but it should pass Circle anyway.  I'll vendor in the protocol changes once the client-side is merged.]

Issue: KBFS-3545
Issue: keybase/client#14319

### DIFF
--- a/libkbfs/kbfs_ops.go
+++ b/libkbfs/kbfs_ops.go
@@ -407,16 +407,7 @@ func (fs *KBFSOpsStandard) getOpsByHandle(ctx context.Context,
 	return ops
 }
 
-// createAndStoreTlfIDIfNeeded creates a TLF ID for a team-backed
-// handle that doesn't have one yet, and associates it in the service
-// with the team.  If it returns a `nil` error, it may have modified
-// `h` to include the new TLF ID.
-func (fs *KBFSOpsStandard) createAndStoreTlfIDIfNeeded(
-	ctx context.Context, h *TlfHandle) error {
-	if h.tlfID != tlf.NullID {
-		return nil
-	}
-
+func (fs *KBFSOpsStandard) resetTlfID(ctx context.Context, h *TlfHandle) error {
 	if !h.IsBackedByTeam() {
 		return errors.New("Can't create TLF ID for non-team-backed handle")
 	}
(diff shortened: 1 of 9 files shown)


simplefs/simplefs.go
```diff
@@ -1897,6 +1897,22 @@ func (k *SimpleFS) SimpleFSFolderEditHistory(
 	return k.config.KBFSOps().GetEditHistory(ctx, fb)
 }
 
+// SimpleFSFolderEditHistory resets the given TLF.
```
songgao: linter screamer here!

libkbfs/kbfs_ops_test.go
```diff
@@ -3904,3 +3904,65 @@ func TestKBFSOpsReadonlyFSNodes(t *testing.T) {
...
+	rootNode, _, err := kbfsOps.GetOrCreateRootNode(ctx, h, MasterBranch)
+	require.NoError(t, err)
+
+	oldID := h.tlfID
+	t.Logf("Make a new revision for TLF ID %s", oldID)
+	_, _, err = kbfsOps.CreateDir(ctx, rootNode, "a")
+	require.NoError(t, err)
+	err = kbfsOps.SyncAll(ctx, rootNode.GetFolderBranch())
+	require.NoError(t, err)
+
```
songgao: I'm a bit confused; There doesn't seem to be a place where we blow away the TLF to simulate the server side reset. How did it work?

Also it'd be nice to test the `FolderNotResetOnServer` case too if it's not too complicated.

libkbfs/kbfs_ops.go
```diff
@@ -426,10 +417,20 @@ func (fs *KBFSOpsStandard) createAndStoreTlfIDIfNeeded(
...
 	}
 
+	matches, epoch, err := h.tlfID.GetEpochFromTeamTLF(teamID)
+	if err != nil {
+		return err
+	}
+	if matches {
+		epoch++
+	} else {
+		epoch = 0
```
songgao: If it doesn't match, does it mean the TLF is not migrated to impl yet?

libkbfs/kbfs_ops_test.go
strib: The local mdserver implementation doesn't store the handle -> ID mapping for implicit team TLFs, so it doesn't have the ability to do the same checks as the real mdserver.  So every `GetForHandle` call returns `ServerErrorClassicTLFDoesNotExist{}`.

If you think it's important I can change the disk-based mdserver to track those mappings, and have the test delete it explicitly.  I'm a little worried it'll break other tests, so I can't say how easy it will be.

(I have tested this manually against the real mdserver, for what it's worth.)

libkbfs/kbfs_ops.go
strib: Yeah, exactly.

libkbfs/kbfs_ops_test.go
songgao: Ah got it. As you've already tested against the real mdserver, I'm good with this. Thanks for the context!

PR #1870, opened 2018-10-23
libkbfs: actually keep a favorites cache

jakob223:
This commit changes the favorites cache to not refresh until it is invalidated instead of refreshing every time.

Relevant PRs:
 - https://github.com/keybase/client/pull/14370
 - https://github.com/keybase/keybase/pull/3059

### DIFF
--- a/libkbfs/favorites.go
+++ b/libkbfs/favorites.go
@@ -120,9 +120,7 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 	// Fetch a new list if:
 	//  * The user asked us to refresh
 	//  * We haven't fetched it before
-	//  * The user wants the list of favorites.  TODO: use the cached list
-	//    once we have proper invalidation from the server.
-	if req.refresh || f.cache == nil || req.favs != nil {
+	if req.refresh || f.cache == nil {
 		folders, err := kbpki.FavoriteList(req.ctx)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


libkbfs/favorites.go
```diff
@@ -340,8 +338,9 @@ func (f *Favorites) RefreshCache(ctx context.Context) {
 	}
 }
 
-// Get returns the logged-in users list of favorites. It
-// doesn't use the cache.
+// Get returns the logged-in user's list of favorites. It uses the cache.
+// TODO: decide on behavior if we've gotten an invalidation but not an updated
```
strib: What are the choices here?  If we've gotten an invalidation, I think the `Get` should block on fetching the favorites (which is what should be happening in this PR if I'm reading the code right.)

jakob223:
That's what we're currently doing - I was just concerned that we don't want
to block on fetching the favorites if we're offline - there's a situation
whereby we get invalidated and then get offline before the new favorites
arrive and then can't get the favorites at all. But it's a pretty unlikely
event.

jakob223:
Oops, when I said "forthcoming" I really meant "there's another PR for that that I've now updated.

The newest commit invalidates the favorites cache after a week, for reasons described here: https://github.com/keybase/client/pull/14554

libkbfs/favorites.go
```diff
@@ -120,15 +123,16 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
-	if req.refresh || f.cache == nil || req.favs != nil {
+	//  * It's stale
+	if req.refresh || f.cache == nil || time.Now().After(f.cacheExpireTime) {
 		folders, err := kbpki.FavoriteList(req.ctx)
 		if err != nil {
 			return err
 		}
 
 		f.cache = make(map[Favorite]bool)
+		f.cacheExpireTime = libkb.ForceWallClock(time.Now()).Add(time.
```
strib: Please make the expire time interval into a global const in this file, for easier tweaking in the future.

libkbfs/favorites.go
strib: > That's what we're currently doing - I was just concerned that we don't want to block on fetching the favorites if we're offline - there's a situation whereby we get invalidated and then get offline before the new favorites arrive and then can't get the favorites at all. But it's a pretty unlikely event.

In this case, the `FavoriteList` call will just fail, and the cache will remain in place.  Maybe in that case, when `req.refresh == true`, you can just set the expiration time to a few minutes in the future?  (And then get rid of this TODO.)

PR #1871, opened 2018-10-23
folder_branch_ops: commit MDs to disk after reading/prefetching

strib:
`mdserverRemote` now reads the latest TLF revision from the disk when possible, and stages commits that are get/put from the remote server.

FBO nows commits staged MDs for synced TLFs to disk once one of the following conditions is met:
  1) The new revision is fully prefetched.
  2) The new revision has been read by the user.

In addition, FBO won't even show a downloaded revision to the user until at least the root block for the revision has been prefetched.

All of this is to ensure that a synced TLF contains at least some data if the user suddenly loses connectivity, or restarts in offline mode.

Also fix a few bugs, and initialize the disk MD cache on startup.

Issue: KBFS-3505

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -1430,7 +1430,7 @@ func (c *ConfigLocal) MakeDiskMDCacheIfNotExists() error {
 	if c.diskMDCache != nil {
 		return nil
 	}
-	return c.resetDiskBlockCacheLocked()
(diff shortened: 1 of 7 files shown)


strib:
[Note: don't review yet, I'm still writing tests.]

strib:
Closing, wrong branch name.  See #1872 instead.

PR #1873, opened 2018-10-24
disk_quota_cache: cache quotas to disk, and use when needed

strib:
This introduces a new disk-based cache for quota infos (keyed by `UserOrTeamID`).  The eventually-consistent quota fetcher uses the disk-cached value, if the current quota can't be fetched from the server within a small amount of time (500ms).

Issue: KBFS-3506

### DIFF
--- a/libkbfs/quota_usage.go
+++ b/libkbfs/quota_usage.go
@@ -15,6 +15,13 @@ import (
 	"golang.org/x/net/context"
 )
 
+const (
+	// quotaServerTimeoutWhenCached defines how long we wait for the
+	// quota usage to return from the server, when we've already read
+	// it from the disk cache.
+	quotaServerTimeoutWhenCached = 500 * time.Millisecond
+)
+
 // ECQUCtxTagKey is the type for unique ECQU background operation IDs.
 type ECQUCtxTagKey struct{}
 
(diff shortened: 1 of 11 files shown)


libkbfs/quota_usage.go
```diff
@@ -70,18 +77,69 @@ func NewEventuallyConsistentTeamQuotaUsage(
...
+	case nil:
+	case context.DeadlineExceeded:
+		if quotaInfoFromCache != nil {
+			q.log.CDebugf(ctx, "Can't contact server; using cached quota")
+			quotaInfo = quotaInfoFromCache
+			doCacheToDisk = false
+		} else {
+			return err
+		}
+	default:
```
songgao: In offline mode, we'll probably hit this `default` case right? Seems we shouldn't return error in that case.

libkbfs/quota_usage.go
```diff
@@ -70,18 +77,69 @@ func NewEventuallyConsistentTeamQuotaUsage(
...
+	dqc := q.config.DiskQuotaCache()
+	if dqc != nil {
+		qi, err := dqc.Get(ctx, id)
+		if err == nil {
+			q.log.CDebugf(ctx, "Read quota for %s from disk cache", id)
+			quotaInfoFromCache = &qi
+			var cancel context.CancelFunc
+			getCtx, cancel = context.WithTimeout(
+				ctx, quotaServerTimeoutWhenCached)
+			defer cancel()
```
songgao: 500ms is pretty short. Would it be better to let it keep getting it in background, until a longer timeout?

libkbfs/quota_usage.go
strib: I don't think so, we should hit the `DeadlineExceeded` case in offline mode.  Remember the RPC layer just waits for the connection to come up, it doesn't return any other TCP disconnected errors.

libkbfs/quota_usage.go
strib: I couldn't figure out an easy way to keep the original request going after a timeout, so instead commit 96453b4 launches a new background fetch with no timeout and updates the cache if it returns safely.  Please give it a review!

strib:
Ping @songgao on a review of 96453b4 when you get a chance, thanks!

strib:
Still looking for a review of aee5e6e (newly rebased commit), thanks!

songgao:
Oops sorry I totally missed these emails. Looking now!

PR #1874, opened 2018-10-26
timeout in kbfs before service

songgao:


songgao:
Sigh; looks like this might have introduced flakiness in tests on Windows

songgao:
But the failed tests seem unrelated to SimpleFS. Hmm ...

songgao:
OK merging!

PR #1875, opened 2018-10-26
kbfs_ops: share a quota usage instance between non-team TLFs

strib:
So they don't each have to call the bserver it get the same info when
filling in status objects.

Also, allow quotas to be cached for 10s.  Not sure why we previously insisted on hitting the server every time, but when the user has a lot of non-team TLFs we should be able to share a single cached value between all of them.

Issue: KBFS-3558

### DIFF
--- a/libkbfs/folder_branch_status.go
+++ b/libkbfs/folder_branch_status.go
@@ -93,12 +93,14 @@ type folderBranchStatusKeeper struct {
 }
 
 func newFolderBranchStatusKeeper(
-	config Config, nodeCache NodeCache) *folderBranchStatusKeeper {
+	config Config, nodeCache NodeCache,
+	quotaUsage *EventuallyConsistentQuotaUsage) *folderBranchStatusKeeper {
 	return &folderBranchStatusKeeper{
 		config:     config,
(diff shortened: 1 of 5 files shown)


libkbfs/folder_branch_status.go
```diff
@@ -240,7 +242,8 @@ func (fbsk *folderBranchStatusKeeper) getStatusWithoutJournaling(
 		}
 		_, usageBytes, archiveBytes, limitBytes,
 			gitUsageBytes, gitArchiveBytes, gitLimitBytes, quErr :=
-			fbsk.quotaUsage.GetAllTypes(ctx, 0, 0)
+			fbsk.quotaUsage.GetAllTypes(
+				ctx, quotaUsageStaleTolerance/2, quotaUsageStaleTolerance)
```
songgao: There's a TODO above this at line 233; seems we could remove that now? Also it'd be nice to have a comment explain that for non-team TLFs we are passing in a `quotaUsage` object.

strib: The TODO comment still applies because it's talking about the team TLFs.  I'll make it better and include your suggestion.

PR #1876, opened 2018-10-27
disk_block_cache: add a db for tracking the last unref revision

strib:
The sync cache will keep track of the last revision for which all the unreferenced blocks have been deleted from the disk cache.  A future PR will introduce a background goroutine that uses this to clean the cache, without relying on being online to process MD updates in real time.

Issue: KBFS-3507

### DIFF
--- a/libkbfs/disk_block_cache.go
+++ b/libkbfs/disk_block_cache.go
@@ -222,7 +222,7 @@ func newDiskBlockCacheLocalFromStorage(
 	// cache will block until this is done. The log will contain the beginning
 	// and end of this sync.
 	go func() {
-		err := cache.syncBlockCountsFromDb()
+		err := cache.syncBlockCountsAndUnrefsFromDb()
 		if err != nil {
 			close(startErrCh)
 			closer()
@@ -340,9 +340,9 @@ func (cache *DiskBlockCacheLocal) encodeLastUnref(rev kbfsmd.Revision) (
 	return cache.config.Codec().Encode(&entry)
 }
 
-func (cache *DiskBlockCacheLocal) syncBlockCountsFromDb() error {
-	cache.log.Debug("+ syncBlockCountsFromDb begin")
-	defer cache.log.Debug("- syncBlockCountsFromDb end")
+func (cache *DiskBlockCacheLocal) syncBlockCountsAndUnrefsFromDb() error {
+	cache.log.Debug("+ syncBlockCountsAndUnrefsFromDb begin")
+	defer cache.log.Debug("- syncBlockCountsAndUnrefsFromDb end")
 	// We take a write lock for this to prevent any reads from happening while
 	// we're syncing the block counts.
 	cache.lock.Lock()
@@ -370,6 +370,8 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsFromDb() error {
 	cache.numBlocks = numBlocks
 	cache.tlfSizes = tlfSizes
 	cache.currBytes = totalSize
+
+	cache.log.Debug("+ syncBlockCountsAndUnrefsFromDb block counts done")
 
 	tlfLastUnrefs := make(map[tlf.ID]kbfsmd.Revision)
 	lastUnrefIter := cache.lastUnrefDb.NewIterator(nil, nil)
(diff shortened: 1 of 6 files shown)


libkbfs/disk_block_cache.go
```diff
@@ -320,6 +370,25 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsFromDb() error {
...
+			return err
+		}
+
+		rev, err := cache.decodeLastUnref(lastUnrefIter.Value())
+		if err != nil {
+			return err
+		}
+		tlfLastUnrefs[tlfID] = rev
+	}
+	cache.tlfLastUnrefs = tlfLastUnrefs
```
songgao: Do we wanna rename this function not that it syncs the Unref MDs too?

libkbfs/disk_block_cache.go
```diff
@@ -320,6 +370,27 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsFromDb() error {
 	cache.numBlocks = numBlocks
 	cache.tlfSizes = tlfSizes
 	cache.currBytes = totalSize
+
+	cache.log.Debug("+ syncBlockCountsAndUnrefsFromDb block counts done")
```
jzila: Shouldn't this be `| syncBlockCounts...`?

libkbfs/disk_block_cache.go
```diff
@@ -851,6 +922,55 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
...
+	if err != nil {
+		return kbfsmd.RevisionUninitialized, err
+	}
+
+	rev, ok := cache.tlfLastUnrefs[tlfID]
+	if !ok {
+		// No known unref'd revision.
+		return kbfsmd.RevisionUninitialized, nil
+	}
+	return rev, nil
```
jzila: Did you purposely not check the underlying disk cache since it's synced to memory?

libkbfs/disk_block_cache.go
```diff
@@ -851,6 +922,55 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
 	return cache.evictSomeBlocks(ctx, numBlocks, blockIDs)
 }
 
+// GetLastUnrefRev implements the DiskBlockCache interface for
+// DiskBlockCacheLocal.
+func (cache *DiskBlockCacheLocal) GetLastUnrefRev(
+	ctx context.Context, tlfID tlf.ID) (kbfsmd.Revision, error) {
+	cache.lock.Lock()
+	defer cache.lock.Unlock()
```
jzila: `RLock`/`RUnlock`

libkbfs/disk_block_cache_remote.go
```diff
@@ -124,6 +125,20 @@ func (dbcr *DiskBlockCacheRemote) UpdateMetadata(ctx context.Context,
 		})
 }
 
+// GetLastUnrefRev implements the DiskBlockCache interface for
+// DiskBlockCacheRemote.
+func (dbcr *DiskBlockCacheRemote) GetLastUnrefRev(
+	_ context.Context, _ tlf.ID) (kbfsmd.Revision, error) {
+	panic("GetLastUnrefRev() not implemented in DiskBlockCacheRemote")
```
jzila: I know it's not in use yet in the actual code, but is there a chance that minimal mode could hit this? If so, this should be an error instead of a panic.

libkbfs/disk_block_cache.go
strib: I don't know what that syntax means,I was just copy-pasting.  But happy to change it if there's some meaning behind it.

libkbfs/disk_block_cache.go
strib: Yeah exactly.  No reason to go to disk.

libkbfs/disk_block_cache_remote.go
strib: There is zero chance a remote user will intentionally hit this, so better to panic in my mind.  (Was following the `Status()` function.)

libkbfs/disk_block_cache.go
jzila: The client applies this more strictly, but `+` refers to a method starting, and `-` refers to a method ending. `|` refers to it continuing, but that's not used as strictly.

libkbfs/disk_block_cache_remote.go
jzila: Cool thx.

PR #1878, opened 2018-10-29
add CORS and custom error pages to .kbp_config

songgao:
These are per-path settings, just like the ACLs. So rename `acls` to `per_path_configs`, and just add them there.

Both commits are independently review-able. The first commit is just the rename. The second commit adds the new fields to the config.

Note that this doesn't actually do anything for those fields. I'll have another PR up (which should be much smaller than this) for actually using them when I get more time. Will hold off merging this until that one is up and reviewed.

Thanks for looking!

cc @buoyad

### DIFF
--- a/libpages/config/config_v1.go
+++ b/libpages/config/config_v1.go
@@ -18,10 +18,10 @@ import (
 // V1 defines a V1 config. Public fields are accessible by `json`
 // encoders and decoder.
 //
-// On first call to GetPermission* methods, it initializes an internal ACL
-// checker. If the object is constructed from ParseConfig, its internal ACL
-// checker is initialized automatically. Any changes to the ACL fields
-// afterwards have no effect.
+// On first call to GetPermission* methods, it initializes an internal per-path
+// config reader. If the object is constructed from ParseConfig, its internal
+// per-path config reader is initialized automatically. Any changes to the
+// PerPathConfigs fields afterwards have no effect.
 type V1 struct {
 	Common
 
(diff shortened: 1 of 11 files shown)


libpages/config/per_path_configs_v1.go
```diff
@@ -19,15 +19,27 @@ const (
...
+	// AccessControlAllowOrigin, if set, causes the setting the
+	// Access-Control-Allow-Origin header when serving requests under the
+	// corresponding path.
+	AccessControlAllowOrigin string `json:"Access-Control-Allow-Origin,omitempty"`
+	// Custom403Forbidden specifies a path (relative to site root) to a html
+	// file to be served when 403 errors happen.
+	Custom403Forbidden string `json:"custom_403_forbidden,omitempty"`
+	// Custom404NotFound specifies a path (relative to site root) to a html
+	// file to be served when 404 errors happen.
+	Custom404NotFound string `json:"custom_404_not_found,omitempty"`
```
songgao: These are the three added fields in per-path config.

libpages/config/config_v1.go
```diff
@@ -130,15 +156,19 @@ func (c *V1) Encode(w io.Writer, prettify bool) error {
 // Validate checks all public fields of c, and returns an error if any of them
 // is invalid, or a nil-error if they are all valid.
 //
-// Although changes to ACL fields have no effect to ACL checkings once the
-// internal ACL checker is intialized (see comment on V1), this method still
-// checks the updated ACL feilds. So it's OK to use Validate directly on a
-// *V1 that has been modified since it was initialized.
+// Although changes to per-path config fields have no effect to per-path config
+// checkings once the internal per-path config reader is intialized (see
+// comment on V1), this method still checks the updated per-path config feilds.
```
strib: "feilds" -> "fields"

libpages/config/config_v1_test.go
```diff
@@ -348,7 +348,44 @@ func TestV1EncodeObjectKeyOrder(t *testing.T) {
...
+		Common: Common{
+			Version: Version1Str,
+		},
+		Users: map[string]string{
+			"alice": string(generateBcryptPasswordHashForTestOrBust(t, "12345")),
+		},
+		ACLs:           perPathConfigs,
+		PerPathConfigs: perPathConfigs,
+	}).EnsureInit()
+	require.Error(t, err)
```
strib: Redundant with the check below?

libpages/config/per_path_configs_v1.go
```diff
@@ -71,22 +83,48 @@ type accessControlV1 struct {
...
+	}
+	return cleaned, nil
+}
+
+func checkCustomPagePath(p string) (cleaned string, err error) {
+	if len(p) == 0 {
+		return "", nil
+	}
+	cleaned = path.Clean(p)
+	if strings.HasPrefix(cleaned, "..") || !strings.HasSuffix(cleaned, ".html") {
```
strib: Is the suffix restriction needed?  Why not allow `.txt` suffixes, or other static file types?

jzila:
Closing for archival.

PR #1879, opened 2018-10-29
reporter: suppress "user not found" errors

strib:
Now that we show errors again, we've gotten some complaints about shell trickery that triggers these errors by looking up certain files along the path of the current working directory.  We've been getting along fine without this type of errors for a while, so suppress them again.

### DIFF
--- a/libkbfs/reporter_kbpki.go
+++ b/libkbfs/reporter_kbpki.go
@@ -7,7 +7,6 @@ package libkbfs
 import (
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/keybase/client/go/logger"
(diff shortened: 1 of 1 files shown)


songgao:
Could you explain how `external` is used? A quick grep and look-around didn't reveal anything for me.

strib:
As far as I know it isn't used.  It was probably just there for informational purposes.  It was just the format suggested by coyne and nojima starting here (even before SBS was supported for real, I think):

https://keybase.atlassian.net/browse/KBFS-598?focusedCommentId=25091&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25091

strib:
> Out of curiosity though, what was the @: check in the removed code?

It's checking for unresolved SBS names, which contain either a `@` or a `:`.

PR #1880, opened 2018-10-29
libkbfs: set default sync block cache size to 100%

jakob223:

### DIFF
--- a/libkbfs/backpressure_disk_limiter.go
+++ b/libkbfs/backpressure_disk_limiter.go
@@ -877,7 +877,9 @@ func (bdl *backpressureDiskLimiter) onSimpleByteTrackerEnable(ctx context.Contex
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
 	bdl.overallByteTracker.onEnable(diskCacheBytes)
-	tracker.onEnable(diskCacheBytes)
+	if typ != syncCacheLimitTrackerType {
+		tracker.onEnable(diskCacheBytes)
+	}
 }
 
 func (bdl *backpressureDiskLimiter) onSimpleByteTrackerDisable(ctx context.Context,
@@ -889,7 +891,9 @@ func (bdl *backpressureDiskLimiter) onSimpleByteTrackerDisable(ctx context.Conte
 	bdl.lock.Lock()
 	defer bdl.lock.Unlock()
 	tracker.onDisable(diskCacheBytes)
-	bdl.overallByteTracker.onDisable(diskCacheBytes)
+	if typ != syncCacheLimitTrackerType {
+		bdl.overallByteTracker.onDisable(diskCacheBytes)
+	}
 }
 
 func (bdl *backpressureDiskLimiter) getDelayLocked(
@@ -1000,7 +1004,9 @@ func (bdl *backpressureDiskLimiter) commitOrRollback(ctx context.Context,
 		}
 		tracker.commitOrRollback(blockBytes, shouldCommit)
 	}
-	bdl.overallByteTracker.commitOrRollback(blockBytes, shouldCommit)
+	if typ != syncCacheLimitTrackerType {
+		bdl.overallByteTracker.commitOrRollback(blockBytes, shouldCommit)
+	}
 }
 
 func (bdl *backpressureDiskLimiter) onBlocksFlush(
@@ -1024,7 +1030,9 @@ func (bdl *backpressureDiskLimiter) release(ctx context.Context,
 		}
 		tracker.release(blockBytes)
 	}
-	bdl.overallByteTracker.release(blockBytes)
+	if typ != syncCacheLimitTrackerType {
+		bdl.overallByteTracker.release(blockBytes)
+	}
 }
 
 func (bdl *backpressureDiskLimiter) reserveBytes(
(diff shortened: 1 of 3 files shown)


jakob223:
I'm having trouble grokking the CI failure.

strib:
> I'm having trouble grokking the CI failure.

Hmm just seems like a random, unrelated appveyor issue.  I restarted it for you.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -1053,10 +1053,9 @@ func (bdl *backpressureDiskLimiter) reserveBytes(
 	if count < 0 {
 		return count, nil
 	}
-	// We calculate the total free bytes by adding the reported free bytes and
-	// the non-`tracker` used bytes.
-	tracker.updateFree(freeBytes + bdl.overallByteTracker.used -
-		tracker.usedResources())
+	// TODO: verify there's not any other kind of bytes we care about
+	// subtracting too.
+	tracker.updateFree(freeBytes)
```
jzila: Are you sure you want to do this? This will affect the working set block cache too, such that it'll now be tracking global free space. It throws off the fraction calculation.

jakob223: I think so?

My intent is to let the sync block cache take over the entire disk if it wants to and starve the working set cache, as discussed on https://keybase.atlassian.net/browse/KBFS-3517?focusedCommentId=59212&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-59212

But I am not confident that I fully understand all of the block cache machineries.

jakob223: maybe it would be better to do it the old way if the disk percentages add up to less than 100%, and this way if they add up to more.

jakob223:
seems like CI failed in the same way as before..

libkbfs/backpressure_disk_limiter.go
strib: I would have expected to have seen something that finds the places that updates the `overallByteTracker`, and NOT update it when the type of cached bytes is for the sync cache.

strib: Actually, I guess you could get rid of `overallByteTracker` completely, and just rely on the working set cache in those places that use it?

jzila: Hmm. I was hoping to keep the `overallByteTracker` code in place, in case we ever get other disk caches that need to compete for space.

jzila: I'll think about it and write back here.

strib: Fine with me.  Then the first thing I said.  The sync cache shouldn't update the overall tracker if it shouldn't count towards the total bytes used by KBFS.

jakob223: @jzila are you still thinking about it or should I go with strib's most recent suggestion?

jakob223:
Maybe we're accidentally filling up the entire disk and this is causing strange problems?

strib:
> Maybe we're accidentally filling up the entire disk and this is causing strange problems?

As far as I know, none of the block caches should be enabled for these dokan tests.  On top of that, for the sync block cache, it would have to explicitly opt-in to storing blocks in the cache.  I guess it's possible the journal is filling up the disk due to your changes to the tracking algorithm?

Let's just wait for the next version based on the feedback, and see if that helps.

libkbfs/backpressure_disk_limiter.go
jzila: Go ahead with strib's most recent suggestion.

libkbfs/backpressure_disk_limiter.go
```diff
@@ -1049,17 +1053,23 @@ func (bdl *backpressureDiskLimiter) reserveBytes(
...
+		count := bdl.overallByteTracker.tryReserve(blockBytes)
+		if count < 0 {
+			return count, nil
+		}
+		// We calculate the total free bytes by adding the reported free bytes and
+		// the non-`tracker` used bytes.
+		tracker.updateFree(freeBytes + bdl.overallByteTracker.used -
+			tracker.usedResources())
+	} else {
+		// TODO: this allows the sync cache to take up 100% of free space
```
strib: Is this really a "TODO"?  Seems like an explanatory comment.

jakob223: lol, that would be me spacing out when deleting a TODO and accidentally keeping the TODO part of the comment and simply replacing its contents with the following comment.

jakob223:
(new commits are just rebasing on top of master; will merge if they pass CI)

jakob223:
Tests pass locally... most of the time? `testJournalCoalescingConflictingCreates` failed once but when I run it alone it passes. Then I reran all the tests twice and they passed. Jenkins results seem to have TONS of goroutines for some reason?

strib:
I'm having a bunch of timeout issues on another PR too.  I wonder if something's going on right now where the CI machines are oversubscribed?

And yeah, looks like there's a goroutine leak somewhere, I'll look into it when I have time.

jakob223:
Tests passing locally, failing on CI again..

strib:
Hmm that new failure looks like a legit test flake (unrelated to the PR) that I haven't seen before.  I'm going to file a ticket for it and retry your CI run on appveyor.

strib:
Sigh looks like another random error. Trying to figure out how it could be related, or if just random.  One thing I noticed (that I think must be unrelated to the error) is that we could not be calling `overallByteTracker.onEnable()` or `overallByteTracker.onDisable()` if the type is the sync cache (lines 879 and 891).  Similarly, there's the `overallByteTracker.reserve()` call at line 983.

Again, these are likely unrelated since these tests don't even use the sync cache.  I'll keep digging and try another CI run.

jakob223:
Line 983 I didn't modify because that function is only ever called for the
journal tracker.

The others I didn't think to since I assumed the sync cache would only ever
be enabled if the other caches were also enabled - do you think that might
change?

strib:
> Line 983 I didn't modify because that function is only ever called for the
journal tracker.

Ah, I see now it errors out if called with the wrong type.  Perfect, thanks!

> The others I didn't think to since I assumed the sync cache would only ever
be enabled if the other caches were also enabled - do you think that might
change?

I don't follow the logic; why does it matter if other caches are enabled?  Those functions add or subtract the cache's bytes to the overall sum, which is exactly what we don't want for the sync cache.

jakob223:
I misunderstood what those methods do / are used for. Will push a fix.

PR #1882, opened 2018-10-30
kbfsgit: resolve sym refs and use dest ref names in ref data map

strib:
When we construct the ref-data-by-name map to `libgit.UpdateRepoMD`, we were erroring out when trying to look up the hash corresponding to symbolic references like "HEAD".  We should resolve those before we look up the hash.

Also, we were putting the _source_ reference name as the key to the map, rather than the dest name.  That meant if someone pushed from a local branch directly to master, that would not be recorded as a master push in the server's DB.  Fix that by always using the destination ref name instead.

Issue: keybase/client#14423

### DIFF
--- a/kbfsgit/runner.go
+++ b/kbfsgit/runner.go
@@ -1436,13 +1436,7 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
 }
 
 func dstNameFromRefString(refStr string) plumbing.ReferenceName {
-	start := strings.Index(refStr, ":") + 1
-	dst := refStr[start:]
-	return plumbing.ReferenceName(dst)
-}
-
-func dstNameFromRefSpec(refspec gogitcfg.RefSpec) plumbing.ReferenceName {
-	return dstNameFromRefString(refspec.String())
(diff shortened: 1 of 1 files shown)


kbfsgit/runner.go
```diff
@@ -1435,6 +1435,16 @@ func (r *runner) pushAll(ctx context.Context, fs *libfs.FS) (err error) {
...
 }
 
+func dstNameFromRefString(refStr string) plumbing.ReferenceName {
+	start := strings.Index(refStr, ":") + 1
+	dst := refStr[start:]
+	return plumbing.ReferenceName(dst)
+}
+
+func dstNameFromRefSpec(refspec gogitcfg.RefSpec) plumbing.ReferenceName {
+	return dstNameFromRefString(refspec.String())
```
jzila: `return refspec.Dst("")`, after which you don't need the above.

strib: Hmm good point.  I wonder why I had it fully implemented before.  Maybe I was worried about the wildcard behavior?  But I don't think the remote helper gets wildcard specs over the local protocol anymore, so it's probably ok to do it the way you suggested.

PR #1885, opened 2018-10-30
update vendored go-winio

zanderz:
to include our second PR against microsoft/go-winio: Microsoft/go-winio#101

### DIFF
--- a/vendor/github.com/keybase/go-winio/file.go
+++ b/vendor/github.com/keybase/go-winio/file.go
@@ -10,12 +10,55 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	"unsafe"
 )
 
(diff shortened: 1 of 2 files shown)


zanderz:
Isn't it? Just not on master? I have taken the approach of a 3rd branch, since we have 2 PRs against microsoft/go-winio
https://github.com/keybase/go-winio/commit/41191563760e6b6b3374d77fc0a3b1cb58094433

strib:
Sorry, yes I meant it's not in master.  We prefer only checking in master commits here, since the branches will be deleted eventually.  Is there a reason you can't merge that repo first, and then vendor here?

zanderz:
ok @strib it vendors a hash from master. Also verified this client no longer crashes for the user with the conflicting AV installed.

PR #1886, opened 2018-10-30
test: clean up some goroutine leaks

strib:
* Properly cancel FUSE background context in DSL tests; the parent context wasn't getting canceled.

* state_checker: when checking GC last data, don't create new ops. Because this happens during shut down, the new ops will never be cleaned up!  We only care about ones that actually ran GC here, so no need to create new ones.

cc: @jakob223 these are a few leaks I found.  Maybe there are more, but this should be a good start.  (I don't think they were related to the timeouts we saw earlier though.)

### DIFF
--- a/test/run_fuse_test.go
+++ b/test/run_fuse_test.go
@@ -59,14 +59,20 @@ func createUserFuse(tb testing.TB, ith int, config *libkbfs.ConfigLocal,
 	}
 	tb.Logf("FUSE HasInvalidate=%v", mnt.Conn.Protocol().HasInvalidate())
 
-	ctx := libkbfs.BackgroundContextWithCancellationDelayer()
+	ctx, cancelFn := context.WithCancel(context.Background())
+	ctx, err = libkbfs.NewContextWithCancellationDelayer(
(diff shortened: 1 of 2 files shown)


test/run_fuse_test.go
```diff
@@ -59,15 +59,21 @@ func createUserFuse(tb testing.TB, ith int, config *libkbfs.ConfigLocal,
 	}
 	tb.Logf("FUSE HasInvalidate=%v", mnt.Conn.Protocol().HasInvalidate())
 
-	ctx := libkbfs.BackgroundContextWithCancellationDelayer()
+	ctx, cancelFn := context.WithCancel(context.Background())
```
jzila: Is this canceled anywhere? Looks like it could leak...

strib: @jzila Yep, it's canceled here: https://github.com/keybase/kbfs/blob/f5d4ec2036bb405200148d293a06d5fc74e7adea/test/engine_fs_test.go#L57

(Via the test shutdown.)

jzila: Thanks!

PR #1892, opened 2018-11-01
kbfscrypto: add a cause to a few DecryptionErrors

strib:
I don't think this was causing any actual errors, but @maxtaco says it's a bug not to set the `Cause` on a `DecryptionError`.

Issue: KBFS-3588


jzila:
Oops approved too quickly. I think we need to change tests to check for the type of error rather than equality to `DecryptionError{}`.

strib:
Oops, I only ran the `kbfscrypto` tests.  Will fix the `libkbfs` tests, thanks!

Issue #1888, opened 2018-10-31
keybase chat commands get slower and slower


mortenkjarulff:
Attached script do:

```
while true ; do

  sleep 15

  keybase chat send ...

  ls /keybase/team/...

  echo '{"method": "read", ... | keybase chat api ...

  if 6 hours has gone
    run_keybase
  fi

done
```

After a while, `chat send` and `chat read` gets slower and slower, going from less that 1 second to more than 60 seconds.
`sleep` and `ls` times seems constant.
When 6 hours has gone and the `run_keybase` is executed, times get low again (but build up over the next 6 hours).

If I outcomment the `ls` command, all times are constant low.

These logs are from the same run, WITH the `ls` command:
  a4f72ae29df5ba7e9a4c2a1c
  0e3959537531bc60a74ac21c

These logs are from the same run, WITHOUT the `ls` command:
  41c5342a84ccee599571831c
  b7cc0500989c6a67920c0a1c

/Morten

mortenkjarulff:
[test.sh.txt](https://github.com/keybase/kbfs/files/2534988/test.sh.txt)

strib:
@mortenkjarulff  did you mean to file this in keybase/client?  Chat is pretty much completely separate from kbfs at this point.

mortenkjarulff:
Hi,

Appears not to be so separate 😀

Shall I file it there instead?

/Morten

strib:
Oh, I guess I glossed over or didn't understand the line about "outcomment the ls".

Ok, here's what appears to be the issue.  The team folder you're `ls`ing has been written to by a device that has since been revoked.  When you `ls` it, that triggers kbfs to attempt to "garbage-collect" the old data in it, by iterating through the old revisions of the folder.  When it looks at the revisions that were written by the now-revoked device, it triggers a bunch of extra checking to verify that the device was valid at the time of the revision.

It looks like, in the case of this folder, the revisions by that revoked device are so expensive that they are slowing down chat.

Can you just `ls` that folder once, and maybe let the garbage collection run overnight and see if it finishes?  You can tell if it made progress overnight by looking at the `LastGCRevision` field in the team folder's `.kbfs_status` file.

If it doesn't make any progress, even after many hours of work, I'll try to figure out a custom solution for you.

And obviously we need to figure out a way to reduce these extra checks without affecting security...

mortenkjarulff:
Not sure I understand. I can reproduce at any time, even long time after I revoked.

[...]

PR #1893, opened 2018-11-01
md_ops: add caching to improve validation of MDs written by revoked devices

strib:
In #1888, a user is doing way more work than necessary to validate a bunch of MDs written by a revoked device, during garbage collection.  We can improve that a small amount of caching:

This PR:
* Moves verification of KBFS Merkle nodes into the mdserver remote, so they can be cached more effectively.
* Adds a limited size LRU cache for the (validated) results of `FindNextMD`.
* Tracks the chain of MDs that have been validated up to a given revision (the one immediately following the revoke in the Merkle tree).

This should eliminate most of the redundant work that was happening when verifying lots of MDs written by the same revoked device, as in garbage collection.

Issue: KBFS-3581
Issue: #1888

### DIFF
--- a/libkbfs/md_ops.go
+++ b/libkbfs/md_ops.go
@@ -55,11 +55,25 @@ func init() {
 type MDOpsStandard struct {
 	config Config
 	log    logger.Logger
+
+	lock sync.Mutex
+	// For each TLF, maps an MD revision representing the next MD
+	// after a device revoke, with the minimum revision number that's
+	// been validated in chain up to the given MD revision.  That is,
+	// for TLF 1, if we have a next revision of 1000, and we've
+	// validated that MDs 100-1000 form a valid chain, then the map
+	// would contain: {1: {1000: 100}}
+	leafChainsValidated map[tlf.ID]map[kbfsmd.Revision]kbfsmd.Revision
 }
 
 // NewMDOpsStandard returns a new MDOpsStandard
 func NewMDOpsStandard(config Config) *MDOpsStandard {
-	return &MDOpsStandard{config, config.MakeLogger("")}
+	return &MDOpsStandard{
+		config: config,
+		log:    config.MakeLogger(""),
+		leafChainsValidated: make(
+			map[tlf.ID]map[kbfsmd.Revision]kbfsmd.Revision),
+	}
 }
 
 // convertVerifyingKeyError gives a better error when the TLF was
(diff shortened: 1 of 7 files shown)


libkbfs/md_ops.go
```diff
@@ -421,13 +456,26 @@ func (md *MDOpsStandard) checkRevisionCameBeforeMerkle(
 	}
 
 	// Otherwise it's valid, as long as there's a valid chain of MD
-	// revisions between the two, and the global root info checks out
-	// as well.  By using `ctx`, we'll avoid infinite loops in the
+	// revisions between the two.  First, see if what chain we've
```
songgao: Did you mean s/if what/what?

libkbfs/md_ops.go
```diff
@@ -55,11 +55,25 @@ func init() {
...
 	log    logger.Logger
+
+	lock sync.Mutex
+	// For each TLF, maps an MD revision representing the next MD
+	// after a device revoke, with the minimum revision number that's
+	// been validated in chain up to the given MD revision.  That is,
+	// for TLF 1, if we have a next revision of 1000, and we've
+	// validated that MDs 100-1000 form a valid chain, then the map
+	// would contain: {1: {1000: 100}}
+	leafChainsValidated map[tlf.ID]map[kbfsmd.Revision]kbfsmd.Revision
```
songgao: Is the idea that we are always checking on the revision right after a device revoke?

libkbfs/md_ops.go
```diff
@@ -421,13 +456,26 @@ func (md *MDOpsStandard) checkRevisionCameBeforeMerkle(
...
 	// Otherwise it's valid, as long as there's a valid chain of MD
-	// revisions between the two, and the global root info checks out
-	// as well.  By using `ctx`, we'll avoid infinite loops in the
+	// revisions between the two.  First, see if what chain we've
+	// already validated, and if this revision falls in that chain,
+	// we're done.  Otherwise, just fetch the part of the chain we
+	// haven't validated yet.
+	newChainEnd := md.startOfValidatedChainForLeaf(irmd.TlfID(), leaf.Revision)
+	if newChainEnd == kbfsmd.RevisionUninitialized {
+		newChainEnd = leaf.Revision
```
songgao: WDYT just have `startOfValidatedChainForLeaf` return this, so we don't have to rely on the caller to remember to check if the returned revision is `kbfsmd.RevisionUninitialized`?

libkbfs/md_ops.go
strib: > Is the idea that we are always checking on the revision right after a device revoke?

Yep.  When we see something signed by a revoked key, we ask the service what revision came back right after the revoke, and always check that one.

libkbfs/md_ops.go
strib: Done.

PR #1895, opened 2018-11-02
govendor update leveldb

joshblum:
mirrors https://github.com/keybase/client/pull/14557


joshblum:
yup! All of the changes looked like minor bugfixes

PR #1896, opened 2018-11-02
simplefs MOTW

zanderz:
copied from chat. Tested on Windows; mac is likely ok too

image


simplefs/simplefs.go
```diff
@@ -922,6 +922,10 @@ func (k *SimpleFS) doCopyFromSource(
 	}
 	defer dst.Close()
 
+	if pathType, _ := destPath.PathType(); pathType == keybase1.PathType_LOCAL {
+		defer Quarantine(ctx, destPath.Local())
```
songgao: Maybe we should set the error to what this function returns, if the error is nil? i.e. something like:

```
defer func() {
  qerr := Quarantine(ctx, ...)
  if *err == nil {
    err = qerr
  }
}(&err)
```

zanderz:
@songgao I copied the mac file straight over from client but it doesn't build, do you have any ideas?
```
+ go build --tags=production ./...

# github.com/keybase/kbfs/simplefs

Undefined symbols for architecture x86_64:

  "_NSURLQuarantinePropertiesKey", referenced from:

      _quarantineFile in _x002.o

  "_OBJC_CLASS_$_NSDictionary", referenced from:

      objc-class-ref in _x002.o

  "_OBJC_CLASS_$_NSString", referenced from:

      objc-class-ref in _x002.o

  "_OBJC_CLASS_$_NSURL", referenced from:

      objc-class-ref in _x002.o

etc.
```

PR #1897, opened 2018-11-02
macOS/fuse: set quarantine xattr for TLFs whith other writers

songgao:
Note that this breaks running *.app directly from such KBFS TLFs. I've
tried several things including modifying the
com.apple.LaunchServices.QuarantineEventsV2 sqlite DB, but I just can't
get macOS to pass the integrity check when the com.apple.quarantine
xattr is set. I think it might require setting xattr for stuff inside
the .app individually. Not sure. But this seems to work for scripts
pretty well at least. It triggers a prompt when user tries to run a
script in a TLF where we are not the only writer. If user chooses to
open it anyway, next time user tries to open it the prompt won't show.

This PR doesn't persist it yet though. In the next PR I'll have the
xattr stored in a local DB so it persists through restarts.


tlf/name.go
```diff
@@ -158,3 +158,20 @@ func CanonicalToPreferredName(username kbname.NormalizedUsername,
...
+func UserIsOnlyWriter(username kbname.NormalizedUsername, canon CanonicalName) bool {
+	tlfname := string(canon)
+	if len(username) == 0 {
+		return false
+	}
+	ws, _, _, err := SplitName(tlfname)
+	if err != nil {
+		return false
+	}
+	fmt.Printf("SONGGAO %s %s %#+v\n", username, canon, ws)
```
jzila: 🔥

libfuse/quarantine_darwin.go
```diff
@@ -0,0 +1,135 @@
...
+// Removexattr implements the fs.NodeRemovexattrer interface.
+func (h *QuarantineXattrHandler) Removexattr(ctx context.Context, req *fuse.RemovexattrRequest) error {
+	if req.Name != quarantineXattrName {
+		// The request is not about quarantine. Let the OS fallback to ._ file
+		// based method.
+		return fuse.ENOTSUP
+	}
+
+	h.bytes = nil
+	h.bytesSet = true
```
jzila: Should this be false?

songgao: I think this should be `true`, so next time Get happens, we'd return `ENOATTR`. If a syscall came in to explicitly remove that attr, we should remember it.

PR #1900, opened 2018-11-06
md_ops: don't use application keygen for team keygen

strib:
The merkle tree doesn't yet record which team keygen is used to encrypt a merkle leaf, so just use 1 as the min number and let the service scan.  In the future, we should have the server record the keygen in the merkle leaf header and use that instead.  (Note that "team keygen" is completely separate from "application keygen", so we can't just use `rmd.LatestKeyGeneration()` here.)

Issue: KBFS-3605


strib:
cc: @xgess @mmaxim

PR #1901, opened 2018-11-06
Disk-based favorites cache 

aaazalea:
Still needs
 [x] favorites cache merged
 [ ] client PR merged
 [ ] more tests

### DIFF
--- a/libkbfs/favorites.go
+++ b/libkbfs/favorites.go
@@ -5,8 +5,11 @@
 package libkbfs
 
 import (
+	"fmt"
 	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/kbfs/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -18,9 +21,22 @@ import (
 )
 
 const (
-	disableFavoritesEnvVar       = "KEYBASE_DISABLE_FAVORITES"
-	favoritesCacheExpirationTime = time.Hour * 24 * 7 // one week
+	disableFavoritesEnvVar           = "KEYBASE_DISABLE_FAVORITES"
+	favoritesCacheExpirationTime     = time.Hour * 24 * 7 // one week
+	kbfsFavoritesCacheSubfolder      = "kbfs.favorites.cache"
+	favoritesDiskCacheVersion        = 1
+	favoritesDiskCacheStorageVersion = 1
 )
+
+type errIncorrectFavoritesCacheVersion struct {
+	cache   string
+	version int
+}
+
+func (e errIncorrectFavoritesCacheVersion) Error() string {
+	return fmt.Sprintf("decoding of %s favorites cache failed: version was %d",
+		e.cache, e.version)
+}
 
 type favToAdd struct {
 	Favorite
@@ -42,6 +58,7 @@ func (f favToAdd) ToKBFolder() keybase1.Folder {
 // given ctx is used for all network operations.
 type favReq struct {
 	// Request types
+	clear   bool
 	refresh bool
 	toAdd   []favToAdd
 	toDel   []Favorite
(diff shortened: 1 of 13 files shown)


jakob223:
Client PR: https://github.com/keybase/client/pull/14608

jakob223:
Expect CircleCI here to fail until client is merged. I definitely think this should get a `kbfsdocker` test or two, not sure if any tests in this repo would be useful.

libkbfs/favorites.go
```diff
@@ -107,6 +112,87 @@ func NewFavorites(config Config) *Favorites {
 	return newFavoritesWithChan(config, make(chan *favReq, 100))
 }
 
+func (f *Favorites) readCacheFromDisk(ctx context.Context) error {
+	// Read the encrypted cache from disk
+	folder := f.config.StorageRoot()
+	subfolder := "kbfs.favorites.cache"
```
strib: This is screaming for a global constant.

libkbfs/favorites.go
```diff
@@ -107,6 +112,87 @@ func NewFavorites(config Config) *Favorites {
...
+	}
+
+	// Send the data to the service to be decrypted
+	decryptedData, err := f.config.KeybaseService().DecryptFavorites(ctx, data)
+	if err != nil {
+		return err
+	}
+
+	// Decode the data into the a map
+	decoder := gob.NewDecoder(bytes.NewBuffer(decryptedData))
```
strib: Why use gob?  As far as I know this is the only place we use this.

I think a more standard way would be to define a struct that holds the cache map, and encode that using `f.config.Codec()` (which would use msgpack). Besides being more standard, that would let us add other fields in the future.

Also it might be good to add a version number to the struct holding the map as well, in case the format (or interpretation of the format) of the cache needs to change in the future.  And after decrypting, we check that the version number matches what we expect or else we panic.

libkbfs/favorites.go
```diff
@@ -107,6 +112,87 @@ func NewFavorites(config Config) *Favorites {
...
+	folder := f.config.StorageRoot()
+	subfolder := "kbfs.favorites.cache"
+	session, err := f.config.KBPKI().GetCurrentSession(ctx)
+	if err != nil {
+		return err
+	}
+	user := string(session.UID)
+	path := filepath.Join(folder, subfolder, user)
+	// TODO: is this the right permissions
+	return ioutil.WriteFile(path, data, 0600)
```
strib: Besides using msgpack for the encoded data we send over to the service (as mentioned above), I think we probably want to also use it here for the contents of the file on disk (i.e., a struct that contains a `[]byte`).  That's because we should also add a version number to that struct also, in case we ever change the encryption scheme of the cache in the future.  And when reading the file, check the version number and panic if it's unexpected.

libkbfs/favorites.go
```diff
@@ -107,6 +112,87 @@ func NewFavorites(config Config) *Favorites {
...
+	go f.loop()
+}
+
+// Initialize loads the favorites cache from disk and starts listening for
+// requests asynchronously.
+func (f *Favorites) Initialize(ctx context.Context) {
+	// load cache from disk
+	err := f.readCacheFromDisk(ctx)
+	if err != nil {
+		f.config.MakeLogger("favorites").CDebugf(nil,
```
strib: Maybe `CWarningf` instead?

libkbfs/favorites.go
```diff
@@ -125,15 +211,18 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
 	//  * The user asked us to refresh
 	//  * We haven't fetched it before
 	//  * It's stale
-	if req.refresh || f.cache == nil || time.Now().After(f.cacheExpireTime) {
+	if req.refresh || f.cache == nil || time.Now().After(f.
```
strib: Please split the line after the last `(`.

strib: Please use `f.config.Clock().Now()` instead of `time.Now()`.  (I missed that in an unreviewed update to an old PR, sorry.)

libkbfs/favorites.go
```diff
@@ -125,15 +211,18 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
 		folders, err := kbpki.FavoriteList(req.ctx)
 		if err != nil {
 			if req.refresh {
 				// if we're supposed to refresh the cache and it's not
-				// working, try again in a few minutes.
-				in5Minutes := time.Now().Add(5 * time.Minute)
-				if in5Minutes.After(f.cacheExpireTime) {
-					f.cacheExpireTime = in5Minutes
+				// working, try again in a minute.
+				in1Minute := time.Now().Add(time.Minute)
```
strib: Why the change here?  Mostly just curious.

libkbfs/favorites.go
jakob223: We do expect this to happen on the first login of each user; does that qualify for a warning?

libkbfs/favorites.go
jakob223: I don't remember. I think one minute seems more reasonable to me if the reason for failure is that the computer is starting up and network doesn't exist yet.

libkbfs/favorites.go
strib: You mean we expect it to happen _before_ the first login of each user?  In that case, debug is fine.

libkbfs/favorites.go
```diff
@@ -107,6 +112,87 @@ func NewFavorites(config Config) *Favorites {
...
+
+	// Write the encrypted cache to disk
+	folder := f.config.StorageRoot()
+	subfolder := "kbfs.favorites.cache"
+	session, err := f.config.KBPKI().GetCurrentSession(ctx)
+	if err != nil {
+		return err
+	}
+	user := string(session.UID)
+	path := filepath.Join(folder, subfolder, user)
```
strib: Just realized that this path has to have the UID in there somewhere.  Multiple Keybase accounts should be able to cache their favorites to disk.

strib: Nevermind, I'm dumb.

libkbfs/favorites.go
```diff
@@ -107,6 +113,118 @@ func NewFavorites(config Config) *Favorites {
...
+	}
+
+	// decode the data from the file and ensure its version is correct
+	var decodedData favoritesCacheEncryptedForDisk
+	err = f.config.Codec().Decode(data, &decodedData)
+	if err != nil {
+		return err
+	}
+	if decodedData.version != favoritesDiskCacheStorageVersion {
+		panic("Favorites cache storage version is wrong.")
```
strib: I think I changed my mind about panicking here (and below).  Just return an error.

Oh, and you might as well include the unexpected version number in the error string.

libkbfs/favorites.go
```diff
@@ -125,15 +243,18 @@ func (f *Favorites) handleReq(req *favReq) (err error) {
...
 		folders, err := kbpki.FavoriteList(req.ctx)
 		if err != nil {
 			if req.refresh {
 				// if we're supposed to refresh the cache and it's not
-				// working, try again in a few minutes.
-				in5Minutes := time.Now().Add(5 * time.Minute)
-				if in5Minutes.After(f.cacheExpireTime) {
-					f.cacheExpireTime = in5Minutes
+				// working, try again in a minute.
+				in1Minute := f.config.Clock().Now().Add(time.Minute)
```
strib: Please also update the `time.Now()` in the `ForceWallClock` call below.

PR #1902, opened 2018-11-08
tlf_history: treat files to ignore as prefixes

strib:
Otherwise, conflicted copies will show up.

Issue: KBFS-3609

### DIFF
--- a/kbfsedits/tlf_history.go
+++ b/kbfsedits/tlf_history.go
@@ -242,7 +242,17 @@ var filesToIgnore = map[string]bool{
 
 func ignoreFile(filename string) bool {
 	_, base := path.Split(filename)
-	return filesToIgnore[base] || strings.HasPrefix(base, "._")
+	if filesToIgnore[base] || strings.HasPrefix(base, "._") {
+		return true
+	}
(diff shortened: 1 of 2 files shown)


kbfsedits/tlf_history.go
```diff
@@ -242,7 +242,17 @@ var filesToIgnore = map[string]bool{
 
 func ignoreFile(filename string) bool {
 	_, base := path.Split(filename)
-	return filesToIgnore[base] || strings.HasPrefix(base, "._")
+	if filesToIgnore[base] || strings.HasPrefix(base, "._") {
+		return true
+	}
+	// Treat the files to ignore as prefixes, since if they ever
+	// conflict they'll have the conflict suffix.
```
songgao: Is there an easy way to check for conflict suffix to avoid false positives here?

strib: We could with regular expressions, but that seems too expensive.  I can't imagine a valid file with these prefixes.

strib:
> Also I wonder if we should exclude stuff like vim .swp files too.

Let's do anything like that in separate issues/PRs.

PR #1904, opened 2018-11-12
Jakob/kbfs 3511 prefetch in simple fs stat

jakob223:
(tests will presumably fail until client code is merged)

Why does it seem like every file I check is TRIGGERED_PREFETCH?


libfs/file_info.go
```diff
@@ -82,39 +88,62 @@ type fileInfoSys struct {
...
-var _ LastWriterGetter = fileInfoSys{}
+type KBFSMetadataForSimpleFS struct {
+	LastWriter     keybase1.User
+	PrefetchStatus keybase1.PrefetchStatus
+}
+
+var _ KBFSMetadataForSimpleFSGetter = fileInfoSys{}
 
-func (fis fileInfoSys) LastWriter() (keybase1.User, error) {
+func (fis fileInfoSys) KBFSMetadataForSimpleFS() (KBFSMetadataForSimpleFS,
```
strib: Please break the line right before `KBFSMetadataForSimpleFS`.

libfs/file_info.go
```diff
@@ -82,39 +88,62 @@ type fileInfoSys struct {
...
+	default:
+		return KBFSMetadataForSimpleFS{}, ErrUnknownPrefetchStatus
+	}
+
 	lastWriterName := md.LastWriterUnverified
 	if lastWriterName == "" {
 		// This can happen in old, buggy team folders where the writer
 		// isn't properly set.  See KBFS-2939.
-		return keybase1.User{}, nil
+		return KBFSMetadataForSimpleFS{PrefetchStatus: prefetchStatus}, nil
```
strib: Might be cleaner if you do this above:
```
status := KBFSMetadataForSimpleFS{PrefetchStatus: prefetchStatus}
```
Then you can return that anywhere it returns `nil`, and just fill in the `LastWriter` field below when it's ready.

libfs/file_info.go
```diff
@@ -82,39 +88,62 @@ type fileInfoSys struct {
...
 	}
 
 	_, id, err := fis.fi.fs.config.KBPKI().Resolve(
 		fis.fi.fs.ctx, lastWriterName.String())
 	if err != nil {
-		return keybase1.User{}, err
+		return KBFSMetadataForSimpleFS{PrefetchStatus: prefetchStatus}, nil
```
strib: I think this and the next `return` should return `KBFSMetadataForSimpleFS{}, err`. Otherwise, the error is lost.  And I think it's ok to lose the prefetch status in those rare cases.

strib:
Also, rebase to master and fix the lint: https://ci.keyba.se/job/kbfs/job/PR-1904/1/execution/node/123/log/

strib:
> Why does it seem like every file I check is TRIGGERED_PREFETCH?

You have to manually enable syncing on a folder to get it to be anything but that, I think.  For example:

```sh
echo 1 > /keybase/private/jakob223/.kbfs_enable_sync
```

jzila:
@jakob223: small files should get to `FinishedPrefetch` by themselves:
```
Johns-iMac.local 16:04:56 jzila ➜ pwd
/keybase/private/jzila
Johns-iMac.local 16:05:03 jzila ➜ cat .kbfs_fileinfo_id_rsa.pub
{
  "LastWriterUnverified": "jzila",
  "BlockInfo": {
    "ID": "01d9dc3a04ffb34826aeaaa072f66515aee4ea68e1f63df11d60d42fd31782dc9c",
    "KeyGen": 6,
    "DataVer": 1,
    "DirectType": 0,
    "Creator": "69da56f622a2ac750b8e590c3658a700",
    "RefNonce": [
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0
    ],
    "BlockType": 0,
    "EncodedSize": 569
  },
  "PrefetchStatus": "FinishedPrefetch",
  "PrevRevisions": null
}
```

libfs/file_info.go
```diff
@@ -82,39 +88,62 @@ type fileInfoSys struct {
...
 	}
 	md, err := fis.fi.fs.config.KBFSOps().GetNodeMetadata(
 		fis.fi.fs.ctx, fis.fi.node)
 	if err != nil {
-		return keybase1.User{}, err
+		return KBFSMetadataForSimpleFS{}, err
 	}
+	var prefetchStatus keybase1.PrefetchStatus
+	switch md.PrefetchStatus {
+	case "NoPrefetch":
```
jzila: I wonder if we should refactor `NodeMetadata` to contain a `PrefetchStatus` instead of a string. Not a big deal though.

jakob223: Oddly, as far as I can tell, there's no existing references to prefetch status on the node metadata? Which would make that change very straightforward. Though it might cause confusion w/r/t the *other* prefetch status struct (in `data_types.go`). Maybe we should just make everything be `keybase1.PrefetchStatus`?

jzila: Existing references are when the `NodeMetadata` is output directly  in `cat /keybase/{private,public,team}/tlf/.kbfs_fileinfo_foo.txt`. But if you change it to a `libkbfs.PrefetchStatus`, it'll inherit its `String()` method, which should hopefully also output nicely.

jzila: I'd prefer to avoid using `keybase1.PrefetchStatus` in `libkbfs` code since it's string-based instead of integer-based.

PR #1906, opened 2018-11-13
macOS/quarantine: persist xattr values

songgao:
~Create a levelDb storage for local xattr values. Index them with block ID~ Create a levelDb storage for block metadata, and use that for local `com.apple.quarantine` xattr values. The `com.apple.quarantine` xattr data is index them with block ID so if they get moved around, the quarantine xattr follows them as long as content doesn't change.~

@strib and I had a short discussion offline about how to garbage collect these values and thought we could store the canonical path in levelDB too, and just check if canonical path exists the same when determining if a xattr is orphaned. I went with that approach first, but then realized we could have more than one block references for one block, so checking canonical path won't work. ~I guess having the block ID is enough. If this becomes a problem in the future, we can check with bserver to see if the block is still referenced. How does it sound?~ In the future we can probably hook this into the block deletion mechanism for the sync cache, if we need garbage collection for these block metadata.


songgao:
Ugh; untagging review now; will fix tests tomorrow before retagging you all again.

songgao:
Per discussion yesterday in the sprint call, I'm gonna rename the new levelDB into metadata instead of being xattr-specific. Untagging review for now.

libkbfs/interfaces.go
```diff
@@ -1285,6 +1295,19 @@ type DiskMDCache interface {
...
 
+// DiskBlockMetadataStore defines a type that stores xattrs.
+type DiskBlockMetadataStore interface {
+	// GetXattr looks for and returns the Xattr value of xattrType for blockID
+	// if it's found, and ldberrors.ErrNotFound if it's not found.
+	GetXattr(ctx context.Context,
+		blockID kbfsblock.ID, xattrType XattrType) ([]byte, error)
+	// SetXattr sets xattrType Xattr to xattrValue for blockID.
+	SetXattr(ctx context.Context,
+		blockID kbfsblock.ID, xattrType XattrType, xattrValue []byte) error
```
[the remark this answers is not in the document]
jzila: Why wouldn't you return a `blockMetadata` struct instead of an `interface{}` or `[]byte`? I mean ideally you wouldn't even need the `Xattr` methods at all, and you could pull them out of the `blockMetadata` struct. But I understand that you want the `Noop` shim, which is why I went with this recommendation.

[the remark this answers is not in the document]
songgao: Ah that's right; that'd work. Not sure what I was thinking.

There's one more thing though. In the "set" path, it's actually an update. With `SetXattr`, caller just need to pass in the xattr value. Otherwise, caller need to always do a `GetMetadata` first, amend the value, then call `SetMetadata`. This is what happens under-the-hood anyways, but I'd love the encapsulation that `SetXattr` provides.

Sounds like the suggested approach from your original comment works the best then. Will update; thanks!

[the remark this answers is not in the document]
songgao: Hmm actually in the "setXattr" path, getMetadata and setMetadata need to happen in the same locking session. So the presumed `SetMetadata` would probably have an interface that looks like the `setMetadata` right now. Does it sound good to you? @jzila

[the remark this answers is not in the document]
jzila: Yep, I'm fine with that.

songgao:
@jzila ready for another look; thanks!

libkbfs/disk_block_metadata_store.go
```diff
@@ -80,135 +80,146 @@ func newDiskBlockMetadataStore(
...
-}
-
-func (c *diskBlockMetadataStore) getMetadata(ctx context.Context,
-	blockID kbfsblock.ID) (value DiskBlockMetadataValue, err error) {
-	c.lock.Lock()
-	defer c.lock.Unlock()
+// GetMetadata implements the BlockMetadataStore interface.
+func (s *diskBlockMetadataStore) GetMetadata(ctx context.Context,
+	blockID kbfsblock.ID) (value BlockMetadataValue, err error) {
+	s.lock.Lock()
```
jzila: `RLock`/`RUnlock()`

libkbfs/disk_block_metadata_store.go
```diff
@@ -80,135 +80,146 @@ func newDiskBlockMetadataStore(
...
+	s.hitMeter.Shutdown()
+	s.missMeter.Shutdown()
+	s.putMeter.Shutdown()
 }
 
-// ErrDiskBlockMetadataStoreShutdown is returned when methods are called on
+// MarkMiss implements the BlockMetadataStore interface.
+func (s *diskBlockMetadataStore) MarkMiss(num int64) {
+	s.missMeter.Mark(num)
+}
```
jzila: I don't think we want this to be exported.

libkbfs/disk_block_metadata_store.go
```diff
@@ -80,135 +80,146 @@ func newDiskBlockMetadataStore(
...
-	blockMetadata, err := c.getMetadata(ctx, blockID)
+	blockMetadata, err := s.store.GetMetadata(ctx, blockID)
 	if err != nil {
 		return nil, err
 	}
 
 	v, ok := blockMetadata.Xattr[xattrType]
 	if !ok {
-		c.missMeter.Mark(1)
+		s.store.MarkMiss(1)
```
jzila: This should be tracking its own miss meter, rather than mutating the `MetadataStore` miss meter. They represent different information.

libkbfs/interfaces.go
```diff
@@ -1343,17 +1341,36 @@ type DiskQuotaCache interface {
...
+	// found.
+	GetMetadata(ctx context.Context, blockID kbfsblock.ID) (BlockMetadataValue, error)
+	// UpdateMetadata updates the block metadata for blockID using updater.
+	// Specifically, it looks for existing block metdata for blockID. If it's
+	// found, it's passed into updater. Otherwise, a zero value of
+	// BlockMetadataValue is passed into the updater. After if updater returns
+	// nil, the updated metadata is stored.
+	UpdateMetadata(ctx context.Context, blockID kbfsblock.ID, updater BlockMetadataUpdater) error
+	// MarkMiss marks miss(es) in the store's miss meter.
+	MarkMiss(num int64)
```
jzila: As above, I don't think this should be exported.

libkbfs/disk_block_metadata_store.go
songgao: GOod point!

PR #1910, opened 2018-11-14
mdserver_remote: get latest handle from cache when disconnected

strib:
This RPC is on the critical path for parsing implicit team handles, so needs to be available offline.  But in the offline case, the latest cached MD should be good enough.  The call is only there to prevent using an old TLF ID from a TLF that's already been finalized.

With this PR, the offline mdserver test gets farther when enabled, though it still doesn't pass.

Issue: KBFS-3579

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -1027,28 +1027,33 @@ func (md *MDServerRemote) GetLatestHandleForTLF(ctx context.Context, id tlf.ID) 
 	}()
 
 	handle, handleErr := md.getLatestHandleFromCache(ctx, id)
-	if !md.IsConnected() && handleErr == nil {
+	if handleErr == nil {
+		if !md.IsConnected() {
+			md.log.CDebugf(ctx,
+				"Got latest handle for %s from cache when mdserver is "+
+					"disconnected", id)
+			return handle, nil
+		}
 		md.log.CDebugf(ctx,
-			"Got latest handle for %s from cache when mdserver is "+
-				"disconnected", id)
-		return handle, nil
-	} else if handleErr == nil {
-		md.log.CDebugf(ctx,
-			"Setting a quick timeout when a cached handle is available for "+
-				"TLF %s", id)
+			"Setting a quick timeout when a cached handle is available "+
+				"for TLF %s", id)
 		var cancel context.CancelFunc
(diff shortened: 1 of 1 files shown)


libkbfs/mdserver_remote.go
```diff
@@ -1003,8 +1025,30 @@ func (md *MDServerRemote) GetLatestHandleForTLF(ctx context.Context, id tlf.ID)
...
+				"disconnected", id)
+		return handle, nil
+	} else if handleErr == nil {
+		md.log.CDebugf(ctx,
+			"Setting a quick timeout when a cached handle is available for "+
+				"TLF %s", id)
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, mdServerLatestHandleTimeout)
+		defer cancel()
+	}
```
jzila:
```go
if handleErr == nil {
	if !md.IsConnected() {
		md.log.CDebugf(ctx,
			"Got latest handle for %s from cache when mdserver is "+
				"disconnected", id)
		return handle, nil
	} else {
		md.log.CDebugf(ctx,
			"Setting a quick timeout when a cached handle is available for "+
				"TLF %s", id)
		var cancel context.CancelFunc
		ctx, cancel = context.WithTimeout(ctx, mdServerLatestHandleTimeout)
		defer cancel()
	}
}
```

libkbfs/mdserver_remote.go
```diff
@@ -1003,8 +1025,30 @@ func (md *MDServerRemote) GetLatestHandleForTLF(ctx context.Context, id tlf.ID)
...
 	buf, err := md.getClient().GetLatestFolderHandle(ctx, id.String())
 	if err != nil {
+		if err == context.DeadlineExceeded && handleErr == nil {
+			md.log.CDebugf(ctx,
+				"Got latest handle for %s from cache when mdserver can't "+
+					"be reached quickly", id)
+			return handle, nil
+		}
 		return tlf.Handle{}, err
 	}
```
jzila:
```go
switch err {
case nil:
case context.DeadlineExceeded:
	if handleErr == nil {
		md.log.CDebugf(ctx,
			"Got latest handle for %s from cache when mdserver can't "+
				"be reached quickly", id)
		return handle, nil
	}
	return tlf.Handle{}, err
default:
	return tlf.Handle{}, err
}
```

PR #1912, opened 2018-11-14
config_local: use a struct for values in the synced TLF DB

strib:
In the future, we'll need to add other fields in here to support partial syncing on a subfolder basis.

Issue: KBFS-3516

### DIFF
--- a/libkbfs/config_local.go
+++ b/libkbfs/config_local.go
@@ -1519,7 +1519,21 @@ func (c *ConfigLocal) loadSyncedTlfsLocked() (err error) {
 			deleteBatch.Delete(iter.Key())
 			continue
 		}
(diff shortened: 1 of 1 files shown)


strib:
> I guess you'll have a list of paths for each TLF?

The struct itself is per-TLF, with one per row in the DB, so yeah each struct will have a list of paths.

strib:
> Are you planning on encrypting the list of paths?

Ugggh, great question, hadn't thought about that yet.  Probably should.

jzila:
Yeah, I agree. Technically offline access means that people will have the ability to get at the data, but a `keybase logout` should still kill your ability to access any unencrypted data, even offline.

PR #1915, opened 2018-11-16
prefetcher: reschedule prefetches with backoff when disk is full

strib:
Instead of canceling prefetches when the disk is almost full, reschedule them (starting with the top-most block) using an exponential-backoff timer.  This way, the user can see a warning that the disk is full, make some space, and the prefetches will automatically resume (eventually).

If needed in the future, we can add a background process that checks for more disk space and then reschedules them immediately.  But I worry about racing with other processes needing the disk more immediately, so let's see how this works for now.

To get a reliable test, I had to move the space function into the `DiskBlockCache` interface.

@jzila this one could use your attention specifically!  I'm not 100% sure this is the best approach.

Issue: KBFS-3519

### DIFF
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -188,6 +188,11 @@ func (p *blockPrefetcher) applyToParentsRecursive(
 	for b := range pre.parents {
 		parent, ok := p.prefetches[b]
 		if !ok {
+			// Note that the parent (or some other ancestor) might be
+			// rescheduled for later and have been removed from
+			// `prefetches`.  In that case still delete it from the
+			// `parents` list as normal; the reschedule will add it
+			// back in later as needed.
 			delete(pre.parents, b)
 			continue
 		}
@@ -452,7 +457,7 @@ func (p *blockPrefetcher) handlePrefetch(pre *prefetch, isPrefetchNew,
 }
 
 func (p *blockPrefetcher) rescheduleTopBlock(
-	blockID kbfsblock.ID, pp *prefetch, req *prefetchRequest) {
+	blockID kbfsblock.ID, pp *prefetch) {
 	if len(pp.parents) > 0 {
 		p.cancelPrefetch(blockID, pp)
 		return
@@ -474,8 +479,10 @@ func (p *blockPrefetcher) rescheduleTopBlock(
 		// Prefetch already scheduled.
 		return
 	}
-	req.ptr = BlockPointer{ID: blockID}
+	// Copy the req but use a new empty block.
+	req := *pp.req
 	req.block = pp.req.block.NewEmpty()
+	pp.req = &req
 	d := rp.off.NextBackOff()
 	if d == backoff.Stop {
 		p.log.Debug("Stopping rescheduling of %s due to stopped backoff timer",
@@ -484,7 +491,7 @@ func (p *blockPrefetcher) rescheduleTopBlock(
 	}
 	p.log.Debug("Rescheduling prefetch of %s in %s", blockID, d)
 	rp.timer = time.AfterFunc(d, func() {
-		p.triggerPrefetch(req)
+		p.triggerPrefetch(pp.req)
 	})
 }
 
@@ -595,12 +602,11 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				// one.
 				pre = p.newPrefetch(1, false, req)
 				p.prefetches[blockID] = pre
+			} else {
+				pre.req = req
 			}
 			p.log.Debug("rescheduling top-block prefetch for block %s", blockID)
-			p.applyToParentsRecursive(
-				func(blockID kbfsblock.ID, pp *prefetch) {
-					p.rescheduleTopBlock(blockID, pp, req)
-				}, blockID, pre)
+			p.applyToParentsRecursive(p.rescheduleTopBlock, blockID, pre)
 		case reqInt := <-p.prefetchRequestCh.Out():
 			req := reqInt.(*prefetchRequest)
 			pre, isPrefetchWaiting := p.prefetches[req.ptr.ID]
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -672,6 +672,20 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 					"%s", req.ptr.ID)
 				continue
 			}
+
+			// Bail out early if we know the sync cache is already
+			// full, to avoid enqueuing the child blocks when they
+			// aren't able to be cached.
+			if p.rescheduleIfNeeded(ctx, req) {
+				// This is inefficient since it'd be better to know if
+				// the `subtreeBlockCount` below is 0, or if `isTail`
+				// below is true before needlessly rescheduling this.
+				// But currently that requires some complexity to
+				// figure out, so for now just do this early and
+				// revisit if it becomes a problem.
+				continue
+			}
+
 			if isPrefetchWaiting {
 				if pre.subtreeTriggered {
 					p.log.CDebugf(ctx, "prefetch subtree already triggered "+
@@ -724,15 +738,6 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				ctx = pre.ctx
 				p.log.CDebugf(ctx, "created new prefetch for block %s",
 					req.ptr.ID)
-			}
-
-			if p.rescheduleIfNeeded(ctx, req) {
-				// This is inefficient since it'd be better to know if
-				// `isTail` below is true before needlessly
-				// rescheduling this.  But currently that requires
-				// some complexity to figure out, so for now just do
-				// this early and revisit if it becomes a problem.
-				continue
 			}
 
 			// TODO: There is a potential optimization here that we can
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -463,6 +463,8 @@ func (p *blockPrefetcher) rescheduleTopBlock(
 		return
 	}
 
+	// Effectively below we are transferring the request for the top
+	// block from `p.prefetches` to `p.rescheduled`.
 	delete(p.prefetches, blockID)
 	pp.Close()
 
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -672,9 +672,6 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 					"%s", req.ptr.ID)
 				continue
 			}
-			if p.rescheduleIfNeeded(ctx, req) {
-				continue
-			}
 			if isPrefetchWaiting {
 				if pre.subtreeTriggered {
 					p.log.CDebugf(ctx, "prefetch subtree already triggered "+
@@ -728,6 +725,16 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				p.log.CDebugf(ctx, "created new prefetch for block %s",
 					req.ptr.ID)
 			}
+
+			if p.rescheduleIfNeeded(ctx, req) {
+				// This is inefficient since it'd be better to know if
+				// `isTail` below is true before needlessly
+				// rescheduling this.  But currently that requires
+				// some complexity to figure out, so for now just do
+				// this early and revisit if it becomes a problem.
+				continue
+			}
+
 			// TODO: There is a potential optimization here that we can
 			// consider: Currently every time a prefetch is triggered, we
 			// iterate through all the block's child pointers. This is short
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/eapache/channels"
+	"github.com/keybase/backoff"
 	"github.com/keybase/client/go/logger"
 	"github.com/keybase/kbfs/kbfsblock"
 	"github.com/keybase/kbfs/tlf"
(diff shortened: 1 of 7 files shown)


libkbfs/prefetcher.go
```diff
@@ -425,6 +451,68 @@ func (p *blockPrefetcher) handlePrefetch(pre *prefetch, isPrefetchNew,
...
+		}
+		p.rescheduled[blockID] = rp
+	}
+
+	if rp.timer != nil {
+		// Prefetch already scheduled.
+		return
+	}
+	req.ptr = BlockPointer{ID: blockID}
+	req.block = pre.req.block.NewEmpty()
```
jzila: if you use `pre.req` here, then the `ptr` should already be correct, and you can instantiate a new block in place: `pre.req.block = pre.req.block.NewEmpty()`. Then you can pass `pre.req` to the `triggerPrefetch` below, and avoid passing `req` into this method at all.

libkbfs/prefetcher.go
```diff
@@ -496,6 +586,19 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
...
+			blockID := req.ptr.ID
+			pre, ok := p.prefetches[blockID]
+			if !ok {
+				pre = p.newPrefetch(1, false, req)
+				p.prefetches[blockID] = pre
+			}
+			p.log.Debug("rescheduling top-block prefetch for block %s", blockID)
+			p.applyToParentsRecursive(
+				func(blockID kbfsblock.ID, pp *prefetch) {
+					p.rescheduleTopBlock(blockID, pp, req)
```
jzila: I'm a bit confused as to why you're passing in `req` here. Can't you just create a new one inside the function? You seem to be reusing the `req` struct and reassigning stuff inside it as you walk up the tree.

libkbfs/prefetcher.go
```diff
@@ -496,6 +586,19 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 			p.log.Debug("canceling prefetch for block %s", blockID)
 			// Walk up the block tree and delete every parent.
 			p.applyToParentsRecursive(p.cancelPrefetch, blockID, pre)
+		case reqInt := <-p.prefetchRescheduleCh.Out():
+			req := reqInt.(*prefetchRequest)
+			blockID := req.ptr.ID
+			pre, ok := p.prefetches[blockID]
+			if !ok {
```
jzila: Comment here that you need to create and cache a new prefetch to stop other causes from creating a new one as well, while this one is waiting.

Also please use `isPrefetchWaiting` instead of `ok`, to stay consistent with the rest of the prefetcher.

libkbfs/prefetcher.go
```diff
@@ -496,6 +586,19 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
...
+				pre = p.newPrefetch(1, false, req)
+				p.prefetches[blockID] = pre
+			}
```
jzila: Doing the following will allow you to avoid passing `req` into the `rescheduleTopBlock` method.
```go
} else if (pre.req == nil) {
	pre.req = req
}
```

libkbfs/prefetcher.go
```diff
@@ -559,6 +664,9 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 					"%s", req.ptr.ID)
 				continue
 			}
+			if p.rescheduleIfNeeded(ctx, req) {
+				continue
+			}
```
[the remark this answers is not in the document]
strib: @jzila I'm not sure if this is the right place!  What do you recommend?  Should it be just above the `handlePrefetch()` call below?

I don't really follow your point about the downstream fetches, but I'll just put this wherever you tell me is right.

libkbfs/prefetcher.go
strib: The problem is all the other fields in the `req`: priority, lifetime, isDeepSync, etc.  I didn't really want to pass all of those around individually, and I didn't know how to figure them out from within the `rescheduleTopBlock` function.  If you know what the right rules should be, I'll gladly change it.

libkbfs/prefetcher.go
strib: As below, I'm not sure if `pre.req` has all the right parameters in it.  If we can assume it does, that would be great!

strib:
Thanks for the initial look.  I pushed some fixes but still unsure what to do about the request and the exact placement of the `rescheduleIfNeeded` call.

libkbfs/prefetcher.go
```diff
@@ -496,6 +586,19 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
...
 			p.applyToParentsRecursive(p.cancelPrefetch, blockID, pre)
+		case reqInt := <-p.prefetchRescheduleCh.Out():
+			req := reqInt.(*prefetchRequest)
+			blockID := req.ptr.ID
+			pre, ok := p.prefetches[blockID]
+			if !ok {
+				pre = p.newPrefetch(1, false, req)
+				p.prefetches[blockID] = pre
+			}
+			p.log.Debug("rescheduling top-block prefetch for block %s", blockID)
```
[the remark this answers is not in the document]
jzila: > But I guess that old parent backpointers don't get cleaned up on a cancel

I don't think this is true--`CancelPrefetch` does an `applyToParentsRecursive`.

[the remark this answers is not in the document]
strib: But if you have a block path `a1 -> b`, and then we get an update where `a1 -> a2` but `b` doesn't change, then `CancelPrefetch(a1)` is called.  But that doesn't update the prefetch state for `b` to take away `a1` as a parent pointer, right?

libkbfs/prefetcher.go
jzila: I think `pre.req` is more likely to have the correct parameters than `req`, since `pre.req` pertains directly to the `blockID` being traversed in `applyToParentsRecursive`.

Ultimately though this depends on what you want:
1. If you're trying to reschedule the top block prefetch as-is, then use `pre.req`.
2. If you're trying to reschedule the top block based on the parameters in `req`, then use `req`.

I suggested the first approach based on my assumption that we want 1, and that 1 also results in cleaner code.

libkbfs/prefetcher.go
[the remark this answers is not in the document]
jzila: @strib correct, it doesn't take away `a1` as a parent pointer, but then as soon as `b`'s prefetch happens, it'll hit this line, which will remove it from its parents:
https://github.com/keybase/kbfs/blob/9291860e2465a194d9916d53cc683caf6243d919/libkbfs/prefetcher.go#L171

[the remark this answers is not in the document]
jzila: Which actually brings to mind another point: I wonder if that should now be:
```go
parent, ok := p.prefetches[b]
if !ok {
	parent, ok = p.rescheduled[b]
	if !ok {
		delete(pre.parents, b)
	}
}
```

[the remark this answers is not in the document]
strib: > I wonder if that should now be:

Hmm, I don't think so.  If the prefetch has been legitimately cancel, as in the parent-being-upgraded case, we don't want to bother rescheduling it.  So with your clarification and the behavior of `folderBlockOps` to cancel old prefetches, I think it's very unlikely that a block will actually have multiple parents for any meaningful amount of time.  So it seems to me like everything here should work correctly -- in most cases, it'll reschedule the prefetch of the top-most block it can find in the tree.

Please correct me if I've missed something.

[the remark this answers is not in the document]
jzila: I agree that it should be rare.

> If the prefetch has been legitimately cancel, as in the parent-being-upgraded case, we don't want to bother rescheduling it.

100% agree about that case. But now when you reschedule, you move a block out of `p.prefetches` and into `p.rescheduled`. We probably shouldn't lose parent tracking in that case...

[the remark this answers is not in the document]
jzila: Then again, maybe we should. Since the top-most block will be doing the prefetching in parent tracking, and we will have lost the full chain.

E.g. `a -> b -> c -> d -> e`. We reschedule `c`, which blasts away `c`, and `b`, and reschedules `a`. Then we prefetch `e`, which reschedules to `d`. Not sure that's the right answer here.

libkbfs/prefetcher.go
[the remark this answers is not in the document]
jzila: The downstream fetches point relates to the one I just posted in my other comment: that you want to give a chance to the lowest-level block to reschedule, such that its parent tree doesn't get blasted away. Even in that case, however, we'll end up orphaning and potentially rescheduling a bunch of subtrees.

Regarding the positioning of this rescheduling code, we should chat so I understand exactly the scenarios when we want to reschedule.

libkbfs/prefetcher.go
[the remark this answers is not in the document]
strib: Yeah I think the scenario you outline just above is a good-enough one to start with for now.  I don't think we can do much better without trying to keep the `*prefetchRequest` for rescheduled requests around somewhere.  I propose just keeping `applyToParentsRecursive` as it is for now, and revisiting later if needed.  I'll add a comment there though.

libkbfs/prefetcher.go
```diff
@@ -425,6 +456,69 @@ func (p *blockPrefetcher) handlePrefetch(pre *prefetch, isPrefetchNew,
...
+		p.rescheduled[blockID] = rp
+	}
+
+	if rp.timer != nil {
+		// Prefetch already scheduled.
+		return
+	}
+	// Copy the req but use a new empty block.
+	req := *pp.req
+	req.block = pp.req.block.NewEmpty()
```
jzila: Actually here's another concern. If you're using a `NewEmpty` block, then when you go to reschedule this prefetch, then you'll actually have to _fetch_ it again (probably from the cache).

Right now the prefetcher assumes that if it encounters a `prefetchRequest` from `prefetchRequestCh`, then that means that the block is populated. So if you want to do `NewEmpty` here, then you'll need to fetch it.

Here with the rescheduling you're trying to simulate the scenario that a prefetch request didn't exist in the prefetcher before, but upon reschedule is treated as if it'd just gotten fetched. So there must be `Block` content.

I think the right way to do this is:

```
req := *pp.req
// At this point you are done with `pp`.
...
rp.timer = time.AfterFunc(d, func() {
	p.triggerPrefetch(&req)
})
```

jzila: The way I think of it:
`prefetch` instances cached inside the prefetcher might not have a `*prefetchRequest` yet.
But a `prefetchRequest` should always have populated `Block` data.

strib: Sounds good thanks. Done in the latest commit.

libkbfs/prefetcher.go
```diff
@@ -425,6 +456,67 @@ func (p *blockPrefetcher) handlePrefetch(pre *prefetch, isPrefetchNew,
...
+			off: p.makeNewBackOff(),
+		}
+		p.rescheduled[blockID] = rp
+	}
+
+	if rp.timer != nil {
+		// Prefetch already scheduled.
+		return
+	}
+	// Copy the req but use a new empty block.
```
jzila: This comment is no longer accurate.

libkbfs/prefetcher.go
```diff
@@ -505,6 +616,8 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				pre.req = req
 			}
 
+			p.clearRescheduleState(req.ptr.ID)
```
jzila: Why do you `delete(p.reschedules, blockID)` in other situations you call this, but not here and below?

strib: Because here we are doing the actual prefetch and if it fails, we need to reschedule it with the next backoff duration.  If I deleted the reschedule state here, the backoff timer would start over on every re-fetch.

libkbfs/prefetcher.go
```diff
@@ -612,6 +725,16 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 				p.log.CDebugf(ctx, "created new prefetch for block %s",
 					req.ptr.ID)
 			}
+
+			if p.rescheduleIfNeeded(ctx, req) {
```
strib: @jzila we can't actually put this call here, because of these lines above: https://github.com/keybase/kbfs/blob/0d0a1c8f5c24918d8c1fa745efd2c9ababecde44/libkbfs/prefetcher.go#L713-L715

The block could have already been decremented and marked as subtree-triggered, but then when we reschedule it here we don't undo any of that, and some prefetches will never be completed due to the incorrect state.

So I'm inclined to move it back up to where it was.  Or I could try to refactor this code a bit so that the single `continue` above happens, then we do this check, and then we do everything else that might modify the prefetch request state.  Let me know if you have a preference.

jzila: If we have `a -> b -> c`, and we are handling `c`, then we'll:
* decrement up to `a`
* set `subtreeTriggered`
* reschedule `c`
  * lose `subtreeTriggered` (which we actually want, since it hasn't happened until `handlePrefetch`)
  * cancel `c` and `b`, rescheduling `a`.

Seems to me like we do actually want to lose this information. What seems to be going wrong?

strib: @jzila Here's the exact situation:

1. Path `root -> {a -> {aa, ab}, b}`.
2. Cache fills up after `a` and `ab` are fetched.
3. When handling the request for `aa`, the parent path is decremented, bring `a` down to a count of 0.  But `completePrefetch` is never called, so `a` never gets the status `FinishedPrefetch`.
4. Then `aa` is queued to be rescheduled.
5. The test explicitly requests `a`.
6. The prefetcher sees that `a` has a `subtreeBlockCount` of 0, and cancel its prefetch request.
7. When `aa` is rescheduled, its parent doesn't have an entry in `p.prefetches` because it was just canceled, and so `aa` is seen as the top-block to be rescheduled.
8. Once `aa` is fetched, it completes itself, but never completes `a` since it's no longer listed as a parent.

So neither `a` nor `root` are ever marked as `FinishedPrefetch`.  And it happens because the block counts were decremented before a reschedule.

jzila: I see, okay. We can move it back and leave the slight efficiency losses alone.

PR #1919, opened 2018-11-20
bserver_remote: don't cache archived data

jakob223:


libkbfs/bserver_remote.go
```diff
@@ -376,8 +376,15 @@ func (b *BlockServerRemote) Get(ctx context.Context, tlfID tlf.ID, id kbfsblock.
...
 	ctx = rpc.WithFireNow(ctx)
+	var res keybase1.GetBlockRes
 	b.log.LazyTrace(ctx, "BServer: Get %s", id)
+
+	// Once the block has been retrieved, cache it.
 	defer func() {
+		// But don't cache it if it's archived data.
+		if res.Status == keybase1.BlockStatus_ARCHIVED {
+			return
+		}
```
songgao: Should this be inside the `else` below and right after `CDebugf` so that we still get the logs?

strib:
> Also I wonder if we need to explicitly remove them from the disk cache if we find out it's archived. The block could have previously been put into the cache. But maybe that's already covered in the eviction (for cache) and the block deletion logic (for sync). Just thought I'd bring it up.

@songgao yeah that's already covered by work I previously did on eviction.  In fact that's what inspired this PR, because otherwise archived blocks can get in the cache through archived browsing and then never get evicted.

libkbfs/bserver_remote.go
strib: I agree we still want to have the log in this case.

PR #1920, opened 2018-11-20
Dokan alternate streams

taruti:
This is support for iterating alternate streams.

Opening them in CreateFile is not yet done.


jzila:
cc @zanderz as well for Windows code review.

PR #1921, opened 2018-11-20
folder_branch_ops: support partial sync config

strib:
[CircleCI depends on keybase/client#14780, which is vendored here and will be updated with the correct hash when it is merged.]

We want to cache the list of partially-synced paths in an encrypted block on disk.  Since that has to happen in folderBranchOps, this moves the main interface for setting/getting sync config into the KBFSOps interface.  (Of course the paths, including this exact sync config list, is still logged, but at least logs can be easily deleted without side effects, while the sync config DB cannot.)

In the lower-level interface, we move from storing the keybase1 sync config object in the DB (which contains the plaintext path list) to storing a new struct which contains a block pointer, buffer, and key half.  This struct is backwards compatible with any current sync configs stored using the old struct.

We don't yet act on this partial sync config; that will happen in a separate PR.

Issue: KBFS-3522

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -7,6 +7,8 @@ package libkbfs
 import (
 	"fmt"
 	"os"
+	stdpath "path"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
@@ -105,6 +107,7 @@ const (
 	fboMDWriter fboMutexLevel = 1
 	fboHead     fboMutexLevel = 2
 	fboBlock    fboMutexLevel = 3
+	fboSync     fboMutexLevel = 4
 )
 
 func (o fboMutexLevel) String() string {
@@ -115,6 +118,8 @@ func (o fboMutexLevel) String() string {
 		return "headLock"
 	case fboBlock:
 		return "blockLock"
+	case fboSync:
+		return "syncLock"
 	default:
 		return fmt.Sprintf("Invalid fboMutexLevel %d", int(o))
 	}
@@ -272,6 +277,8 @@ type folderBranchOps struct {
 	// Has this folder ever been cleared?
 	hasBeenCleared bool
 
+	syncLock leveledRWMutex
+
 	blocks  folderBlockOps
 	prepper folderUpdatePrepper
 
@@ -414,6 +421,7 @@ func newFolderBranchOps(
 	mdWriterLock := makeLeveledMutex(mutexLevel(fboMDWriter), &sync.Mutex{})
 	headLock := makeLeveledRWMutex(mutexLevel(fboHead), &sync.RWMutex{})
 	blockLockMu := makeLeveledRWMutex(mutexLevel(fboBlock), &sync.RWMutex{})
+	syncLock := makeLeveledRWMutex(mutexLevel(fboSync), &sync.RWMutex{})
 
 	forceSyncChan := make(chan struct{})
 
@@ -428,6 +436,7 @@ func newFolderBranchOps(
 			config, nodeCache, quotaUsage),
 		mdWriterLock: mdWriterLock,
 		headLock:     headLock,
+		syncLock:     syncLock,
 		blocks: folderBlockOps{
 			config:        config,
 			log:           log,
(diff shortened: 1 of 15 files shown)


libkbfs/interfaces.go
```diff
@@ -541,6 +542,20 @@ type KBFSOps interface {
 	// called after explicit user confirmation.  After the call,
 	// `handle` has the new TLF ID.
 	Reset(ctx context.Context, handle *TlfHandle) error
+
+	// GetSyncConfig returns the sync state configuration for the
+	// given TLF.
+	GetSyncConfig(ctx context.Context, tlfID tlf.ID) (
+		keybase1.FolderSyncConfig, error)
+	// SetSyncConfig set the sync state configuration for the given
```
songgao: s/set/sets

libkbfs/interfaces.go
```diff
@@ -541,6 +542,20 @@ type KBFSOps interface {
...
+	// TLF to either fully enabled, or fully disabled.  If syncing is
```
songgao: doesn't this cover "partially enabled" too?

libkbfs/kbfs_ops_test.go
```diff
@@ -4214,3 +4220,105 @@ func TestKBFSOpsSyncedMDCommit(t *testing.T) {
...
+	for _, p := range syncConfig.Paths {
+		require.True(t, pathsMap[p])
+		delete(pathsMap, p)
+	}
+
+	t.Log("Test some failure scenarios")
+	syncConfig.Paths = []string{"a/b/c", "a/b/c"}
+	_, err = kbfsOps.SetSyncConfig(ctx, h.tlfID, syncConfig)
+	require.Error(t, err)
+	syncConfig.Paths = []string{"/a/b/c", "a/b/c"}
```
songgao: This is to test abs paths are not allowed, rather than duplicate paths right? It'd be clearer if the second path is removed.

libkbfs/folder_branch_ops.go
```diff
@@ -7424,6 +7433,173 @@ func (fbo *folderBranchOps) Reset(
...
+	if err != nil {
+		return FolderSyncEncryptedPartialPaths{}, err
+	}
+	if oldConfig.Mode == keybase1.FolderSyncMode_ENABLED {
+		return FolderSyncEncryptedPartialPaths{},
+			errors.Errorf("TLF %s is already fully synced", fbo.id())
+	}
+
+	// Make sure the new path list doesn't contain duplicates,
+	// contains no duplicates, and each path is cleaned.
```
jzila: Department of redundancy department.

libkbfs/folder_branch_ops.go
```diff
@@ -7424,6 +7433,173 @@ func (fbo *folderBranchOps) Reset(
...
+		p = stdpath.Clean(filepath.ToSlash(p))
+		if seenPaths[p] {
+			return FolderSyncEncryptedPartialPaths{}, errors.Errorf(
+				"%s is in the paths list more than once", p)
+		}
+		if stdpath.IsAbs(p) {
+			return FolderSyncEncryptedPartialPaths{}, errors.Errorf(
+				"Absolute paths like %s are not allowed", p)
+		}
+		if strings.HasPrefix(p, "..") {
```
jzila: you can make a relative path out of the TLF even if the prefix isn't `..`. You probably want `strings.Contains(p, "..")`

libkbfs/folder_branch_ops.go
```diff
@@ -7424,6 +7433,173 @@ func (fbo *folderBranchOps) Reset(
...
+	chargedTo, err := chargedToForTLF(
+		ctx, fbo.config.KBPKI(), fbo.config.KBPKI(), kmd.GetTlfHandle())
+	if err != nil {
+		return FolderSyncEncryptedPartialPaths{}, err
+	}
+
+	info, _, readyBlockData, err :=
+		ReadyBlock(ctx, fbo.config.BlockCache(), fbo.config.BlockOps(),
+			fbo.config.Crypto(), kmd, b, chargedTo,
+			fbo.config.DefaultBlockType())
```
jzila: So since we don't have indirect blocks for path storage, we implicitly have a cap for the number of partial paths that we can sync. IMO it's high enough that we can pretty much ignore this, but it might be good to call it out somewhere.

Also worth calling out that the block is never synced remotely--that the partial sync state is per-client only.

libkbfs/folder_branch_ops.go
strib: @jzila I don't think you can, because of the `stdpath.Clean` call above.  It should get rid of all the unnecessary ".." components except at the beginning.

libkbfs/folder_branch_ops.go
strib: > So since we don't have indirect blocks for path storage, we implicitly have a cap for the number of partial paths that we can sync. IMO it's high enough that we can pretty much ignore this, but it might be good to call it out somewhere.

Since we're not sending these blocks to the server, there's really no limit on the size of the block, except the user's memory.  Is that what you meant?   Not sure if that's worth mentioning or not, since anything that hits the user's memory limit will probably just crash things before it's even committed to disk.

PR #1924, opened 2018-11-20
config_local: set codec before loading sync TLFs

strib:


jakob223:
Should we add a test that reads data from the sync block cache? Maybe in `kbfsdocker`?

strib:
> Should we add a test that reads data from the sync block cache? Maybe in kbfsdocker?

This is actually reading from the sync config db, not the cache itself.  But yes, probably when offline mode is all done, we should add a more complete test that uses the sync cache and sets something in this db.

PR #1930, opened 2018-11-22
disk_block_cache: the caller of Get/Put specifies the cache type

strib:
In anticipation of partial syncing, where the `*folderBranchOps` instance is the only place that knows whether a given block needs to be in the sync cache or not, this commit removes usage of `config.IsSyncedTlf` from the disk block cache.

Instead, it plumbs through a "cache type" with every `Get` and `Put` call.  For `Get` calls, if the block isn't already in the desired cache type, it will be moved there (which will be useful when a path is marked as syncing after some of its blocks are already in the working set cache).

This commit plumbs the cache type down from the places where the get/put calls originate from.  Future work will need to make decisions in those places about whether a particular block goes into the sync cache; for now, it just uses `config.IsSyncedTlf` there.

This depends on #1925 and #1932.

Issue: KBFS-3523

### DIFF
--- a/libkbfs/disk_block_cache_test.go
+++ b/libkbfs/disk_block_cache_test.go
@@ -749,7 +749,10 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
 	t.Log("Move the block by getting it with a different preferred cache.")
 	_, _, _, err = cache.Get(ctx, tlf1, block1Ptr.ID, DiskBlockSyncCache)
 	require.NoError(t, err)
+	err = cache.waitForDeletes(ctx)
+	require.NoError(t, err)
 	require.Equal(t, 1, cache.syncCache.numBlocks)
+	require.Equal(t, 0, cache.workingSetCache.numBlocks)
 
 	t.Log("After the move, make sure the prefetch status is right.")
 	_, _, prefetchStatus, err := cache.Get(
--- a/libkbfs/disk_block_cache_test.go
+++ b/libkbfs/disk_block_cache_test.go
@@ -174,14 +174,17 @@ func TestDiskBlockCachePutAndGet(t *testing.T) {
 	ctx := context.Background()
 
 	t.Log("Put a block into the cache.")
-	err := cache.Put(ctx, tlf1, block1Ptr.ID, block1Encoded, block1ServerHalf)
+	err := cache.Put(
+		ctx, tlf1, block1Ptr.ID, block1Encoded, block1ServerHalf,
+		DiskBlockAnyCache)
 	require.NoError(t, err)
 	putMd, err := cache.GetMetadata(ctx, block1Ptr.ID)
 	require.NoError(t, err)
 	config.TestClock().Add(time.Second)
 
 	t.Log("Get that block from the cache. Verify that it's the same.")
-	buf, serverHalf, _, err := cache.Get(ctx, tlf1, block1Ptr.ID)
+	buf, serverHalf, _, err := cache.Get(
+		ctx, tlf1, block1Ptr.ID, DiskBlockAnyCache)
 	require.NoError(t, err)
 	require.Equal(t, block1ServerHalf, serverHalf)
 	require.Equal(t, block1Encoded, buf)
@@ -196,7 +199,8 @@ func TestDiskBlockCachePutAndGet(t *testing.T) {
 	t.Log("Attempt to Get a block from the cache that isn't there." +
 		" Verify that it fails.")
 	ptr2 := makeRandomBlockPointer(t)
-	buf, serverHalf, _, err = cache.Get(ctx, tlf1, ptr2.ID)
+	buf, serverHalf, _, err = cache.Get(
+		ctx, tlf1, ptr2.ID, DiskBlockAnyCache)
 	require.EqualError(t, err, NoSuchBlockError{ptr2.ID}.Error())
 	require.Equal(t, kbfscrypto.BlockCryptKeyServerHalf{}, serverHalf)
 	require.Nil(t, buf)
@@ -219,7 +223,8 @@ func TestDiskBlockCacheDelete(t *testing.T) {
 		tlf := tlf.FakeID(f, tlf.Private)
 		blockPtr, _, blockEncoded, serverHalf := setupBlockForDiskCache(
 			t, config)
-		err := cache.Put(ctx, tlf, blockPtr.ID, blockEncoded, serverHalf)
+		err := cache.Put(
+			ctx, tlf, blockPtr.ID, blockEncoded, serverHalf, DiskBlockAnyCache)
 		require.NoError(t, err)
 	}
 	tlf1 := tlf.FakeID(3, tlf.Private)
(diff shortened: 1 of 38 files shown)


strib:
Almost all of this PR is boring plumbing stuff.  The interesting code that should be reviewed more carefully is in:

* `interfaces.go`
* `disk_block_cache_wrapped.go`
* `disk_block_cache_test.go`

Almost everything else is just boilerplate plumbing.

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -112,23 +111,43 @@ func (cache *diskBlockCacheWrapped) IsSyncCacheEnabled() bool {
 }
 
 // Get implements the DiskBlockCache interface for diskBlockCacheWrapped.
-func (cache *diskBlockCacheWrapped) Get(ctx context.Context, tlfID tlf.ID,
-	blockID kbfsblock.ID) (
+func (cache *diskBlockCacheWrapped) Get(
+	ctx context.Context, tlfID tlf.ID, blockID kbfsblock.ID,
+	preferredCacheType DiskBlockCacheType) (
```
songgao: This is not an atomic operation. Stuff can happen between the Get-Get-Put-UpdateMetadata-Delete sequence. I can't think of an actual bug that would happen with this yet, but thought I'd bring it up. Also, if you decide to not make it atomic, perhaps worth mentioning in comment in `interfaces.go`?

libkbfs/disk_block_cache_test.go
```diff
@@ -687,3 +709,37 @@ func TestDiskBlockCacheUnsyncTlf(t *testing.T) {
...
+	err = cache.UpdateMetadata(ctx, block1Ptr.ID, TriggeredPrefetch)
+	require.NoError(t, err)
+	require.Equal(t, 1, cache.workingSetCache.numBlocks)
+	require.Equal(t, 0, cache.syncCache.numBlocks)
+
+	t.Log("Move the block by getting it with a different preferred cache.")
+	_, _, _, err = cache.Get(ctx, tlf1, block1Ptr.ID, DiskBlockSyncCache)
+	require.NoError(t, err)
+	require.Equal(t, 1, cache.syncCache.numBlocks)
+
```
songgao: Do we want a `require.Equal(t, 0, cache.workingSetCache.numBlocks)` here?

libkbfs/disk_block_cache_wrapped.go
strib: Except for the final delete, it's protected with `cache.mtx.RLock()`.  Did you miss that, or were you worried about two Gets happening at the same time?

libkbfs/disk_block_cache_test.go
strib: Ideally yes, but the Delete happens in another goroutine and I didn't think it was worth adding a bunch of code to track outstanding deletes.

libkbfs/disk_block_cache_wrapped.go
strib: The worst I can think of is undoing a Delete that happens in the middle of the Get operations, so a block that should be deleted ends up moved to a sync cache instead.  That seems pretty unlikely since explicit deletes happen only after archiving, and the block shouldn't be getting accessed after it's archived.

So I think it's probably safe as is, but would like @jzila to weigh in as well.

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -112,23 +111,48 @@ func (cache *diskBlockCacheWrapped) IsSyncCacheEnabled() bool {
...
+		primaryCache.Get(ctx, tlfID, blockID, preferredCacheType)
 	if _, isNoSuchBlockError := err.(NoSuchBlockError); isNoSuchBlockError &&
 		secondaryCache != nil {
-		return secondaryCache.Get(ctx, tlfID, blockID)
+		buf, serverHalf, prefetchStatus, err = secondaryCache.Get(
+			ctx, tlfID, blockID, preferredCacheType)
+		if err != nil {
+			return nil, kbfscrypto.BlockCryptKeyServerHalf{}, NoPrefetch, err
+		}
+		if preferredCacheType != DiskBlockAnyCache {
```
jzila: It seems to me that by using `DiskBlockAnyCache` everywhere, then if a block was once synced (and then later unsynced), we won't ever move it out of the sync cache. Is that desired behavior?

libkbfs/disk_block_cache_wrapped.go
strib: @jzila Yes agreed. But what do you think about the race possibilities here?  Do you think it's worth having locks to protect against the unlikely delete race I mention above?  Or can you think of any other races to be worried about?

libkbfs/disk_block_cache_wrapped.go
```diff
@@ -112,23 +111,48 @@ func (cache *diskBlockCacheWrapped) IsSyncCacheEnabled() bool {
...
 	}
 	// Check both caches if the primary cache doesn't have the block.
 	buf, serverHalf, prefetchStatus, err =
-		primaryCache.Get(ctx, tlfID, blockID)
+		primaryCache.Get(ctx, tlfID, blockID, preferredCacheType)
 	if _, isNoSuchBlockError := err.(NoSuchBlockError); isNoSuchBlockError &&
 		secondaryCache != nil {
-		return secondaryCache.Get(ctx, tlfID, blockID)
+		buf, serverHalf, prefetchStatus, err = secondaryCache.Get(
+			ctx, tlfID, blockID, preferredCacheType)
```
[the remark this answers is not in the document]
strib: Agreed, but it is part of the `Get` interface that both `diskBlockCacheWrapped` and `DiskBlockCacheLocal` have to implement.

libkbfs/disk_block_cache_wrapped.go
strib: If syncing on a TLF is completely unsynced, all the blocks of that TLF are removed from the sync cache.  I still need to work out a strategy for deleting blocks corresponding to unsynced partial paths -- there's a ticket for that and I'll handle it separately.

And if a block goes in the sync cache and the subsequently accessed with `DiskBlockAnyCache`, I think it's correct to leave it in the sync cache, since that was the most recent preferred cache for it.

But let me know if I misunderstood the question.

libkbfs/disk_block_cache_wrapped.go
jzila: When this was first built, I think we didn't care about that particular race--the worst case is that sometimes we have a block in both caches, or we have a block in a cache when it shouldn't be any more. And as you say, an archived block shouldn't be accessed any more so this is extremely unlikely.

libkbfs/disk_block_cache_wrapped.go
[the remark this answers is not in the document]
jzila: Yeah. I was thinking about that--maybe we need an explicit interface for the wrapped cache, so that we don't have this extraneous parameter in the specific caches.

libkbfs/disk_block_cache_wrapped.go
jzila: I guess I'm saying that we never use `DiskBlockWorkingSetCache`, and it's unclear to me whether we ever plan on using it. And the fact that we use `DiskBlockAnyCache` in any situation where we even might use `DiskBlockWorkingSetCache`, means that we don't ever get into a situation where a block is moved from the sync cache to the working set cache (but we do get the reverse).

libkbfs/disk_block_cache_wrapped.go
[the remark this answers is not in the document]
strib: > Yeah. I was thinking about that--maybe we need an explicit interface for the wrapped cache, so that we don't have this extraneous parameter in the specific caches.

True, but that's not really a new thing introduced in this PR.  For example, the unref stuff and tlf-clearning stuff isn't applied to both caches.  It might make sense to do as a separate cleanup, but I don't want to shoehorn it into this PR.

libkbfs/disk_block_cache_wrapped.go
strib: Correct.  In fact I just pushed a fix that avoids deleting a synced block if `DiskBlockAnyCache` is passed in, because that was a bug.  I have no use case in mind for `DiskBlockWorkingSetCache` -- I can delete that constant if you find it confusing.

jzila: No, it's fine. Just wanted to make sure I understood everything.

PR #1932, opened 2018-11-27
block_retrieval_queue: decouple priorities from prefetching

strib:
Introduce new action types to avoid overloading the queuing priority to be a prefetching indicator.  Now there are action types for the different combinations of:

* Prefetching.
* Writing blocks to the sync cache.
* Tail-prefetching
  * That is, blocks that are tracked by the prefetcher, but do
    not themselves trigger prefetches.

This will also be useful for partial prefetching, when we will need to fetch and sync parent blocks of partially-synced paths, without triggering a full deep-sync prefetch from those parents.

Suggested by jzila.

Issue: KBFS-3523
Issue: #1925

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -982,6 +982,27 @@ const (
 	BlockRequestWithDeepSync BlockRequestAction = blockRequestTrackedInPrefetch | blockRequestPrefetch | blockRequestSync | blockRequestDeepSync
 )
 
+func (bra BlockRequestAction) String() string {
+	if bra.DeepSync() {
+		return "deep-sync"
+	}
+	if bra == BlockRequestSolo {
+		return "solo"
+	}
+
+	attrs := make([]string, 0, 2)
+	if bra.Prefetch() {
+		attrs = append(attrs, "prefetch")
+	} else if bra.PrefetchTracked() {
+		attrs = append(attrs, "prefetch-tracked")
+	}
+
+	if bra.Sync() {
+		attrs = append(attrs, "sync")
+	}
+	return strings.Join(attrs, "|")
+}
+
 // Combine returns a new action by taking `other` into account.
 func (bra BlockRequestAction) Combine(
 	other BlockRequestAction) BlockRequestAction {
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -949,6 +949,7 @@ const (
 	blockRequestTrackedInPrefetch BlockRequestAction = 1 << iota
 	blockRequestPrefetch
 	blockRequestSync
+	blockRequestDeepSync
 
 	// BlockRequestSolo indicates that no action should take place
 	// after fetching the block.  However, a TLF that is configured to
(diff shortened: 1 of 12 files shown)


libkbfs/data_types.go
```diff
@@ -935,3 +935,129 @@ func syncPathListFromBlock(codec kbfscodec.Codec, b *FileBlock) (
...
+	BlockRequestWithSyncAndPrefetch
+	// BlockRequestPrefetchTail indicates that the block is being
+	// tracked in the prefetcher, but shouldn't kick off any more
+	// prefetches.
+	BlockRequestPrefetchTail
+	// BlockRequestPrefetchTailWithSync indicates that the block is
+	// being tracked in the prefetcher and goes in the sync cache, but
+	// shouldn't kick off any more prefetches.
+	BlockRequestPrefetchTailWithSync
+)
```
jzila: We seem to have 3 somewhat orthogonal boolean values here: sync or not, trigger prefetches or not, tracked in prefetcher or not. Sounds like a good candidate for a bitfield, in which case `Combine` below could be simplified a lot.

libkbfs/prefetcher.go
```diff
@@ -227,9 +227,9 @@ func (p *blockPrefetcher) completePrefetch(
 			b := pp.req.block.NewEmpty()
 			// TODO: after we split out priority from whether to prefetch, make
 			// this a much higher priority.
```
jzila: This `TODO` can go away.

libkbfs/prefetcher.go
```diff
@@ -296,8 +296,8 @@ top:
 // calculatePriority returns either a base priority for an unsynced TLF or a
 // high priority for a synced TLF.
 func (p *blockPrefetcher) calculatePriority(
-	basePriority int, isDeepSync bool) int {
-	if isDeepSync {
+	basePriority int, action BlockRequestAction) int {
+	if action.DeepSync() {
 		return defaultOnDemandRequestPriority - 1
```
jzila: I think this should probably just be `basePriority` now, where each level sets the priority of triggered requests to 1 less than the parent block. The root block of a deep sync will have a much lower base priority than working set triggered prefetches, which will just start at `defaultOnDemandRequestPriority`.

libkbfs/prefetcher.go
```diff
@@ -423,35 +421,41 @@ func (p *blockPrefetcher) prefetchDirectDirBlock(ctx context.Context,
...
-				b, req.kmd, req.lifetime, isPrefetchNew, isDeepSync)
+			numBlocks, isTail = p.prefetchIndirectDirBlock(
+				pre.ctx, req.ptr.ID, b, req.kmd, req.lifetime, isPrefetchNew,
+				childAction)
 		} else {
-			numBlocks, isTail = p.prefetchDirectDirBlock(pre.ctx, req.ptr.ID,
-				b, req.kmd, req.lifetime, isPrefetchNew, isDeepSync)
+			numBlocks, isTail = p.prefetchDirectDirBlock(
+				pre.ctx, req.ptr.ID, b, req.kmd, req.lifetime, isPrefetchNew,
+				childAction)
```
jzila: I think all 3 of these should now accept `pre.req.priority` as their base priority. And instead of doing the fancy `startingPriority` decrementing thing, all triggered prefetches should just have a base priority of `pre.req.priority - 1`.

libkbfs/data_types.go
strib: Yeah I should have done that to start with, but I didn't originally see how the combos would play out.  Much cleaner now in commit 2c47267b.

strib:
@jzila I fixed up the priority stuff in commit 39dcb0b, but I'm not 100% sure I followed what you were suggesting.  Please review that commit when you get a chance!

jzila:
@strib almost there--hopefully my comments cleared up what I meant.

strib:
Thanks @jzila.  I fixed up the priorities in 9cbd51a.  I also fixed some bugs here that I uncovered while testing partial syncs in a later PR.  I put up fixes in 244a743.  Would be nice if you could give it a quick look, but it's pretty straightforward I think so no worries if you can't.  If not, I'll probably merge in the morning unless you ask me to hold off.

PR #1933, opened 2018-11-27
Update hash for client in vendor.json

jakob223:
Since I forgot to re-vendor a thing after a merge in client squashed it.


jakob223:
Tried to tell github to let me merge this and accidentally created an empty merge commit instead..

PR #1935, opened 2018-11-29
folder_branch_ops: support partial syncs

strib:
Any time `folderBranchOps` learns about a newly merged revision, and the TLF is configured in partial-sync mode, kick off a partial sync. Also, kick one off when the sync config is changed.

For each partially-synced path, do a solo-sync-fetch of each parent block, and then deep-sync the final element in the path.

Deleting blocks from unsynced paths will happen in a future PR.

Issue: KBFS-3523


strib:
OOOPS.  Wrong PR.  Will undo.

PR #1937, opened 2018-11-29
folder_branch_ops: support partial syncs

strib:
Any time `folderBranchOps` learns about a newly merged revision, and the TLF is configured in partial-sync mode, kick off a partial sync. Also, kick one off when the sync config is changed.

For each partially-synced path, do a solo-sync-fetch of each parent block, and then deep-sync the final element in the path.

Deleting blocks from unsynced paths will happen in a future PR.

Issue: KBFS-3523

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -865,6 +865,8 @@ func (fbo *folderBranchOps) syncOneNode(
 	}
 }
 
+// doPartialSync iterates through the paths, deep-syncing them and
+// also syncing their parent directories up to the root node.
 func (fbo *folderBranchOps) doPartialSync(
 	ctx context.Context, syncConfig keybase1.FolderSyncConfig,
 	latestMerged ImmutableRootMetadata) (err error) {
@@ -969,11 +971,11 @@ func (fbo *folderBranchOps) kickOffPartialSync(
 	ctx context.Context, lState *lockState,
 	syncConfig keybase1.FolderSyncConfig, rmd ImmutableRootMetadata) {
 	// Kick off a background partial sync.
-	partialSyncCtx := fbo.ctxWithFBOID(context.Background())
+	partialSyncCtx, cancel := context.WithCancel(
+		fbo.ctxWithFBOID(context.Background()))
 	fbo.log.CDebugf(
 		ctx, "Partial sync with a new context: FBOID=%s",
 		partialSyncCtx.Value(CtxFBOIDKey))
-	partialSyncCtx, cancel := context.WithCancel(partialSyncCtx)
 	fbo.partialSyncs.Add(1)
 	go func() {
 		defer cancel()
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -361,6 +361,7 @@ type folderBranchOps struct {
 	forcedFastForwards kbfssync.RepeatedWaitGroup
 	merkleFetches      kbfssync.RepeatedWaitGroup
 	editActivity       kbfssync.RepeatedWaitGroup
+	partialSyncs       kbfssync.RepeatedWaitGroup
 	launchEditMonitor  sync.Once
 
 	muLastGetHead sync.Mutex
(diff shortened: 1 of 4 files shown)


strib:
Note: this PR now depends on #1943, so you might want to review that one first.

libkbfs/folder_branch_ops.go
```diff
@@ -886,6 +1119,26 @@ func (fbo *folderBranchOps) commitFlushedMD(
 
 		fbo.log.CDebugf(ctx, "Prefetch for revision %d complete; commiting",
 			rev)
+	case keybase1.FolderSyncMode_PARTIAL:
+		// For partially-synced TLFs, wait for the partial sync to
+		// complete, or for an update to happen.
+		lState := makeFBOLockState()
+		fbo.kickOffPartialSyncIfNeeded(ctx, lState, rmd)
```
jzila: So if this is about to happen while a partial sync is already happening, then `fbo.latestMergedUpdated` would be closed and the previous sync thus gets automatically canceled. Is my understanding correct?

libkbfs/folder_branch_ops.go
```diff
@@ -886,6 +1119,26 @@ func (fbo *folderBranchOps) commitFlushedMD(
...
+		go func() {
+			select {
+			case <-updatedCh:
+				cancel()
+			case <-fbo.shutdownChan:
+				cancel()
+			case <-ctx.Done():
+			}
+		}()
+		err := fbo.partialSyncs.Wait(ctx)
```
jzila: So we don't return from this commit until the partial sync completes?

libkbfs/folder_branch_ops.go
```diff
@@ -6193,6 +6466,10 @@ func (fbo *folderBranchOps) SyncFromServer(ctx context.Context,
 	if err := fbo.fbm.waitForSyncCacheCleans(ctx); err != nil {
 		return err
 	}
+	if err := fbo.partialSyncs.Wait(ctx); err != nil {
```
jzila: Same question here--we have to wait for the whole partial sync?

libkbfs/folder_branch_ops.go
```diff
@@ -791,6 +792,236 @@ func (fbo *folderBranchOps) startMonitorChat(tlfName tlf.CanonicalName) {
...
+
+		parentPath, syncedElem := stdpath.Split(p)
+		parents := strings.Split(strings.TrimSuffix(parentPath, "/"), "/")
+		currNode := rootNode
+		for _, parent := range parents {
+			if len(parent) == 0 {
+				continue
+			}
+			// TODO: parallelize the parent fetches and lookups.
+			currNode, _, err = fbo.Lookup(ctx, currNode, parent)
```
jzila: Won't this already trigger a prefetch? I guess it's fine, though, since you escalate the prefetch action with the explicit `syncOneNode` call later.

libkbfs/folder_branch_ops.go
```diff
@@ -791,6 +792,236 @@ func (fbo *folderBranchOps) startMonitorChat(tlfName tlf.CanonicalName) {
...
+
+func (fbo *folderBranchOps) kickOffPartialSync(
+	ctx context.Context, lState *lockState,
+	syncConfig keybase1.FolderSyncConfig, rmd ImmutableRootMetadata) {
+	// Kick off a background partial sync.
+	partialSyncCtx := fbo.ctxWithFBOID(context.Background())
+	fbo.log.CDebugf(
+		ctx, "Partial sync with a new context: FBOID=%s",
+		partialSyncCtx.Value(CtxFBOIDKey))
+	partialSyncCtx, cancel := context.WithCancel(partialSyncCtx)
```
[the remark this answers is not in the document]
strib: Sure I will combine them.

libkbfs/folder_branch_ops.go
```diff
@@ -791,6 +792,236 @@ func (fbo *folderBranchOps) startMonitorChat(tlfName tlf.CanonicalName) {
...
+		fbo.headLock.Lock(lState)
+		defer fbo.headLock.Unlock(lState)
+		if rmd.Revision() != fbo.latestMergedRevision {
+			fbo.log.CDebugf(
+				partialSyncCtx, "Latest merged changed is now %d, not %d; "+
+					"aborting partial sync", fbo.latestMergedRevision,
+				rmd.Revision())
+			return nil
+		}
+		return fbo.latestMergedUpdated
```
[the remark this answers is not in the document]
strib: I don't think it can be `nil` once there's an `rmd` that can be passed to this method.  It's always set by `setLatestMergedRevisionLocked()`, which is called on all the MD initialization paths.

libkbfs/folder_branch_ops.go
strib: Correct.  As soon as `setLatestMergedRevisionLocked` is called (which happens before this method is called in both cases), the previous sync gets canceled because of the channel close.

libkbfs/folder_branch_ops.go
strib: Not necessarily, because the `ctx` can be canceled in a number of cases above, such as when the latest merged version is updated.  But yeah in the absence of any other revisions or a shut down, we wait for the partial sync to complete before commiting the MD to disk.  (It also gets committed to disk if the MD is accessed for reading.)

libkbfs/folder_branch_ops.go
strib: Yes, but this method is only used by tests and git operation (which wouldn't have partial syncing).

PR #1938, opened 2018-11-29
libfs: MountInterrupter should be done after trying to unmount fails

taruti:
Currently if unmounting fails the MountInterrupter is not marked as done.
This means that shutdown can get blocked when unmounting fails
again and again.

### DIFF
--- a/libfs/mount_interrupter.go
+++ b/libfs/mount_interrupter.go
@@ -59,7 +59,8 @@ func (mi *MountInterrupter) MountAndSetUnmount(mounter Mounter) error {
 
 // Done signals Wait and runs the unmounter if set by MountAndSetUnmount.
 // It can be called multiple times with no harm. Each call triggers a call to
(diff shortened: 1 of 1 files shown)


taruti:
@songgao could be the right person for review.

songgao:
If I remember correctly, that was intentional, so that one can keep hitting Ctrl-C to call unmount multiple times.

songgao:
Perhaps we could do something like making it close if we get two signals within a second, but not if they are more than one second apart?

taruti:
Or having a counter there. Because now if unmount fails each time then it will keep spinning forever.

songgao:
Hmm wdym keep spinning forever?

Re counter, the one second thing I mentioned above was basically a counter, but with a timeout. It seems just having a counter without ever reset would generate some surprises as over a long time it just seems inconsistent.

PR #1939, opened 2018-11-29
Fix offline mode when mdserver is disconnected

strib:
Three things here:
1. Allow kbfsdocker tests to turn on journaling via an environment variable, because...
2. When journaling is turned on, we can skip checking the mdserver for unmerged MDs (and we don't/can't cache the absence of these, so otherwise it blocks TLF access if the mdserver if offline)
3. Use a simpler mdserver connectivity check in `SyncFromServer` to avoid extra checks while offline, because the "services" map takes much longer to be updated.

With these changes, the mdserver offline test now passes.

Issue: KBFS-3622

### DIFF
--- a/kbfsfuse/kbfsfuse.sh
+++ b/kbfsfuse/kbfsfuse.sh
@@ -26,12 +26,13 @@ fi
 git config --global user.email "you@example.com"
 git config --global user.name "Your Name"
 
+# Journaling is turned off by default for all tests, since some tests
+# depend on sync semantics.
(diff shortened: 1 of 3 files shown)


strib:
> For my own sake of understanding, could you elaborate in what situation would journal go Unmerged? I thought it'd always be "Master" locally.

"Unmerged" here just means that the local view of the folder is conflicting with the server view. So it happens any time the TLF forks locally, until the conflict is resolved.

PR #1940, opened 2018-11-30
keybase_service_util: don't block on auth token refreshes

strib:
We don't want to block a `CurrentSession` call just because one of the KBFS servers in unreachable.  This fixes one of the offline tests.

Issue: KBFS-3653

### DIFF
--- a/libkbfs/keybase_service_util.go
+++ b/libkbfs/keybase_service_util.go
@@ -57,14 +57,17 @@ func serviceLoggedIn(ctx context.Context, config Config, session SessionInfo,
 			"%+v", err)
 	}
 
+	// Launch auth refreshes in the background, in case we are
+	// currently disconnected from one of these servers.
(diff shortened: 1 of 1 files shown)


libkbfs/keybase_service_util.go
```diff
@@ -57,14 +57,17 @@ func serviceLoggedIn(ctx context.Context, config Config, session SessionInfo,
...
+	// currently disconnected from one of these servers.
 	mdServer := config.MDServer()
 	if mdServer != nil {
-		mdServer.RefreshAuthToken(ctx)
+		go mdServer.RefreshAuthToken(context.Background())
 	}
 	bServer := config.BlockServer()
 	if bServer != nil {
-		bServer.RefreshAuthToken(ctx)
+		go bServer.RefreshAuthToken(context.Background())
```
songgao: Do we want maybe a 1min timeout?

PR #1942, opened 2018-12-01
ops: show a special error on timeouts for unsynced TLFs

strib:
If we are about to return a timeout error to the user when they are reading data from an unsynced TLF, return a special error type that lets the GUI show a message explaining that the TLF is unsynced.

Unlike in the case of archived data (where we first check if we're disconnected and then immediately return the error), in this case we give the system time to recover from the connectivity issue during the operation, to avoid spurious notifications that might be recoverable.

Issue: KBFS-3585


libkbfs/kbfs_ops.go
```diff
@@ -476,10 +476,27 @@ func (fs *KBFSOpsStandard) createAndStoreTlfIDIfNeeded(
...
+	}
+
+	if fs.config.IsSyncedTlf(h.tlfID) {
+		fs.log.CWarningf(ctx, "Got a read timeout on a synced TLF: %+v", err)
+		return err
+	}
+
+	// For unsynced TLFs, return a specific error to let the system
+	// know to show a sync recommendation.
+	return errors.WithStack(OfflineUnsyncedError{h})
```
jzila: These could stack up with the fbo transform, right? I guess it's harmless though since we'll just have multiple log lines.

strib: Not sure what you mean by "stack up", but yeah one `OfflineUnsyncedError` could be replaced by another one here, for no real reason.  I'll add a quick check to avoid that.

PR #1944, opened 2018-12-04
disk_block_cache: prioritize the user's home TLFs for non-eviction

jakob223:
TODO: write a test

### DIFF
--- a/libkbfs/disk_block_cache.go
+++ b/libkbfs/disk_block_cache.go
@@ -929,21 +929,24 @@ type weightedByCount struct {
 // weighting by per-TLF block count.
 func (cache *DiskBlockCacheLocal) shuffleTLFsAtPriorityWeighted(
 	priority evictionPriority) []weightedByCount {
-	weightedSlice := make([]weightedByCount,
+	weightedSlice := make([]weightedByCount, 0,
 		len(cache.priorityTlfMap[priority]))
 	idx := 0
 	// Use an exponential distribution to ensure the weights are
 	// correctly used.
 	// See http://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf
-	for tlfId, count := range cache.priorityTlfMap[priority] {
-		weightedSlice[idx] = weightedByCount{
-			key:   rand.ExpFloat64() * float64(count),
-			value: tlfId,
+	for tlfID, count := range cache.priorityTlfMap[priority] {
+		if count == 0 {
+			continue
 		}
+		weightedSlice = append(weightedSlice, weightedByCount{
+			key:   math.Pow(rand.Float64(), 1.0/float64(count)),
+			value: tlfID,
+		})
 		idx++
 	}
 	sort.Slice(weightedSlice, func(i, j int) bool {
-		return weightedSlice[i].key < weightedSlice[j].key
+		return weightedSlice[i].key > weightedSlice[j].key
 	})
 	return weightedSlice
 }
@@ -1015,10 +1018,13 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
 					blockIDs = append(blockIDs, lruEntry{blockID, lru})
 				}
 			}()
+			if len(blockIDs) == numElements {
+				break
+			}
 		}
 		// Evict some of the selected blocks.
 		currNumRemoved, currSizeRemoved, err := cache.evictSomeBlocks(ctx,
-			numBlocks, blockIDs)
+			numBlocks-numRemoved, blockIDs)
 		if err != nil {
 			return numRemoved, sizeRemoved, err
 		}
--- a/libkbfs/disk_block_cache.go
+++ b/libkbfs/disk_block_cache.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"path/filepath"
 	"sort"
 	"sync"
@@ -48,6 +49,17 @@ const (
 	workingSetCacheName             string = "WorkingSetBlockCache"
 )
 
+var errTeamOrUnknownTLFAddedAsHome = errors.New(
+	"Team or Unknown TLF added to disk block cache as home TLF")
+
+type evictionPriority int
+
+const (
+	priorityNotHome evictionPriority = iota
+	priorityPublicHome
+	priorityPrivateHome
+)
+
 // DiskBlockCacheLocal is the standard implementation for DiskBlockCache.
 type DiskBlockCacheLocal struct {
 	config     diskBlockCacheConfig
@@ -80,11 +92,18 @@ type DiskBlockCacheLocal struct {
 	// Track the number of blocks in the cache per TLF and overall.
 	tlfCounts map[tlf.ID]int
 	numBlocks int
+	// Track the number of blocks in the cahce per eviction priority,
+	// for easy eviction counting.
+	priorityBlockCounts map[evictionPriority]int
+	priorityTlfMap      map[evictionPriority]map[tlf.ID]int
 	// Track the aggregate size of blocks in the cache per TLF and overall.
 	tlfSizes  map[tlf.ID]uint64
 	currBytes uint64
 	// Track the last unref'd revisions for each TLF.
 	tlfLastUnrefs map[tlf.ID]kbfsmd.Revision
+	// Don't evict files from the user's private or public home directory.
+	// Higher numbers are more important not to evict.
+	homeDirs map[tlf.ID]evictionPriority
 
 	startedCh  chan struct{}
 	startErrCh chan struct{}
@@ -218,18 +237,25 @@ func newDiskBlockCacheLocalFromStorage(
 		evictSizeMeter:           NewCountMeter(),
 		deleteCountMeter:         NewCountMeter(),
 		deleteSizeMeter:          NewCountMeter(),
+		homeDirs:                 map[tlf.ID]evictionPriority{},
 		log:                      log,
 		blockDb:                  blockDb,
 		metaDb:                   metaDb,
 		tlfDb:                    tlfDb,
 		lastUnrefDb:              lastUnrefDb,
 		tlfCounts:                map[tlf.ID]int{},
-		tlfSizes:                 map[tlf.ID]uint64{},
-		tlfLastUnrefs:            map[tlf.ID]kbfsmd.Revision{},
-		startedCh:                startedCh,
-		startErrCh:               startErrCh,
-		shutdownCh:               make(chan struct{}),
-		closer:                   closer,
+		priorityBlockCounts:      map[evictionPriority]int{},
+		priorityTlfMap: map[evictionPriority]map[tlf.ID]int{
+			priorityPublicHome:  {},
+			priorityPrivateHome: {},
+			priorityNotHome:     {},
+		},
+		tlfSizes:      map[tlf.ID]uint64{},
+		tlfLastUnrefs: map[tlf.ID]kbfsmd.Revision{},
+		startedCh:     startedCh,
+		startErrCh:    startErrCh,
+		shutdownCh:    make(chan struct{}),
+		closer:        closer,
 	}
 	// Sync the block counts asynchronously so syncing doesn't block init.
 	// Since this method blocks, any Get or Put requests to the disk block
@@ -369,6 +395,12 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsAndUnrefsFromDb() error {
 
 	tlfCounts := make(map[tlf.ID]int)
 	tlfSizes := make(map[tlf.ID]uint64)
+	priorityBlockCounts := make(map[evictionPriority]int)
+	priorityTlfMap := map[evictionPriority]map[tlf.ID]int{
+		priorityNotHome:     {},
+		priorityPublicHome:  {},
+		priorityPrivateHome: {},
+	}
 	numBlocks := 0
 	totalSize := uint64(0)
 	iter := cache.metaDb.NewIterator(nil, nil)
@@ -382,6 +414,8 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsAndUnrefsFromDb() error {
 		size := uint64(metadata.BlockSize)
 		tlfCounts[metadata.TlfID]++
 		tlfSizes[metadata.TlfID] += size
+		priorityBlockCounts[cache.homeDirs[metadata.TlfID]]++
+		priorityTlfMap[cache.homeDirs[metadata.TlfID]][metadata.TlfID]++
 		numBlocks++
 		totalSize += size
 	}
@@ -389,6 +423,8 @@ func (cache *DiskBlockCacheLocal) syncBlockCountsAndUnrefsFromDb() error {
 	cache.numBlocks = numBlocks
 	cache.tlfSizes = tlfSizes
 	cache.currBytes = totalSize
+	cache.priorityTlfMap = priorityTlfMap
+	cache.priorityBlockCounts = priorityBlockCounts
 
 	cache.log.Debug("| syncBlockCountsAndUnrefsFromDb block counts done")
 
@@ -645,6 +681,8 @@ func (cache *DiskBlockCacheLocal) Put(
 		cache.config.DiskLimiter().commitOrRollback(ctx, cache.cacheType,
 			encodedLen, 0, true, "")
 		cache.tlfCounts[tlfID]++
+		cache.priorityBlockCounts[cache.homeDirs[tlfID]]++
+		cache.priorityTlfMap[cache.homeDirs[tlfID]][tlfID]++
 		cache.numBlocks++
 		encodedLenUint := uint64(encodedLen)
 		cache.tlfSizes[tlfID] += encodedLenUint
@@ -765,6 +803,8 @@ func (cache *DiskBlockCacheLocal) deleteLocked(ctx context.Context,
 	// Update the cache's totals.
 	for k, v := range removalCounts {
 		cache.tlfCounts[k] -= v
+		cache.priorityBlockCounts[cache.homeDirs[k]] -= v
+		cache.priorityTlfMap[cache.homeDirs[k]][k] -= v
 		cache.numBlocks -= v
 		cache.tlfSizes[k] -= removalSizes[k]
 		cache.currBytes -= removalSizes[k]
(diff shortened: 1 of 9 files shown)


libkbfs/disk_block_cache.go
```diff
@@ -896,35 +902,52 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
...
-
-	for i := 0; i < numElements; i++ {
-		if !iter.Next() {
-			break
+	maxPriorityForEviction := 0
+
+	// Repeatedly double the search space until there are enough evictable
+	// blocks. If we reach the whole cache size, then give up and relax which
+	// blocks can be evicted.
+	for ; blockIDs.Len() < numBlocks && maxPriorityForEviction < 2; numElements *= 2 {
```
strib: Since 2 is a valid priority, don't we want `maxPriorityForEviction <= 2`?  Also, these priority levels should be global constants probably, for readability.

libkbfs/disk_block_cache.go
```diff
@@ -1102,6 +1125,30 @@ func (cache *DiskBlockCacheLocal) DoesSyncCacheHaveSpace(
 	return limiterStatus.SyncCacheByteStatus.UsedFrac <= .99
 }
 
+// AddHomeTLF adds a TLF marked as "home" so that the blocks from it are
+// less likely to be evicted, as well as whether this is their public or
+// private TLF, where the public TLF's files are more likely to be evicted
+// than the private one's.
+func (cache *DiskBlockCacheLocal) AddHomeTLF(ctx context.Context,
+	tlfID tlf.ID, tlfType tlf.Type) error {
```
strib: No need to pass in `tlfType` -- you can get it from `tlfID.Type()`.

libkbfs/disk_block_cache.go
```diff
@@ -1102,6 +1125,30 @@ func (cache *DiskBlockCacheLocal) DoesSyncCacheHaveSpace(
 	return limiterStatus.SyncCacheByteStatus.UsedFrac <= .99
 }
 
+// AddHomeTLF adds a TLF marked as "home" so that the blocks from it are
+// less likely to be evicted, as well as whether this is their public or
+// private TLF, where the public TLF's files are more likely to be evicted
+// than the private one's.
+func (cache *DiskBlockCacheLocal) AddHomeTLF(ctx context.Context,
```
strib: Both this function and `ClearHomeTlf` need to take `cache.lock`.

libkbfs/interfaces.go
```diff
@@ -1312,6 +1312,14 @@ type DiskBlockCache interface {
...
 	// true.
 	DoesSyncCacheHaveSpace(ctx context.Context) bool
+	// AddHomeTLF adds a TLF marked as "home" so that the blocks from it are
+	// less likely to be evicted, as well as whether this is their public or
+	// private TLF, where the public TLF's files are more likely to be evicted
+	// than the private one's.
+	AddHomeTLF(ctx context.Context, tlfID tlf.ID, tlfType tlf.Type) error
+	// ClearHomeTLF should be called on logout so that the old user's TLFs
+	// are not still marked as home.
+	ClearHomeTLF(ctx context.Context) error
```
strib: Should this be plural, e.g. `ClearHomeTLFs`?

libkbfs/disk_block_cache.go
```diff
@@ -1102,6 +1125,30 @@ func (cache *DiskBlockCacheLocal) DoesSyncCacheHaveSpace(
 	return limiterStatus.SyncCacheByteStatus.UsedFrac <= .99
 }
 
+// AddHomeTLF adds a TLF marked as "home" so that the blocks from it are
```
strib: You can use the ".. implements the DiskBlockCache interface..." style of comment here and below, for functions that have comments in the interface definition.  Otherwise we have to remember to update the comment in two places if it ever changes.

libkbfs/disk_block_cache.go
```diff
@@ -79,6 +82,9 @@ type DiskBlockCacheLocal struct {
 	currBytes uint64
 	// Track the last unref'd revisions for each TLF.
 	tlfLastUnrefs map[tlf.ID]kbfsmd.Revision
+	// Don't evict files from the user's private or public home directory.
+	// Higher numbers are more important not to evict.
+	homeDirs map[tlf.ID]int
```
strib: This should probably be initialized in the constructor, just in case someone tries to write to it before they are logged in.

libkbfs/disk_block_cache.go
```diff
@@ -896,35 +902,52 @@ func (cache *DiskBlockCacheLocal) evictLocked(ctx context.Context,
...
 	blockIDs := make(blockIDsByTime, 0, numElements)
-
-	for i := 0; i < numElements; i++ {
-		if !iter.Next() {
-			break
+	maxPriorityForEviction := 0
+
+	// Repeatedly double the search space until there are enough evictable
+	// blocks. If we reach the whole cache size, then give up and relax which
+	// blocks can be evicted.
```
jzila: I'm worried that this will be woefully slow in the situations that we're trying to cover. Especially since we have to iterate through `numElements` each time. In the worst case, we're talking iterating through 2*n elements per max priority. Additionally, you have one loop here iterating over 2 iteration variables: 1 for the `numElements` doubling, and 1 for `maxPriorityForEviction`. So we'll actually iterate over 2*n elements the first time, then n elements 2 more times, 1 per priority.

The pathological case might be quite common--I'm assuming people's caches will be mostly taken up by their own TLF's blocks.

The worst part is: this will happen every time an eviction is attempted, so once you're in the pathological scenario, you'll get slow evictions all the time.

libkbfs/disk_block_cache.go
jzila: +1 to constant.

libkbfs/disk_block_cache.go
strib: I'm to blame for the initial idea for the algorithm.  It was just the easiest, fastest thing I could think of.

Maybe we can we use `cache.tlfDb` to pick candidate keys that aren't in the home TLFs?  That should be pretty fast and straightforward.

I want to avoid re-doing the cache layout if possible, since that sounds hard.

strib: Cool.  Sorry, I should have read your overall comment more closely.

I think it should be pretty easy.  We can just make a tlf key with a random ID + the zero block key, and see what the next TLF is, skipping the home directories.  Do that `numElements` times. 
 And we can use `cache.tlfCounts` to check ahead of time if the only TLFs we know about are the home dirs, so we don't get stuck doing a bunch of skip-aheads for no reason.

jzila: Yes, the `cache.tlfCounts` solution is the right way to go, I think.

libkbfs/disk_block_cache.go
jakob223: So that's the rationale for those comments. I always find them mildly annoying as they introduce an extra step, but I guess it makes sense since we'd forget to update them.

jakob223:
@jzila @strib redid this, addressed all the comments, there's still a few TODOs in there that I'm not sure which way to go on if anyone has opinions.

libkbfs/disk_block_cache.go
```diff
@@ -879,12 +895,42 @@ func (cache *DiskBlockCacheLocal) evictFromTLFLocked(ctx context.Context,
...
-// evictionConsiderationFactor block IDs.  We sort the resulting blocks by
-// value (LRU time) and pick the minimum numBlocks. We then call cache.Delete()
-// on that list of block IDs.
+func (cache *DiskBlockCacheLocal) randomTlfForEvictionLocked(minEvictBlocks int) (tlf.
+	ID, error) {
+	totalCount := 0
+	var priorityToEvict evictionPriority
+	for priorityToEvict = priorityNotHome; (totalCount <= minEvictBlocks) &&
+		(priorityToEvict <= priorityPrivateHome); priorityToEvict++ {
+		for tlfId, count := range cache.tlfCounts {
```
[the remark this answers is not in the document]
jakob223: This doesn't order them by ID, it's basically constructing your `priorityBlockCounts` struct each time in order to choose one to evict from below. But I agree that your strategy is better because it's more flexible and doesn't redo work.

libkbfs/disk_block_cache.go
```diff
@@ -879,53 +919,115 @@ func (cache *DiskBlockCacheLocal) evictFromTLFLocked(ctx context.Context,
...
+	priority evictionPriority) []weightedByCount {
+	weightedSlice := make([]weightedByCount,
+		len(cache.priorityTlfMap[priority]))
+	idx := 0
+	// Use an exponential distribution to ensure the weights are
+	// correctly used.
+	// See http://utopia.duth.gr/~pefraimi/research/data/2007EncOfAlg.pdf
+	for tlfId, count := range cache.priorityTlfMap[priority] {
+		weightedSlice[idx] = weightedByCount{
+			key:   rand.ExpFloat64() * float64(count),
```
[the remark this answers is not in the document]
jzila: Also if this is correct, then you'll need to make sure that tlfs are removed from the priority map when their count reaches 0.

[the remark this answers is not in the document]
jakob223: 1. A comment on a stackoverflow post somewhere led be to believe this was equivalent, but I just did some math and was unable to convince myself of that. I'll change it to what you wrote.
2. Unless you think that enough empty TLFs will accumulate to slow down this process, it shouldn't be a problem for them to be empty.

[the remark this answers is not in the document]
jzila: The emptiness will be a problem because you'll get division by zero.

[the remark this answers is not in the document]
jakob223: Dividing by `float64(0)` gives you `+Inf`, which works in a sane way.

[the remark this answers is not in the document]
jzila: Does it? https://play.golang.org/p/_V1VH1BoWwK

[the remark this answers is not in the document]
jakob223: That's a compile-time error when the compiler knows you're going to divide by zero and tries to stop you. It works at runtime though.

https://play.golang.org/p/7Qgr8vsSTmL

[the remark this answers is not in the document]
jzila: Neat, TIL! Seems like it'll work in that expression context as well.

libkbfs/disk_block_cache.go
```diff
@@ -879,53 +919,118 @@ func (cache *DiskBlockCacheLocal) evictFromTLFLocked(ctx context.Context,
...
-// evictLocked evicts a number of blocks from the cache.  We choose a pivot
-// variable b randomly. Then begin an iterator into cache.metaDb.Range(b,
-// MaxBlockID) and iterate from there to get numBlocks *
-// evictionConsiderationFactor block IDs.  We sort the resulting blocks by
-// value (LRU time) and pick the minimum numBlocks. We then call cache.Delete()
-// on that list of block IDs.
+// weightedByCount is used to shuffle TLF IDs, weighting by per-TLF block count.
+type weightedByCount struct {
+	key   float64
+	value tlf.ID
```
jzila: Please name this `tlfID`.

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +760,51 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+	cache.AddHomeTLF(ctx, homePublicTLF)
+
+	t.Log("Seed the cache with blocks")
+	numTlfs := 12
+	numBlocksPerTlf := 5
+	numBlocks := numTlfs * numBlocksPerTlf * 2
+	seedDiskBlockCacheForTest(t, ctx, cache, config, numTlfs,
+		numBlocksPerTlf, tlf.Private)
+	seedDiskBlockCacheForTest(t, ctx, cache, config, numTlfs,
+		numBlocksPerTlf, tlf.Public)
```
jzila: public/private doesn't matter for these because they aren't home TLFs, right?

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +760,51 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+	cache, config := initDiskBlockCacheTest(t)
+	defer shutdownDiskBlockCacheTest(cache)
+
+	ctx := context.Background()
+
+	t.Log("Set home directories on the cache")
+	homeTLF := tlf.FakeID(11, tlf.Private)
+	cache.AddHomeTLF(ctx, homeTLF)
+	homePublicTLF := tlf.FakeID(11, tlf.Public)
+	cache.AddHomeTLF(ctx, homePublicTLF)
```
jzila: Where do you seed blocks for these two TLFs?

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +760,51 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+	numTlfs := 12
+	numBlocksPerTlf := 5
+	numBlocks := numTlfs * numBlocksPerTlf * 2
+	seedDiskBlockCacheForTest(t, ctx, cache, config, numTlfs,
+		numBlocksPerTlf, tlf.Private)
+	seedDiskBlockCacheForTest(t, ctx, cache, config, numTlfs,
+		numBlocksPerTlf, tlf.Public)
+
+	countToEvict := numBlocks - numBlocksPerTlf*2 - 2
+	t.Log("Evict most of the blocks in the cache")
```
jzila: This is confusing. IMO the home TLFs should also be seeded with `numBlocksPerTlf` blocks, then this can be simpler.

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +760,51 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+	t.Log("Set home directories on the cache")
+	homeTLF := tlf.FakeID(11, tlf.Private)
+	cache.AddHomeTLF(ctx, homeTLF)
+	homePublicTLF := tlf.FakeID(11, tlf.Public)
+	cache.AddHomeTLF(ctx, homePublicTLF)
+
+	t.Log("Seed the cache with blocks")
+	numTlfs := 12
+	numBlocksPerTlf := 5
+	numBlocks := numTlfs * numBlocksPerTlf * 2
```
jzila: I don't follow this--why is this number true?

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +760,51 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, numBlocksPerTlf, cache.syncCache.tlfCounts[homeTLF])
+	require.Equal(t, 2, cache.syncCache.tlfCounts[homePublicTLF])
+
+	// evict the last 2 public-home-dir blocks and n-2 private homedir blocks
+	countToEvict = numBlocksPerTlf
+	t.Log("Evict most of the remaining blocks in the cache")
+	_, _, err = cache.syncCache.evictLocked(ctx, countToEvict)
+	require.NoError(t, err)
+	require.Equal(t, 2, cache.syncCache.tlfCounts[homeTLF])
```
jzila: IMO you're on the right track for this test, but you should explicitly have these be true:

1. Make sure ever TLF is seeded, including the 2 home TLFs.
2. Distribute the blocks exponentially over the non-home TLFs. Use LOTS of blocks to get better statistical behavior.
3. Evict half the non-home TLF blocks using small eviction sizes (each eviction is 10% of the starting size of the _smallest_ TLF in your distribution). (tlf sizes 100, 200, 400, 800, 1600, eviction size 10). Verify that the non-home TLFs have been reduced in size by about half at the end of this step. (give or take 20%, so 30-70%)
4. Evict the rest of the non-home TLF blocks in ~2 evictions.
5. Evict enough blocks to get rid of the public home TLF.
6. Evict enough blocks to get rid of the private home TLF.

jzila: If step 2 leaves you a bit flaky, run the whole thing 3 times and take the average.

libkbfs/disk_block_cache_test.go
jakob223: One of them from each is a home TLF.

libkbfs/disk_block_cache_test.go
jakob223: The helper that seeds blocks for tests seeds the first N TLF IDs, which include these.

libkbfs/disk_block_cache_test.go
jakob223: They are seeded with that many blocks, but I should add comments making that clear.

libkbfs/disk_block_cache_test.go
jakob223: Each of the two seed calls adds `numTlfs * numBlocksPerTlf` blocks.

libkbfs/disk_block_cache_test.go
jzila: Hmm...not thrilled about how opaque that is.

Additionally, you can't have different TLFs have the same ID, even if the TLF type is different. The cache only indexes on TLF ID, not on type.
https://github.com/keybase/kbfs/blob/2ebbe681777cc1e03dbcc2ae84cefc6488290f3c/libkbfs/disk_block_cache.go#L418-L420
https://github.com/keybase/kbfs/blob/2ebbe681777cc1e03dbcc2ae84cefc6488290f3c/tlf/id.go#L175-L177

libkbfs/disk_block_cache_test.go
```diff
@@ -753,6 +754,127 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+
+	_, _, err = cache.syncCache.evictLocked(ctx, evictionSize2)
+	require.NoError(t, err)
+
+	// Make sure the home TLFs are minimally touched.
+	require.Equal(t, homeTLFBlocksEach, cache.syncCache.tlfCounts[homeTLF])
+	require.Equal(t, homeTLFBlocksEach-publicHomeEvict,
+		cache.syncCache.tlfCounts[homePublicTLF])
+
+	t.Log(" Evict enough blocks to get rid of the public home TLF.")
```
jzila: Extra space at the beginning of this log line.

jzila:
Closing in favor of keybase/client#15329

PR #1945, opened 2018-12-04
prefetcher: track prefetch progress 

strib:
(This PR includes the changes of #1948, so please review that one first and wait until it's merged to review this one.)

This tracks, for each block being tracked in the prefetcher, the number of known bytes rooted at that block, and the number of bytes fetched so far.  This is a bit crude because the total doesn't represent ALL of the bytes rooted at the block; just the bytes the prefetcher has seen so far on its traversal.  So for example if a lot of a TLF has already been synced and the prefetcher avoids traversing down that branch of the TLF, those bytes won't be included in the total.  Also, the total can increase during the traversal.

But it's a good start I think.  If the GUI needs to know the full total for a TLF, we can get that easily from the MD object.  But it's hard to compute a progress percentage from that, so for now I say we go this way and see how far it gets us.

Issue: KBFS-3575

### DIFF
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -44,6 +44,7 @@ type prefetchRequest struct {
 	prefetchStatus PrefetchStatus
 	action         BlockRequestAction
 	sendCh         chan<- <-chan struct{}
+	statusCh       chan<- PrefetchByteStatus
 }
 
 type ctxPrefetcherTagKey int
@@ -57,17 +58,16 @@ const (
 )
 
 type prefetch struct {
-	subtreeBlockCount   int
-	subtreeBytesFetched uint64
-	subtreeBytesTotal   uint64
-	start               time.Time
-	subtreeTriggered    bool
-	req                 *prefetchRequest
+	subtreeBlockCount int
+	subtreeTriggered  bool
+	req               *prefetchRequest
 	// Each refnonce for this block ID can have a different set of parents.
 	parents map[kbfsblock.RefNonce]map[BlockPointer]bool
 	ctx     context.Context
 	cancel  context.CancelFunc
 	waitCh  chan struct{}
+
+	PrefetchByteStatus
 }
 
 func (p *prefetch) Close() {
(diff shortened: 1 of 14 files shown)


libkbfs/prefetcher.go
```diff
@@ -203,9 +216,10 @@ func (p *blockPrefetcher) applyToParentsRecursive(
 // hit get marked complete and deleted.
 // TODO: If we ever hit a lower number than the child, panic.
 func (p *blockPrefetcher) completePrefetch(
-	numBlocks int) func(kbfsblock.ID, *prefetch) {
+	numBlocks int, numBytes uint64) func(kbfsblock.ID, *prefetch) {
```
jakob223: Maybe add an explanation of `numBytes` in the comment? I get what it does now but it took a minute of looking over the code which meant I forgot where I came from to look at what it meant in the first place.

libkbfs/prefetcher.go
```diff
@@ -906,6 +996,31 @@ func (p *blockPrefetcher) WaitChannelForBlockPrefetch(
...
+		return PrefetchProgress{}, errors.New("Already shut down")
+	case <-ctx.Done():
+		return PrefetchProgress{}, ctx.Err()
+	}
+	// Wait for response.
+	select {
+	case status := <-c:
+		return status, nil
+	case <-p.shutdownCh:
+		return PrefetchProgress{}, errors.New("Already shut down")
```
jakob223: Should this error be a variable?

libkbfs/prefetcher.go
```diff
@@ -165,6 +174,10 @@ func (p *blockPrefetcher) newPrefetch(count int, triggered bool,
 		ctx:               ctx,
 		cancel:            cancel,
 		waitCh:            make(chan struct{}),
+		PrefetchProgress: PrefetchProgress{
+			SubtreeBytesTotal: bytes,
+			Start:             p.config.Clock().Now(),
```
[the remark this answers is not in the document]
strib: The tests often need to fake out the time, so we always use an interface for the clock instead of directly using `time`, to have more control over it during testing.

libkbfs/prefetcher.go
strib: Sure!

libkbfs/prefetcher.go
strib: Done, thanks.

strib:
@jzila this has other reviews, but I thought you'd like a say too.  If not, let me know and I'll merge as is.

jzila:
You can actually merge this at your own pace.

strib:
> You can actually merge this at your own pace.

Sadly, I still need to figure out a fix for some flaky prefetcher behavior, which I detailed in the jira issue.

jzila:
Closing in favor of keybase/client#15328

PR #1950, opened 2018-12-05
cache dirChildren in autogit and fast mode in libpages

songgao:
@strib I added a fast mode for Pages, hopefully temporary before we figure out a better solution that can benefit autogit itself too.

### DIFF
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -613,9 +613,10 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 	}()
 
 	if fs.empty && (filename == "" || filename == ".") {
-		// TODO: use FileInfoFast.
-		return &FileInfo{
-			fs: fs,
+		// Always use FileInfoFast here for this no matter fast mode is enabled
+		// or not.
+		return &FileInfoFast{
+			name: filename,
 			ei: libkbfs.EntryInfo{
 				Type: libkbfs.Dir,
 			},
@@ -781,9 +782,10 @@ func (fs *FS) Lstat(filename string) (fi os.FileInfo, err error) {
 	}()
 
 	if fs.empty && (filename == "" || filename == ".") {
-		// TODO: use FileInfoFast.
-		return &FileInfo{
-			fs: fs,
+		// Always use FileInfoFast here for this no matter fast mode is enabled
+		// or not.
+		return &FileInfoFast{
+			name: filename,
 			ei: libkbfs.EntryInfo{
 				Type: libkbfs.Dir,
 			},
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -600,7 +600,7 @@ func (fs *FS) makeFileInfo(
 		fs:   fs,
 		ei:   ei,
 		node: node,
-		name: node.GetBasename(),
+		name: name,
 	}
 }
 
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -241,8 +241,8 @@ func NewUnwrappedFS(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
 	return newFS(
-		ctx, config, tlfHandle, branch,
-		subdir, uniqID, priority, true, readwrite)
+		ctx, config, tlfHandle, branch, subdir, uniqID, priority, true,
+		readwrite)
 }
 
 // NewReadonlyFS returns a new FS instance, chroot'd to the given TLF
@@ -260,8 +260,8 @@ func NewReadonlyFS(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
 	return newFS(
-		ctx, config, tlfHandle, branch,
-		subdir, uniqID, priority, false, readonly)
+		ctx, config, tlfHandle, branch, subdir, uniqID, priority, false,
+		readonly)
 }
 
 // NewFSIfExists returns a new FS instance, chroot'd to the given TLF
@@ -279,8 +279,8 @@ func NewFSIfExists(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
 	return newFS(
-		ctx, config, tlfHandle, branch,
-		subdir, uniqID, priority, false, readwriteNoCreate)
+		ctx, config, tlfHandle, branch, subdir, uniqID, priority, false,
+		readwriteNoCreate)
 }
 
 // NewFS returns a new FS instance, chroot'd to the given TLF and
@@ -293,8 +293,8 @@ func NewFS(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
 	return newFS(
-		ctx, config, tlfHandle, branch,
-		subdir, uniqID, priority, false, readwrite)
+		ctx, config, tlfHandle, branch, subdir, uniqID, priority, false,
+		readwrite)
 }
 
 // lookupOrCreateEntryNoFollow looks up the entry for a file in a
--- a/libfs/fs.go
+++ b/libfs/fs.go
@@ -241,8 +241,8 @@ func NewUnwrappedFS(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
 	return newFS(
-		ctx, config, tlfHandle, branch, subdir, uniqID, priority, true,
-		readwrite)
+		ctx, config, tlfHandle, branch,
+		subdir, uniqID, priority, true, readwrite)
 }
 
 // NewReadonlyFS returns a new FS instance, chroot'd to the given TLF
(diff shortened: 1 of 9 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -2109,6 +2109,7 @@ func (fbo *folderBranchOps) getDirChildren(ctx context.Context, dir Node) (
 		if err != nil {
 			return nil, err
 		}
+		fbo.log.CDebugf(ctx, "Got children using an FS")
```
strib: Needed?

libgit/browser.go
```diff
@@ -39,6 +40,12 @@ type Browser struct {
 	tree  *object.Tree
 	root  string
 	mtime time.Time
+
+	lock sync.RWMutex
+	// Caches ReadDir results. Since the *Browser object gets blown away when
+	// HEAD changes, we don't need to worry about invalidation -- unless memory
+	// footprint becomes a problem.
+	dirChildrenCache map[string][]os.FileInfo
```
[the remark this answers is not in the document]
strib: Also, it's a shame this doesn't help improve `LStat/Stat` commands.  Can we give it another level indexed by entry name, instead of having it as an array?  Or would reconstructing the array from that be too expensive?

libgit/browser.go
```diff
@@ -228,9 +236,23 @@ func (b *Browser) Join(elem ...string) string {
 
 // ReadDir implements the billy.Filesystem interface for Browser.
 func (b *Browser) ReadDir(p string) (fis []os.FileInfo, err error) {
+	if p == "" {
+		p = "."
+	}
+
+	b.lock.RLock()
```
strib: All this locking is ridiculous.  Would be much clearer with a helper function that returns a bool if the returned list should be used or not.

libfs/fs.go
```diff
@@ -208,8 +208,8 @@ outer:
 func NewUnwrappedFS(ctx context.Context, config libkbfs.Config,
 	tlfHandle *libkbfs.TlfHandle, branch libkbfs.BranchName, subdir string,
 	uniqID string, priority keybase1.MDPriority) (*FS, error) {
-	return newFS(
-		ctx, config, tlfHandle, branch, subdir, uniqID, priority, true, false)
+	return newFS(ctx, config, tlfHandle, branch,
```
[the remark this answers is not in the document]
songgao: Oops; I went with a `NewFastFS` initially, but realized it wouldn't work for new git FS'es created inside, and switched to the ctx approach, then forgot to change this back. Will do!

libkbfs/folder_branch_ops.go
songgao: nope; good catch!

libgit/browser.go
songgao: > ridiculous

That's what "R" in RLock stands for!

songgao:
@strib Thanks for looking! I'll deploy this branch for now (so @maxtaco can go ahead and bump client version requirement), and address the feedback in the evening.

libgit/browser.go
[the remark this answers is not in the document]
songgao: I'm thinking we should probably plumb through a global `*lru.Cache` to be shared among all `*Browser` instances. For `kbpagesd` I think this would make a big difference for more active sites that are over the capacity, compared to having per-browser caches. WDYT?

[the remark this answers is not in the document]
songgao: > Can we give it another level indexed by entry name, instead of having it as an array? Or would reconstructing the array from that be too expensive?

Good idea; I think it should be fine, but will do some simple benchmark to be sure.

songgao:
@strib please take another look; thanks! I added a LRU cache shared among all libgit.Browser instances, and made it work for both `LStat` and `ReadDir`. Also resolved the two TODOs regarding `FileInfoFast` from #1955.

I'm realizing the branch name is off-by-2-months, but I guess that's fine ...

songgao:
Also, from some of my basic tests, this is even faster for `kbpagesd` to list directories! likely due to the optimization for LStat.

libfs/fs.go
```diff
@@ -597,9 +613,10 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 	}()
 
 	if fs.empty && (filename == "" || filename == ".") {
-		// TODO: use FileInfoFast.
-		return &FileInfo{
-			fs: fs,
+		// Always use FileInfoFast here for this no matter fast mode is enabled
```
strib: "no matter" -> "whether"

libgit/shared_browser_cache.go
```diff
@@ -0,0 +1,150 @@
...
+
+	lru "github.com/hashicorp/golang-lru"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+type browserCacheEntryType int
+
+const (
+	_ browserCacheEntryType = iota
+	fileInfoBowserCacheEntry
```
strib: "Bowser" -> "Browser"

libgit/shared_browser_cache.go
```diff
@@ -0,0 +1,150 @@
...
+	childrenPathsBowserCacheEntry
```
strib: "Bowser" -> "Browser"

libgit/shared_browser_cache.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type sharedInBrowserCache interface {
+	setFileInfo(
+		commitHash plumbing.Hash,
+		entryPathRelativeToRepoRoot string,
+		fi os.FileInfo,
```
strib: We don't usually do method definitions like this, with a trailing `,` on the last element and a line break before the `)`.  Took me a while to figure out what these were.  (Same with the function definitions below.)

libgit/browser.go
```diff
@@ -197,7 +203,11 @@ func (b *Browser) Lstat(filename string) (fi os.FileInfo, err error) {
 
 	// Git doesn't keep track of the mtime of individual files
 	// anywhere, so just use the timestamp from the commit.
-	return &browserFileInfo{entry, size, b.mtime}, nil
+	fi = &browserFileInfo{entry, size, b.mtime}
+
+	b.sharedCache.setFileInfo(b.commitHash, path.Join(b.root, filename), fi)
```
strib: I don't see a corresponding `getFileInfo` here or in `Stat()`.

libgit/browser.go
```diff
@@ -239,13 +260,18 @@ func (b *Browser) ReadDir(p string) (fis []os.FileInfo, err error) {
...
 	}
 
+	childrenPathsToCache := []string(nil)
 	for _, e := range dirTree.Entries {
 		fi, err := b.Lstat(e.Name)
 		if err != nil {
 			return nil, err
 		}
 		fis = append(fis, fi)
+		childrenPathsToCache = append(childrenPathsToCache, path.Join(b.root, e.Name))
```
strib: Should this be `path.Join(cachePath, e.Name)`?

libgit/browser.go
```diff
@@ -228,9 +238,20 @@ func (b *Browser) Join(elem ...string) string {
 
 // ReadDir implements the billy.Filesystem interface for Browser.
 func (b *Browser) ReadDir(p string) (fis []os.FileInfo, err error) {
+	if p == "" {
+		p = "."
+	}
+
+	cachePath := path.Join(b.root, p)
```
strib: Also `path.Clean` this path?

libgit/browser.go
```diff
@@ -239,13 +260,18 @@ func (b *Browser) ReadDir(p string) (fis []os.FileInfo, err error) {
 		}
 	}
 
+	childrenPathsToCache := []string(nil)
```
strib:
```go
childrenPathsToCache := make([]string, 0, len(dirTree.Entries))
```

libgit/browser.go
songgao: I think it should be joined against b.root. The b.Lstat above is relative to the root, so e.Name must already include the directory it's in right?

libgit/browser.go
songgao: good catch!

libgit/browser.go
strib: You might be thinking of `os.File.Name()`, which is the whole path passed to `Open`?  For an `os.FileInfo.Name` (https://golang.org/pkg/os/#FileInfo), it's described as just the "base name".  And that's what `libkbfs.Node.GetBasename()` returns as well (which is what's used to construct the file info returned by `Lstat()`.

libgit/browser.go
songgao: `path.Join` already does a `Clean` before returning: https://golang.org/src/path/path.go?s=4034:4066#L145 so no need here

libgit/browser.go
songgao: You know, sometimes when you write too much JavaScript, you tend to forget the performance sweetness like this from Go 😂

libgit/browser.go
songgao: As talked offline, `b.Lstat` is supposed to take `path.Join(p, e.Name)` too. Fixed!

songgao:
@strib ready for another look thanks!

libgit/browser.go
```diff
@@ -260,14 +264,14 @@ func (b *Browser) ReadDir(p string) (fis []os.FileInfo, err error) {
...
+	childrenPathsToCache := make([]string, 0, len(dirTree.Entries))
 	for _, e := range dirTree.Entries {
-		fi, err := b.Lstat(e.Name)
+		fi, err := b.Lstat(path.Join(p, e.Name))
 		if err != nil {
 			return nil, err
 		}
 		fis = append(fis, fi)
-		childrenPathsToCache = append(childrenPathsToCache, path.Join(b.root, e.Name))
+		childrenPathsToCache = append(childrenPathsToCache, path.Join(b.root, p, e.Name))
```
strib: Can replace `b.root, p` with `cachePath`, since it already has them joined together.

libpages/root.go
```diff
@@ -129,7 +129,7 @@ func (r *Root) MakeFS(
 		if err != nil {
 			return CacheableFS{}, tlf.ID{}, nil, err
 		}
-		tlfFS, err := libfs.NewReadonlyFS(
+		tlfFS, err := libfs.NewFS(
```
strib: How does this fix the test?  Doesn't making this readonly improve the performance?

songgao: I haven't looked into it carefully yet, but I think it's related to the fact that `SubscribeToObsolete` requires it to not be empty, but NewReadonlyFS can get an empty one initially. I was reluctant to simply remove that nonEmpty check for this method, because I was not sure if we need to do something special if the folder is empty. Didn't have a chance to make sure of that yet.

songgao: Will merge this for now and try to figure this out later!

PR #1952, opened 2018-12-05
disk_block_cache: refactor interface to use cache types more

strib:
And remove cache types from `DiskBlockCacheLocal`, which doesn't need to implement the main interface.

This will be useful in future PRs when tracking last-unref'd revision in the working set cache, and checking for space in the working set cache.

Issue: KBFS-3525
Issue: KBFS-3527

### DIFF
--- a/libkbfs/disk_block_cache.go
+++ b/libkbfs/disk_block_cache.go
@@ -5,6 +5,7 @@
 package libkbfs
 
 import (
+	"fmt"
 	"io"
 	"math"
 	"path/filepath"
@@ -86,8 +87,6 @@ type DiskBlockCacheLocal struct {
 
 	closer func()
 }
-
-var _ DiskBlockCache = (*DiskBlockCacheLocal)(nil)
 
 // DiskBlockCacheStartState represents whether this disk block cache has
 // started or failed.
@@ -521,8 +520,8 @@ func (cache *DiskBlockCacheLocal) checkCacheLocked(method string) error {
 }
 
 // Get implements the DiskBlockCache interface for DiskBlockCacheLocal.
-func (cache *DiskBlockCacheLocal) Get(ctx context.Context, tlfID tlf.ID,
-	blockID kbfsblock.ID, _ DiskBlockCacheType) (buf []byte,
+func (cache *DiskBlockCacheLocal) Get(
+	ctx context.Context, tlfID tlf.ID, blockID kbfsblock.ID) (buf []byte,
 	serverHalf kbfscrypto.BlockCryptKeyServerHalf,
 	prefetchStatus PrefetchStatus, err error) {
 	cache.lock.RLock()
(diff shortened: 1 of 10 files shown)


libkbfs/disk_block_cache_wrapped.go
```diff
@@ -97,14 +97,28 @@ func newDiskBlockCacheWrapped(config diskBlockCacheConfig,
...
-		return false
+func (cache *diskBlockCacheWrapped) getCacheLocked(
+	cacheType DiskBlockCacheType) (*DiskBlockCacheLocal, error) {
+	if cacheType == DiskBlockSyncCache {
+		if cache.syncCache == nil {
+			return nil, errors.New("Sync cache not enabled")
+		}
+		return cache.syncCache, nil
+	}
+	return cache.workingSetCache, nil
```
jakob223: Is the working set cache never nil?

libkbfs/disk_block_cache.go
```diff
@@ -589,10 +588,9 @@ func (cache *DiskBlockCacheLocal) evictUntilBytesAvailable(
 }
 
 // Put implements the DiskBlockCache interface for DiskBlockCacheLocal.
-func (cache *DiskBlockCacheLocal) Put(ctx context.Context, tlfID tlf.ID,
-	blockID kbfsblock.ID, buf []byte,
-	serverHalf kbfscrypto.BlockCryptKeyServerHalf,
-	_ DiskBlockCacheType) (err error) {
```
jakob223: why did this used to take an unused cache type?

jakob223:
I accidentally clicked "review" instead of "approve" and now I don't see how to fix it..

libkbfs/disk_block_cache.go
```diff
@@ -87,8 +88,6 @@ type DiskBlockCacheLocal struct {
 	closer func()
 }
 
-var _ DiskBlockCache = (*DiskBlockCacheLocal)(nil)
```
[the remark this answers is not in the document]
strib: Correct.  Since `DiskBlockCache` has so many methods that specify a cache type, it doesn't make sense for the local cache, which is already one of a specific type, to take that parameter.  And it's not used anywhere the generic interface is expected, so this makes everything cleaner.

libkbfs/disk_block_cache_test.go
```diff
@@ -608,7 +604,7 @@ func TestSyncBlockCacheStaticLimit(t *testing.T) {
 	t.Log("Test that disk cache eviction works when we hit the static limit.")
 	cache, config := initDiskBlockCacheTest(t)
 	standardCache := cache.syncCache
-	defer shutdownDiskBlockCacheTest(standardCache)
```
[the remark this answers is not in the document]
strib: Good question, I don't know either.  Probably just someone typed it that way without thinking, and it just worked since the local cache implemented the generic interface.

libkbfs/disk_block_cache_wrapped.go
strib: Correct.  As far as I know in production, neither is ever nil, but probably for historical reasons we still usually check the sync cache to make sure.  But I don't think there's any need to check for the working set cache -- if a `DiskBlockCache` exists at all, we can assume the working set cache is there.

libkbfs/disk_block_cache.go
strib: Because it had to satisfy the `DiskBlockCache` interface before, and now it doesn't have to anymore.

strib:
> I accidentally clicked "review" instead of "approve" and now I don't see how to fix it..

I think you can go back to the "Files changed" page to change it.

PR #1955, opened 2018-12-06
simplefs: use `NewFSIfExists` for read-only operations

strib:
We don't want to accidentally create a TLF when reading an empty TLF, so use `libfs.NewFSIfExists` in that case.

It would be nice to also use a read-only FS too, but if a `libkfbs.Node` gets created in read-only mode, it will be served to subsequent calls as read-only as well, even if they're done through a read-write file system.  Maybe that's something we want to fix in the future, but for now just make sure the TLF isn't created.


libfs/fs.go
```diff
@@ -108,31 +109,60 @@ func (e ErrNotADirectory) Error() string {
 	return fmt.Sprintf("%s is not a directory", e.Name)
 }
 
+type accessType int
+
+const (
+	readwrite accessType = iota
```
jakob223: nit: capitalization here confuses me.
also, maybe add a comment on `readwriteNoCreate` since it's non-obvious to me what create means without looking at the context of the PR.

libfs/fs.go
```diff
@@ -108,31 +109,60 @@ func (e ErrNotADirectory) Error() string {
...
+	if rootNode == nil {
+		if len(subdir) > 0 {
+			return nil, errors.Errorf(
+				"Subdir %s doesn't exist in empty TLF", subdir)
+		}
+		log.CDebugf(ctx,
+			"Returning empty FS for empty TLF=%s", tlfHandle.GetCanonicalName())
+		return &FS{
+			ctx: ctx,
+			fsInner: &fsInner{
```
jakob223: should this behave differently in the event that it's being opened read-only?

libfs/fs.go
```diff
@@ -526,6 +593,18 @@ func (fs *FS) Stat(filename string) (fi os.FileInfo, err error) {
 		err = translateErr(err)
 	}()
 
+	if fs.empty && (filename == "" || filename == ".") {
+		// TODO: use FileInfoFast.
```
jakob223: What's `FileInfoFast`?

libfs/fs.go
```diff
@@ -459,6 +515,13 @@ func (fs *FS) ensureParentDir(filename string) error {
 	return nil
 }
 
+func (fs *FS) checkEmpty() error {
```
jakob223: Unless there's a convention I'm unaware of, I feel like this would be more readable if it were called `requireNonempty` or something.

libfs/fs.go
strib: Re: capitalization: yeah, I was trying to be consistent with the `Readonly` capitalization we have (which maybe we inherited from somewhere else?  I don't remember.)  Anyway, I can change it if people feel strongly about it.

Will add a comment.

libfs/fs.go
strib: I don't think so.  An empty file system is inherently read-only already, since it doesn't allow any write operations.

libfs/fs.go
strib: That's something in an unmerged PR from @songgao: #1950.  It avoids needing to pass in a node here.

strib:
@jakob223 thanks for the review!  Added kbfs-hackers back on as a reviewer so either jzila or songgao can take a look too.

libfs/fs.go
songgao: Sorry I haven't got it merged yet; will try to get it done today or tomorrow!

simplefs/simplefs_test.go
```diff
@@ -174,14 +175,40 @@ func TestList(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	checkPendingOp(
+		ctx, t, sfs, opid, keybase1.AsyncOps_LIST, path1, keybase1.Path{}, true)
+	err = sfs.SimpleFSWait(ctx, opid)
+	require.NoError(t, err)
+	listResult, err = sfs.SimpleFSReadList(ctx, opid)
+	require.NoError(t, err)
+	require.Len(t, listResult.Entries, 0,
+		"Expected 0 directory entries in listing")
```
songgao: Do you think it's worthwhile to have a test for SimpleFSCopyRecursive too, to cover both the copy and the recursive path?

strib: Sure, easy enough.  Done!

strib:
Oops right, sorry, my brain messed up.  I'll look quickly to see if `CopyRecursive` would be easy to add too.

songgao:
Huh; good thing it revealed that then. Thanks!

PR #1957, opened 2018-12-08
libkbfs: prefetch recently-edited files for non-synced TLFs

strib:
This PR enables the deep-prefetching of recently-edited files on TLFs into the working-set cache, as per design's request.  To make this work:

* Introduce a new block request action flag that stops a prefetch completely as soon as the cache is full.  This is different from how deep-sync caches work, which need to be rescheduled.
* These new "stop-if-full" requests trigger deep-prefetches, just like a "deep-sync" request does. (At least until the cache fills up.)
* Whenever loading a TLF for prefetching (including for synced folders), suppress identify popups and errors so that the user doesn't get random popups.  If the user tries to access data in these TLFs, they should get the popup and errors.
* Change the partial sync code so that if it receives a config with a DISABLED mode, but with partial paths, it syncs them into the working-set cache with a "stop-if-full" action.

I'm a bit scared of some of the implications here, since it means someone you don't know can create a folder with you and start making you download stuff.  In a future PR I'll make sure the edit history only includes favorited folders, but that will have to wait for Jakob's work on the favorites cache.  Also, I've observed this using quite a bit of CPU for the first few minutes after startup.

Issue: KBFS-3525

### DIFF
--- a/libkbfs/prefetcher.go
+++ b/libkbfs/prefetcher.go
@@ -781,7 +781,8 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
 			if req.prefetchStatus == TriggeredPrefetch &&
 				!req.action.DeepSync() &&
 				(isPrefetchWaiting &&
-					req.action.Sync() == pre.req.action.Sync()) {
+					req.action.Sync() == pre.req.action.Sync() &&
+					req.action.StopIfFull() == pre.req.action.StopIfFull()) {
 				p.log.CDebugf(ctx, "prefetch already triggered for block ID "+
 					"%s", req.ptr.ID)
 				continue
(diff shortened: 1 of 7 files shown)


libkbfs/prefetcher.go
```diff
@@ -783,13 +801,16 @@ func (p *blockPrefetcher) run(testSyncCh <-chan struct{}) {
...
 			// full, to avoid enqueuing the child blocks when they
 			// aren't able to be cached.
-			if p.rescheduleIfNeeded(ctx, req) {
+			if doStop, doCancel := p.stopIfNeeded(ctx, req); doStop {
 				// This is inefficient since it'd be better to know if
 				// the `subtreeBlockCount` below is 0, or if `isTail`
 				// below is true before needlessly rescheduling this.
 				// But currently that requires some complexity to
 				// figure out, so for now just do this early and
 				// revisit if it becomes a problem.
```
jzila: I think this comment is stale now, or belongs in the `stopIfNeeded` method.

PR #1960, opened 2018-12-10
disk_block_cache: clean out working set cache regularly

jakob223:
Modify the background cleanout of the sync cache to also cleanout the working set cache.

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -114,11 +114,11 @@ type folderBlockManager struct {
 	// merged revision for this TLF.
 	latestMergedChan chan struct{}
 
-	// cleanSyncCacheGroup tracks the outstanding sync-cache cleanings.
-	cleanSyncCacheGroup kbfssync.RepeatedWaitGroup
+	// cleanDiskCachesGroup tracks the outstanding disk-cache cleanings.
+	cleanDiskCachesGroup kbfssync.RepeatedWaitGroup
 
-	cleanSyncCacheCancelLock sync.Mutex
-	cleanSyncCacheCancel     context.CancelFunc
+	cleanDiskCacheCancelLock sync.Mutex
+	cleanDiskCacheCancel     context.CancelFunc
 
 	helper fbmHelper
 
@@ -167,7 +167,7 @@ func newFolderBlockManager(
 	go fbm.deleteBlocksInBackground()
 	if qrEnabled {
 		go fbm.reclaimQuotaInBackground()
-		go fbm.cleanSyncCacheInBackground()
+		go fbm.cleanDiskCachesInBackground()
 	}
 	return fbm
 }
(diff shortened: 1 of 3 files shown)


libkbfs/interfaces.go
```diff
@@ -1274,6 +1274,19 @@ const (
...
+	case DiskBlockSyncCache:
+		return "sync block cache"
+	case DiskBlockWorkingSetCache:
+		return "working set cache"
+	case DiskBlockAnyCache:
+		return "any disk block cache"
+	default:
+		return "unknown disk block cache"
+	}
+}
```
songgao: Maybe just use the variable name in the returned string for easy grepping?

libkbfs/folder_block_manager.go
```diff
@@ -1306,10 +1306,21 @@ func (fbm *folderBlockManager) clearLastQRData() {
...
+	}
+	return fbm.doCleanDiskCache(DiskBlockWorkingSetCache)
+}
+
+func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
+	err error) {
 	syncConfig := fbm.config.GetTlfSyncState(fbm.id)
-	if syncConfig.Mode == keybase1.FolderSyncMode_DISABLED {
+	if syncConfig.Mode == keybase1.FolderSyncMode_DISABLED &&
+		cacheType == DiskBlockSyncCache {
```
songgao: It seems the syncConfig is only used for determining this. Would it make sense to move this check into `doCleanDiskCaches` so if `DiskBlockSyncCache` is `DISABLED` we'd just not call this function?

libkbfs/interfaces.go
jakob223: 👍

libkbfs/folder_block_manager.go
jakob223: done.

jakob223:
retagging review per song's request

songgao:
👍🏼

libkbfs/folder_block_manager.go
```diff
@@ -1306,25 +1306,36 @@ func (fbm *folderBlockManager) clearLastQRData() {
 	fbm.lastReclamationTime = time.Time{}
 }
 
-func (fbm *folderBlockManager) doCleanSyncCache() (err error) {
-	defer fbm.cleanSyncCacheGroup.Done()
+func (fbm *folderBlockManager) doCleanDiskCaches() (err error) {
+	defer fbm.cleanDiskCachesGroup.Done()
+	// TODO: determine if these can be parallelized
```
strib: From a correctness standpoint, parallelizing them should be fine.  Not sure if it will make the disk performance extra sad though.  I'd err on the side of keeping it serialized for now, unless we have a need to speed it up.

libkbfs/folder_block_manager.go
```diff
@@ -1306,25 +1306,36 @@ func (fbm *folderBlockManager) clearLastQRData() {
...
+	if syncConfig.Mode != keybase1.FolderSyncMode_DISABLED {
+		err = fbm.doCleanDiskCache(DiskBlockSyncCache)
+		if err != nil {
+			return err
+		}
 	}
+	return fbm.doCleanDiskCache(DiskBlockWorkingSetCache)
+}
+
+func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
```
strib: Please move this above where it is first used.

PR #1962, opened 2018-12-11
disk_block_cache: support mark-and-sweep and clear prefetch on delete

strib:
This PR adds mark-and-sweep support to the disk block cache.

It also clears the prefetch status from the in-memory block cache when a block is deleted from the disk cache, to keep them consistent.  This is important when clearing all the blocks from a TLFs, and for when we start doing mark-and-sweep in the next PR.

Also, `BlockCacheStandard.DeleteTransient` doesn't need the whole `BlockPointer`.  Not really related, just something I noticed.

Issue: KBFS-3644


libkbfs/bcache.go
```diff
@@ -42,6 +42,9 @@ type BlockCacheStandard struct {
 
 	bytesLock       sync.Mutex
 	cleanTotalBytes uint64
+
+	// Covers the `prefetchStatus` field for all `blockContainer`s.
```
jzila: Doesn't this interfere with other locks in `cleanTransient`? Blocks can be removed/overwritten even while this lock is held.

libkbfs/disk_block_cache_helper.go
```diff
@@ -37,6 +37,8 @@ type DiskBlockCacheMetadata struct {
 	TriggeredPrefetch bool `codec:"HasPrefetched"`
 	// whether the block's triggered prefetches are complete
 	FinishedPrefetch bool
+	// the last tag with which the block was marked
+	Tag string
```
jzila: Wouldn't it be more space efficient to make this a `byte` enum or bitfield?

libkbfs/disk_block_cache.go
```diff
@@ -1104,6 +1102,121 @@ func (cache *DiskBlockCacheLocal) DoesCacheHaveSpace(
...
+			len(blockIDs), tag, tlfID)
+		_, _, err = cache.deleteLocked(ctx, blockIDs)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nextKey, nil
+}
+
+// DeleteUnmarked deletes all the blocks without the given tag.
```
jzila: It might make more sense for this to return a `<-chan error`, allowing either sync or async usage.

libkbfs/disk_block_cache_test.go
```diff
@@ -758,3 +764,52 @@ func TestDiskBlockCacheMoveBlock(t *testing.T) {
...
+
+	t.Log("Mark a couple blocks.")
+	tag := "mark"
+	err := cache.Mark(ctx, ids[1], tag, DiskBlockSyncCache)
+	require.NoError(t, err)
+	err = cache.Mark(ctx, ids[3], tag, DiskBlockSyncCache)
+	require.NoError(t, err)
+
+	t.Log("Delete all unmarked blocks.")
+	err = cache.DeleteUnmarked(ctx, tlfID, tag, DiskBlockSyncCache)
```
jzila: Looks like this doesn't test the timed functionality since it doesn't reach the batching threshold.

Ideally the timer would be a tunable parameter anyway, in which case tests can pass in a zero delay.

libkbfs/bcache.go
strib: I don't understand what you mean by "interfere".  It's not taken in the `onEvict` handler, so I don't think there's any chance of deadlock.  And this isn't trying to cover blocks being removed or overwritten.  It's only to protect against a data race on the `prefetchStatus` field of each individual `blockContainer` struct in the cache.

libkbfs/disk_block_cache_helper.go
strib: I erred on the side of debugability.  Not sure how a bitfield would work in practice -- it seemed hard to make the upper layer generate a new one for every mark-and-sweep if it's anything but a random string.

libkbfs/disk_block_cache.go
strib: As it is now, it's mirroring `ClearAllTlfBlocks`, and I'd prefer to keep it consistent.  If the user wants async, they can just use it in a goroutine themselves, pretty easy.  I'd rather not have to manage the goroutines in the cache if it's not needed.

libkbfs/disk_block_cache_test.go
strib: Done!

libkbfs/disk_block_cache_test.go
```diff
@@ -717,6 +717,9 @@ func TestDiskBlockCacheUnsyncTlf(t *testing.T) {
 	seedDiskBlockCacheForTest(t, ctx, cache, config, numTlfs, numBlocksPerTlf)
 	require.Equal(t, numBlocks, standardCache.numBlocks)
 
+	standardCache.clearTickerDuration = 0
+	standardCache.numBlocksToEvictOnClear = 1
```
jzila: Will setting these race with the loop?

strib: No, because this happens before the clearing loop starts (it's kicked off by the `SetTlfSyncState` below.

jzila: Cool thanks.

libkbfs/bcache.go
jzila: OK sounds good to me.

libkbfs/disk_block_cache.go
jzila: OK fair enough. You can leave it as is then.

libkbfs/disk_block_cache_helper.go
jzila: I guess it's not obvious to me what you plan the tags to look like. Any reason it can't just be a timestamp?

strib: Oops, sorry for merging this with this discussion still unresolved.  I used the FBOID debug tag as the marking tag in the next PR: https://github.com/keybase/kbfs/pull/1963/files#diff-f6470cb87de71869397b1e97dd1ee43dR1100

That way, it's easy to debug the DB given the logs.  A millisecond timestamp would probably work too, but wouldn't show up in the debug tags.  Let me know if you want me to change it before I merge #1963.

jzila: FBOID + a counter? I guess that's short enough that it shouldn't be a big deal.

PR #1963, opened 2018-12-11
folder_branch_ops: mark and sweep for partially-synced TLFs

strib:
(Depends on #1962.)

Partially-synced TLFs need to clean up the sync cache in the background, in the cases where a) some previously-synced path is now unsynced manually by the user, or b) the user moves some previously-synced directory out of a synced path.  In the latter case, we don't have an explicit config action from the user, so it needs to be done periodically in the background.

This adds a new background goroutine to FBO for partially-synced folders, that triggers a mark-and-sweep operation either when a path is removed from the sync config, or on a periodic timer (once an hour).  It iterates through each synced path and marks each block that should still be synced with a tag.  After that, it removes all blocks from that TLF that aren't marked with that tag from the sync cache.

This process is stopped whenever the latest merged MD is updated.

Issue: KBFS-3644

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -8269,6 +8269,11 @@ func (fbo *folderBranchOps) SetSyncConfig(
 			return nil, errors.New(
 				"Unexpected sync config; mark-and-sweep already started")
 		}
+		if oldConfig.Mode == keybase1.FolderSyncMode_ENABLED {
+			return nil, errors.New(
+				"Cannot enable partial syncing while fully-synced")
+		}
+
 		fbo.log.CDebugf(ctx, "Entering partial mode, starting mark-and-sweep")
 		// `kickOffPartialSync` call above will start the mark and sweep.
 	} else if oldPartial && newPartial {
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -6,6 +6,7 @@ package libkbfs
 
 import (
 	"fmt"
+	"math/rand"
 	"os"
 	stdpath "path"
 	"path/filepath"
@@ -99,6 +100,8 @@ const (
 	// If there are more than this many new revisions, fast forward
 	// rather than downloading them all.
 	fastForwardRevThresh = 50
+	// Period between mark-and-sweep attempts.
+	markAndSweepPeriod = 1 * time.Hour
 )
 
 type fboMutexLevel mutexLevel
@@ -277,7 +280,8 @@ type folderBranchOps struct {
 	// Has this folder ever been cleared?
 	hasBeenCleared bool
 
-	syncLock leveledRWMutex
+	syncLock            leveledRWMutex
+	markAndSweepTrigger chan<- struct{}
 
 	blocks  folderBlockOps
 	prepper folderUpdatePrepper
@@ -1024,6 +1028,15 @@ func (fbo *folderBranchOps) kickOffPartialSync(
 			}
 		}()
 	}
+
+	// Kick off amark-and-sweep if one doesn't exist yet.
+	fbo.syncLock.Lock(lState)
+	defer fbo.syncLock.Unlock(lState)
+	if fbo.markAndSweepTrigger == nil {
+		trigger := make(chan struct{}, 1)
+		fbo.markAndSweepTrigger = trigger
+		go fbo.partialMarkAndSweepLoop(trigger)
+	}
 }
 
 func (fbo *folderBranchOps) makeRecentFilesSyncConfig(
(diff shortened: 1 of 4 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -1024,6 +1037,296 @@ func (fbo *folderBranchOps) kickOffPartialSyncIfNeeded(
...
+
+			err = fbo.blocks.MarkNode(
+				ctx, lState, currNode, latestMerged, tag, cacheType)
+			if err != nil {
+				return err
+			}
+		}
+
+		// Now mark everything rooted at this path.
+		currNode, _, err = fbo.Lookup(ctx, currNode, syncedElem)
```
jzila: It strikes me that this loop will destroy the effectiveness of the in-memory cache, since it'll keep thrashing the synced blocks into the memory cache. It'll also be doing constant reading from the disk. Not sure if we care all that much about that part.

libkbfs/folder_branch_ops.go
```diff
@@ -1024,6 +1037,296 @@ func (fbo *folderBranchOps) kickOffPartialSyncIfNeeded(
...
+			fbo.log.CDebugf(ctx, "Mark-and-sweep timer fired")
+		case <-fbo.shutdownChan:
+			fbo.log.CDebugf(ctx, "Shutdown")
+			return
+		}
+
+		if currMarkAndSweepCancel != nil {
+			currMarkAndSweepCancel()
+		}
+		timer.Stop()
```
jzila: why stop the timer here?

libkbfs/folder_branch_ops.go
```diff
@@ -7901,6 +8224,57 @@ func (fbo *folderBranchOps) SetSyncConfig(
...
+	newPartial := newConfig.Mode == keybase1.FolderSyncMode_PARTIAL
+	if oldPartial && !newPartial {
+		if fbo.markAndSweepTrigger == nil {
+			return nil, errors.New(
+				"Unexpected sync config; mark-and-sweep already started")
+		}
+
+		fbo.log.CDebugf(ctx, "Exiting partial mode, stopping mark-and-sweep")
+		close(fbo.markAndSweepTrigger)
+		fbo.markAndSweepTrigger = nil
```
jzila: I'm confused. Shouldn't we be marking and sweeping out all old blocks if `newPartial` is false?

libkbfs/folder_branch_ops.go
```diff
@@ -7901,6 +8224,57 @@ func (fbo *folderBranchOps) SetSyncConfig(
...
+			return nil, errors.New(
+				"Unexpected sync config; mark-and-sweep already started")
+		}
+		if oldConfig.Mode == keybase1.FolderSyncMode_ENABLED {
+			return nil, errors.New(
+				"Cannot enable partial syncing while fully-synced")
+		}
+
+		fbo.log.CDebugf(ctx, "Entering partial mode, starting mark-and-sweep")
+		// `kickOffPartialSync` call above will start the mark and sweep.
```
jzila: above where?

libkbfs/folder_branch_ops.go
strib: I don't think we can avoid the "constant reading" part given that we have to do mark-and-sweep.  Good point about the in-memory cache getting thrashed.  Do you think it's worth building some mechanism to avoid memory-caching during this process?  I lean towards no but I could be convinced.

For what it's worth, the `blockTree` is smart enough to avoid fetching any direct leaf blocks in this process, so for example the block for a small file in a directory won't be read into the cache when it's marked. (Same with all the leaf nodes of big files.)

libkbfs/folder_branch_ops.go
strib: That's for the case where we got a manual trigger.  No need to let the timer keep going in that case.

libkbfs/folder_branch_ops.go
jzila: Some alternatives:
1. Get rid of the in-memory transient cache completely, or make it tiny (<=10MB). It won't be valuable once we have a mark and sweep trashing it.
2. Avoid caching from mark-and-sweep.
3. Have a separate cache for marking and sweeping (dirty cache, clean transient cache, and mark-sweep cache).

I'm inclined to go with 1, as the working set disk cache gives us most of the benefit of the clean transient LRU cache.

jzila:
Just had another thought: are we expecting this to run on mobile?

libkbfs/folder_branch_ops.go
strib: I'm ok with 1, especially after some performance tests, but do you think it belongs in this PR?  Seems pretty unrelated.

If you want to do 2 as part of this PR, maybe we can just hack in a context value for it.

Either way, let me know what you think this PR should block on.

strib:
> Just had another thought: are we expecting this to run on mobile?

Seems like we'd have to run it, once we start allowing partially-synced paths, or else periodically clear the cache and resync.  We can tweak timer values to make it more palatable if needed.  Open to alternatives though.

libkbfs/folder_branch_ops.go
strib: If the new config is enabling syncing, then a full sync will be kicked off, subsuming the currently synced blocks. 
 If the new config is disabling syncing completely, then `ConfigLocal.SetTlfSyncState()` calls `ClearAllTlfBlocks()` and gets rid of everything, without needing mark-and-sweep.

libkbfs/folder_branch_ops.go
strib: In the `defer` way higher up.

jzila:
Closing in favor of keybase/client#15327

PR #1965, opened 2018-12-12
mdserver_remote: use short timeout to check server for latest TLF

strib:
This waits 500ms for a server response before giving up and using the cached MD if there is one.

In most cases `folderBranchOps` will roll forward to the latest revision without this fix, but in the meantime bad stuff can happen: if it's an unsynced TLF, the user can get "block not found" errors, or rekeys/upgrades can fail because of conflicts.

Issue: KBFS-3665
Issue: KBFS-3676

### DIFF
--- a/libkbfs/mdserver_remote.go
+++ b/libkbfs/mdserver_remote.go
@@ -38,6 +38,10 @@ const (
 	// server for the latest handle before we use the cached value
 	// instead.
 	mdServerLatestHandleTimeout = 500 * time.Millisecond
+	// mdServerTimeoutWhenMDCached defines how long we wait for the
+	// latest MD to return from the server, when we've already read it
+	// from the disk cache.
+	mdServerTimeoutWhenMDCached = 500 * time.Millisecond
 )
 
 // MDServerRemote is an implementation of the MDServer interface.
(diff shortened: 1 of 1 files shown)


songgao:
Two questions:

1. It seems this (GetForTLF) is only called when user first time access a TLF, and when suer tries to write into it. So the 500ms delay won't affect reads. Is this correct?

2. In slow network where it times out after 500ms, we'd still have that "block not found" issue right? Is there a way to detect that error and do the get after that?

strib:
> It seems this (GetForTLF) is only called when user first time access a TLF, and when suer tries to write into it. So the 500ms delay won't affect reads. Is this correct?

Not sure where you got that impression, but no, this is used before the first read of a TLF during an kbfs instance's lifetime too: https://github.com/keybase/kbfs/blob/f7c9b5de11e410df6d1f4963d1c74177840b15d8/libkbfs/kbfs_ops.go#L546

But it's not really right to say that it "slows down" reads, because getting the MD on the first TLF access is how KBFS has worked for years.  So it just brings performance back to closer to what it was before I introduced the cache (and broke a bunch of stuff).

> In slow network where it times out after 500ms, we'd still have that "block not found" issue right? Is there a way to detect that error and do the get after that?

We will immediately register with the mdserver to be notified about new MDs, and start applying them as quickly as we can.  But yes whatever we return from here controls the view of the folder until that catch-up can happen.  We can try to figure out something crazy to do in `folderBranchOps` to add another blocking step to wait on a fast-forward if there's a "block not found" error, but it would be pretty nasty to implement and I personally don't think it's worth spending much time on.  I'd rather increase the timeout here to a point you're more comfortable with, at the cost of slowing down offline access, than re-architecting the `folderBranchOps` behavior.

songgao:
Thanks for the detailed explaination!

> We will immediately register with the mdserver to be notified about new MDs, and start applying them as quickly as we can. But yes whatever we return from here controls the view of the folder until that catch-up can happen.

Does this mean the "block not found" error we've been seeing were all transient and would have been resolved as soon as that notification came in? I guess that's why it was flaky.

libkbfs/mdserver_remote.go
```diff
@@ -622,13 +631,28 @@ func (md *MDServerRemote) GetForTLF(ctx context.Context, id tlf.ID,
...
+	_, rmdses, err := md.get(getCtx, arg)
+	switch errors.Cause(err) {
+	case nil:
+	case context.DeadlineExceeded:
+		if cachedRmds != nil {
+			md.log.CDebugf(ctx, "Can't contact server; using cached MD")
+			return cachedRmds, nil
+		}
+		return nil, err
+	default:
```
songgao: In offline mode, we might get something like "network unreachable" here right? Do we need to capture those networking errors here?

songgao:
> Not sure where you got that impression, 

Ah I only grepped `folder_branch_ops.go` for it. Now I remember. Thanks.

libkbfs/mdserver_remote.go
songgao: Ah that's right! I'm having a Deja Vu for this question and answer. Sorry for my brain fart.

PR #1966, opened 2018-12-12
Unused return value etc bugs

taruti:
Some nits found in libkbfs while doing an unrelated things.

This contains:
+ Handle some unhandled error returns
+ Remove a unused variable in folder_branch_ops
+ Fix bug in select in folder_block_ops, this original code broke only from the inner select.


taruti:
Spotted one suspect, then used staticcheck to find the rest.

PR #1970, opened 2018-12-13
chat_rpc: only send edit notifications on favorited TLFs

jakob223:
Issue: KBFS-3084

### DIFF
--- a/libkbfs/chat_rpc.go
+++ b/libkbfs/chat_rpc.go
@@ -410,6 +410,17 @@ func (c *ChatRPC) GetGroupedInbox(
 		return nil, err
 	}
 
+	favorites, err := c.config.KBFSOps().GetFavorites(ctx)
+	if err != nil {
+		c.log.CWarningf(ctx,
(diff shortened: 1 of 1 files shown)


libkbfs/chat_rpc.go
```diff
@@ -421,15 +428,24 @@ func (c *ChatRPC) GetGroupedInbox(
...
-
 		tlfType := tlf.Private
 		if info.Visibility == keybase1.TLFVisibility_PUBLIC {
 			tlfType = tlf.Public
 		} else if info.MembersType == chat1.ConversationMembersType_TEAM {
 			tlfType = tlf.SingleTeam
 		}
 
+		tlfIsFavorite := false
+		for _, fav := range favorites {
```
strib: Probably best to build a map of this once right after fetching favorites, so we don't have to iterate it every time.  Something like `map[Favorite]bool`.

jakob223:
Tested manually, it works. Implemented the optimization you requested.

Re the extra feature, my testing this made that seem quite necessary, otherwise it is ~impossible to remove favorites with recent edits.

PR #1972, opened 2018-12-14
modes: use one prefetch worker on mobile

strib:
But add a new mode setting that will trigger prefetches on normal block fetches, and turn that off on mobile.

Issue: keybase/client#15087

### DIFF
--- a/libkbfs/data_types.go
+++ b/libkbfs/data_types.go
@@ -1099,10 +1099,3 @@ func (bra BlockRequestAction) CacheType() DiskBlockCacheType {
 	}
 	return DiskBlockAnyCache
 }
-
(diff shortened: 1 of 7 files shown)


libkbfs/data_types.go
```diff
@@ -1099,3 +1099,10 @@ func (bra BlockRequestAction) CacheType() DiskBlockCacheType {
 	}
 	return DiskBlockAnyCache
 }
+
+func defaultBlockRequestAction(mode InitMode) BlockRequestAction {
```
jzila: why isn't this a function on `Mode()`?

strib: Because I didn't think about it.  Fixed.

PR #1973, opened 2018-12-14
folder_block_manager: check blocks for liveness before deleting from sync cache

strib:
If a block is deduped and has multiple references to it (as can happen when the writer of a TLF isn't using the journal), we need to check with the server before deleting it from the sync cache.  Otherwise, we could be missing a block from the sync cache if the user goes offline.

Also, this changes the disk block cache to only delete from a specific cache if a specific type is passed in, since the sync cache and working set cache now have different delete policies.  (It's fine and more performant to evict dedup'd blocks from the working set cache even if they still have references left.)

Issue: KBFS-3661

### DIFF
--- a/libkbfs/folder_block_manager.go
+++ b/libkbfs/folder_block_manager.go
@@ -1364,9 +1364,30 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
 		}
 
 		ptrs := getUnrefPointersFromMD(rmd)
-		ids := make([]kbfsblock.ID, len(ptrs))
-		for i, ptr := range ptrs {
-			ids[i] = ptr.ID
+
+		ids := make([]kbfsblock.ID, 0, len(ptrs))
+		if cacheType == DiskBlockSyncCache {
+			// For sync caches, we need to make sure that all of the
+			// references to this block are really gone.
+			liveCounts, err := fbm.config.BlockOps().GetLiveCount(
+				ctx, fbm.id, ptrs)
+			if err != nil {
+				return err
+			}
+
+			for id, count := range liveCounts {
+				if count == 0 {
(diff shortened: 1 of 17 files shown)


libkbfs/folder_block_manager.go
```diff
@@ -1306,6 +1308,74 @@ func (fbm *folderBlockManager) clearLastQRData() {
...
+	nonLiveBlocks []kbfsblock.ID, err error) {
+	fbm.log.CDebugf(ctx, "Get live count for %d pointers", len(ptrs))
+	bops := fbm.config.BlockOps()
+
+	// Round up to find the number of chunks.
+	numChunks := (len(ptrs) + numPointersToDowngradePerChunk - 1) /
+		numPointersToDowngradePerChunk
+	numWorkers := numChunks
+	if numWorkers > maxParallelBlockPuts {
+		numWorkers = maxParallelBlockPuts
```
jzila: Shouldn't this be its own constant? It doesn't really have anything to do with block put parallelism.

libkbfs/folder_block_manager.go
```diff
@@ -1365,12 +1435,40 @@ func (fbm *folderBlockManager) doCleanDiskCache(cacheType DiskBlockCacheType) (
...
+			// code again with the `gcOp`, and we'll delete the block
+			// then.  (Note there's always a chance for a race here,
+			// since the client could crash before archiving the
+			// blocks.  But the GC should always catch it eventually.)
+			err := fbm.waitForArchives(ctx)
+			if err != nil {
+				return err
+			}
+
+			ids, err = fbm.doChunkedGetNonLiveBlocks(ctx, ptrs)
```
jzila: Comment that this blocks, but that's fine because this method is called in a background goroutine.

PR #1974, opened 2018-12-17
folder_branch_ops: don't rekey team folders

strib:
A user could have an old rekey request hanging around for a TLF that was migrated to a team, and would crash if it tries to rekey it.

Issue: KBFS-3694

### DIFF
--- a/libkbfs/folder_branch_ops.go
+++ b/libkbfs/folder_branch_ops.go
@@ -6247,6 +6247,12 @@ func (fbo *folderBranchOps) rekeyLocked(ctx context.Context,
 				return RekeyResult{}, err
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


libkbfs/folder_branch_ops.go
```diff
@@ -1792,6 +1794,10 @@ func (fbo *folderBranchOps) getMDForRekeyWriteLocked(
 		return nil, kbfscrypto.VerifyingKey{}, false, err
 	}
 
+	if md.TypeForKeying() == tlf.TeamKeying {
+		return nil, kbfscrypto.VerifyingKey{}, false, err
```
jzila: `err` should be `nil` here just to be explicit.

strib: Oops, that was intended, thanks.

PR #1977, opened 2018-12-18
keybase_service_base: don't try to use a nil ID on identify error

strib:


libkbfs/keybase_service_base.go
```diff
@@ -490,6 +490,7 @@ func (k *KeybaseServiceBase) Identify(ctx context.Context, assertion, reason str
 		k.log.CDebugf(ctx,
 			"Ignoring error (%s) for user %s with no sigchain; "+
 				"error type=%T", err, res.Ul.Name, err)
+		ei.onError(ctx)
```
jakob223: Looks like we're not actually ignoring the error here anymore, so that debug line is incorrect?

strib: We're still ignoring the error in the sense that we're not returning it from this function.  The `onError` call is just making sure that some cleanup code runs (in place of the `userBreak` or `teamBreak` calls below, which won't get run now).

strib:
> Worth adding a test that would have failed on the bug that was causing the keybase.pub problem?

Can't really test this file in this repo because it depends heavily on interacting with a real service.  So the test would need to be in kbfsdocker.  It would be nice to have such a test, but it would take some effort and I'm not sure about the priority of it.

jzila:
Yeah I'm inclined to agree with Strib--KBFSDocker isn't ideal for regression tests because each tests has such high overhead.

PR #1979, opened 2018-12-18
favorites: remove unfavorited TLFs from edit history

jakob223:
When KBFS becomes aware that a TLF is no longer a favorite, tell the
kbfsedits.UserHistory to delete it.

This is important because it prevents a recently edited file from appearing in the
history in the widget, which causes that TLF to be readded to favorites.

Issue: KBFS-3689


jakob223:
I did test manually; here's a unit test.

PR #1986, opened 2018-12-19
favorites: create directory for favorites disk cache

jakob223:
I realized this was never being created since the directory does not exist while working on another thing stored on disk.

### DIFF
--- a/libkbfs/favorites.go
+++ b/libkbfs/favorites.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/keybase/client/go/libkb"
+	"github.com/keybase/client/go/logger"
 	"github.com/syndtr/goleveldb/leveldb"
 	"os"
 	"strings"
@@ -80,6 +81,7 @@ type favReq struct {
 type Favorites struct {
 	config   Config
 	disabled bool
+	log      logger.Logger
 
 	// Channels for interacting with the favorites cache
 	reqChan chan *favReq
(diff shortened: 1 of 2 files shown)


jzila:
Is this the versioned path?

jakob223:
I don't know what you mean by that.

jzila:
The folder for all the disk caches needs to be versioned, in case we want to change the on-disk format of anything that is being cached.

e.g. `~/Library/Application\ Support/Keybase/kbfs_block_cache/v1/<leveldb root>`

jakob223:
There is a version in each of the files being read. Do you think we need a version on the directory too? If so, I can add a version there too.

In the meantime, I'm gonna fix the gomock complaining.

jzila:
Oh hmm. Can you sketch out the directory tree for me?

jakob223:
config.StorageRoot():
  \--- kbfs.favorites:
            |--- user1 (file containing a struct containing a version and
some serialized data)
            \--- user2 (etc)

jakob223:
I've refactored the cache to use levelDB and added a test for the disk-based cache.

libkbfs/favorites.go
```diff
@@ -134,18 +139,27 @@ type favoritesCacheEncryptedForDisk struct {
 }
 
 func (f *Favorites) readCacheFromDisk(ctx context.Context) error {
+	log := f.config.MakeLogger("Favorites")
```
jzila: You call this multiple times. It might be better to store a logger instead. `f.config.MakeLogger("FAV")` (we use 3 letter codes to keep the log prefix line length the same).

PR #1987, opened 2018-12-20
bcache: stop tracking prefetch status in memory

strib:
Now that blocks can move from the working set cache to the sync cache (which changes the block's prefetch status), it's hard to keep a synchronized view of the prefetch status in memory.  So, just rely on the disk cache entirely for the prefetch status.

Note this contains a bit of a hack for storing the prefetch status in the block retrieval queue when there's no disk block cache.  I couldn't find a more elegant way of doing it that didn't require splitting up the disk cache lookup into two calls, so I thought this was better.

Issue: KBFS-3644

### DIFF
--- a/libkbfs/bcache.go
+++ b/libkbfs/bcache.go
@@ -16,11 +16,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-type blockContainer struct {
-	block          Block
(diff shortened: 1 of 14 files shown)


strib:
@jzila I will merge this into #1962 before I merge that one.

Issue #1958, opened 2018-12-08
Enormous KBFS memory usage when not doing anything (Arch Linux)


quantum-dan:
So right now I'm not uploading or downloading anything (and the Keybase app agrees, it says "Done" with regards to files) and my memory usage looks like this:
![2018-12-08_12 02 21](https://user-images.githubusercontent.com/8014837/49689648-45c62800-fae1-11e8-980a-f5b07e080c64.png)

Log ID: bb31cd3b22499f77350fce1c

That's a fraction of my memory, but it's enough to cause noticeable slowdown here and there.

Yesterday I started making a Veracrypt container but changed my mind and cancelled it, so that's not being uploaded.  I just finished making a backup to my private directory using Borg Backup a couple of hours ago, so I could understand the memory usage back then, but it's still high a few hours later.

strib:
2 things:

* I don't know much about Arch, but you're sure that's reporting _resident_ memory usage, and not just virtual memory usage, right?  Our programming language Go has an issue releasing the latter I think, but it wouldn't actually affect available memory.
* I think I just found and fixed a memory leak, the PR is still in review: #1956.  I see a bunch of prefetcher stuff in your logs so it could very much be related.

quantum-dan:
* I'm not entirely certain which, but my system is definitely short on available memory as a result.  This is the first time in years that it's actually started using swap storage.
* That could explain things.  Thanks.

DonaldKBrown:
This issue just appeared in the latest update on an Ubuntu 18.04 system as well. Computer slowed down to a halt, couldn't even send a Slack message until I killed the main `kbfsfuse` process. See screenshot:

![2018-12-20-081821](https://user-images.githubusercontent.com/20977880/50293853-c5cd8580-0431-11e9-88f3-e6f6db884706.png)

strib:
I'm very confused, there should only be one `kbfsfuse` process running at a time.  Where are the other ones coming from?  cc: @heronhaye 

Anyway @DonaldKBrown please run `keybase log send`, and if you're still having the issue while kbfs is running and mounted, please also save the contents of the `/keybase/.kbfs_profiles` directory.

heronhaye:
I believe it's the different threads, you can turn off `hide userland process threads` in `F2`.

DonaldKBrown:
@strib : My log id is `b3e1deb590f3212f0f10411c`. Thank you for looking into this.
[...]

PR #1988, opened 2018-12-20
prefetcher: avoid mem leaks by not storing block in request

strib:
We don't actually use the contents of the block in the request; we only use it to generate a new block.  So just store the function for generating the new block instead, to be more explicit.

Also, don't pass the same block used to make that function to the `Request()` function, which fills in the block.  I suspect that was a memory leak.

Issue: #1958

### DIFF
--- a/libkbfs/block_types.go
+++ b/libkbfs/block_types.go
@@ -148,6 +148,11 @@ func (cb *CommonBlock) NewEmpty() Block {
 	return NewCommonBlock()
 }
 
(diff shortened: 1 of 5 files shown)


strib:
@jzila i changed this around a bit because I was worried that keeping around `b.NewEmpty` would keep a reference to `b` around as well.  So now there's a `b.NewEmptier()` function which just returns a static function that doesn't need to keep the old block around.

PR #1989, opened 2018-12-20
Enforce minimum Dokan version

zanderz:


dokan/loaddll.go
```diff
@@ -32,9 +34,49 @@ func (ep *errorPrinter) Printf(s string, os ...interface{}) {
 	fmt.Fprintf(&ep.buf, s, os...)
 }
 
+// adapted from keybase/client/go/install/fuse_status_windows.go
+func isDokanCurrent(path string) (bool, error) {
```
taruti: Can we ever return "false, nil" ?

If not then perhaps just ```func isDokanCurrent(path string) error```.

dokan/loaddll.go
```diff
@@ -32,9 +34,49 @@ func (ep *errorPrinter) Printf(s string, os ...interface{}) {
...
 // loadLibrary calls win32 LoadLibrary and logs the result.
 func loadLibrary(epc *errorPrinter, path string) (windows.Handle, error) {
 	hdl, err := windows.LoadLibrary(path)
+	if err == nil {
+		current, err2 := isDokanCurrent(path)
+		if !current {
+			windows.Close(hdl)
+		}
+		err = err2
+	}
```
taruti:
```
func loadLibrary(epc *errorPrinter, path string) (windows.Handle, error) {
	hdl, err := windows.LoadLibrary(path)
	epc.Printf("LoadLibrary(%q) -> %v,%v\n", path, hdl, err)
	if err!=nil {
		return hdl, err
	}
	err = isDokanCurrent(path)
	if err!=nil {
			windows.Close(hdl)
		}
	}
	return hdl, err
}
```

zanderz: ok

taruti:
https://github.com/dokan-dev/dokany/issues/765 might come early enough for us.

taruti:
https://github.com/dokan-dev/dokany/releases/tag/v1.2.1.2000 is out with a bumbed DOKAN_VERSION. That should be used instead.

taruti:
i.e. check the return value of ```C.kbfsLibDokan_GetVersion(dokanVersionProc) >= 121```.

PR #1996, opened 2019-01-03
vendor: Update github.com/xanzy/ssh-agent for Go 1.11 on Windows

taruti:
Update vendor because of Windows.

Compilation fails on Windows with Go 1.11.4 without this because the vendored package used to use "import . unsafe; import . syscall" in Windows specific code. This was broken by the addition of syscall.Pointer.

### DIFF
--- a/vendor/github.com/xanzy/ssh-agent/pageant_windows.go
+++ b/vendor/github.com/xanzy/ssh-agent/pageant_windows.go
@@ -29,8 +29,8 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	. "syscall"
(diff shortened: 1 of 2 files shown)


jakob223:
Aren't we still using Go 1.10?

taruti:
Yes, but fixing things proactively is a good thing. Updated a laptop, noticed & fixed this.

jzila:
@taruti This one isn't as trivial to port over, so please reapply it to vendor.json in keybase/client.

taruti:
Done as https://github.com/keybase/client/pull/15333

PR #1998, opened 2019-01-03
Surya/core 9529/kbfs systemd replace run keybase

heronhaye:
TODO: make at least some effort for bsd compat

### DIFF
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -80,6 +80,10 @@ func start() *libfs.Error {
 	} else {
 		mountDir = flag.Arg(0)
 	}
+	err := os.MkdirAll(mountDir, os.ModePerm)
+	if err != nil {
+		fmt.Println(err)
+	}
 
 	if len(flag.Args()) > 1 {
 		fmt.Print(getUsageString(ctx))
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -10,6 +10,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"runtime"
 
 	"bazil.org/fuse"
 
@@ -83,6 +84,16 @@ func start() *libfs.Error {
 	err := os.MkdirAll(mountDir, os.ModePerm)
 	if err != nil {
 		fmt.Println(err)
+		switch os := runtime.GOOS; os {
+		case "linux":
+			fmt.Println("failed to create mount directory")
+		case "darwin":
+			// If the user did not override the defaults this isn't expected to work.
+			// The default is in a directory only writable by root.
+			fmt.Println("mount directory does not exist and should have been created by system service")
+		case "windows":
+			fmt.Println("I honestly have no idea what to say here, maybe someone can tell me what is supposed to create this on Windows")
+		}
 	}
 
 	if len(flag.Args()) > 1 {
--- a/kbfsfuse/main.go
+++ b/kbfsfuse/main.go
@@ -10,7 +10,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"runtime"
 
 	"bazil.org/fuse"
 
@@ -74,26 +73,15 @@ func start() *libfs.Error {
 		if err != nil {
 			return libfs.InitError(err.Error())
 		}
+		// If a mountdir was not set by keybase config set mountdir,
+		// the service returns a default value which may or may not
+		// exist yet.
 		if len(mountDir) == 0 {
 			fmt.Print(getUsageString(ctx))
 			return libfs.InitError("no mount specified")
 		}
 	} else {
 		mountDir = flag.Arg(0)
-	}
-	err := os.MkdirAll(mountDir, os.ModePerm)
-	if err != nil {
-		fmt.Println(err)
-		switch os := runtime.GOOS; os {
-		case "linux":
-			fmt.Println("failed to create mount directory")
-		case "darwin":
-			// If the user did not override the defaults this isn't expected to work.
-			// The default is in a directory only writable by root.
-			fmt.Println("mount directory does not exist and should have been created by system service")
-		case "windows":
-			fmt.Println("I honestly have no idea what to say here, maybe someone can tell me what is supposed to create this on Windows")
-		}
 	}
 
 	if len(flag.Args()) > 1 {
--- a/libfuse/mounter.go
+++ b/libfuse/mounter.go
@@ -7,6 +7,7 @@ package libfuse
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path"
@@ -92,7 +93,7 @@ func (m *mounter) Unmount() (err error) {
 		_, err = exec.Command("fusermount", "-u", dir).Output()
 		// Only clean up mountdir on a clean unmount
 		if err == nil {
-			defer m.DeleteMountdir()
+			defer m.DeleteMountdirIfEmpty()
 		}
 	default:
 		err = fuse.Unmount(dir)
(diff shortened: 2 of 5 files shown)


packaging/linux/systemd/keybase-redirector.service
```diff
@@ -0,0 +1,13 @@
...
+Description=Keybase Root Redirector for KBFS
+Wants=keybase.service
+
+[Service]
+Type=simple
+
+# Abort if root redirector is explicitly disabled
+ExecStartPre=/usr/bin/keybase -c /etc/keybase/config.json config get --direct --assert-false --assert-ok-on-nil disable-root-redirector
+
+# TODO Do we need any special thing for nohup here?
```
heronhaye: can have same behavior with loginctl linger but not sure if that's what we want

libfuse/mounter.go
```diff
@@ -87,7 +87,8 @@ func (m *mounter) Unmount() (err error) {
 			_, err = exec.Command(
 				"/usr/sbin/diskutil", "unmountDisk", "force", dir).Output()
 		case "linux":
-			_, err = exec.Command("fusermount", "-ul", dir).Output()
+			// Lazy unmount; will unmount when KBFS is no longer in use
+			_, err = exec.Command("fusermount", "-u", "-z", dir).Output()
```
heronhaye: @taruti is this ok? my fusermount(1) shows the option is `-z`, which is pretty confusing since `umount` uses `-l`. hopefully it isn't os-dependent

libkbfs/init.go
```diff
@@ -510,7 +511,7 @@ func InitWithLogPrefix(
 	log logger.Logger, logPrefix string) (cfg Config, err error) {
 	done := make(chan struct{})
 	interruptChan := make(chan os.Signal, 1)
-	signal.Notify(interruptChan, os.Interrupt)
+	signal.Notify(interruptChan, os.Interrupt, syscall.SIGTERM)
```
heronhaye: maybe also SIGHUP?

songgao: I think this is why we only did `os.Interrupt` initially (from [os doc](https://golang.org/pkg/os/)):

> The only signal values guaranteed to be present in the os package on all systems are Interrupt (send the process an interrupt) and Kill (force the process to exit). Interrupt is not implemented on Windows; using it with os.Process.Signal will return an error.

But apparently this passed CI, so perhaps SIGTERM is fine.

SIGHUP sounds reasonable to me, though I'm interested in what others think too.

heronhaye: oops, I missed that. could we do another switch on `GOOS`?

heronhaye: weird that it doesn't fail to build instead of throwing an error

libfuse/mounter.go
songgao: It'd be nice to use the [syscall unmount](https://github.com/keybase/fuse/commit/e3996deee1b03bec702b6a4d4f5b6f2154895e41#diff-eb35e1f2018531b3563fe5d935fa14d6) like [here](https://github.com/keybase/kbfs/blob/c2afd3aee19dbc113edd2b148db63834c0d08ef1/redirector/main.go#L287), but I guess can be in a separate PR.

packaging/linux/systemd/keybase-redirector.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Keybase Root Redirector for KBFS
+Wants=keybase.service
+
+[Service]
+Type=simple
+
+# Abort if root redirector is explicitly disabled
+ExecStartPre=/usr/bin/keybase -c /etc/keybase/config.json config get --direct --assert-false --assert-ok-on-nil disable-root-redirector
```
songgao: This looks cool! Didn't seem to work on macOS when I played with it though. Does it give an exit code 0 only if assert passes, i.e. `disable-root-redirector==false`?

libkbfs/init.go
songgao: Hmm that might be different. It's talking about error on sending the signal on Windows, not subscribing to it. `signal.Notify` doesn't even return an error. So perhaps it's fine as long as it builds. We'd just miss them if some are not supported on the platform. Also it seems `os.Interrupt` is actually [supported](https://golang.org/pkg/os/signal/#hdr-Windows) by `signal.Notify` on Windows. So I think we're good!

The [`signal` doc](https://golang.org/pkg/os/signal/#hdr-Default_behavior_of_signals_in_Go_programs) mentions more signals. WDYT about those? I haven't tried catching them myself, so not sure about the behaviors around those.

kbfsfuse/main.go
```diff
@@ -73,6 +73,9 @@ func start() *libfs.Error {
 		if err != nil {
 			return libfs.InitError(err.Error())
 		}
+		// If a mountdir was not set by keybase config set mountdir,
+		// the service returns a default value which may or may not
+		// exist yet.
```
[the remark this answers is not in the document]
heronhaye: Thanks for the review.
Yes, I believe that should never happen. Should I get rid of that branch then?

libfuse/mounter.go
```diff
@@ -98,6 +116,14 @@ func (m *mounter) Unmount() (err error) {
 	return
 }
 
+func (m *mounter) DeleteMountdir() (err error) {
+	err = os.RemoveAll(m.options.MountPoint)
```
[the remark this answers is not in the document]
heronhaye: Hmm, the directory above `kbfs` is used by service for other things in the default mountdirs. But you're right, this could leave additional directories if someone chooses a custom mountdir. I'm not sure how you would know which ones to clean up though. Maybe by saving the filenames on MkdirAll (probably need to write some custom traversal code)?

DeleteMountdir should only be called if the mount exited cleanly. Is it possible for that to happen but the mount is not empty afterwards?

libkbfs/init.go
heronhaye: To be honest I've never heard of most of them, but SIGPIPE, SIGPWR seem ok. Possibly also SIGQUIT and SIGABRT?

> Starting with Oracle Solaris 10 operating system, this option is ignored, as the operating system reserves two additional signals (called SIGJVM1 and SIGJVM2).

libfuse/mounter.go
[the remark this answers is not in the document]
heronhaye: Thought about it and I think it's a good idea just in case, added a commit. There's a slight TOCTTOU issue? Should we flock it first or is it ok?

libkbfs/init.go
heronhaye: Put up a (tentative) commit for syscall.SIGTERM, syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGABRT, syscall.SIGPWR in addition to os.Interrupt. SIGPIPE seems to only be a crash if the program couldn't print to stdout/stderr, otherwise the File.Write will return an error, so I thought it would be incorrect to catch that case as well.

>This means that, by default, command line programs will behave like typical Unix command line programs, while other programs will not crash with SIGPIPE when writing to a closed network connection.

jzila:
Moved to keybase/client#15321

packaging/linux/systemd/keybase-redirector.service
songgao: Got it; thanks for the explanation. That's neat! Both names sound OK to me as long as it's explained in the CLI doc.

libkbfs/init.go
songgao: Thanks Surya! Could you please make sure sending SIGABRT (and its [friends](https://golang.org/pkg/os/signal/#hdr-Default_behavior_of_signals_in_Go_programs)) to the process still causes it to print stacktraces? The doc doesn't seem to mention what happens when they are caught.

libfuse/mounter.go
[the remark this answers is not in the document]
songgao: Yeah I was thinking recording the created directories. We do need to check to make sure they are empty before removing, so not sure if it's worth it.

Re TOCTTOU, by "rename dance", I meant atomically renaming the directory atomically before doing anything, but I guess there's still no way to make it contingent to the empty check. Flock sounds good to me, but I don't know enough about it. I think Strib did something similar (checking if mount target is empty). @strib do you remember where it was?

PR #2000, opened 2019-01-04
dokan: Update some cgo handling for Go 1.11 changes

taruti:
cgo rules changed very slightly with Go 1.11. One laptop has it and dokan binding refused to compile.


jzila:
Closing in favor of keybase/client#15323

PR #2001, opened 2019-01-04
libdokan: Refuse to execute files

taruti:


jzila:
Closing in favor of keybase/client#15322

